PR #32, opened 2018-02-14
[WIP] api: an early sketch for kata libary api

laijs:
A api set that intends for bridging cri(fratki,hyperd) and
oci(kata-runtime) with kata.

Signed-off-by: Lai Jiangshan <jiangshanlai@gmail.com>

### DIFF
--- a/api/agent.go
+++ b/api/agent.go
@@ -1,0 +1,55 @@
+package api
+
+import (
+	"encoding/json"
+	"syscall"
+
+	ocispecs "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+type AgentConfig struct {
+	// ConnProto: the method to connect to agent(or proxy)
+	// "unix": the paths are unix socket
+	// "vsock": use cid and port pair to connect to the agent (kata-only)
+	// "kvmtool": the paths are kvmtool driver created pty slaver with escaping character
+	ConnProto string
+	// AgentProto: the agent protocol
+	// "hyperstart": orginal hyperstart protocol
+	// "kata": kata agent protocol
+	// "kata-yamux": kata agent protocol over yamux
+	AgentProto string
+
+	HyperstartCtlPath, HyperstartStreamPath string
+	KataPath                                string
+	KataVsockCid, KataVsockPort             uint32
+}
+
+type Agent interface {
+	Config() AgentConfig
+	Dump() json.RawMessage
+	Release() error
+	Destroy()
+
+	PauseSync() error
+	Unpause() error
+
+	CreateContainer(container string, user *UserGroupInfo, storages []*Storage, c *ocispecs.Spec) error
+	StartContainer(container string) error
+	ExecProcess(container, process string, user *UserGroupInfo, p *ocispecs.Process) error
+	SignalProcess(container, process string, signal syscall.Signal) error
+	WaitProcess(container, process string) int
+
+	StdioPipe(container, process string) (stdinPipe io.WriteCloser, stdout io.Reader, stderr io.Reader, err error)
+	TtyWinResize(container, process string, row, col uint16) error
+
+	StartSandbox(sb *SandboxNetworkConfig, storages []*Storage) error
+	DestroySandbox() error
+	UpdateRoute(r []Route) error
+	UpdateInterface(ifs []Interface) error
+	//OnlineCpuMem() error
+}
+
+// proxies, shim should be implemented via AgentBuilder (and Agent.CreateContainer() Agent.ExecProcess())
+type AgentBuilder interface {
+	Build(AgentConfig) Agent
+}
--- a/api/agent.go
+++ b/api/agent.go
@@ -1,12 +1,5 @@
 package api
 
-import (
-	"encoding/json"
-	"syscall"
-
-	ocispecs "github.com/opencontainers/runtime-spec/specs-go"
-)
-
 type AgentConfig struct {
 	// ConnProto: the method to connect to agent(or proxy)
 	// "unix": the paths are unix socket
@@ -26,30 +19,31 @@ type AgentConfig struct {
 
 type Agent interface {
 	Config() AgentConfig
-	Dump() json.RawMessage
-	Release() error
-	Destroy()
 
-	PauseSync() error
-	Unpause() error
-
-	CreateContainer(container string, user *UserGroupInfo, storages []*Storage, c *ocispecs.Spec) error
-	StartContainer(container string) error
-	ExecProcess(container, process string, user *UserGroupInfo, p *ocispecs.Process) error
-	SignalProcess(container, process string, signal syscall.Signal) error
-	WaitProcess(container, process string) int
-
-	StdioPipe(container, process string) (stdinPipe io.WriteCloser, stdout io.Reader, stderr io.Reader, err error)
-	TtyWinResize(container, process string, row, col uint16) error
-
-	StartSandbox(sb *SandboxNetworkConfig, storages []*Storage) error
-	DestroySandbox() error
-	UpdateRoute(r []Route) error
-	UpdateInterface(ifs []Interface) error
-	//OnlineCpuMem() error
+	// methods are private in the lib
 }
 
-// proxies, shim should be implemented via AgentBuilder (and Agent.CreateContainer() Agent.ExecProcess())
 type AgentBuilder interface {
 	Build(AgentConfig) Agent
 }
+
+// return the default AgentBuilder provided by the lib
+func DirectConnect() AgentBuilder {
+}
+
+// Example code for how to use AgentBuilder with proxy
+// The way how to luanch proxy is user (an oci-runtime-cli implementation) specific.
+// So this code can be put in runtime code.
+//
+// Proxy is an implementation of kata-agent-api, it fits for AgentBuilder
+//
+// struct proxyAgent {
+//    proxyBinary string
+//    // other fields
+// }
+// func (p proxyAgent) Build(configToVM AgentConfig) Agent {
+//    var configToProxy AgentConfig
+//    configToProxy, err := p.launchProxy(configToVM)
+//    if err != nil { return nil }
+//    return DirectConnect().Build(configToProxy)
+// }
--- a/api/network.go
+++ b/api/network.go
@@ -1,0 +1,45 @@
+package api
+
+type InterfaceConfig struct {
+	Id      string
+	Name    string
+	Lo      bool
+	Bridge  string
+	IP      string
+	MAC     string
+	MTU     uint64
+	Gateway string
+	TapName string
+	Options string
+}
+
+type InterfaceDevice struct {
+}
+
+type InterfaceDescription struct {
+}
+
+type PortDescription struct {
+	HostPort      int32
+	ContainerPort int32
+	Protocol      string
+}
+
+type NeighborNetworks struct {
+	InternalNetworks []string
+	ExternalNetworks []string
+}
+
+type SandboxNetworkConfig struct {
+	Hostname   string
+	Dns        []string
+	Neighbors  *NeighborNetworks
+	DnsOptions []string
+	DnsSearch  []string
+}
+
+type Network interface {
+	ID() string
+	Add(InterfaceConfig) (InterfaceDvice, error)
+	Remove(InterfaceDeive) error
+}
--- a/api/persistence.go
+++ b/api/persistence.go
@@ -1,0 +1,27 @@
+package api
+
+import (
+	"encoding/json"
+	"syscall"
+)
+
+// Save changes that needs to be persistent. (save the config of inserted volume and used scsi id for example)
+// Rename it to ResourceStorage?
+type Persistence interface {
+	// new resources [and lock the filelock]
+	CreateSandbox(id string) error
+	// store the sandbox (call sandbox.Dump() and save)
+	StoreSandbox(*Sandbox) error
+	// TODO more finegrain store operations
+
+	// [lock the filelock and] build sandbox from existing running sandbox
+	ObtainSandbox(id string, builder SandboxBuilder) (*Sandbox, error)
+	// release sandbox [and unlock the filelock]
+	ReleaseSandbox(*Sandbox) error
+
+	// queries
+	ListSandboxs() []SandboxStatus
+	ListContainers() []ContainerStatus
+	SandboxOfContainer(string) string
+	// TODO more finegrain query/fetch operations
+}
--- a/api/provision.go
+++ b/api/provision.go
@@ -1,0 +1,84 @@
+package api
+
+import (
+	"encoding/json"
+)
+
+type VMConfig struct {
+	CPU    int
+	Memory int
+	Kernel string
+	Initrd string // it can be nil, boot image should be added via AddDevice() in this case
+	//Bios   string
+	//Cbfs   string
+
+	CPUMax        int
+	MemSlot       int
+	MemMax        int
+	MemPath       string
+	MemPathShared bool
+
+	// to be filled
+}
+
+type QosConfig struct {
+	CgroupPath string
+
+	// For network QoS (kilobytes/s)
+	InboundAverage  string
+	InboundPeak     string
+	OutboundAverage string
+	OutboundPeak    string
+
+	// to be filled
+}
+
+type Device interface {
+}
+
+type VM interface {
+	// start to a fine state: example hotplug is workable, vm console is workable.
+	Start()
+	//Shutdown()
+	Destroy()
+	Snapshot() (Snapshot, error)
+	Dump() json.RawMessage
+	Release() error
+
+	Pause() error
+	Resume() error
+
+	//ConsoleUrl() string
+	//Console() io.Reader
+	Stats() (*VMStats, error)
+	// update qos config,cpu/memory,...
+	Update( /*todo*/ ) error
+
+	Capabilities() Capabilities
+
+	// add cpu/memory/storage/network and other devices.
+	// if the vm is not started, Device might be added on boot instead of hutpluging
+	AddDevice(Device) (kata.Device, error)
+	RemoveDevice(Device) error
+
+	SetCpu(cpu int)
+	SetMemory(cpu int)
+	// add storage.
+	// The returned storage is use for agent api, and it is
+	// not the same content of the original one.
+	// exmple1: source path is different between host and vm.
+	// exmple2: when a storage with driver=qcow2 is added, the dirver is
+	// handled on the runtime instead of the agent. And the agent can use
+	// the resulted device as normal block device (without dirver field).
+	AddStorage(Storage) (Storage, error)
+	RemoveStorage(Storage) error
+	AddNIC(NetowrkDevice) (Interface, error)
+	RemoveNIC(NetowrkDevice) error
+}
+
+type VMBuilder interface {
+	Build(VMConfig, AgentConfig, QosConfig) (VM, error)
+	BuildFromReleased(json.RawMessage) (VM, error)
+	BuildFromSnapshot(Snapshot, VMConfig, AgentConfig, QosConfig) (VM, error)
+	Capabilities() BuilderCapabilities
+}
--- a/api/sandbox.go
+++ b/api/sandbox.go
@@ -1,0 +1,100 @@
+package api
+
+import (
+	"context"
+	"syscall"
+
+	"github.com/hyperhq/runv/api"
+	ocispecs "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+type ContainerConfig struct {
+	Id string
+
+	UGI    *UserGroupInfo
+	Mounts map[string]Mount
+
+	OciSpec ocispecs.Spec
+}
+
+type Mount struct {
+	StorageID string
+	Path      string
+	ReadOnly  bool
+}
+
+type UsersAndGroups struct {
+	User             string
+	Group            string
+	AdditionalGroups []string
+}
+
+type ProcessConfig struct {
+	Container      string
+	Id             string
+	UsersAndGroups *UsersAndGroups
+
+	OciProcess ocispecs.Process
+}
+
+//example implementation
+//type Sandbox struct {
+//	VM    VM
+//	Agent Agent
+//	Persistence Persistence // save it when every major operation
+//
+//	Devices map[string]AddedDevice
+//
+//	Containers map[string]Container
+//}
+
+type SandboxBuilder struct {
+	VMBuilder
+	AgentBuider
+	Persistence
+	netowrks map[string]Network
+}
+
+func SynthesizeSandbox(VM, Agent, Persistence, map[string]Network) *Sandbox {
+}
+
+func CreateSandbox(SandboxBuider, VMConfig, AgentConfig, SandboxNetworkConfig) *Sandbox {
+
+}
+
+// most sandbox api will call into vm api, and them agent api
+type Sandbox interface {
+	Start(b *BootConfig) (err error)
+	Dump() json.RawMessage
+	Release() error
+	Shutdown() error
+	Destroy() error
+	// get notified when vm is shutdown/killed/destroyed.
+	SandboxDone() context.Context
+
+	UpdateRoute( /*todo*/ ) error
+	AddNic( /*todo*/ ) error
+	AllNics() []*InterfaceConfig
+	DeleteNic(id string) error
+	UpdateNic( /*todo*/ ) error
+	SetCpus(cpus int) error
+	SetMem(totalMem int) error
+	AddStorage(s Storage) error
+	RemoveStorage(name string) error
+
+	AddContainer(c *ContainerConfig) error
+	RemoveContainer(id string) error
+	StartContainer(id string) error
+	AddProcess(process *ProcessConfig) error
+	WaitProcess(container, process string, ctx context.Context) (int, error)
+	SignalProcess(container, process string, signal syscall.Signal) error
+
+	TerminalResize(containerId, execId string, row, column int) error
+	// called when non-shim(fratki/hyperd)
+	StdioPipe(container, process string) (stdinPipe io.WriteCloser, stdout io.Reader, stderr io.Reader, err error)
+
+	Stats() *SandboxStats
+	ContainerList() []string
+	Pause() error
+	Resume() error
+}
--- a/api/storage.go
+++ b/api/storage.go
@@ -1,0 +1,10 @@
+package api
+
+type Storage struct {
+	Name          string
+	Driver        string
+	DriverOptions map[string]string
+	Source        string
+	Fstype        string
+	MoutOptions   []string
+}


api/agent.go
```diff
@@ -0,0 +1,59 @@
...
+	AgentProto string
+
+	HyperstartCtlPath, HyperstartStreamPath string
+	KataPath                                string
+	KataVsockCid, KataVsockPort             uint32
+}
+
+type Agent interface {
+	Config() AgentConfig
+	Realese() (json.RawMessage, error)
```
jodh-intel: typo: `Release()`.

api/agent.go
```diff
@@ -0,0 +1,59 @@
...
+	Destory()
```
jodh-intel: typo: `Destroy()`.

api/agent.go
```diff
@@ -0,0 +1,59 @@
...
+
+	PauseSync() error
+	Unpause() error
+
+	APIVersion() (uint32, error)
```
jodh-intel: How about making this use semantic versioning (https://semver.org/) as that would be more flexible for clients I think:

```
APIVersion() (string, error)
```

api/agent.go
```diff
@@ -0,0 +1,59 @@
...
+	ReadStdout(container, process string, data []byte) (int, error)
+	ReadStderr(container, process string, data []byte) (int, error)
+	CloseStdin(container, process string) error
+	TtyWinResize(container, process string, row, col uint16) error
+
+	StartSandbox(sb *SandboxNetworkConfig, storages []*Storage) error
+	DestroySandbox() error
+	UpdateRoute(r []Route) error
+	UpdateInterface(ifs []Interface) error
+	//OnlineCpuMem() error
```
jodh-intel: Remove or uncomment?

api/network.go
```diff
@@ -0,0 +1,45 @@
+package api
+
+type InterfaceConfig struct {
+	Id      string
+	Name    string
+	Lo      bool
+	Bridge  string
+	Ip      string
+	Mac     string
+	Mtu     uint64
```
jodh-intel: I'd prefer `IP`, `MAC` and `MTU` here as they are all abbreviations.

api/network.go
```diff
@@ -0,0 +1,45 @@
...
+	Gw      string
```
jodh-intel: Maybe `Gateway` for maximum clarity.

api/network.go
```diff
@@ -0,0 +1,45 @@
...
+	TapName string
+	Options string
+}
+
+type InterfaceDeive struct {
```
jodh-intel: `InterfaceDevice` ?

api/network.go
```diff
@@ -0,0 +1,45 @@
...
+	Hostname   string
+	Dns        []string
+	Neighbors  *NeighborNetworks
+	DnsOptions []string
+	DnsSearch  []string
+}
+
+type Network interface {
+	ID() string
+	Add(InterfaceConfig) (InterfaceDeive, error)
```
jodh-intel: `InterfaceDevice`?

api/persistence.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+
+type Persistence interface {
+	// [new and lock the filelock and] store the just created sandbox
+	NewSandbox(*Sandbox) error
+	// NewContainer(*Container) error
+	// [lock the filelock and] build sandbox from existing running sandbox
+	GetSandbox(id string, SandboxBuilder) (*Sandbox, error)
+	// realese sandbox [and unlock the filelock]
```
jodh-intel: typo: "release".

api/persistence.go
```diff
@@ -0,0 +1,22 @@
...
+	"syscall"
+)
+
+
+type Persistence interface {
+	// [new and lock the filelock and] store the just created sandbox
+	NewSandbox(*Sandbox) error
+	// NewContainer(*Container) error
+	// [lock the filelock and] build sandbox from existing running sandbox
+	GetSandbox(id string, SandboxBuilder) (*Sandbox, error)
```
jodh-intel: Maybe just `Sandbox(...)` (https://golang.org/doc/effective_go.html#Getters).

api/persistence.go
```diff
@@ -0,0 +1,22 @@
...
+	// realese sandbox [and unlock the filelock]
+	PutSandbox(*Sandbox) error
```
jodh-intel: `ReleaseSandbox()` is clearer than `PutSandbox()` imho.

api/persistence.go
```diff
@@ -0,0 +1,22 @@
...
+
+	// queries
+	ListSandboxs()[]SandboxStatus
```
jodh-intel: `ListSandboxes()`

api/persistence.go
```diff
@@ -0,0 +1,22 @@
+package api
+
+import (
+	"encoding/json"
+	"syscall"
+)
+
+
+type Persistence interface {
```
jodh-intel: I don't really understand why this is called `Persistence`.

api/persistence.go
```diff
@@ -0,0 +1,22 @@
...
+	// NewContainer(*Container) error
+	// [lock the filelock and] build sandbox from existing running sandbox
+	GetSandbox(id string, SandboxBuilder) (*Sandbox, error)
+	// realese sandbox [and unlock the filelock]
+	PutSandbox(*Sandbox) error
+
+	// queries
+	ListSandboxs()[]SandboxStatus
+	ListContainers()[]ContainerStatus
+	SandboxOfContainer(string)string
```
jodh-intel: What does this do?

api/sandbox.go
```diff
@@ -0,0 +1,114 @@
...
+type Sandbox struct {
+	VM    VM
+	Agent Agent
+
+	Devices map[string]AddedDevice
+
+	Containers map[string]Container
+}
+
+type SandboxBuider struct {
```
jodh-intel: `SandboxBuilder`.

api/sandbox.go
```diff
@@ -0,0 +1,114 @@
...
+func CreateSandbox(SandboxBuider, VMConfig, AgentConfig, SandboxNetworkConfig) *Sandbox {
+
+}
+
+// most sandbox api will call into vm api, and them agent api
+func (sandbox *Sandbox) Start(b *BootConfig) (err error) {
+}
+func (sandbox *Sandbox) Release() ([]byte, error) {
+}
+func (sandbox *Sandbox) SanboxDone() context.Context {
```
jodh-intel: `SandboxDone()`.

api/sandbox.go
```diff
@@ -0,0 +1,114 @@
...
+}
+func (sandbox *Sandbox) Shutdown() api.Result {
+}
+func (sandbox *Sandbox) Destory() {
```
jodh-intel: `Destroy()`.

api/storage.go
```diff
@@ -0,0 +1,10 @@
+package api
+
+type Storage struct {
+	Name          string
+	Driver        string
+	DriverOptions map[string]string
+	Source        string
+	Fstype        string
+	MoutOption    []string
```
jodh-intel: `MountOptions` (plural?)

api/provision.go
```diff
@@ -0,0 +1,83 @@
...
+	Pause() error
+	Resume() error
+
+	//ConsoleUrl() string
+	//Console() io.Reader
+	Stats() (*VMStats, error)
+	// update qos config,cpu/memory,...
+	Update( /*todo*/ ) error
+
+	Cappabilities() Cappabilities
```
jodh-intel: `Capabilities()`.

api/provision.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+type Device interface {
+}
+
+type VM interface {
+	// start to a fine state: example hotplug is workable, vm console is workable.
+	Start()
+	//Shutdown()
+	Destory()
```
jodh-intel: `Destroy()`.

api/provision.go
```diff
@@ -0,0 +1,83 @@
...
+	//Console() io.Reader
+	Stats() (*VMStats, error)
+	// update qos config,cpu/memory,...
+	Update( /*todo*/ ) error
+
+	Cappabilities() Cappabilities
+
+	// add cpu/memory/storage/network and other devices.
+	// if the vm is not started, Device might be added on boot instead of hutpluging
+	AddDeivce(Device) (kata.Device, error)
```
jodh-intel: `AddDevice()`

api/provision.go
```diff
@@ -0,0 +1,83 @@
...
+	// add storage.
+	// The returned storage is use for agent api, and it is
+	// not the same content of the original one.
+	// exmple1: source path is different between host and vm.
+	// exmple2: when a storage with driver=qcow2 is added, the dirver is
+	// handled on the runtime instead of the agent. And the agent can use
+	// the resulted device as normal block device (without dirver field).
+	AddStorage(Storage) (Storage, error)
+	RemoveStorage(Storage) error
+	AddNic(NetowrkDevice) (Interface, error)
```
jodh-intel: `AddNIC(NetworkDevice)`?

api/provision.go
```diff
@@ -0,0 +1,83 @@
...
+	RemoveNic(NetowrkDevice) error
```
jodh-intel: `RemoveNIC(NetworkDevice)`?

api/provision.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+type VMBuilder interface {
+	Build(VMConfig, AgentConfig, QosConfig) (VM, error)
+	BuildFromReleased(json.RawMessage) (VM, error)
+	BuildFromSnapshot(Snapshot, VMConfig, AgentConfig, QosConfig) (VM, error)
+	Cappabilities() BuilderCappabilities
```
jodh-intel: `Capabilities() BuilderCapabilities`.

api/sandbox.go
```diff
@@ -0,0 +1,114 @@
...
+	"syscall"
+
+	"github.com/hyperhq/runv/api"
+	ocispecs "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+type ContainerConfig struct {
+	Id string
+
+	UGI    *UserGroupInfo
```
jodh-intel: Maybe `UsersAndGroups` would make this a little clearer.

jodh-intel:
Thanks for putting this together @laijs. Just a few initial typo corrections mostly.

sameo:
@laijs If we go with proposal #33 , this PR will have to be redefined as the current virtcontainers API has a lot of overlap with what what you're proposing here.

sameo:
@laijs @gnawux The virtcontainers 1.0 API is now documented here:

https://github.com/containers/virtcontainers/blob/master/documentation/api/1.0/api.md

bergwolf:
@sameo I thought if the virtcontainers API has a lot of overlap with the API in this PR, it will be quite easy for the virtcontainers framework to support the API defined here. Why do you say it has to be redefined?

api/agent.go
```diff
@@ -0,0 +1,59 @@
...
+	ConnProto string
+	// AgentProto: the agent protocol
+	// "hyperstart": orginal hyperstart protocol
+	// "kata": kata agent protocol
+	// "kata-yamux": kata agent protocol over yamux
+	AgentProto string
+
+	HyperstartCtlPath, HyperstartStreamPath string
+	KataPath                                string
+	KataVsockCid, KataVsockPort             uint32
```
bergwolf: These also need documentation especially `KataPath` which is not a self-explaining name.

api/persistence.go
laijs: the 'get' is from 'get/put' pair. It is not a pure "read" operation, it might lock the filelock or get a reference.

api/persistence.go
laijs: `kata-runtime kill container-name`
we need to find the sandbox from the container-name at first.
In runv, there is a symlink in the container's state root.

api/persistence.go
laijs: In cc-runtime, the same functionality is called "filesystem". And even in runv-cli, the data is also saved on the local filesystem. But the name of the media for saving the data doesn't represent the functionality well. In other cases, we might save the data in the DB(hyperd,fratki) or k8s-etcd.

The functionality is "retrieving/dumping the data from running sandbox + saving it on some places + restructuring the running sandbox (golang objects) for later operations". `Persistence` is competent at describing it. I don't insist on this name if some other name is proposed. And I can change it to "Filesystem" if it gets some votes.

sameo:
@bergwolf 
> I thought if the virtcontainers API has a lot of overlap with the API in this PR, it will be quite easy for the virtcontainers framework to support the API defined here. Why do you say it has to be redefined?

I think there's some confusion here. There is indeed a lot of overlap between what's proposed here and the current virtcontainers API. So are you proposing that we implement the kata containers runtime API **on top** of the virtcontainers API, or would the (potentially modified) virtcontainers API become the runtime API? Given the overlap between the 2, I think the first option would mostly be redundant.

bergwolf:
@sameo I certainly do not want to see two sets of APIs in the kata runtime. I think the goal of the API discussion here is to have a unified API to support both kata command line and CRI native implementations.

There are already many overlaps between @laijs's PR and the current virtconatiners API. So please comment on what the gaps you have in mind (both design-wise and implementation-wise) and we can work together to improve the API definition to a state that all stakeholders are satisfied. When we all agree that the API is good enough, we should be able to implement the API with virtcontainers framework and use it to support both use cases.

sameo:
@bergwolf I'll do an analysis of what's currently missing from the current virtcontainers API to implement something that would satisfy this PR, at least from what I understand from it.

api/provision.go
```diff
@@ -0,0 +1,83 @@
...
+
+	CPUMax        int
+	MemSlot       int
+	MemMax        int
+	MemPath       string
+	MemPathShared bool
+
+	// to be filled
+}
+
```
sameo: A combination of those virtcontainers structures are equivalent to the above one:

```Go
// Resources describes VM resources configuration.
type Resources struct {
	// VCPUs is the number of available virtual CPUs.
	VCPUs uint

	// Memory is the amount of available memory in MiB.
	Memory uint
}

// HypervisorConfig is the hypervisor configuration.
type HypervisorConfig struct {
	// KernelPath is the guest kernel host path.
	KernelPath string

	// ImagePath is the guest image host path.
	ImagePath string

	// FirmwarePath is the bios host path
	FirmwarePath string

	// MachineAccelerators are machine specific accelerators
	MachineAccelerators string

	// HypervisorPath is the hypervisor executable host path.
	HypervisorPath string

	// DisableBlockDeviceUse disallows a block device from being used.
	DisableBlockDeviceUse bool

	// KernelParams are additional guest kernel parameters.
	KernelParams []Param

	// HypervisorParams are additional hypervisor parameters.
	HypervisorParams []Param

	// HypervisorMachineType specifies the type of machine being
	// emulated.
	HypervisorMachineType string

	// Debug changes the default hypervisor and kernel parameters to
	// enable debug output where available.
	Debug bool

	// DefaultVCPUs specifies default number of vCPUs for the VM.
	// Pod configuration VMConfig.VCPUs overwrites this.
	DefaultVCPUs uint32

	// DefaultMem specifies default memory size in MiB for the VM.
	// Pod configuration VMConfig.Memory overwrites this.
	DefaultMemSz uint32

	// DefaultBridges specifies default number of bridges for the VM.
	// Bridges can be used to hot plug devices
	DefaultBridges uint32

	// MemPrealloc specifies if the memory should be pre-allocated
	MemPrealloc bool

	// HugePages specifies if the memory should be pre-allocated from huge pages
	HugePages bool

	// Realtime Used to enable/disable realtime
	Realtime bool

	// Mlock is used to control memory locking when Realtime is enabled
	// Realtime=true and Mlock=false, allows for swapping out of VM memory
	// enabling higher density
	Mlock bool

	// DisableNestingChecks is used to override customizations performed
	// when running on top of another VMM.
	DisableNestingChecks bool

	// customAssets is a map of assets.
	// Each value in that map takes precedence over the configured assets.
	// For example, if there is a value for the "kernel" key in this map,
	// it will be used for the pod's kernel path instead of KernelPath.
	customAssets map[assetType]*asset
}
```

api/provision.go
```diff
@@ -0,0 +1,83 @@
...
+	CgroupPath string
+
+	// For network QoS (kilobytes/s)
+	InboundAverage  string
+	InboundPeak     string
+	OutboundAverage string
+	OutboundPeak    string
+
+	// to be filled
+}
```
sameo: virtcontainers does not have this kind of setting, that is a potential gap. Could you please explain where that is used and what it's for?

api/sandbox.go
```diff
@@ -0,0 +1,114 @@
...
+	Mounts map[string]Mount
+
+	OciSpec ocispecs.Spec
+}
```
sameo: virtcontainers `ContainerConfig`:

```Go
// ContainerConfig describes one container runtime configuration.
type ContainerConfig struct {
	ID string

	// RootFs is the container workload image on the host.
	RootFs string

	// ReadOnlyRootfs indicates if the rootfs should be mounted readonly
	ReadonlyRootfs bool

	// Cmd specifies the command to run on a container
	Cmd Cmd

	// Annotations allow clients to store arbitrary values,
	// for example to add additional status values required
	// to support particular specifications.
	Annotations map[string]string

	Mounts []Mount

	// Device configuration for devices that must be available within the container.
	DeviceInfos []DeviceInfo
}
```

api/sandbox.go
```diff
@@ -0,0 +1,114 @@
...
+
+type Mount struct {
+	StorageID string
+	Path      string
+	ReadOnly  bool
+}
+
```
sameo: virtcontainer's `Mount`:

```Go
// Mount describes a container mount.
type Mount struct {
	Source      string
	Destination string

	// Type specifies the type of filesystem to mount.
	Type string

	// Options list all the mount options of the filesystem.
	Options []string

	// HostPath used to store host side bind mount path
	HostPath string

	// ReadOnly specifies if the mount should be read only or not
	ReadOnly bool
}
```

api/sandbox.go
```diff
@@ -0,0 +1,114 @@
...
+type Sandbox struct {
+	VM    VM
+	Agent Agent
+
+	Devices map[string]AddedDevice
+
+	Containers map[string]Container
+}
+
```
sameo: virtcontainers does not export the `Pod` structure directly to the top level API, but the API caller should use the `PodConfig` structure instead:

```Go
// PodConfig is a Pod configuration.
type PodConfig struct {
	ID string

	Hostname string

	// Field specific to OCI specs, needed to setup all the hooks
	Hooks Hooks

	// VMConfig is the VM configuration to set for this pod.
	VMConfig Resources

	HypervisorType   HypervisorType
	HypervisorConfig HypervisorConfig

	AgentType   AgentType
	AgentConfig interface{}

	ProxyType   ProxyType
	ProxyConfig ProxyConfig

	ShimType   ShimType
	ShimConfig interface{}

	NetworkModel  NetworkModel
	NetworkConfig NetworkConfig

	// Volumes is a list of shared volumes between the host and the Pod.
	Volumes []Volume

	// Containers describe the list of containers within a Pod.
	// This list can be empty and populated by adding containers
	// to the Pod a posteriori.
	Containers []ContainerConfig

	// Annotations keys must be unique strings and must be name-spaced
	// with e.g. reverse domain notation (org.clearlinux.key).
	Annotations map[string]string
}
```

api/sandbox.go
```diff
@@ -0,0 +1,114 @@
...
+type ProcessConfig struct {
+	Container string
+	Id        string
+	UGI       *UserGroupInfo
+
+	OciProcess ocispecs.Process
+}
+
```
sameo: I think the virtcontainers equivalent for this one is `Cmd`:

```Go
// Cmd represents a command to execute in a running container.
type Cmd struct {
	Args    []string
	Envs    []EnvVar
	WorkDir string

	// Note that these fields *MUST* remain as strings.
	//
	// The reason being that we want runtimes to be able to support CLI
	// operations like "exec --user=". That option allows the
	// specification of a user (either as a string username or a numeric
	// UID), and may optionally also include a group (groupame or GID).
	//
	// Since this type is the interface to allow the runtime to specify
	// the user and group the workload can run as, these user and group
	// fields cannot be encoded as integer values since that would imply
	// the runtime itself would need to perform a UID/GID lookup on the
	// user-specified username/groupname. But that isn't practically
	// possible given that to do so would require the runtime to access
	// the image to allow it to interrogate the appropriate databases to
	// convert the username/groupnames to UID/GID values.
	//
	// Note that this argument applies solely to the _runtime_ supporting
	// a "--user=" option when running in a "standalone mode" - there is
	// no issue when the runtime is called by a container manager since
	// all the user and group mapping is handled by the container manager
	// and specified to the runtime in terms of UID/GID's in the
	// configuration file generated by the container manager.
	User                string
	PrimaryGroup        string
	SupplementaryGroups []string

	Interactive     bool
	Console         string
	Detach          bool
	NoNewPrivileges bool
	Capabilities    LinuxCapabilities
}
```

api/sandbox.go
```diff
@@ -0,0 +1,114 @@
...
+func (sandbox *Sandbox) Stats() *SandboxStats {
+}
+func (sandbox *Sandbox) ContainerList() []string {
+}
+func (sandbox *Sandbox) Pause() error {
+}
+func (sandbox *Sandbox) Resume() error {
+}
+func (sandbox *Sandbox) Save(path string) error {
+}
```
sameo: The virtcontainers containers and pod API implement a large part of the above API:

#### `CreatePod`
```Go
// CreatePod is the virtcontainers pod creation entry point.
// CreatePod creates a pod and its containers. It does not start them.
func CreatePod(podConfig PodConfig) (VCPod, error)
```

#### `DeletePod`
```Go
// DeletePod is the virtcontainers pod deletion entry point.
// DeletePod will stop an already running container and then delete it.
func DeletePod(podID string) (VCPod, error)
```

#### `StartPod`
```Go
// StartPod is the virtcontainers pod starting entry point.
// StartPod will talk to the given hypervisor to start an existing
// pod and all its containers.
func StartPod(podID string) (VCPod, error)
```

#### `StopPod`
```Go
// StopPod is the virtcontainers pod stopping entry point.
// StopPod will talk to the given agent to stop an existing pod
// and destroy all containers within that pod.
func StopPod(podID string) (VCPod, error)
```

#### `RunPod`
```Go
// RunPod is the virtcontainers pod running entry point.
// RunPod creates a pod and its containers and then it starts them.
func RunPod(podConfig PodConfig) (VCPod, error)
```

#### `ListPod`
```Go
// ListPod is the virtcontainers pod listing entry point.
func ListPod() ([]PodStatus, error)
```

#### `StatusPod`
```Go
// StatusPod is the virtcontainers pod status entry point.
func StatusPod(podID string) (PodStatus, error)
```

#### `PausePod`
```Go
// PausePod is the virtcontainers pausing entry point which pauses an
// already running pod.
func PausePod(podID string) (VCPod, error)
```

#### `ResumePod`
```Go
// ResumePod is the virtcontainers resuming entry point which resumes
// (or unpauses) and already paused pod.
func ResumePod(podID string) (VCPod, error)
```

#### `CreateContainer`
```Go
// CreateContainer is the virtcontainers container creation entry point.
// CreateContainer creates a container on a given pod.
func CreateContainer(podID string, containerConfig ContainerConfig) (VCPod, VCContainer, error)
```

#### `DeleteContainer`
```Go
// DeleteContainer is the virtcontainers container deletion entry point.
// DeleteContainer deletes a Container from a Pod. If the container is running,
// it needs to be stopped first.
func DeleteContainer(podID, containerID string) (VCContainer, error)
```

#### `StartContainer`
```Go
// StartContainer is the virtcontainers container starting entry point.
// StartContainer starts an already created container.
func StartContainer(podID, containerID string) (VCContainer, error)
```

#### `StopContainer`
```Go
// StopContainer is the virtcontainers container stopping entry point.
// StopContainer stops an already running container.
func StopContainer(podID, containerID string) (VCContainer, error)
```

#### `EnterContainer`
```Go
// EnterContainer is the virtcontainers container command execution entry point.
// EnterContainer enters an already running container and runs a given command.
func EnterContainer(podID, containerID string, cmd Cmd) (VCPod, VCContainer, *Process, error)
```

#### `StatusContainer`
```Go
// StatusContainer is the virtcontainers container status entry point.
// StatusContainer returns a detailed container status.
func StatusContainer(podID, containerID string) (ContainerStatus, error)
```

#### `KillContainer`
```Go
// KillContainer is the virtcontainers entry point to send a signal
// to a container running inside a pod. If all is true, all processes in
// the container will be sent the signal.
func KillContainer(podID, containerID string, signal syscall.Signal, all bool) error
```

#### `ProcessListContainer`
```Go
// ProcessListContainer is the virtcontainers entry point to list
// processes running inside a container
func ProcessListContainer(podID, containerID string, options ProcessListOptions) (ProcessList, error)
```

api/storage.go
```diff
@@ -0,0 +1,10 @@
...
+}
```
sameo: This `Storage` is partly covered by virtcontainers `Mount` but some fields are missing, and the name might be misleading. A few questions:

- In which context is this structure used?
- How is it different than a generic container Mount
- What is the `Driver` field about? Do you have one example to share?

gnawux:
> Compared to the virtcontainers API, this PR is more imperative. It exposes more structures and interfaces to the caller while virtcontainers provides a declarative only API.

@sameo when talking about "imperative" and "declarative", I assume you are talking about style or flavor rather than which is better or worse. 

Firstly, the two concepts are relative, there is no absolute declarative/imperative in most of the cases. Actually, comparing to kubernetes APIs, which is a typical declarative API set, CRI is a more "imperative" interface -- for example, having launch `PodSandbox`, it will create containers in sandbox independently, which implies the devices may be added to a sandbox after the `RunPodSandbox` operation.

In fact, the higher level APIs tend to be more declarative, which is application friendly and easier for operating a big application. On the other hand, the lower level APIs tend to be not so declarative, which is more efficient and close to the nature of how the devices work.

[Will add more explanation tomorrow... sorry it's too late for me]

api/agent.go
```diff
@@ -0,0 +1,59 @@
...
+	DestroySandbox() error
+	UpdateRoute(r []Route) error
+	UpdateInterface(ifs []Interface) error
+	//OnlineCpuMem() error
+}
+
+// proxies, shim should be implemented via AgentBuilder (and Agent.CreateContainer() Agent.ExecProcess())
+type AgentBuilder interface {
+	Build(AgentConfig) Agent
+}
```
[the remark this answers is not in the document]
laijs: The agent package and the sandbox package will be different packages. Therefore, Agent and its Methods have to be exposed so that the sandbox package can use the agent package.

sameo:
@gnawux 
> In fact, the higher level APIs tend to be more declarative, which is application friendly and easier for operating a big application. On the other hand, the lower level APIs tend to be not so declarative, which is more efficient and close to the nature of how the devices work.

Yes, definitely agree with that. I did not mean that one is better than the other. However I do believe that our lower level APIs should/could be internal and not external. The smaller the external API, the better (for security, UX, and design reasons).

gnawux:
@sameo 

> However I do believe that our lower level APIs should/could be internal and not external. The smaller the external API, the better (for security, UX, and design reasons).

Yes, I understand what you think about, and I agree we should keep a small set external APIs for applications. Here we have some thoughts on why we put some low-level interfaces here, and it is appreciated if you could figure out a better way for them:

The `Persistence` and `Sandbox` are the top level interfaces for customer application.

- `Sandbox`: there are three interfaces behind the `Sandbox`
  - `VM` gives the control of hypervisor to sandbox;
  - `Agent` gives the control of containers/processes via the agent in sandbox;
  - `Network` (multiple) 
- `Persistence`: is a resource storage engine, which could store the persistence data of sandboxes, and could be used for reconnecting to a running Sandbox.

A Sandbox is created with `CreateSandbox` API, which takes a `SandboxBuilder` together with sandbox configurations. The `SandboxBuilder` include a `VMBuilder` for creating `VM`, an `AgentBuilder` for creating `Agent`, and the `Persistence` for saving the `Sandbox`.

The Kata lib provides pre-defined builders for difference cases/drivers:

- VMBuilder may support different drivers for VMs, as well as factories.
  - Customer applications can also have its own implementations of VM via VMBuilder. 
- AgentBuilder may support generate `Agent` which
  - Connects agent with external proxy and shim processes;
  - Connects agent directly with gRPC API over vsock or yamux.

The application could provide their own implementation or wrapper for the builders, which provide the flexibility for the different use case.

Customer applications may implement their own Persistence for storing sandbox info in local filesystem, local database, or somewhere else.

gnawux:
Here is the detailed document for how the proposed APIs work: https://docs.google.com/document/d/1BJeIFWykgmPIr45GRPNoawPPMr-3ekTIpJwU-g5QsHs/edit?usp=sharing

jessfraz:
I think I would prefer modifying the virtcontainers API with whatever extra it needs. It seems like this is a bit over the top with regards to what it needs I would prefer minimal and then see what else is necessary from there.

gnawux:
@jessfraz The API is based on runV API (the `Sandbox` part) and with some low-level interfaces (which also exist in virtcontainers, and the debate is if we need to make them public here). 

It is obvious we are not a project from the scratch. With so much existing use cases from each runtime, we put a bigger API set for taking all we considered, and we'd like to learn your and other's comments on

- are these requirements valid, and 
- if some of them are valid, are there any better designs for the requirements.

I understand @sameo are seeking a minimal API set and don't want to expose the low-level interfaces like `Agent`, `VM`, etc. **I want to hide them as well.**  Having discussed with @laijs, I think we'd better put all these out, and collect what do you think about the requirements behind the interfaces as we state in https://github.com/kata-containers/runtime/pull/32#issuecomment-368357212 (and in the document: https://docs.google.com/document/d/1BJeIFWykgmPIr45GRPNoawPPMr-3ekTIpJwU-g5QsHs/edit?usp=sharing). 

What are the low-level interfaces we put here:

- `VM` gives the control of hypervisor to sandbox;
- `Agent` gives the control of containers/processes via the agent in sandbox;
- `Persistence`: is a resource storage engine, which could store the persistence data of sandboxes, and could be used for reconnecting to a running Sandbox.

Why we make the interfaces public though we hate to publish the detail:

- The application could either select the builders provided by kata, or provide their own implementation or wrapper for the builders, which provide the flexibility for the different use case.
  - Customer applications can also have its own implementations of VM via VMBuilder. VMBuilder may support different drivers for VMs, as well as factories.
  - AgentBuilder may support generate Agent which
    - Connects agent with external proxy and shim processes;
    - Connects agent directly with gRPC API over vsock or yamux.
- Customer applications may implement their own Persistence for storing sandbox info in local filesystem, local database, or somewhere else.

sameo:
@gnawux 
> What are the low-level interfaces we put here:
>
> `VM` gives the control of hypervisor to sandbox;

One of the key point of virtcontainers is to abstract this from the caller. If the VM lifecycle is handled outside of virtcontainers, we leave much of the complexity in the caller's hand. That's not what we want imho. We want to make the caller's life simple, and provide them with a container/sandbox API that's easy to use and close to what they're already familiar with. We want to make hardware virtualized containers just containers and hide the VM management away from them.

> `Agent` gives the control of containers/processes via the agent in sandbox;

If virtcontainers is missing containers management APIs, I'd be happy to consider extending the API. But we need to know which ones they are.
Providing a complete, low level agent API is basically saying that we can't define a proper container level API and are passing that up to the caller. Eventually we let the caller do a myriad of things that we don't control or understand, and that we can't validate at all.

> `Persistence`: is a resource storage engine, which could store the persistence data of sandboxes, and could be used for reconnecting to a running Sandbox.

So I see something a little concerning with this kind if API: We're basically letting the caller define its own persistent data storage format while in my opinion, the storage format is a part of the API. It *may* make sense to let callers implement their own backend driver but that should live at the transport/IO level: `ReadData`, `WriteData`, etc...virtcontainers should maintain the storage format and layout to be able to move from one backend to the other transparently.

That being said, do we have practical (i.e. existing) examples of a need for using anything else than a POSIX filesystem? Do we realistically want to use something like etcd or an SQL database to store virtcontainers data? I usually prefer to extend an API to support a real life use case rather than extending it up front and having to maintain it in the long term.

api/sandbox.go
laijs: @sameo  These APIs are just wrappers over Sandbox & Persistence APIs, there are too much oci-runtime specific. What we need is just to expose the Sandbox & Persistence APIs. (The Sandbox APIs are the APIs that take Sandbox as the receiver, not  sandbox name).

laijs:
@all The methods in the agent interface are changed to be private.

It can be reviewed in the second commit:
https://github.com/kata-containers/runtime/pull/32/commits/f491d42de1e65dae0d2bc2ad39c76c5222a1105b

sboeuf:
@laijs thanks for the change.
More globally, I think we might miss things in virtcontainers such as being able to add a new network interface on an existing pod, and there is no pod or container lifecycle event related to such a thing. For this kind of case, I would expect the virtcontainers API to be extended such as:
```go
// api.go
...

func AddNetwork(podID, interfaceName string) error {
        ...
}

...
```
This example shows that you could try highlight what is really missing by extending the API, instead of trying to redefine everything.

WeiZhang555:
I would suggest that we can provide a plugin framework to let everything with many choices pluggable, so that for something like persistence, we provide some choices like database and raw file, or something like an etcd plugin

gnawux:
@WeiZhang555 yes, I agree. The proposed low-level interfaces are designed for plugin. It is better to explicitly split plugin interfaces from the customer-facing interface.

api/sandbox.go
sameo: @bergwolf could you elaborate on how `Mount` would reference a `Storage`? Who decides that a container's mount is referring to a specific storage definition?

api/storage.go
```diff
@@ -0,0 +1,10 @@
+package api
+
+type Storage struct {
+	Name          string
+	Driver        string
+	DriverOptions map[string]string
+	Source        string
+	Fstype        string
+	MoutOptions   []string
+}
```
sameo: So `Storage` represents e.g. an RBD or ceph volume attached to a VM/Pod? It's a pod resource that can be then mounted in one or more containers? Is that a correct description?
And so `sandbox.AddStorage()` is the part that attaches the storage to the VM/Pod, without mounting it in any container?

api/sandbox.go
bergwolf: It is referenced by letting `OCI.Mount.Source` == `pb.Storage.Mountpoint` (or a subdirectory). The kata runtime API caller needs to craft proper `OCI.Mount.Source` to establish the mapping. It is a bit awkward than the original runv API but the change is needed because we want to use libcontainer in kata agent.

sboeuf: This does not need to be specified at the API level, this is very OCI specific. The kata agent implementation in virtcontainers `kata_agent.go` can take care of translating whatever is expected by the Kata agent itself.

api/storage.go
bergwolf: Yes, your description is correct. Just to be clear, `Storage` is not limited to RBD. `sandbox.AddStorage()` is supposed to call the hypervisor hotplug code. It won't mount the storage in any container because `sandbox.AddStorage()` can be called before `sandbox.AddContainer` thus there might be no containers in the sandbox at all.

sameo: So do you see any reason why it could not be handled as part of the sandbox/pod creation step? Or is `addStorage` potentially being called outside of the pod/sandbox creation step?

api/sandbox.go
bergwolf: @sboeuf we want to pass the OCI spec around through the API to better help kata runtime communicate with libcontainer. The OCI spec field in the agent's API is there for the same purpose -- so that we do not convert from OCI spec to kata spec and then back the OCI spec. Although we do need to make necessary modifications, the most of the OCI spec gets through untouched.

OTOH, the main purpose to have `sandbox.AddStorage()` is to support explicit hotplug and let upper layer prepare container resources (nics/volumes/rootfs etc.) in parallel. If we eventually decide not to pass the OCI spec around, we still need an `Mount` field in kata runtime's container spec to properly reference `Storage`.

api/storage.go
bergwolf: I'm not sure if I understand your question. In CRI, sandbox creation does not include all container specs. `sandbox.AddStorage` is supposed to be called when handling the CRI `CreateContainer()` request.

sameo: Right, so CRI `CreateContainer()` will eventually call into virtcontainers' `CreateContainer()` or `pod.CreateContainer()`: Can we not export `sandbox.AddStorage()` and make it an internal call of virtcontainers `CreateContainer()` or `pod.CreateContainer()`?

api/sandbox.go
bergwolf: The difference is that the container rootfs is prepared outside of the runtime library. By letting upper layer initiate the preparation in parallel, we can save hundreds of milliseconds when starting a container (actual time of saving depends on which storage is used though).

If we hide them in sandbox/container creation flow as in current virtcontainers API, we can do nothing but waiting when preparing the container rootfs.

I can live with OCI spec annotation that is really not different from passing through API. IMHO it gives better type checking to pass it through API. Just my two cents though.

api/storage.go
bergwolf: As I mentioned [above](https://github.com/kata-containers/runtime/pull/32#discussion_r173491512) to reply to @sboeuf, exporting these hotplug APIs will allow upper layer to prepare rootfs/volumes/nics in parallel that can save us hundreds of milliseconds when creating a new container.

Another reason to not reject `AddStorage` is that we will need to export `AddNic` anyway to support CNI even in the kata runtime CLI case, assuming @WeiZhang555 still wants the `kata-cli interface` subcommand similar to `runv interface` he implemented in runv. IOW, the hotplug APIs are not only useful for CRI native support. They are useful for kata-cli as well.

egernst:
With the merge of https://github.com/kata-containers/documentation/blob/master/design/kata-api-design.md, can this be closed, @laijs  @bergwolf ?

WeiZhang555:
I'm closing this because it's not needed any more as @egernst , feel free to reopen it if I'm wrong. @laijs @bergwolf

PR #54, opened 2018-03-14
[DNM] plugin: Implement an hypervisor plugin POC

### LINKED ISSUE #51: virtcontainers: add support for plugins
Add support for dynamically adding external virtcontainers interface implementations.  This allows for a clean way to allow for proprietary kata runtime extensions.

Suggest that we leverage golang's plugin [package](https://golang.org/pkg/plugin/).
Plugins to call into the virtcontainers plugin API (`RegisterPlugin()`)
Potentially supported plugin types:
* Hypervisors
* Metadata backend (For storing/fetching Kata runtime’s metadata)
* VM caching/cloning

Existing implementations (for example hypervisors) will be registered as plugins but still built-in, while external plugin implementations will be presented as dynamic libraries.

Let's use this issue to define the API.  This issue can be closed once the design is documented and plugin framework is merged.

sboeuf:
This pull request tries to address the needs for an hypervisor plugin into virtcontainers through a simple proof of concept.

Fixes #51

### DIFF
--- a/virtcontainers/hypervisor_plugin.go
+++ b/virtcontainers/hypervisor_plugin.go
@@ -1,0 +1,96 @@
+//
+// Copyright (c) 2018 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package virtcontainers
+
+import (
+	"fmt"
+	"plugin"
+)
+
+var (
+	getHypervisorPluginFuncName = "GetHypervisorPlugin"
+)
+
+type pluginHypervisor struct {
+	HypervisorPlugin
+}
+
+func (h *pluginHypervisor) init(pod *Pod) error {
+	plugIn, err := plugin.Open(pod.config.HypervisorConfig.PluginPath)
+	if err != nil {
+		return fmt.Errorf("Failed to open plugin path %q: %v",
+			pod.config.HypervisorConfig.PluginPath, err)
+	}
+
+	getHypervisorPluginFunc, err := plugIn.Lookup(getHypervisorPluginFuncName)
+	if err != nil {
+		return fmt.Errorf("Failed to lookup function %q: %v",
+			getHypervisorPluginFuncName, err)
+	}
+
+	h.HypervisorPlugin = getHypervisorPluginFunc.(func() HypervisorPlugin)()
+	if err != nil {
+		return fmt.Errorf("%s() failed: %v",
+			getHypervisorPluginFuncName, err)
+	}
+
+	return h.HypervisorPlugin.Init(pod)
+}
+
+func (h *pluginHypervisor) createPod(podConfig PodConfig) error {
+	return h.HypervisorPlugin.CreatePod(podConfig)
+}
+
+func (h *pluginHypervisor) startPod() error {
+	return h.HypervisorPlugin.StartPod()
+}
+
+func (h *pluginHypervisor) waitPod(timeout int) error {
+	return h.HypervisorPlugin.WaitPod(timeout)
+}
+
+func (h *pluginHypervisor) stopPod() error {
+	return h.HypervisorPlugin.StopPod()
+}
+
+func (h *pluginHypervisor) pausePod() error {
+	return h.HypervisorPlugin.PausePod()
+}
+
+func (h *pluginHypervisor) resumePod() error {
+	return h.HypervisorPlugin.ResumePod()
+}
+
+func (h *pluginHypervisor) addDevice(devInfo interface{}, devType DeviceType) error {
+	return h.HypervisorPlugin.AddDevice(devInfo, devType)
+}
+
+func (h *pluginHypervisor) hotplugAddDevice(devInfo interface{}, devType DeviceType) error {
+	return h.HypervisorPlugin.HotplugAddDevice(devInfo, devType)
+}
+
+func (h *pluginHypervisor) hotplugRemoveDevice(devInfo interface{}, devType DeviceType) error {
+	return h.HypervisorPlugin.HotplugRemoveDevice(devInfo, devType)
+}
+
+func (h *pluginHypervisor) getPodConsole(podID string) string {
+	return h.HypervisorPlugin.GetPodConsole(podID)
+}
+
+func (h *pluginHypervisor) capabilities() Capabilities {
+	return h.HypervisorPlugin.Capabilities()
+}
--- a/virtcontainers/pod.go
+++ b/virtcontainers/pod.go
@@ -478,6 +478,32 @@ func (p *Pod) ID() string {
 	return p.id
 }
 
+// RunStoragePath returns the pod runtime storage path corresponding to this
+// specific pod.
+func (p *Pod) RunStoragePath() string {
+	return filepath.Join(runStoragePath, p.id)
+}
+
+// Config returns the pod configuration.
+func (p *Pod) Config() *PodConfig {
+	return p.config
+}
+
+// FetchHypervisorState fetches the hypervisor state related to this pod.
+func (p *Pod) FetchHypervisorState() (HypervisorState, error) {
+	state := HypervisorState{}
+	if err := p.storage.fetchHypervisorState(p.id, &state); err != nil {
+		return HypervisorState{}, err
+	}
+
+	return state, nil
+}
+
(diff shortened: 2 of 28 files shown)


sboeuf:
@sameo PTAL

virtcontainers/hypervisor_plugin.go
```diff
@@ -0,0 +1,209 @@
...
+			pluginPath, err)
+	}
+
+	initFunc, err := h.handler.Lookup(initFuncName)
+	if err != nil {
+		return fmt.Errorf("Failed to lookup function %q: %v",
+			initFuncName, err)
+	}
+
+	h.hypervisor, err = initFunc.(func(*Pod) (interface{}, error))(pod)
```
jodh-intel: Using `interface{}` is going to have a significant performance impact over using a concrete type so I think we should:

- Create a new package defining the hypervisor interface.
  ```
  package hypervisor

  type Hypervisor interface {
      CreatePod(config PodConfig) error
        : 
  }
  ```
- Import the package into the plugin and this file.

- Have each hypervisor plugin define something like the following:
  ```
  func GetHypervisorPlugin() (*hypervisor.Hypervisor, error)
  ```
- Call `h, err := plugin.Lookup("GetHypervisorPlugin")`

- Check that what you got back is actually a `hypervisor.Hypervisor` with a type assertion.

- Make calls on the returned interface:
   ```
   err := h.CreatePod(config)
   ```

That approach:

- should be a lot faster.
- will simplify the code:
  - as you only need to do the lookup once.
  - you can just call the interface methods without all the type assertions.

sboeuf: I will have to try this out on a simple example first because I don't think this is going to work.
My understanding of the way this Go plugin was working was that we needed to lookup for every function or variable that we want to use from this plugin.
Basically, I think that even if you introduce this external hypervisor package, you will have to lookup for every function before you can call into them.

I'll try that and let you know.

jodh-intel: It works - I have example code if you're interested.

sboeuf: Just tested it on my side and it works :)
This is great and I agree your solution looks way better than mine. Let me rework this PR to apply this solution !

sboeuf: @jodh-intel there is one big no-go with your solution for plugin... We're gonna end up having a cycle dependency. `hypervisor_plugin.go` in virtcontainers will import the new `hypervisor` external package, but this same package will need to import virtcontainers in order to know about types like `vc.Pod`, `vc.PodConfig`, ...

jodh-intel: Yes, that is distressing (aka a pita). The only way round that I can think of is to sub-divide virtcontainers into smaller packages as suggested by @sameo on https://github.com/containers/virtcontainers/issues/631. That could be rather time-consuming to do at this point as it would require rework of the code and the tests.

sboeuf: Yes, this is the only option (breaking virtcontainers into smaller packages) to be able to implement your solution (that I like btw). This will involve a lot of work about the code and tests, I agree, and this will likely end up with other cyclic dependency issues...

jodh-intel: I guess another approach would be to convert all virtcontainer types into seralised strings so that we could have:

```
package hypervisor

type PodConfig string

type Hypervisor interface {
    CreatePod(config PodConfig) error
      : 
}
```

That would result in a lot of serialising/deserialising though which won't help performance.

bergwolf: I'd prefer splitting since we already want to split the api package.

virtcontainers/pod.go
```diff
@@ -378,9 +378,9 @@ func (podConfig *PodConfig) valid() bool {
 		return false
 	}
 
-	if _, err := newHypervisor(podConfig.HypervisorType); err != nil {
-		podConfig.HypervisorType = QemuHypervisor
-	}
+	// HACK: This forces the plugin to be used.
+	podConfig.HypervisorType = PluginHypervisor
+	podConfig.HypervisorConfig.PluginPath = "$GOPATH/src/github.com/kata-containers/runtime/plugins/hypervisor/qemu/qemu.so"
```
bergwolf: Can we have a concept of in-tree plugins that still works in a plugin option manner but we do not load them by dynamic libraries?

I thought we wanted that in the meeting and confirmed with @sameo in his slide (https://docs.google.com/presentation/d/1dbGrD1h9cpuqAPooiEgtiwWDGCYhVPdatq7owsKHDEQ/edit#slide=id.g35183e1ff3_0_25)

jodh-intel: Agreed. This is starting to sound like the kernel ;)

I've just had a play around a little with this concept and have some very crude code to handle both static and dynamic registration using something like this:

```
type PluginHandler interface {
    Foo(rope string) error
    Bar(mad bool) error
    Baz(snowballs float64) error
}


// PluginDetails allows a package to register itself as either a:                              
//                                                                                             
// - static plugin (sets Handler)                                                              
// - dynamic plugin (Sets Dynamic+Path)                                                        
type PluginDetails struct {                                                                    
        // Name of the plugin                                                                  
        Name string                                                                            
                                                                                               
        // set if plugin is dynamically loadable                                               
        Dynamic bool                                                                           
                                                                                               
        // Full path to shared object containing dynamically-loadable module                   
        Path string                                                                            
                                                                                               
        // set if Dynamic is false. Used for handling statically registered                    
        // modules.                                                                            
        Handler PluginHandler                                                                
}

// called from `main` for all modules to use (static and dynamic)
func RegisterPlugin(details PluginDetails) (PluginHandler, error)
```

There is one thing worth noting: the plugin package *only* loads code into the `main` package (although the dynamic loading can itself be performed in another package). This is mentioned (all too briefly) on https://golang.org/pkg/plugin/. fwics, this means we'll need something like:

  ```
  plugins/stub_foo.go

    package main
    import "pkg/foo"

  plugins/stub_bar.go

      package main
      import "pkg/bar"

  pkg/foo/foo.go

    package foo

  pkg/bar/bar.go

    package bar

  ```
  ... and the `main` code will then call (probably through another package):
  ```
  plugin.Open("stub_foo.so")
  plugin.Open("stub_bar.so")
  ```

So, code in the `main` package can only load new code into the `main` package. Hence, to dynamically load a new package, you need to create a stub that is part of `main` and have *that* `import` the real package you want to load.

sboeuf:
@bergwolf I am okay to try splitting virtcontainers into smaller packages and try solving this plugin issue as mentioned by @jodh-intel, but I am not clear about the way you plan to support internal implementation as plugins too. I mean I don't see the need for that since it won't change anything to have the `hypervisor` interface being described as an external interface instead of an internal one.

egernst:
@sboeuf the concern is regarding it being a dynamic library versus built in.  We can register and use the interface, but want it built in.  Does this help clarify?

sboeuf:
Well I might be confused but it really does not matter the way we handle this either through a `RegisterPlugin()` API or directly through each component implementation. Eventually, we'll need to have dynamic libs (`google_private_hypervisor.so` for instance) being able to implement the same `Hypervisor` interface than internal implementation such as `qemu.go` or `mock_hypervisor.go` that we have now.

sboeuf:
Ok sorry for the whole confusion here guys (@egernst @jodh-intel @bergwolf ). This PR is not meant to be merged as it is. It was only to open the discussion about the way this should be done, but I definitely don't want to externalize `qemu.go`.
I only expect proprietary implementation to follow the plugin model, otherwise, I expect any new implementation to go into virtcontainers the same way it is done right now with qemu.go for instance.

virtcontainers/hypervisor_plugin.go
sboeuf: @bergwolf I think splitting is really not worth it, see my comment here: https://github.com/kata-containers/runtime/issues/66

@jodh-intel I have just realized that we actually have a way to implement what you have described without having to split anything here. We only need to export the interface `hypervisor` as `Hypervisor`. I think that as long as we are okay with having a plugin for our hypervisors, this means that we should not mind having this interface exported. In this case, this is a good reason to export one of our interface.
Following this solution, we would end up writing a dynamic library relying on virtcontainers only (no need for an intermediate external package). And we would still need an `Hypervisor` implementation `hypervisor_plugin.so` to do the glue with the actual implementation coming from the dynamic lib.
Does this make sense ??

sboeuf:
@bergwolf @jodh-intel updated in a way that allows for your solution @jodh-intel

virtcontainers/hypervisor_plugin.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return fmt.Errorf("%s() failed: %v",
+			getHypervisorPluginFuncName, err)
+	}
+
+	return h.HypervisorPlugin.Init(pod)
+}
+
+func (h *pluginHypervisor) createPod(podConfig PodConfig) error {
+	return h.HypervisorPlugin.CreatePod(podConfig)
```
jodh-intel: Nice.

sboeuf:
Closing as this was a POC, and when we'll need to implement it, we can simply follow the template of what has been done here.

PR #62, opened 2018-03-15
utils: Use go env to discover the GOPATH

markdryan:
Since Go 1.8 users no longer need to have GOPATH set in their environment
for the Go tool chain to work.  If GOPATH is set, Go will use it.  Otherwise
it defaults to ~/go on linux.  As most users store their code in ~/go, they
don't bother setting GOPATH any more.  virtcontainers-setup.sh, in its
current form, fails for those users, as it requires GOPATH to be set.
This commit fixes the issue by calling go env "GOPATH" to determine the
correct location of the user's go code.  go env "GOPATH" will always
return the correct location, whether GOPATH is set, or not.

Signed-off-by: Mark Ryan <mark.d.ryan@intel.com>

### DIFF
--- a/virtcontainers/utils/virtcontainers-setup.sh
+++ b/virtcontainers/utils/virtcontainers-setup.sh
@@ -19,10 +19,12 @@ set -e
 
 SCRIPT_PATH=$(dirname $(readlink -f $0))
 
-if [[ -z "$GOPATH" ]]; then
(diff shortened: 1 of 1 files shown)


grahamwhaley:
Hi @markdryan - you'll need a `Fixes` line as detailed in [here](https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#patch-format) to pass the CI.

jodh-intel:
lgtm

sboeuf:
@markdryan Please rebase this PR on master, the `jenkins-ci-ubuntu-16-04` should be passing now.

PR #73, opened 2018-03-19
vendor: Vendor github.com/safchain/ethtool

### LINKED ISSUE #71: Determine if ethtool package can be dropped in favour of an upstream vendor
The `ethtool` package:

- https://github.com/kata-containers/runtime/blob/master/virtcontainers/pkg/ethtool/ethtool.go

... originally created under:

- https://github.com/containers/virtcontainers/pull/463

Comes from an upstream project:

- https://github.com/safchain/ethtool

We should determine if we can drop this package an instead vendor-in the official package above.

The upstream appears to have more functionality now than when we imported it.

amshinde:
We were using code copied from github.com/safchain/ethtool.
Vendor in upstream package instead to use additional
functionality added in.

Fixes #71


sboeuf:
LGTM

jodh-intel:
Thanks @amshinde ! ;)

PR #75, opened 2018-03-20
make: Add make target for virtcontainers in top-level Makefile

amshinde:
Running make from top source directory should build virtcontainers
library and the cli.
Add new `install-all` target that will install virtcontainers test
binaries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -198,12 +198,18 @@ define SHOW_ARCH
   $(shell printf "\\t%s%s\\\n" "$(1)" $(if $(filter $(ARCH),$(1))," (default)",""))
 endef
 
-all: runtime
+all: vc runtime
 
 runtime: $(TARGET_OUTPUT) $(CONFIG)
 .DEFAULT: default
 
 build: default
+
+vc:
+	make -C virtcontainers
+
+install-vc:
+	make -C virtcontainers install
 
 define GENERATED_CODE
 // WARNING: This file is auto-generated - DO NOT EDIT!
@@ -306,6 +312,7 @@ $(TARGET_OUTPUT): $(EXTRA_DEPS) $(SOURCES) $(GENERATED_GO_FILES) $(GENERATED_FIL
 	coverage \
 	default \
 	install \
+	install-vc \
 	show-header \
 	show-summary \
 	show-variables


Makefile
```diff
@@ -14,6 +14,15 @@ install :
 	cd cli; \
 	make install
 
+vc:
+	cd virtcontainers; \
```
jodh-intel: This is fine, but it would be cleaner if all these rules (including the existing one on a separate commit) use the magic `-C` flag instead:

```
vc:
    make -C virtcontainers
```

amshinde: @jodh-intel Done.

jodh-intel:
lgtm

sboeuf:
@amshinde could you rebase this PR ?

Makefile
```diff
@@ -14,6 +14,12 @@ install :
 	cd cli; \
 	make install
 
+vc:
+	make -C virtcontainers
+
+install-all:
+	make -C cli install
+	make -C virtcontainers install
```
bergwolf: The artifacts in virtcontainers are for testing purpose. I don't think we want to package and ship them globally. Maybe we can add `install-vc` that installs the binaries from virtcontainers directory and the test script can make use of it when necessary?

amshinde: @bergwolf Missed this comment. I have added an `install-vc` target for the test binaries.

Makefile
```diff
@@ -198,13 +198,20 @@ define SHOW_ARCH
...
 
 runtime: $(TARGET_OUTPUT) $(CONFIG)
 .DEFAULT: default
 
 build: default
 
+vc:
+	make -C virtcontainers
+
+install-vc:
```
bergwolf: add `install-vc` to the `.PHONY` list?

amshinde: Done

Makefile
```diff
@@ -198,13 +198,19 @@ define SHOW_ARCH
   $(shell printf "\\t%s%s\\\n" "$(1)" $(if $(filter $(ARCH),$(1))," (default)",""))
 endef
 
-all: runtime
+all: vc runtime
```
jodh-intel: A couple of points:

- `vc` and `install-vc` need to be added to `show-usage`.

- `make clean` doesn't undo `make vc`.

Issue #95, opened 2018-03-22
ci: tests repo path is hard wired and not over-rideable


grahamwhaley:
In [lib.sh](https://github.com/kata-containers/runtime/blob/master/.ci/lib.sh#L6), the path (url) to the tests repo is hard wired, and not over-rideable:
```
export tests_repo="github.com/kata-containers/tests"
```

This makes it very hard to develop and test new functionality in a tests repo clone/branch, as the call out from the tests repo to the repo under test (in this case the runtime), then switches back to using files from the main tests repo (and not the ones in development)...
Proposal is to change to:
```
export tests_repo="${tests_repo:-github.com/kata-containers/tests}"
```

Which will then pick up any overwride setting of `tests_repo` that has come from higher up the chain (we already have that functionality in the [tests repo](https://github.com/kata-containers/tests/blob/master/.ci/jenkins_job_build.sh#L16) itself for instance, but it is not trickling down to the other repos).

If folks agree, then we should do this in all repos (and/or come up with another way to share a lib.sh across all repos, which may not be ideal either).

jodh-intel:
I had a crazy idea that we could add a submodule to each repo for CI scripts. But clearly I need to go and have a cuppa and a biccy to clear my head! :smile:

PR #96, opened 2018-03-22
ci: allow path to tests repo to be over-wridden

grahamwhaley:
The github path to the tests repo is hard wired in lib.sh.
This makes it difficult to test any test repo WIP, as it ends
up calling back to the main repo.
Fix by using any value already set, and only setting to the main
repo if unset.

Fixes: #95

### DIFF
--- a/.ci/lib.sh
+++ b/.ci/lib.sh
@@ -3,7 +3,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-export tests_repo="github.com/kata-containers/tests"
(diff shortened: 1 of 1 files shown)


grahamwhaley:
marked as DNM - really more an RFC.
If we agree, the easiest fix is to do the same in all repos lib.sh's - there may not be a trivial or suitable way to share a lib.sh across all repos, or @jodh-intel would have done that in the first place :-)

sboeuf:
@grahamwhaley always down for that :)

Issue #108, opened 2018-03-23
Makefile: remove git hooks


bergwolf:
They only exist to remove the generated file `cli/generated.go` but the removal is unnecessary because:
1. it is ignored in .gitignore
2. it get rebuilt every time when we run `make`

PR #109, opened 2018-03-23
Makefile: remove git hooks

bergwolf:
They only exist to remove the generated file `cli/generated.go` but the
removal is unnecessary because:
1. it is ignored in .gitignore
2. it get rebuilt every time when we run `make`

Fixes: #108

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -194,22 +194,9 @@ define SHOW_ARCH
   $(shell printf "\\t%s%s\\\n" "$(1)" $(if $(filter $(ARCH),$(1))," (default)",""))
 endef
 
-# Only install git hooks if working in a git clone
-ifneq (,$(call DIR_EXISTS,.git))
-	HANDLE_GIT_HOOKS = install-git-hooks
-endif
-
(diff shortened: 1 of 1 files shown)


jodh-intel:
lgtm

Hi @bergwolf - looks like you've "hit" the [Travis bug](https://github.com/travis-ci/travis-ci/issues/7830). Can you try force re-basing (change `pick` to `reword`) and then force re-pushing to see if this fixes it?

grahamwhaley:
iirc, we added the hooks so if you swap branch in your git repo then it will rebuild the file - is that case now not relevant?

bergwolf:
@grahamwhaley It does not rebuild the file -- it only removes the file. Also, it will be quite confusing (to me) if the `git checkout/merge` etc. have a side effect of changing a source file. I would really appreciate that we just let the git command do what it is asked to. If these hooks are thought useful, we can add a script to create them and people can choose to install them or not on their own (instead of forcing every one to use them with `make` as it is now.)

PR #114, opened 2018-03-23
versions: Add versions database

### LINKED ISSUE #11: Handling of non-golang dependency versions
In Clear Containers, we have come to rely quite heavily on a file we maintain called `versions.txt`:

- https://github.com/clearcontainers/runtime/blob/master/versions.txt

That file is used to define the versions of various depenencies for Clear Containers. For example:

- supported version of golang.
- hypervisor version details.
- docker version.
- k8s version.
- rootfs image version from Clear Linux.

This file avoids hard-coding versions details in lots of different files (which makes maintenance very difficult). It is a simple database file in `name=value` format (crucially *with comments*) that can be sourced as a shell script or `grep`'d.

It's important to state that this file is not only used by the Clear Containers `runtime` repository: it is also used by the [tests](https://github.com/clearcontainers/tests) and [packaging](https://github.com/clearcontainers/packaging) repositories to query version details.

We are going to need an equivalent for Kata Containers, but I'm wondering if we want to consider refining the approach slightly. The current approach is simple and proven, but it doesn't have much structure. A way of grouping variables would be very useful, for example for the CentOS package versions here:

- https://github.com/clearcontainers/runtime/blob/master/versions.txt#L42..L49

Hence, I think we should consider creating a [versions.yaml](https://github.com/kata-containers/runtime/blob/master/versions.yaml) for Kata. We could then use a tool like https://github.com/mikefarah/yq to query the file:

```
$ go get github.com/mikefarah/yq
$ yq read versions.yaml centos.deps.glib.version
2.46.2
$ yq read versions.yaml docker.swarm.version
1.12.1
```

The advantage of YAML over JSON being a lot more flexibility in terms of structure, and -- a "killer" feature -- it supports comments! :smile:

What do you think @sameo, @grahamwhaley, @chavafg, @gnawux, @bergwolf ?

jodh-intel:
Add a YAML format database that is the equivalent of the Clear
Containers `versions.txt` file [1].

The file defines the versions of important non-golang dependencies used
by this and other Kata repositories particularly for testing and packaging.

Defining all version details centrally in this file avoids duplication
and "bitrot" when versions need to be changed.

[1] - https://github.com/clearcontainers/runtime/blob/master/versions.txt

Fixes #11.

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -1,0 +1,145 @@
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+---
+description: |
+  This file contains version details that are used by various
+  repositories for setting up the correct environment to run
+  tests and package components.
+
+format: |
+  Each entry in this file MUST conform to the following format:
+
+  <group>:
+   description: "<brief-description>"
+   notes: "<notes>"
+
+   <project>:
+     description: "<brief-description>"
+     notes: "<notes>"
+     url: "<project-url>"
+     issue: "<bug-url>"
+     commit: "<commit>"
+     version: "<version>"
+     release: "<version>"
+     meta:
+       <key-1>: "<value-1>"
+       <key-n>: "<value-n>"
+
+  Notes:
+
+  - All sections (except "meta") MUST include a description where
+    applicable. This is expected to be a brief summary.
+
+  - A section MAY specify a "notes" section which may be multi-line.
+    It is expected to be expand on the information specified in
+    "description".
+
+  - All sections (except "meta") MUST include a URL where applicable.
+
+  - A section MAY specify a bug URL using the "issue" field.
+
+  - A section MAY define a "meta" section to store additional
+    information about a project or group.
+
+  - Each entry MUST specify ATLEAST one of "commit", "version" and "release".
+
+  - WARNING: Gotcha alert! Remember to double-quote all strings
+    (except multi-line strings)! This avoids the possibility of a
+    version containing a period being treated as a floating point
+    number (and truncated!)
+
+assets:
+  description: "Additional required system elements"
+
+  hypervisor:
+    description: "Component used to create virtual machines"
+
+    qemu:
+      description: "VMM that uses KVM"
+      url: "https://github.com/kata-containers/qemu"
+      version: "741f430a960b5b67745670e8270db91aeb083c5f-29"
+      release: "19360"
+
+  image:
+    description: |
+      Root filesystem disk image used to boot the guest virtual
+      machine.
+    url: "https://github.com/kata-containers/osbuilder"
+    release: "20640"
+    meta:
+      image-type: "clearlinux"
+
+  kernel:
+    description: "Linux kernel optimised for virtual machines"
+    url: "https://github.com/kata-containers/linux"
+    version: "v4.14.22-86.container"
+    release: "19790"
+
+components:
+  description: "Core system functionality"
+
+  agent:
+    description: |
+      Container management service running in the guest virtual machines
+      root context.
+    url: "https://github.com/kata-containers/agent"
+    commit: "6f6e9ecd8aded0783c31968b304a9d6589114363"
+
+externals:
+  description: "Third-party projects used by the system"
+
+  crio:
+    description: |
+      OCI-based Kubernetes Container Runtime Interface implementation
+    url: "https://github.com/kubernetes-incubator/cri-o"
+    version: "v1.9.10"
+
+  docker:
+    description: "Moby project container manager"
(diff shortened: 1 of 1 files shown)


jodh-intel:
This is a "proof-of-concept" for what a YAML version of the Clear Containers `versions.txt` database might look like.

Note that being YAML:

- it is structured
- we can add comments
- we can also add other details like URL's that are part of the database.
- built-in handling of structure changes.
  Since it's possible we'll need to change the structure of this file if we do decide to use YAML, it's worth pointing out that YAML has builtin support for references. What this means is that we can redefine the structure of the file, but *still* keep backwards compatability for "clients" accessing it using [`yq`](https://github.com/kislyuk/yq) for example!

# Example of retaining compatability

If we started off with this fragment for storing golang details...

```
golang:
  versions:
    minimum: "1.8.3"
    version: "1.10"
```

... but then decided we want to move the entire `golang` block (array) below `languages`...

```
# XXX: retained for backwards compat!
golang:
  versions:
    minimum: *golang-versions-minimum
    version: *golang-versions-version

languages:
  golang:
    minimum: &golang-versions-minimum "1.8.3"
    version: &golang-versions-version "1.10"
```

... you can query the minimum golang version using *either* of the following queries:

```
$ yq read versions.yaml languages.golang.minimum
1.8.3
$ yq read versions.yaml golang.versions.minimum
1.8.3
```



/cc @grahamwhaley, @sboeuf, @jcvenegas, @erick0z, @chavafg, @bergwolf.

versions.yaml
```diff
@@ -0,0 +1,82 @@
...
+        tests and package components.
+
+        Notes:
+
+        - All sections *MUST* include a URL where applicable.
+
+        - All sections *SHOULD* include a description where intent is
+          not clear.
+
+        - WARNING: Gotcha alert! Remember to double-quote all versions
```
sboeuf: Maybe you should double quote every version in this file ? No confusion this way

jodh-intel: Good call. It might be overkill, but I've gone for the nuclear option - all strings are quoted for maximum safety.

jodh-intel:
Branch updated. The contents of this file is basically verbatim what we had in CC. Although we might need to adjust (or even remove) some values subsequently, there is no harm in merging this.

jodh-intel:

jcvenegas:
@jodh-intel  this looks great, I wonder if we could generalize the format for all the dependencies and have a template in the comments. If we have an standar format. Then we could use it for our release notes.
```yaml
<group>:
   description:<>
   <project>:
       url:<URL>
       commit:<commit>
       version:<>
       description:"<>"
       meta:
         <key>: "<val>"
```

versions.yaml
```diff
@@ -0,0 +1,146 @@
...
+    version containing a period being treated as a floating point
+    number (and truncated!)
+
+assets:
+  description: "Additional required system elements"
+
+  hypervisor:
+    description: "Component used to create virtual machines"
+
+    qemu:
```
jodh-intel: @sboeuf, @devimc, @erick0z  - should this be named `qemu-lite` dyt? Or would it be better to add something like:

```
meta:
  variant: "qemu-lite"
```

jodh-intel: Related: https://github.com/kata-containers/tests/issues/163.

versions.yaml
```diff
@@ -0,0 +1,146 @@
...
+
+specs:
+  description: "Details of important specifications"
+
+  oci:
+    description: "Open Containers Initiative runtime specification"
+    url: "https://github.com/opencontainers/runtime-spec/releases"
+    version: "v1.0.0-rc5"
+
+# vim:ts=2:expandtab:tw=70
```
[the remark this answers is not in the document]
jodh-intel: ick - I'd forgotten that was there - removing!

jodh-intel:
Another thought... considering the `golang` versions and also docker versions (see #92), I wonder if we want to generalise this further and specify *two* versions for every entry. For example:

```
golang:
  min-version: "1.8.3"
  max-version: "1.10"

docker:
  min-version: "..."
  max-version: "..."
```

**Iff** we do that, we'll have to maintain both those versions though. And that would imply we'd need to test with *both* versions regularly. It also suggests we could abstract "version" into something like a `Revision` object:

```
# Representation of some form of software versions.
# Atleast one of the value must be specified.
!Revision {
  version: "..."
  release: "..."
  commit: "..."
}
```

We could then define entries like this:
```
docker:
  min-version: !Revision{version: "1.8.3"}

kernel:
  min-revision: !Revision {version: "v4.14.22-86.container",  release: "19790"}
```

This might be too much complexity though since the only two entries that currently specify multiple versions are docker (for swarm) and golang (and that meta entry could be dropped).

Issue #118, opened 2018-03-27
Move to qemu 2.11


chavafg:
change `versions.yaml` to have the 2.11 version
change `QEMU_CMD` to `qemu-system-x86_64`
update virtcontainers unit tests to now point to the correct binary name.

PR #119, opened 2018-03-27
qemu: Move to qemu 2.11

chavafg:
This commit modifies version.yaml to now point to the qemu
2.11 stable version.
It modifies the default QEMU_CMD to be qemu-system-x86_64
instead of qemu-lite-system-x86_64.
And modifies virtcontainers unit tests to now point to the
correct QEMU_CMD.

Fixes: #118.

Depends-on: github.com/kata-containers/tests#182

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -60,9 +60,8 @@ assets:
 
     qemu:
       description: "VMM that uses KVM"
-      url: "https://github.com/kata-containers/qemu"
-      version: "741f430a960b5b67745670e8270db91aeb083c5f-29"
-      release: "19360"
+      url: "https://github.com/qemu/qemu"
+      version: "stable-2.11"
 
   image:
     description: |
--- a/virtcontainers/qemu_arch_base_test.go
+++ b/virtcontainers/qemu_arch_base_test.go
@@ -28,7 +28,7 @@ import (
 
 const (
 	qemuArchBaseMachineType = "pc"
-	qemuArchBaseQemuPath    = "/usr/bin/qemu-lite-system-x86_64"
+	qemuArchBaseQemuPath    = "/usr/bin/qemu-system-x86_64"
 )
 
 var qemuArchBaseQemuPaths = map[string]string{
--- a/arch/amd64-options.mk
+++ b/arch/amd64-options.mk
@@ -18,10 +18,4 @@ MACHINETYPE := pc
 KERNELPARAMS :=
 MACHINEACCELERATORS :=
 
-# The CentOS/RHEL hypervisor binary is not called qemu-lite
-ifeq (,$(filter-out centos rhel,$(distro)))
-    QEMUCMD := qemu-system-x86_64
-else
-    QEMUCMD := qemu-lite-system-x86_64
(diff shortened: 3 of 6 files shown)


devimc:
lgtm

versions.yaml
```diff
@@ -61,8 +61,7 @@ assets:
     qemu:
       description: "VMM that uses KVM"
       url: "https://github.com/kata-containers/qemu"
-      version: "741f430a960b5b67745670e8270db91aeb083c5f-29"
-      release: "19360"
+      version: "stable-2.11"
```
sboeuf: Why are we still relying on `url: "https://github.com/kata-containers/qemu"` ? I thought we could point to the official qemu repo since we want a vanilla version here.

chavafg: Just because in `kata-containers/qemu` we have the `configure.sh` script which provides the flags that we use.
https://github.com/kata-containers/qemu/blob/stable-2.11/configure.sh

sboeuf: Ok

jodh-intel: As mentioned https://github.com/kata-containers/tests/pull/182#discussion_r177658454 though, we shouldn't be storing the config script in the qemu branch itself.

So I think we can switch the URL here.

virtcontainers/qemu_arch_base_test.go
```diff
@@ -28,7 +28,7 @@ import (
 
 const (
 	qemuArchBaseMachineType = "pc"
-	qemuArchBaseQemuPath    = "/usr/bin/qemu-lite-system-x86_64"
+	qemuArchBaseQemuPath    = "/usr/bin/qemu-system-x86_64"
```
jodh-intel: If we are dropping `qemu-lite` entirely, there are changes we'll need to make in a few other areas (search for `QemuPCLite`).

egernst: I think the default will be using the vanilla 2.11 QEMU, though we are expecting a pc-lite supporting qemu-lite to be available in the future as well.  So, I think its reasonable to keep here for now?  WDYT @jodh-intel

jodh-intel: Ok - fine with this bit then ;)

sboeuf:
CI should pass now that https://github.com/kata-containers/tests/pull/182 has been merged. @jodh-intel PTAL

chavafg:
CI passed correctly. ping @kata-containers/runtime 
We need to merge this one now that https://github.com/kata-containers/tests/pull/182 has been merged. If not, we will have failures in the CI jobs on other PRs.

sboeuf:
@bergwolf PTAL

sboeuf:
@egernst PTAL

bergwolf:
lgtm

grahamwhaley:
All looking good and green. I believe we have addressed @jodh-intel 's comments, but he will not be around for a few days to ack his requested changes - so pls don't let @jodh-intel 's review change request block this now.

I see enough ack's - so, merging this now...

PR #121, opened 2018-03-27
exec: Allow to exec a process on a ready container

### LINKED ISSUE #120: exec: A process can be run on a created container
If a container is not `running`, but `created`/`ready` instead, this means a container process exists and that we can actually `exec` another process inside this container. The container does not have to be in `running` state.

sboeuf:
If a container is not running, but created/ready instead, this means
a container process exists and that we can actually exec another
process inside this container. The container does not have to be
in running state.

Fixes #120

### DIFF
--- a/cli/exec.go
+++ b/cli/exec.go
@@ -210,9 +210,11 @@ func execute(context *cli.Context) error {
 
 	params.cID = status.ID
 
-	// container MUST be running
-	if status.State.State != vc.StateRunning {
-		return fmt.Errorf("Container %s is not running", params.cID)
+	// container MUST be ready or running.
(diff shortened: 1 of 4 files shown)


bergwolf:
lgtm

jodh-intel:
The change itself seems fine. But we need tests to validate it.

egernst:
For background on this issues -- this was discovered during your CRIO testing, right?   Which test?

sboeuf:
Yes discovered during CRI-O testing of `ctr hostname env` from `ctr.bats`.

sboeuf:
@jodh-intel PR updated with unit tests.

PR #131, opened 2018-03-29
virtcontainers: Do not rollback by deleting container or pod

### LINKED ISSUE #87: error not propagating back to crio and container left in unknown state
Opening issue to track original from: https://github.com/containers/virtcontainers/issues/623

sboeuf:
In case a consumer of virtcontainers tries to start/stop a container,
or stop a pod, and for some reasons, this fails, virtcontainers always
tries to delete everything related to the container or the pod before
it returns the error.

The caller of the runtime is the one responsible for cleaning things
up if something goes wrong, that's why this cleanup call is never
needed.

A real example of that is the case of cc-runtime and CRI-O, where this
cleanup prevented CRI-O from retrieving proper state of the container
after the failure, leading to the inability to stop and remove the
container and the VM afterwards.

Fixes #87

### DIFF
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -164,7 +164,6 @@ func StopPod(podID string) (VCPod, error) {
 	// Stop it.
 	err = p.stop()
 	if err != nil {
-		p.delete()
 		return nil, err
 	}
 
@@ -406,7 +405,6 @@ func StartContainer(podID, containerID string) (VCContainer, error) {
 	// Start it.
 	err = c.start()
 	if err != nil {
-		c.delete()
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


bergwolf:
lgtm

WeiZhang555:
Code looks good to me, but Jenkins failed. I can't find the jenkins retrigger button, can you force push to trigger the Jenkins again?

grahamwhaley:
@WeiZhang555 - the rebuild button only appears once you have logged in to Jenkins (so it knows you have the rights to do that). I've pressed it for you :-)

WeiZhang555:
@grahamwhaley Oh, thanks!

PR #136, opened 2018-03-30
virtcontainers: Properly rollback mounts and hotplugs when container creation failed

### LINKED ISSUE #135: virtcontainers: Properly rollback when container creation fails
We need to make sure the mounts and hotplugs performed during the container creation are completely removed in case of a failure.

sboeuf:
This pull request ensures that when a container creation fails, any mount or hotplug action performed during this creation phase will be undone to leave the system in the state it was before.

Fixes #135

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -416,63 +416,78 @@ func newContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
 	return c, nil
 }
 
+// rollbackFailingContainerCreation rolls back important steps that might have
+// been performed before the container creation failed.
+// - Unplug CPU and memory resources from the VM.
+// - Unplug devices from the VM.
+func rollbackFailingContainerCreation(c *Container, err error) {
+	if err != nil && c != nil {
+		if err2 := c.removeResources(); err2 != nil {
+			c.Logger().WithError(err2).Error("rollback failed removeResources()")
+		}
+		if err2 := c.detachDevices(); err2 != nil {
+			c.Logger().WithError(err2).Error("rollback failed detachDevices()")
+		}
+		if err2 := c.removeDrive(); err2 != nil {
+			c.Logger().WithError(err2).Error("rollback failed removeDrive()")
+		}
+	}
+}
+
 // createContainer creates and start a container inside a Pod. It has to be
 // called only when a new container, not known by the pod, has to be created.
-func createContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
+func createContainer(pod *Pod, contConfig ContainerConfig) (c *Container, err error) {
 	if pod == nil {
 		return nil, errNeedPod
 	}
 
-	c, err := newContainer(pod, contConfig)
+	c, err = newContainer(pod, contConfig)
 	if err != nil {
-		return nil, err
+		return
 	}
 
-	if err := c.createContainersDirs(); err != nil {
-		return nil, err
+	if err = c.createContainersDirs(); err != nil {
+		return
 	}
 
-	if !c.pod.config.HypervisorConfig.DisableBlockDeviceUse {
-		agentCaps := c.pod.agent.capabilities()
-		hypervisorCaps := c.pod.hypervisor.capabilities()
+	// In case the container creation fails, the following takes care
+	// of rolling back all the actions previously performed.
+	defer rollbackFailingContainerCreation(c, err)
 
-		if agentCaps.isBlockDeviceSupported() && hypervisorCaps.isBlockDeviceHotplugSupported() {
-			if err := c.hotplugDrive(); err != nil {
-				return nil, err
-			}
-		}
+	if err = c.hotplugDrive(); err != nil {
+		return
 	}
 
 	// Attach devices
-	if err := c.attachDevices(); err != nil {
-		return nil, err
+	if err = c.attachDevices(); err != nil {
+		return
 	}
 
-	if err := c.addResources(); err != nil {
-		return nil, err
+	if err = c.addResources(); err != nil {
+		return
 	}
 
 	// Deduce additional system mount info that should be handled by the agent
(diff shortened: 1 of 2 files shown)


sboeuf:
@egernst @bergwolf PTAL

virtcontainers/container.go
```diff
@@ -418,61 +418,64 @@ func newContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
...
-		if agentCaps.isBlockDeviceSupported() && hypervisorCaps.isBlockDeviceHotplugSupported() {
-			if err := c.hotplugDrive(); err != nil {
-				return nil, err
-			}
+	// In case the container creation fails, the following takes care
+	// of rolling back all the hotplug and mount operations.
+	defer func() {
+		if err != nil && c != nil {
+			c.removeResources()
+			c.detachDevices()
```
jshachm: If a device is shared with different containers in one pod, should we determine whether the device is out of use or just  hostunplug it.

sboeuf: I don't know if we're trying to support this case yet.
But more than that, we're trying to rollback a container creation failure, the device is not yet shared with other containers, therefore I think it's not really related to this PR.

jshachm: Consider that  trying to create container C  in the pod, and use the `block device` like  `vdc` hotplugged by container B(concttainer A is `pause` contaienr ).  If we just hot unplug `vdc`, can it be a case that container B will face some problems even if it's an creation failure of container C.

If we finally decide not to support this case, maybe it's not really related to this PR.

sboeuf: Ok so I have made some investigations and basically we don't completely support the case you're describing right now. I have opened an issue for that: https://github.com/kata-containers/runtime/issues/171

This is why this PR is not related to this issue since we don't even have the support for properly sharing the same device from inside the agent.

sboeuf: And the roll back would not be an issue since we would detect that no device needs to be attached, meaning the `Container` structure will have no device associated, leading to the function `c.detachDevices()` to be a noop in case of container C.

egernst: Thanks @sboeuf .  Since this is moving forward and fixing existing functionality, I'm okay with this.

sboeuf:
@bergwolf @laijs PTAL

virtcontainers/container.go
```diff
@@ -418,61 +418,64 @@ func newContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
...
-
-		if agentCaps.isBlockDeviceSupported() && hypervisorCaps.isBlockDeviceHotplugSupported() {
-			if err := c.hotplugDrive(); err != nil {
-				return nil, err
-			}
+	// In case the container creation fails, the following takes care
+	// of rolling back all the hotplug and mount operations.
+	defer func() {
+		if err != nil && c != nil {
+			c.removeResources()
```
jodh-intel: Similar to https://github.com/kata-containers/agent/pull/196, could you add some log calls if these "undo" calls fail?

sboeuf: Yes, working on it !

sboeuf:
PR updated !

virtcontainers/kata_agent.go
```diff
@@ -599,7 +599,25 @@ func (k *kataAgent) appendDevices(deviceList []*grpc.Device, devices []Device) [
...
+// - Unmount container rootfs.
+func (k *kataAgent) rollbackFailingContainerCreation(c *Container, err error) {
+	if err != nil {
+		if c != nil {
+			if err2 := c.unmountHostMounts(); err2 != nil {
+				k.Logger().WithError(err2).Error("rollback failed unmountHostMounts()")
+			}
+		}
+
+		if err2 := bindUnmountContainerRootfs(kataHostSharedDir, c.pod.id, c.id); err2 != nil {
```
grahamwhaley: This call uses `c` - should it be inside the `c != nil` check? I guess maybe that function is `nil` safe - but, just checking...
If so, you coded the two nil checks with && in the other file (just to note consistency)

virtcontainers/kata_agent.go
```diff
@@ -667,28 +689,26 @@ func (k *kataAgent) createContainer(pod *Pod, c *Container) (*Process, error) {
 		// (kataGuestSharedDir) is already mounted in the
 		// guest. We only need to mount the rootfs from
 		// the host and it will show up in the guest.
-		if err := bindMountContainerRootfs(kataHostSharedDir, pod.id, c.id, c.rootFs, false); err != nil {
-			bindUnmountAllRootfs(kataHostSharedDir, *pod)
-			return nil, err
+		if err = bindMountContainerRootfs(kataHostSharedDir, pod.id, c.id, c.rootFs, false); err != nil {
+			return
```
grahamwhaley: Slightly numpty question - this func has a mixture of `return` and `return nil, err` - that doesn't read well to me - is that something that should really be consistent and cleaned up?

sboeuf:
@grahamwhaley updated !

virtcontainers/container.go
```diff
@@ -416,63 +416,78 @@ func newContainer(pod *Pod, contConfig ContainerConfig) (*Container, error) {
...
 	}
 
-	if !c.pod.config.HypervisorConfig.DisableBlockDeviceUse {
-		agentCaps := c.pod.agent.capabilities()
-		hypervisorCaps := c.pod.hypervisor.capabilities()
+	// In case the container creation fails, the following takes care
+	// of rolling back all the actions previously performed.
+	defer rollbackFailingContainerCreation(c, err)
 
-		if agentCaps.isBlockDeviceSupported() && hypervisorCaps.isBlockDeviceHotplugSupported() {
```
amshinde: @sboeuf I know this has been merged, but I hadnt taken a look at this PR. Any reason why the checks for block device support were taken out here?

PR #138, opened 2018-03-31
volumes: Attach volumes that are block device files as block devices

### LINKED ISSUE #137: We should be passing volumes that are block device files by actually passing the device with virtio .
We can get away from using 9p in this scenario and get better performance with virtio-scsi/blk.

amshinde:
Check if a volume passed to the container with -v is a block device
file, and if so pass the block device by hotplugging it to the VM
instead of passing this as a 9pfs volume. This would give us
better performance.

Fixes #137

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 )
 
 // Process gathers data related to a container process.
@@ -308,6 +309,32 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
 		// but it does not make sense to pass this as a 9p mount from the host side.
 		// This needs to be handled purely in the guest, by allocating memory for this inside the VM.
 		if m.Destination == "/dev/shm" {
+			continue
+		}
+
+		var stat unix.Stat_t
+		if err := unix.Stat(m.Source, &stat); err != nil {
+			return nil, err
+		}
+
+		// Check if mount is a block device file. If it is, the block device will be attached to the host
+		// instead of passing this as a shared mount.
+		if c.checkBlockDeviceSupport() && stat.Mode&unix.S_IFBLK == unix.S_IFBLK {
+			b := &BlockDevice{
+				DeviceType: DeviceBlock,
+				DeviceInfo: DeviceInfo{
+					HostPath:      m.Source,
+					ContainerPath: m.Destination,
+					DevType:       "b",
+				},
+			}
+
+			// Attach this block device, all other devices passed in the config have been attached at this point
+			if err := b.attach(c.pod.hypervisor, c); err != nil {
+				return nil, err
+			}
+
+			c.mounts[idx].BlockDevice = b
 			continue
 		}
 
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -539,6 +539,35 @@ func (k *kataAgent) replaceOCIMountSource(spec *specs.Spec, guestMounts []Mount)
 	return nil
 }
 
+func (k *kataAgent) replaceOCIMountsForStorages(spec *specs.Spec, volumeStorages []*grpc.Storage) error {
+	ociMounts := spec.Mounts
+	var index int
+	var m specs.Mount
+
+	for i, v := range volumeStorages {
+		for index, m = range ociMounts {
+			if m.Destination != v.MountPoint {
+				continue
+			}
+
+			// Create a temporary location to mount the Storage. Mounting to the correct location
+			// will be handled by the OCI mount structure.
+			filename := fmt.Sprintf("%s-%s", uuid.Generate().String(), filepath.Base(m.Destination))
+			path := filepath.Join(kataGuestSharedDir, filename)
+
+			k.Logger().Debugf("Replacing OCI mount source (%s) with %s", m.Source, path)
+			ociMounts[index].Source = path
+			volumeStorages[i].MountPoint = path
+
+			break
+		}
+		if index == len(ociMounts) {
+			return fmt.Errorf("OCI mount not found for block volume %s", v.MountPoint)
+		}
+	}
+	return nil
+}
+
 func constraintGRPCSpec(grpcSpec *grpc.Spec) {
 	// Disable Hooks since they have been handled on the host and there is
 	// no reason to send them to the agent. It would make no sense to try
@@ -720,6 +749,20 @@ func (k *kataAgent) createContainer(pod *Pod, c *Container) (p *Process, err err
 		return nil, err
 	}
 
+	// Append container devices for block devices passed with --device.
+	ctrDevices = k.appendDevices(ctrDevices, c.devices)
+
+	// Handle all the volumes that are block device files.
+	// Note this call modifies the list of container devices to make sure
+	// all hotplugged devices are unplugged, so this needs be done
+	// after devices passed with --device are handled.
+	volumeStorages := k.handleBlockVolumes(c)
+	if err := k.replaceOCIMountsForStorages(ociSpec, volumeStorages); err != nil {
+		return nil, err
+	}
+
+	ctrStorages = append(ctrStorages, volumeStorages...)
+
 	grpcSpec, err := grpc.OCItoGRPC(ociSpec)
 	if err != nil {
 		return nil, err
@@ -731,9 +774,6 @@ func (k *kataAgent) createContainer(pod *Pod, c *Container) (p *Process, err err
 	// We need to constraint the spec to make sure we're not passing
 	// irrelevant information to the agent.
 	constraintGRPCSpec(grpcSpec)
-
-	// Append container devices for block devices passed with --device.
-	ctrDevices = k.appendDevices(ctrDevices, c.devices)
 
 	req := &grpc.CreateContainerRequest{
 		ContainerId: c.id,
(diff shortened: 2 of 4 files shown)


amshinde:
@sboeuf Can you take a look at this?

egernst:
@bergwolf @laijs  - PTAL?

amshinde:
@sboeuf I have reworked the PR a bit associating a block device to a virtcontainers mount. PTAL.

sboeuf:
@amshinde thx, I will !

virtcontainers/container.go
```diff
@@ -304,6 +305,32 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
...
+				DeviceType: DeviceBlock,
+				DeviceInfo: DeviceInfo{
+					HostPath:      m.Source,
+					ContainerPath: m.Destination,
+					DevType:       "b",
+				},
+			}
+
+			// Attach this block device, all other devices passed in the config have been attached at this point
+			if err := b.attach(c.pod.hypervisor, c); err != nil {
```
sboeuf: This is being attached but where are we detaching (unplugging) this device from the VM ? We need to make sure the device is properly detached when the container is stopped(destroyed).

virtcontainers/kata_agent.go
```diff
@@ -532,6 +532,30 @@ func (k *kataAgent) replaceOCIMountSource(spec *specs.Spec, guestMounts []Mount)
...
+			path := filepath.Join(kataGuestSharedDir, filename)
+
+			k.Logger().Debugf("Replacing OCI mount source (%s) with %s", m.Source, path)
+			ociMounts[index].Source = path
+			volumeStorages[i].MountPoint = path
+
+			break
+		}
+	}
+	return nil
```
sboeuf: Seems there is no need to have this function returning an error since you return only nil at the end.

virtcontainers/mount.go
```diff
@@ -281,6 +281,11 @@ type Mount struct {
 
 	// ReadOnly specifies if the mount should be read only or not
 	ReadOnly bool
+
+	// BlockDevice represents block device that is attached to the
+	// VM in case this mount is a block device file or a directory
+	// backed by a block device.
+	BlockDevice *BlockDevice
```
sboeuf: Just wondering if a volume device is always gonna be a block device ? Or can it be something else ?

WeiZhang555:
> Check if a volume passed to the container with -v is a block device
file

I like the design, so good to me. Will check the implementation later :+1:

jshachm:
@WeiZhang555 @sboeuf 
A silly question: 
Is rollback method needed  when add device failed into container ?

sboeuf:
@jshachm not a silly question! You're right! The device being added is going to be part of the container creation, meaning it will also need to be part of the rollback.

amshinde:
@jshachm @sboeuf I had missed unplugging the device for kata agent, but now I make sure I do so. The block device associated with a volume is unplugged with all other devices associated with a container.

virtcontainers/container.go
amshinde: @sboeuf  This has been taken care of now.

virtcontainers/kata_agent.go
amshinde: I am performing some error checking now to return an error if a mount is not found in the list.

sboeuf:
@amshinde could you rebase this PR please ? There are some conflicts on the first commit.

amshinde:
@sboeuf Rebase done.

sboeuf:
@amshinde thanks I'll do another review!

virtcontainers/mount.go
sboeuf: You didn't answer this question.

virtcontainers/kata_agent.go
```diff
@@ -728,6 +757,17 @@ func (k *kataAgent) createContainer(pod *Pod, c *Container) (p *Process, err err
 	// Append container devices for block devices passed with --device.
 	ctrDevices = k.appendDevices(ctrDevices, c.devices)
 
+	// Handle all the volumes that are block device files.
+	// Note this call modifies the list of container devices to make sure
+	// all hotplugged devices are unplugged, so this needs be done
+	// after devices passed with --device are handled.
+	volumeStorages := k.handleBlockVolumes(c)
+	if err := k.replaceOCIMountsForStorages(ociSpec, volumeStorages); err != nil {
```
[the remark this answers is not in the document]
amshinde: @sboeuf The first function loops over the the virtcontainers mounts and the second one over the OCI mounts. So we would need them in any case if I understand correctly.

virtcontainers/mount.go
amshinde: Volumes will mostly be a block based device, I suppose you can pass a character device here, but then we dont do any special handling, just pass it through 9p.

bergwolf:
Can you let the runtime cli make the decision to pass the mount source as shared 9pfs or as a block device, because when passing it as a block device, it is mounted both in the host and in the guest, which is not supported by any of the upstream Linux block-based file systems, and we potentially face data corruptions and kernel crashes by doing it. So I would expect there is a cli configure option to toggle it and I suggest we set the default as off with a warning if someone wants to enable it.

WeiZhang555:
@bergwolf 
A runtime CLI won't work because the volume is passed through OCI spec.

> because when passing it as a block device, it is mounted both in the host and in the guest, which is not supported by any of ...

And actually this is also not true, we can choose NOT mount the volume in host, and pass the block directly to the kata-runtime. Mounting volume in host isn't  a MUST-DO step for volume to work.

bergwolf:
@WeiZhang555 
> And actually this is also not true, we can choose NOT mount the volume in host, and pass the block directly to the kata-runtime. Mounting volume in host isn't a MUST-DO step for volume to work.

Then you have to call something like `docker -v <block-device>:/path`. I'm fine with it as long as we do not use it to handle something like `docker -v <volume-name-or-path>:/path` that would mount the block device on the host.

virtcontainers/container.go
bergwolf: Do you need to check for duplication? Or is it OK to attach/detach the same device multiple times?

WeiZhang555:
@bergwolf 

> Then you have to call something like docker -v <block-device>:/path. I'm fine with it as long as we do not use it to handle something like docker -v <volume-name-or-path>:/path that would mount the block device on the host.

Exactly, my understanding is `docker run -v <volume-name-or-path>:/path` isn't involved in this PR's scenario. Am I right? @amshinde

virtcontainers/container.go
```diff
@@ -432,6 +459,19 @@ func (c *Container) rollbackFailingContainerCreation() {
 	}
 }
 
+func (c *Container) checkBlockDeviceSupport() bool {
```
bergwolf: So we are going to have different semantics with different underlying hypervisor? Might just reject the usage if the hypervisor does not support passing block devices like this, at least we are sticking with a constant API semantics.

virtcontainers/mount.go
bergwolf: In runv, we also support NFS volumes and ceph rbd volumes. I guess we can add a structure similar to agent's storage that can describe them all. But it does not block this PR and can be left for future improvements.

virtcontainers/container.go
amshinde: Different hypervisors would indicate if they support block devices using the capabilities() interface, whereas the DisableBlockDeviceUse is meant to be a user config to use block devices/ 9p.

amshinde:
@bergwolf @WeiZhang555 Yes this PR only handles `docker -v <block-device>:/path` case.  The other scenario  `docker run -v <volume-name-or-path>:/path` is not involved in this PR.

amshinde:
@bergwolf I am using a cli config option to pass the volume as 9pfs or block-device listed here
https://github.com/kata-containers/runtime/blob/204e40297ca5c675e13e74152d6e1ce330e3def3/cli/config/configuration.toml.in#L66

This is what we used for the rootfs, I suppose it makes sense to introduce another option specific to volumes being passed as 9p/block device.

virtcontainers/container.go
bergwolf: My point is that we need to make sure the kata library API is constant across different hypervisors. E.g., when we get the same OCI spec, we should deliver the same container fs layout, no matter which hypervisor is chosen. So if `DisableBlockDeviceUse` is set to false, and the hypervisor does not support it, we should just fail the API instead of trying to work around it.

bergwolf:
@amshinde I think `disable_block_device_use` is enough for this. No need for another cli option.

bergwolf:
@amshinde we might also want to document the behavior somewhere in the documentation repo, because we are delivering different results when compared to runc for `docker run -v <block-dev>:/path`. kata-containers/documentation/pull/48 might be a candidate place to do it. cc @jodh-intel

virtcontainers/container.go
amshinde: The approach so far has been to fallback to 9p incase the hypervisor does not support it. The block device usage is an optimization after all. You dont agee? The rootfs is also

virtcontainers/container.go
amshinde: @bergwolf I had not seen any issues in attaching/detaching same device files multiple times to a VM. However I tried mounting the device files at two different locations and am seeing some sync issues. I think it would be necessary to keep track at a pod level if the device has already been passed and not pass it again. But I am thinking of opening a separate PR to handle that issue, as this is an existing issue for block devices passed with --device as well. What do you think @bergwolf ?

bergwolf: @amshinde I'm not sure that I followed you. AFAIK, when we mount the same block device at different mount points, they share the same superblock data structure in the kernel, which makes sure data is always consistent. What sync issues did you see with that kind of setup?

amshinde: @bergwolf Right now if a device file say "/dev/sdb" is attached twice, I did not see any issues in attaching it twice and detaching it. However, since the file is attached twice, it may appear twice inside the VM. (say as "/dev/vda" and "/dev/vdb"). Sync issues will arise if you mount those two.

sboeuf: @amshinde @bergwolf I have performed the same testing last week and I can confirm that if you pass a device/volume to a container A, and you modify the content of the mounted filesystem by adding a new file for instance, a new container B created after this will see the change. But unfortunately, after both container are running, any change is only reported back to the host but not sync'ed back to the guest.

sboeuf:
@amshinde this PR needs some rebasing.
@bergwolf what is missing for this PR to be merged ?

amshinde:
@bergwolf @sboeuf  I have opened a separate issue to address the issue of attaching a device just once when the device appears in the devices/volumes of several containers in a pod sandbox. I think this is a complex issue in itself and I plan to handle this in a separate PR.

bergwolf:
I'm fine with handling device/storage sharing in #202. @amshinde please fix CI. Other than that,
LGTM

amshinde:
@sboeuf Can you merge this ?

sboeuf:
Merging as only centos is failing (which is expected for now).

PR #173, opened 2018-04-03
runtime: add builtin proxy and shim capability

bergwolf:
Add a new proxy type `KataProxyBuiltinType` and a new shim type `KataBuiltinShimType`. When they are specified, new proxy or shim processes will not be created. OTOH, the proxy and shim capabilities are provided by the kata runtime library directly.

For proxy, a yamux grpc dialer is used to create connections to the kata agent. And a goroutine is created to watch guest's console.

For shim, [sandbox relay APIs](https://github.com/kata-containers/documentation/blob/master/design/kata-api-design.md#sandbox-relay-api) will be provided. They are left to future PRs though.

Unfortunately due to limitation of the `yamux` library (at most one connection between client and server), this cannot be used with the kata runtime cli. So no cli change is included.

Depends-on: kata-containers/agent/pull/201

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -60,7 +60,7 @@
 
 [[constraint]]
   name = "github.com/kata-containers/agent"
-  revision = "a93071539feee29bfa22b6184380d3fd7c156ef7"
+  revision = "c8199f60759a1d52932df37a1af710cdeada8c81"
 
 [[constraint]]
   name = "github.com/containerd/cri-containerd"
@@ -74,3 +74,7 @@
   non-go = true
   go-tests = true
   unused-packages = true
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/hashicorp/yamux"
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -73,8 +73,9 @@ func (s *kataVSOCK) String() string {
 // KataAgentState is the structure describing the data stored from this
 // agent implementation.
 type KataAgentState struct {
-	ProxyPid int
-	URL      string
+	ProxyPid     int
+	ProxyBuiltIn bool
+	URL          string
 }
 
 type kataAgent struct {
@@ -417,6 +418,7 @@ func (k *kataAgent) startPod(pod Pod) error {
 
 	proxyParams := proxyParams{
 		agentURL: agentURL,
+		logger:   k.Logger().WithField("pod-id", pod.id),
 	}
 
 	// Start the proxy here
@@ -427,12 +429,17 @@ func (k *kataAgent) startPod(pod Pod) error {
 
 	// Fill agent state with proxy information, and store them.
 	k.state.ProxyPid = pid
+	k.state.ProxyBuiltIn = isProxyBuiltIn(pod.config.ProxyType)
 	k.state.URL = uri
 	if err := pod.storage.storeAgentState(pod.id, k.state); err != nil {
 		return err
 	}
 
-	k.Logger().WithField("proxy-pid", pid).Info("proxy started")
+	k.Logger().WithFields(logrus.Fields{
+		"pod-id":    pod.id,
+		"proxy-pid": pid,
+		"proxy-url": uri,
+	}).Info("proxy started")
 
 	hostname := pod.config.Hostname
 	if len(hostname) > maxHostnameLen {
@@ -805,7 +812,7 @@ func (k *kataAgent) connect() error {
 		return nil
 	}
 
-	client, err := kataclient.NewAgentClient(k.state.URL)
+	client, err := kataclient.NewAgentClient(k.state.URL, k.state.ProxyBuiltIn)
 	if err != nil {
 		return err
 	}
--- a/virtcontainers/kata_builtin_proxy.go
+++ b/virtcontainers/kata_builtin_proxy.go
@@ -1,0 +1,103 @@
+//
+// Copyright (c) 2018 HyperHQ Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package virtcontainers
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"net"
+
+	"github.com/sirupsen/logrus"
+)
+
+// This is a kata builtin proxy implementation of the proxy interface. Kata proxy
+// functionality is implemented inside the virtcontainers library.
+type kataBuiltInProxy struct {
+	podID string
+	conn  net.Conn
+}
+
+// start is the proxy start implementation for kata builtin proxy.
+// It starts the console watcher for the guest.
+// It returns agentURL to let agent connect directly.
+func (p *kataBuiltInProxy) start(pod Pod, params proxyParams) (int, string, error) {
+	if p.conn != nil {
+		return -1, "", fmt.Errorf("kata builtin proxy running for pod %s", p.podID)
+	}
+
+	p.podID = pod.id
+	console := pod.hypervisor.getPodConsole(pod.id)
+	err := p.watchConsole(consoleProtoUnix, console, params.logger)
+	if err != nil {
+		return -1, "", err
+	}
+
+	return -1, params.agentURL, nil
+}
+
+// stop is the proxy stop implementation for kata builtin proxy.
+func (p *kataBuiltInProxy) stop(pod Pod, pid int) error {
+	if p.conn != nil {
+		p.conn.Close()
+		p.conn = nil
+		p.podID = ""
+	}
+	return nil
+}
+
+func (p *kataBuiltInProxy) watchConsole(proto, console string, logger *logrus.Entry) (err error) {
+	var (
+		scanner *bufio.Scanner
+		conn    net.Conn
+	)
+
+	switch proto {
+	case consoleProtoUnix:
+		conn, err = net.Dial("unix", console)
+		if err != nil {
+			return err
+		}
+	// TODO: add pty console support for kvmtools
+	case consoleProtoPty:
+		fallthrough
+	default:
+		return fmt.Errorf("unknown console proto %s", proto)
+	}
+
+	p.conn = conn
+
+	go func() {
+		scanner = bufio.NewScanner(conn)
+		for scanner.Scan() {
+			fmt.Printf("[POD-%s] vmconsole: %s\n", p.podID, scanner.Text())
+		}
+
+		if err := scanner.Err(); err != nil {
+			if err == io.EOF {
+				logger.Info("console watcher quits")
+			} else {
+				logger.WithError(err).WithFields(logrus.Fields{
+					"console-protocol": proto,
+					"console-socket":   console,
+				}).Error("Failed to read agent logs")
+			}
+		}
+	}()
+
+	return nil
+}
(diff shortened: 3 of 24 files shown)


virtcontainers/pod.go
```diff
@@ -642,23 +642,36 @@ func newPod(podConfig PodConfig) (*Pod, error) {
 		wg:              &sync.WaitGroup{},
 	}
 
-	if err := p.storage.createAllResources(*p); err != nil {
+	if err = addPod(p); err != nil {
```
sboeuf: You should define the `newPod()` prototype so that `err` is already declared:
```go
func newPod(podConfig PodConfig) (p *Pod, err error) {
```

virtcontainers/pod.go
```diff
@@ -642,23 +642,36 @@ func newPod(podConfig PodConfig) (*Pod, error) {
...
+		}
+	}()
+
+	if err = p.storage.createAllResources(*p); err != nil {
+		return nil, err
+	}
+
+	defer func() {
+		if err != nil {
+			p.storage.deletePodResources(p.id, nil)
```
sboeuf: This is cleaner ;)
But could you also log if the defer has an error.

virtcontainers/pod.go
```diff
@@ -766,6 +784,8 @@ func (p *Pod) delete() error {
 		}
 	}
 
+	removePod(p.id)
```
sboeuf: Please check the error here.

virtcontainers/podlist.go
```diff
@@ -0,0 +1,91 @@
...
+//
+
+package virtcontainers
+
+import (
+	"fmt"
+	"sync"
+)
+
+type podlist struct {
```
sboeuf: s/`podlist`/`podList`/

virtcontainers/podlist.go
```diff
@@ -0,0 +1,91 @@
...
+	lock sync.RWMutex
+	pods map[string]*Pod
+}
+
+// globalPodlist tracks pods globally
+var globalPodlist = podlist{pods: make(map[string]*Pod)}
```
sboeuf: s/`globalPodList`/`globalPodList`/

virtcontainers/podlist.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (pl *podlist) addPod(pod *Pod) (err error) {
```
sboeuf:
```go
func (p *podList) addPod(pod *Pod) (err error) {
```

virtcontainers/podlist.go
```diff
@@ -0,0 +1,91 @@
...
+	defer pl.lock.Unlock()
+	if pl.pods[pod.id] == nil {
+		pl.pods[pod.id] = pod
+	} else {
+		err = fmt.Errorf("pod %s exists", pod.id)
+	}
+	return err
+}
+
+func (pl *podlist) removePod(id string) error {
```
sboeuf:
```go
func (p *podList) removePod(podID string) error {
```

virtcontainers/podlist.go
```diff
@@ -0,0 +1,91 @@
...
+	pl.lock.Lock()
+	defer pl.lock.Unlock()
+	delete(pl.pods, id)
+	return nil
+}
+
+func (pl *podlist) lookupPod(id string) (pod *Pod, err error) {
```
sboeuf:
```go
func (p *podList) lookupPod(podID string) (pod *Pod, err error) {
```

virtcontainers/podlist.go
```diff
@@ -0,0 +1,91 @@
...
+	defer pl.lock.RUnlock()
+	if pl.pods[id] != nil {
+		pod = pl.pods[id]
+	} else {
+		err = fmt.Errorf("pod %s does not exist", id)
+	}
+	return pod, err
+}
+
+func (pl *podlist) foreachPod(f func(*Pod) error) error {
```
sboeuf: Not used from your code here. Might be worth adding it only when it's gonna be called from some real code.

virtcontainers/podlist.go
sboeuf: Also please assign a pointer here.
```go
var globalPodlist = &podList{pods: make(map[string]*Pod)}
```

virtcontainers/podlist.go
```diff
@@ -0,0 +1,91 @@
...
+	return globalPodlist.removePod(id)
+}
+
+func lookupPod(id string) (*Pod, error) {
+	return globalPodlist.lookupPod(id)
+}
+
+func foreachPod(f func(*Pod) error) error {
+	return globalPodlist.foreachPod(f)
+}
```
sboeuf: I would remove those:
```go
func addPod(pod *Pod) error {
	return globalPodlist.addPod(pod)
}

func removePod(id string) error {
	return globalPodlist.removePod(id)
}

func lookupPod(id string) (*Pod, error) {
	return globalPodlist.lookupPod(id)
}

func foreachPod(f func(*Pod) error) error {
	return globalPodlist.foreachPod(f)
}
```

I think they add more confusion where we can call directly from `pod.go`:
```go
globalPodList.addPod(p)
...
```

bergwolf:
@sboeuf thanks for reviewing. changes applied and UT added. PTAL again.

bergwolf:
add `do not merge` label since it depends on kata-containers/agent#201

virtcontainers/kata_builtin_proxy.go
```diff
@@ -0,0 +1,100 @@
...
+		conn    net.Conn
+	)
+
+	switch proto {
+	case CONSOLE_PROTO_UNIX:
+		conn, err = net.Dial("unix", console)
+		if err != nil {
+			return err
+		}
+	// TODO: add pty console support for kvmtools
```
jodh-intel: Is there an open issue for this so we don't forget?

virtcontainers/kata_builtin_proxy.go
```diff
@@ -0,0 +1,100 @@
...
+	default:
+		return fmt.Errorf("unknown console proto %s", proto)
+	}
+
+	p.conn = conn
+
+	go func() {
+		scanner = bufio.NewScanner(conn)
+		for scanner.Scan() {
+			fmt.Printf("[SB-%s] vmconsole: %s\n", p.podID, scanner.Text())
```
jodh-intel: `SB` == "scanner buffer"? Could that prefix be `[pod %s]` for clarity.

virtcontainers/kata_builtin_proxy.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		if err := scanner.Err(); err != nil {
+			if err == io.EOF {
+				logger.Info("console wather quits")
+			} else {
+				logger.Errorf("Failed reading agent logs from socket: %v", err)
```
jodh-intel: How about:

```
logger.WithError(err).WithFields(logrus.Fields{
    "console-protocol" : proto,
    "console-socket" : console,
}).Error("Failed to read agent logs")
```

virtcontainers/kata_builtin_proxy.go
bergwolf: This should be part of the work that introduces kvmtools hypervisor support.

virtcontainers/kata_builtin_proxy.go
bergwolf: `SB` == `Sandbox`

virtcontainers/kata_builtin_proxy.go
bergwolf: Sure, I'll add it. Thanks!

virtcontainers/kata_builtin_proxy.go
jodh-intel: The problem is that almost all other components use `pod` / `pod-id` I think?

bergwolf: I thought we want to switch from `pod` to `sandbox` as agreed in the api design. But it can be done later.

bergwolf:
Dropping `do not merge` label as kata-containers/agent#201 is merged. @sboeuf @jodh-intel PTAL.

virtcontainers/shim.go
```diff
@@ -40,6 +40,9 @@ const (
 
 	// KataShimType is the Kata Containers shim type.
 	KataShimType ShimType = "kataShim"
+
+	// KataBuiltinShimType is the Kata Containers builtin shim type.
+	KataBuiltinShimType ShimType = "kataBuiltinShim"
```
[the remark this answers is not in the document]
bergwolf: The difference is that `no-op` shim is for testing and `kataBuiltinShim` means the shim functionality is still supported. E.g., the relying API would go directly to the agent for builtin shim but can be ignored for `no-op` shims.

virtcontainers/kata_builtin_proxy.go
sboeuf: Yes I think the sandbox naming should be done in a separate PR, meaning that it makes more sense to stick with `[POD-%s]` for now.

virtcontainers/kata_agent.go
```diff
@@ -818,7 +818,6 @@ func (k *kataAgent) sendReq(request interface{}) (interface{}, error) {
 	if err := k.connect(); err != nil {
 		return nil, err
 	}
-	defer k.disconnect()
```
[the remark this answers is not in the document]
sboeuf: Yes you're right but this should not be part of this PR. I think we should go first with what I proposed so that it properly covers the case of the builtin proxy, and later we can submit another PR covering the case of a long running runtime. This should be covered more neatly, by introducing a notion of short running vs long running IMO, and allowing the current use case to still use `disconnect()`. I don't like that we implicitly handle the disconnection because the process ends.

virtcontainers/podlist.go
```diff
@@ -0,0 +1,63 @@
...
+		err = fmt.Errorf("pod %s exists", pod.id)
+	}
+	return err
+}
+
+func (p *podList) removePod(id string) error {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	delete(p.pods, id)
+	return nil
```
sboeuf: No reason to return an error. Let's have this function returning nothing then.

Gopkg.toml
```diff
@@ -60,7 +60,7 @@
 
 [[constraint]]
   name = "github.com/kata-containers/agent"
-  revision = "a93071539feee29bfa22b6184380d3fd7c156ef7"
+  revision = "c8199f60759a1d52932df37a1af710cdeada8c81"
```
sboeuf: Please add the shortlog diff since the last vendoring of this dependency into your commit message. It helps tracking which patches are pulled through this vendoring.

virtcontainers/podlist.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func (p *podList) lookupPod(id string) (pod *Pod, err error) {
+	p.lock.RLock()
+	defer p.lock.RUnlock()
+	if p.pods[id] != nil {
```
devimc: I think this looks better

```go
	if p.pods[id] != nil {
		return p.pods[id], nil
	}
	return nil, fmt.Errorf("pod %s does not exist", id)
```

bergwolf:
@devimc @jodh-intel @sboeuf PR updated. PTAL.

bergwolf:
Jenkins failed on #193

Gopkg.toml
sboeuf: Hey, sorry to be picky here, but we don't really need the merge commits. If you run the following command:
```bash
git log --no-merges --abbrev-commit --pretty=oneline a93071539feee29bfa22b6184380d3fd7c156ef7..origin/master
```
You should only get the __real__ commits.

bergwolf: Taking notes and commit message updated. Thanks @sboeuf!

sboeuf:
Thx @bergwolf 
LGTM

bergwolf:
Jenkins failed on kata-containers/tests/issues/225

sboeuf:
@bergwolf have you rebased on latest master version ? Just want to make sure you have kata-containers/runtime#193 included.

bergwolf:
ping @kata-containers/runtime

PR #179, opened 2018-04-03
virtcontainers: Fix container creation rollback

### LINKED ISSUE #178: Fix rollback container creation
The container creation rollback is not working because the error passed through the defer does not take the latest error returned.

sboeuf:
The rollback does not work as expected because the error has to be
checked from the defer itself.

Fixes #178


sboeuf:
@bergwolf @laijs PTAL

bergwolf:
lgtm!

Issue #182, opened 2018-04-04
update cri-o version to test with latest 1.9 commit


chavafg:
Latest `release-1.9` branch contains a race condition fix [1] that will allow us to unskip a test from the CRI-O `ctr.bats` 
[1] https://github.com/kubernetes-incubator/cri-o/pull/1493

PR #183, opened 2018-04-04
versions: Update CRI-O version

chavafg:
This change uses latest commit from CRI-O
1.9 branch instead of using latest 1.9 release.
This will allow us to unskip a test from the
CRI-O ctr.bats test file, since it contains a race
condition fix.

Fixes: #182.


sboeuf:
@chavafg I've restarted the CI, let's see how it goes.

jodh-intel:
lgtm

PR #191, opened 2018-04-05
virtcontainers: Do not pass /dev/shm as 9p mount

### LINKED ISSUE #190: Do not pass /dev/shm as a 9p mount
/dev/shm is passed in the OCI spec as a bind mount. We need to make sure that this is handled purely in the guest and not passed as 9p mount from the host side.

amshinde:
All bind mounts are now passed to the guest with 9p.
We need to exclude /dev/shm, as this is passed as a bind mount
in the spec. We handle /dev/shm in the guest by allocating
memory for it on the guest side. Passing /dev/shm as a 9p mount
was causing it to be mounted twice.

Fixes #190

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -304,6 +304,13 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
 			continue
 		}
 
+		// We need to treat /dev/shm as a special case. This is passed as a bind mount in the spec,
+		// but it does not make sense to pass this as a 9p mount from the host side.
+		// This needs to be handled purely in the guest, by allocating memory for this inside the VM.
+		if m.Destination == "/dev/shm" {
+			continue
+		}
+
 		randBytes, err := generateRandomBytes(8)
 		if err != nil {
 			return nil, err


grahamwhaley:
Unit tests failed I think:
```
--- FAIL: TestContainerAddResources (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0xc87264]
```

Having a look at the docker docs, I think we can treat `/dev/shm` as special enough to special case.
I will note - we don't support the `--shm-size` option already, and I've just noticed there is also a `--ipc=non` option that enables a 'private ipc namespace' - which, err, is sort of what we have by default with VMs??

jodh-intel:
@amshinde, @sboeuf, @grahamwhaley - yep, that limitation is recorded in https://github.com/kata-containers/kata-containers/issues/21.

jodh-intel:
A quick look suggests that `--shm-size` is not passed in `config.json` fwiw.

grahamwhaley:
ouch, not passed in the `config.json` - I guess it might set the size on the host ... which will make it a little more awkward if we do ever support that setting (well, OK, we look up what is on the host and pass it through to the agent to set it I guess)

amshinde:
@grahamwhaley Yes, we need to inspect the mount on the host to find out the size and pass that to the guest. Had created an issue  for this in CC long back.

chavafg:
@amshinde I think this is related also to: https://github.com/kata-containers/tests/pull/203 which enabled unit tests on jenkins.

amshinde:
@chavafg Oh, so we were not running unit tests before that change.  I see a similar failure on your other PR : http://kata-jenkins-ci.westus2.cloudapp.azure.com/job/kata-containers-runtime-ubuntu-16-04-PR/163/console

@devimc Do you have any input on this? Do you think we should disable those tests till you fix the issue?

devimc:
@amshinde https://github.com/kata-containers/runtime/pull/193 fixes unit tests

virtcontainers/container.go
```diff
@@ -304,6 +304,13 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
 			continue
 		}
 
+		// We need to treat /dev/shm as a special case. This is passed as a bind mount in the spec,
+		// but it does not make sense to pass this as a 9p mount from the host side.
+		// This needs to be handled purely in the guest, by allocating memory for this inside the VM.
+		if m.Destination == "/dev/shm" {
```
bergwolf: Both `runc spec` and `docker-runc spec` creates an OCI spec listing `/dev/shm` as `tmpfs`. How did you get the bind mounted `/dev/shm` in OCI spec?

WeiZhang555: @bergwolf 
It looks like docker will create a bind mount for "/dev/shm", I just checked config file from docker, it adds this item:
```
{"destination":"/dev/shm","type":"bind","source":"/var/lib/docker/containers/fc243bca3002a1264072d07086b77970ab9e6a8da9c5143b0af6c0fbb1f2367d/shm","options":["rbind","rprivate"]}
```
So this change makes sense in this case.

WeiZhang555: Default mounts are here:

```
    "mounts": [
        {
            "destination": "/proc",
            "options": [
                "nosuid",
                "noexec",
                "nodev"
            ],
            "source": "proc",
            "type": "proc"
        },
        {
            "destination": "/dev",
            "options": [
                "nosuid",
                "strictatime",
                "mode=755"
            ],
            "source": "tmpfs",
            "type": "tmpfs"
        },
        {
            "destination": "/dev/pts",
            "options": [
                "nosuid",
                "noexec",
                "newinstance",
                "ptmxmode=0666",
                "mode=0620",
                "gid=5"
            ],
            "source": "devpts",
            "type": "devpts"
        },
        {
            "destination": "/sys",
            "options": [
                "nosuid",
                "noexec",
                "nodev",
                "ro"
            ],
            "source": "sysfs",
            "type": "sysfs"
        },
        {
            "destination": "/sys/fs/cgroup",
            "options": [
                "ro",
                "nosuid",
                "noexec",
                "nodev"
            ],
            "source": "cgroup",
            "type": "cgroup"
        },
        {
            "destination": "/dev/mqueue",
            "options": [
                "nosuid",
                "noexec",
                "nodev"
            ],
            "source": "mqueue",
            "type": "mqueue"
        },
        {
            "destination": "/etc/resolv.conf",
            "options": [
                "rbind",
                "rprivate"
            ],
            "source": "/var/lib/docker/containers/fc243bca3002a1264072d07086b77970ab9e6a8da9c5143b0af6c0fbb1f2367d/resolv.conf",
            "type": "bind"
        },
        {
            "destination": "/etc/hostname",
            "options": [
                "rbind",
                "rprivate"
            ],
            "source": "/var/lib/docker/containers/fc243bca3002a1264072d07086b77970ab9e6a8da9c5143b0af6c0fbb1f2367d/hostname",
            "type": "bind"
        },
        {
            "destination": "/etc/hosts",
            "options": [
                "rbind",
                "rprivate"
            ],
            "source": "/var/lib/docker/containers/fc243bca3002a1264072d07086b77970ab9e6a8da9c5143b0af6c0fbb1f2367d/hosts",
            "type": "bind"
        },
        {
            "destination": "/dev/shm",
            "options": [
                "rbind",
                "rprivate"
            ],
            "source": "/var/lib/docker/containers/fc243bca3002a1264072d07086b77970ab9e6a8da9c5143b0af6c0fbb1f2367d/shm",
            "type": "bind"
        }
    ],
```

In this bind mount case, ignoring `shm` could be enough.

bergwolf: Thanks @WeiZhang555 !

bergwolf:
lgtm!

sboeuf:
@amshinde please rebase on top of kata-containers/runtime#193 so that we can get the CI passing here.

amshinde:
@sboeuf @devimc Can we merge this?

Issue #148, opened 2018-04-02
namespaces: How to enter them in a foolproof way


egernst:
_From @sboeuf on February 27, 2018 20:3_

Following discussion on #613 and implementation #615, I will try to summarize what are the issues we're facing with Golang and what are the options here.
We would like to enter any namespaces from virtcontainers (or any code written in Go) and be able to execute some callbacks inside a set of defined namespaces. The reason of why we need to enter those namespaces does not need to be detailed here.

Now, let's say we want to execute some code in a set of namespaces, we can use the package `nsenter` implemented in the PR #615, which will basically save the current set of namespaces, then enter the targeted namespaces, execute the callback, and switch back to the saved(original) namespaces. But doing so, we have 2 drawbacks that could end up in unwanted behaviors:
- If the callback defines a go routine, the code running into this routine will be running in a different thread (TID will be different), meaning it won't run in the expected set of namespaces.
- If we're using a version < Go 1.10, we might hit a very unlikely issue (but still have a chance), which is inherent to Golang itself. If the code executed from the callback might take too much time, Golang might decide to create a new thread for executing this code, even if `runtime.LockOSThread()` has been previously called, leading our code to run in the wrong namespaces. This is obviously a Golang bug that has been [fixed in Go 1.10](https://golang.org/doc/go1.10#runtime).

Additionally to those potential issues, there is a main constraint that cannot be solved by Golang since this is multithreaded: the ability to enter `mount` or `user` namespaces. Following discussions [here](https://github.com/golang/go/issues/8676), [here](https://lists.linux-foundation.org/pipermail/containers/2013-January/031565.html) and [here](https://stackoverflow.com/questions/25704661/calling-setns-from-go-returns-einval-for-mnt-namespace) give some input.

So now, what should we do ? Well having C code is the solution to the multi-threading problem, and this will solve both the drawbacks and the limitations explained above. So why don't we get started ?
Well, it's not as easy as it seems, because we cannot only define a C function called from Go code, this is already too late, you're already multi-threaded at that time.

The solution would be to invoke a C constructor like this:
```c
void __attribute__((constructor)) init_func(int argc, const char **argv)
```
from our Go code. This constructor is always running before any piece of Go code, ensuring about the single-threading here. In order to reach this code from a Go function, we need to re-execute ourselves with a snippet like this:
```go
package main

/*
#cgo CFLAGS: -Wall
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/types.h>
[...]

PR #194, opened 2018-04-07
[WIP] virtcontainers: netns_watcher: Monitor network changes

### LINKED ISSUE #170: network hotplug: add monitoring process for nets
Related to https://github.com/kata-containers/runtime/issues/113

This will track the implementation of the monitoring process.  virtcontainer changes to be tracked in #113

sboeuf:
This commit introduces a new watcher dedicated to the monitoring
of a specific network namespace in order to detect any change that
could happen to the network.

As a result of such a detection, the watcher should call into the
appropriate runtime path with the proper arguments to modify the
pod network accordingly.

Fixes #170

### DIFF
--- a/virtcontainers/netmon/netmon.c
+++ b/virtcontainers/netmon/netmon.c
@@ -1,0 +1,1301 @@
+/*
+ * Copyright (C) 2018 Intel Corporation
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#define _GNU_SOURCE
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <string.h>
+#include <unistd.h>
+#include <errno.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <sys/wait.h>
+#include <fcntl.h>
+#include <sched.h>
+#include <asm/types.h>
+#include <sys/socket.h>
+#include <net/if.h>
+#include <netinet/in.h>
+#include <linux/netlink.h>
+#include <linux/rtnetlink.h>
+#include <arpa/inet.h>
+#include <ifaddrs.h>
+#include <netdb.h>
+#include <sys/ioctl.h>
+#include <syslog.h>
+#include <getopt.h>
+
+#define PROGRAM_NAME    "netns-watcher"
+#define PROGRAM_VERSION "0.0.1"
+
+#ifndef MAX_IFACES
+#define MAX_IFACES  50
+#endif
+
+#define INIT_IF_IDX -1
+
+bool debug = false;
+
+struct watcher_params {
+	char *sandbox_id;
+	char *runtime_path;
+	char *netns_path;
+};
+
+struct ip_addr {
+	unsigned char  family;
+	char           *addr;
+	struct ip_addr *next;
+};
+
+struct iface {
+	int            idx;
+	char           *hw_addr;
+	char           *name;
+	char           *mtu;
+	struct ip_addr *ip_addrs;
+};
+
+struct route {
+	char *src;
+	char *dst;
+	char *gw;
+	char *dev;
+};
+
+/* 
+ * Global interface list used to keep track of the interfaces and their
+ * parameters.
+ */
+struct iface iface_list[MAX_IFACES];
+
+/*
+ * Print full list of interfaces.
+ */
+void print_iface_list() {
+	int i;
+
+	int max_ifaces = (int)(sizeof(iface_list)/sizeof(iface_list[0]));
+	for (i = 0; i < max_ifaces; i++) {
+		if (iface_list[i].idx == INIT_IF_IDX) {
+			continue;
+		}
+
+		printf("IFACE %d\n", i);
+		printf("\tidx     = %d\n", iface_list[i].idx);
+		printf("\thw_addr = %s\n", iface_list[i].hw_addr);
+		printf("\tname    = %s\n", iface_list[i].name);
+		printf("\tmtu     = %s\n", iface_list[i].mtu);
+
+		struct ip_addr *ipa = iface_list[i].ip_addrs;
+		if (!ipa) {
+			printf("\n");
+			continue;
+		}
+
+		printf("\tip_addrs\n");
+		printf("\t    |---- ip_addr = %s, family = %d\n",
+				ipa->addr, ipa->family);
+		while (ipa->next) {
+			printf("\t    |---- ip_addr = %s, family = %d\n",
+					ipa->next->addr, ipa->next->family);
+			ipa = ipa->next;
+		} 
+
+		printf("\n");
+	}
+}
+
+/*
+ * Print version information.
+ */
+void print_version(void)
+{
+	printf("%s v%s\n", PROGRAM_NAME, PROGRAM_VERSION);
+}
+
+/*
+ * Free internal fields of the watcher_params structure.
+ */
+void free_watcher_params(struct watcher_params *params)
+{
+	if (!params) {
+		return;
+	}
+
+	free(params->netns_path);
+	params->netns_path = NULL;
+
+	free(params->sandbox_id);
+	params->sandbox_id = NULL;
+
+	free(params->runtime_path);
+	params->runtime_path = NULL;
+}
+
+/*
+ * Free internal fields of the ip_addr structure.
+ */
+void free_ip_addr(struct ip_addr *ipa) {
+	struct ip_addr *ipa_next, *ipa_curr;
+
+	if (!ipa) {
+		return;
+	}
+
+	free(ipa->addr);
+	ipa->addr = NULL;
+
+	for (ipa_curr = ipa->next; ipa_curr; ipa_curr = ipa_next) {
+		ipa_next = ipa_curr->next;
+
+		free(ipa_curr->addr);
+		ipa_curr->addr = NULL;
+
+		free(ipa_curr);
+		ipa_curr = NULL;
+	}
+}
+
+/*
+ * Free internal fields of the iface structure.
+ */
+void free_iface(struct iface *nif) {
+	if (!nif) {
+		return;
+	}
+
+	free(nif->hw_addr);
+	nif->hw_addr = NULL;
+
+	free(nif->name);
+	nif->name = NULL;
+
+	free(nif->mtu);
+	nif->mtu = NULL;
+
+	if (nif->ip_addrs) {
+		free_ip_addr(nif->ip_addrs);
+	}
+	free(nif->ip_addrs);
+	nif->ip_addrs = NULL;
+}
+
+/*
+ * Free internal fields of the route structure.
+ */
+void free_route(struct route *rt) {
+	if (!rt) {
+		return;
+	}
+
+	free(rt->src);
+	rt->src = NULL;
+
+	free(rt->dst);
+	rt->dst = NULL;
+
+	free(rt->gw);
+	rt->gw = NULL;
+
+	free(rt->dev);
+	rt->dev = NULL;
+}
+
+/*
+ * Modify the global list of interfaces by inserting a new ip_addr structure
+ * into the linked list of the specified interface.
+ */
+int insert_ip_addr(int idx, const char *addr, unsigned char family)
+{
+	struct ip_addr *ipa;
+	struct ip_addr *tmp;
+
+	/* Checking params */
+	if (idx < 0) {
+		fprintf(stderr, "Wrong index %d provided", idx);
+		return EINVAL;
+	}
+	if (!addr) {
+		fprintf(stderr, "IP address is empty");
+		return EINVAL;
+	}
+
+	ipa = (struct ip_addr *) calloc(1, sizeof(struct ip_addr));
+	if (!ipa) {
+		return ENOMEM;
+	}
+
+	ipa->addr = strdup(addr);
+	if (!ipa->addr) {
+		return ENOMEM;
+	}
+
+	ipa->family = family;
+	ipa->next = NULL;
+
+	if (!iface_list[idx].ip_addrs) {
+		iface_list[idx].ip_addrs = ipa;
+		return 0;
+	}
+
+	tmp = iface_list[idx].ip_addrs;
+	while (tmp->next) {
+		tmp = tmp->next;
+	}
+
+	tmp->next = ipa;
+
+	return 0;
+}
+
+/*
+ * Modify the global list of interfaces by removing an existing ip_addr
+ * structure from the linked list of the specified interface.
+ */
+int delete_ip_addr(int idx, const char *addr)
+{
+	struct ip_addr *tmp;
+	struct ip_addr *prev;
+	bool found = false;
+	int ret = -1;
+
+	/* Checking params */
+	if (idx < 0) {
+		fprintf(stderr, "Wrong index %d provided", idx);
+		return EINVAL;
+	}
+	if (!addr) {
+		fprintf(stderr, "IP address is empty");
+		return EINVAL;
+	}
+
+	if (!iface_list[idx].ip_addrs) {
+		return ENOENT;
+	}
+
+	tmp = iface_list[idx].ip_addrs;
+	if (tmp->addr && !strcmp(tmp->addr, addr)) {
+		iface_list[idx].ip_addrs = iface_list[idx].ip_addrs->next;
+		goto free_node;
+	}
+
+	prev = tmp;
+	while (tmp->next) {
+		prev = tmp;
+		tmp = tmp->next;
+
+		if (tmp->addr && !strcmp(tmp->addr, addr)) {
+			found = true;
+			break;
+		}
+	}
+
+	if (!found) {
+		fprintf(stderr, "No IP address matching %s for interface %s",
+			addr, iface_list[idx].name);
+		return ENOENT;
+	}
+
+	prev->next = tmp->next;
+
+free_node:
+	free(tmp->addr);
+	free(tmp);
+
+	return 0;
+}
+
+/*
+ * Print program usage.
+ */
+void print_usage(void)
+{
+	printf("\nUsage: %s [options]\n\n", PROGRAM_NAME);
+	printf(" -d, --debug        Enable debug output\n");
+	printf(" -h, --help         Display usage\n");
+	printf(" -n, --netns-path   Network namespace path (required)\n");
+	printf(" -p, --sandbox-id   Sandbox ID (required)\n");
+	printf(" -r, --runtime-path Runtime path (required)\n");
+	printf(" -v, --version      Show version\n");
+	printf("\n");
+}
+
+/* 
+ * Validate the parameters and fill the appropriate structure.
+ *
+ * sandbox-id: Sandbox identifier related to the network namespace this
+ * watcher has to listen to.
+ * runtime-path: Runtime path is needed to call into it whenever a
+ * change in the network namespace is detected.
+ * netns-path: Network namespace path that needs to be monitored.
+ *
+ * It returns 0 if the parameters are valid, with a watcher_params structure
+ * properly filled. Otherwise, it returns -1 with a null watcher_params
+ * structure.
+ */
+int parse_options(int argc, const char **argv, struct watcher_params *params)
+{
+	struct option prog_opts[] = {
+		{"debug", no_argument, 0, 'd'},
+		{"help", no_argument, 0, 'h'},
+		{"netns-path", required_argument, 0, 'n'},
+		{"sandbox-id", required_argument, 0, 'p'},
+		{"runtime-path", required_argument, 0, 'r'},
+		{"version", no_argument, 0, 'v'},
+		{ 0, 0, 0, 0},
+	};
+
+	int c;
+	while ((c = getopt_long(argc, (char **)argv, "dhn:p:r:v",
+					prog_opts, NULL)) != -1) {
+		switch (c) {
+		case 'd':
+			debug = true;
+			break;
+		case 'h':
+			print_usage();
+			exit(EXIT_SUCCESS);
+		case 'n':
+			params->netns_path = strdup(optarg);
+			break;
+		case 'p':
+			params->sandbox_id = strdup(optarg);
+			break;
+		case 'r':
+			params->runtime_path = strdup(optarg);
+			break;
+		case 'v':
+			print_version();
+			exit(EXIT_SUCCESS);
+		default:
+			print_usage();
+			exit(EXIT_FAILURE);
+		}
+	}
+
+	if (!params->netns_path) {
+		fprintf(stderr, "Missing network namespace path\n");
+		goto err;
+
+	}
+
+	if (!params->sandbox_id) {
+		fprintf(stderr, "Missing sandbox ID\n");
+		goto err;
+	}
+
+	if (!params->runtime_path) {
+		fprintf(stderr, "Missing runtime path\n");
+		goto err;
+	}
+
+	return 0;
+
+err:
+	print_usage();
+	return EINVAL;
+}
+
+/* 
+ * Enter the network namespace.
+ *
+ * Entering the provided network namespace since the point of this binary is
+ * to monitor any network change happening inside a specific network namespace.
+ */
+int enter_netns(const char *netns_path)
+{
+	int fd;
+
+	if (!netns_path) {
+		fprintf(stderr, "Network namespace path is empty");
+		return EINVAL;
+	}
+
+	fd = open(netns_path, O_RDONLY);
+	if (fd == -1) {
+		fprintf(stderr, "Failed opening network ns %s: %s\n",
+				netns_path, strerror(errno));
+		return -1;
+	}
+
+	if (setns(fd, 0) == -1) {
+		fprintf(stderr, "Failed to join network ns %s: %s\n",
+				netns_path, strerror(errno));
+		return -1;
+	}
+
+	return 0;
+}
+
+/*
+ * Open the netlink socket with appropriate parameters, and return the file
+ * descriptor. It will listen for any message related to a network interface
+ * change (add/delete/update), an IP address (IPv4) change (add/delete/update)
+ * or a route (IPv4) change (add/delete/update).
+ */
+int open_netlink()
+{
+	int sock;
+	struct sockaddr_nl sa;
+
+	memset(&sa, 0, sizeof(sa));
+
+	sa.nl_family = AF_NETLINK;
+	sa.nl_pid = getpid();
+	sa.nl_groups = RTMGRP_LINK |
+		RTMGRP_IPV4_IFADDR |
+		RTMGRP_IPV4_ROUTE;
+
+	sock = socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
+	if (sock == -1) {
+		fprintf(stderr, "Failed creating netlink socket: %s\n",
+				strerror(errno));
+		return -1;
+	}
+
+	if (bind(sock, (struct sockaddr *) &sa, sizeof(sa)) == -1) {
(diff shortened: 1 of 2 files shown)


sboeuf:
@bergwolf @miaoyq @WeiZhang555 Still WIP, but I'd like to get a first review !

bergwolf:
I guess it is better to put this in the cli directory?

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1213 @@
...
+	init_iface_list();
+
+	/* Validate parameters */
+	ret = valid_params(argc, (const char**) argv, &params);
+	if (ret) {
+		return ret;
+	}
+
+	/* Enter network namespace */
+	ret = enter_netns((const char*) params.netns_path);
```
miaoyq: I think we can run the monitor via `network.run(networkNSPath string, cb func() error) error`  instead of `enter_netns((const char*) params.netns_path)`.

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1213 @@
...
+	/* Print content of iface_list */
+	if (debug) {
+		print_iface_list();
+	}
+
+	/* Monitor the network */
+	ret = monitor_netns((const char*) params.pod_id,
+			    (const char*) params.runtime_path);
+	if (ret) {
+		goto exit;
```
miaoyq: Redundant if statement。

sboeuf: Sure I'll remove it.

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: You mean that virtcontainers would start the process directly into the network namespace, and I agree this would work, but I think it is better to have this being performed from the C process directly as entering namespaces is safer in C than Go. See https://github.com/kata-containers/runtime/issues/148 for more details about namespaces issues in Go.

Moreover, this makes this program more generic as it is able to enter a namespace on its own.

sboeuf:
List of tasks to consider:
- Replace `printf` logging with `syslog` logs.
- Provide the `pod_id` through the runtime command line.
- Make sure we don't save the changes to our internal list of interfaces if the runtime call returned with an exit code different from 0.
- Consider valid events only for a list of interface type (like `veth`), but ignore `bridge` for instance. Maybe this could be handled from virtcontainers directly, the monitoring process being more like a simple passthrough.
- Complete the code implementing the calls to the runtime. But this needs the CLI/API extension regarding network hotplug to be implemented. Or at least properly defined.

bergwolf:
@sboeuf There are two things I am thinking differently:
1. The workflow: I think the ns_watcher should be started by kata cli instead of virtcontainers library. Starting it from the virtcontainers library seems to be a layering violation to me, the call chain of which is `kata cli -> virtcontainers -> ns_watcher -> kata cli -> virtcontainers`. I would suggest we avoid such jumping back forth.
2. We should put this in the cli directory. The ns_watcher calls `kata-runtime cli` directly. Putting it in the virtcontainers directory introduces cycle dependency logically.

miaoyq:
>But in case we're adding a new IP address to an existing interface, or even when adding a new interface with IP addresses attached to it, we need to make sure we can provide what the interface should look like. This is part of the decision to go with the declarative way for the network, which means we describe what the interface should look like. Maintaining an accurate description of the network is needed in this case, otherwise we don't have the whole information about the interface through events like RTM_NEWADDR or RTM_DELADDR.

I think we only care the `veth` with ip addr, we should ignore the `veth` without ip addr utill an IP addr is added. Also, monitor only provides pod ID and interface name, and  virtcontainers can get all Network information according to pod ID.
>- Consider valid events only for a list of interface type (like veth), but ignore bridge for instance. Maybe this could be handled from virtcontainers directly, the monitoring process being more like a simple passthrough.

Agree.
>- Complete the code implementing the calls to the runtime. But this needs the CLI/API extension regarding network hotplug to be implemented. Or at least properly defined.

Anybody is working on sandbox API definition related to [the document](https://github.com/kata-containers/documentation/blob/master/design/kata-api-design.md) ? 
>- Replace the table/list of interfaces with a hash table.

What does this mean?  I'm not clear with this. :-p @sboeuf

sboeuf:
@bergwolf 
> The workflow: I think the ns_watcher should be started by kata cli instead of virtcontainers library. Starting it from the virtcontainers library seems to be a layering violation to me, the call chain of which is kata cli -> virtcontainers -> ns_watcher -> kata cli -> virtcontainers. I would suggest we avoid such jumping back forth.

The chain you're describing here is not really different (binary wise) since virtcontainers is a simple library and not a binary(different process) itself.
 
> We should put this in the cli directory. The ns_watcher calls kata-runtime cli directly. Putting it in the virtcontainers directory introduces cycle dependency logically.

There is no such thing as cycle dependency here since the binary is in C and it does not import anything from virtcontainers.

To be honest, I understand your concern and I am balanced between the two different possibilities. The main reason I'd like to keep it inside virtcontainers is that we need to know when we can start this from the CLI perspective. Virtcontainers knows better about this kind of thing. That being said, maybe starting this after the call to `CreateSandbox()` returned might be a way of doing this.
I'd like more input on this. @amshinde @devimc @grahamwhaley @egernst WDYT ?

bergwolf:
@sboeuf If we put this in virtcontainers, we make virtcontainers depend on kata cli because the ns_watcher binary depends on it. The functionality is not working without kata cli. That's why I call it **logical** dependency not golang import dependency.

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+
+			ret = create_iface_from_ifaddrs(ifa, &nif);
+			if (ret) {
+				free_iface(&nif);
+				goto exit;
+			}
+
+			nif.idx = if_idx;
+
+			if (if_idx >= MAX_IFACES) {
```
WeiZhang555: I think MAX_IFACES=50 is far from enough.

if_idx is link index, if host has more than 50 interfaces, then you create a new container, it's interface link index is easily larger than 50.
So I think link index shouldn't be used as the array's index, the array index should be standalone counter.

WeiZhang555: And I get a `segment fault`, to reproduce the error, you can create 50 containers, then try to exec

`$ netns_watcher  -d -n /proc/<pid>/ns/net -p hello -r /usr/local/bin/kata-runtime`

Pid should be process id of last container, and its interface index is larger than 50.

WeiZhang555:
Some inputs from me:

>Either we can make this monitoring process very simple if we consider providing only the interface name and saying that it has to be added/updated/deleted, based on the fact that virtcontainers will scan the network to get the whole interface description.
>Or we stick with the declarative way, and this means the monitoring process has to provide the full description of the network, for which it needs to maintain a full snapshot of the network. For instance, without this snapshot, we cannot determine if an interface needs to be updated or added in case we receive RTM_NEWLINK.

I think latter one is better. ns_watcher should take all responsibilities of watching network changes, and  send all interface infos to kata cli. Let kata cli/virtcontainers take another scan isn't a good idea in my opinion.

And it looks good to me to put this into virtcontainers/ lib.

Most part of the code looks quite cool, only some comments on details.

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: About the table of interfaces, that's exactly what I was mentioning when I said I'd like to replace this with a hash table ;). I will make the change.

Now about the segmentation fault, I am not sure, and I'll try to reproduce !

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+#include <arpa/inet.h>
+#include <ifaddrs.h>
+#include <netdb.h>
+#include <sys/ioctl.h>
+#include <syslog.h>
+#include <getopt.h>
+
+#define PROGRAM_NAME "netns-watcher"
+
+#define MAX_IFACES  50
```
jodh-intel: I'd do the following for maximum built-time flexibility:

```
#ifndef MAX_IFACES
#define MAX_IFACES  50
#endif
```

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
+/*
+ * This file is part of kata-runtime.
+ *
+ * Copyright (C) 2018 Intel Corporation
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
```
jodh-intel: - GPL or Apache 2.0?
- SPDX license header required ;)

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+ */
+struct iface iface_list[MAX_IFACES];
+
+/*
+ * Print full list of interfaces.
+ */
+void print_iface_list() {
+	int i;
+
+	for (i = 0; i < MAX_IFACES; i++) {
```
jodh-intel: I'd rather this was:

```
int max_ifaces = (int)(sizeof(iface_list)/sizeof(iface_list[0]));

for (i = 0; i < max_ifaces; i++) {
```

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+	printf(" -v, --version      Show version\n");
+	printf("\n");
+}
+
+/*
+ * Print version information.
+ */
+void print_version(void)
+{
+	printf("%s v0.1\n", PROGRAM_NAME);
```
jodh-intel: - Hard-coded version number. How about a `#define` for it?
- Version should be in semver format.

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+		free_ip_addr(nif->ip_addrs);
+	}
+	free(nif->ip_addrs);
+	nif->ip_addrs = NULL;
+}
+
+/*
+ * Free internal fields of the route structure.
+ */
+void free_route(struct route *rt) {
```
jodh-intel: I'd add a check on the param:

```
if (! rt) {
    return;
}
```

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+}
+
+/*
+ * Modify the global list of interfaces by inserting a new ip_addr structure
+ * into the linked list of the specified interface.
+ */
+void insert_ip_addr(int idx, char addr[], unsigned char family)
+{
+	struct ip_addr *ipa = (struct ip_addr *)
+		malloc(sizeof(struct ip_addr));
```
jodh-intel: Unless there is a very good reason not to, I'd recommend using `calloc()` always.

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+
+	free(rt->dev);
+	rt->dev = NULL;
+}
+
+/*
+ * Modify the global list of interfaces by inserting a new ip_addr structure
+ * into the linked list of the specified interface.
+ */
+void insert_ip_addr(int idx, char addr[], unsigned char family)
```
jodh-intel: Missing checks on params.

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+{
+	struct ip_addr *ipa = (struct ip_addr *)
+		malloc(sizeof(struct ip_addr));
+
+	ipa->addr = strdup(addr);
```
jodh-intel: Check return value?

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+	}
+
+	tmp->next = ipa;
+}
+
+/*
+ * Modify the global list of interfaces by removing an existing ip_addr
+ * structure from the linked list of the specified interface.
+ */
+void delete_ip_addr(int idx, char addr[])
```
jodh-intel: Check params.

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+		tmp = tmp->next;
+
+		if (tmp->addr && !strcmp(tmp->addr, addr)) {
+			found = true;
+			break;
+		}
+	}
+
+	if (!found) {
+		printf("No IP address matching %s for interface %s",
```
jodh-intel: This seems to be a warning so I'd print to `stderr` and prefix with `WARNING:`.

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+ * has to listen to.
+ * runtime-path: Runtime path is needed to call into it whenever a
+ * change in the network namespace is detected.
+ * netns-path: Network namespace path that needs to be monitored.
+ *
+ * It returns 0 if the parameters are valid, with a watcher_params structure
+ * properly filled. Otherwise, it returns -1 with a null watcher_params
+ * structure.
+ */
+int valid_params(int argc, const char **argv, struct watcher_params *params)
```
jodh-intel: How about `parse_options()` as a clearer name?

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+		} 
+
+		printf("\n");
+	}
+}
+
+/*
+ * Print program usage.
+ */
+void print_usage(void)
```
jodh-intel: I'd move this to be as close to `valid_params()` as possible to make maintenance as easy and reliable as possible.

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+ * properly filled. Otherwise, it returns -1 with a null watcher_params
+ * structure.
+ */
+int valid_params(int argc, const char **argv, struct watcher_params *params)
+{
+	struct option prog_opts[] = {
+		{"debug", no_argument, 0, 'd'},
+		{"help", no_argument, 0, 'h'},
+		{"netns-path", required_argument, 0, 'n'},
+		{"pod-id", required_argument, 0, 'p'},
```
jodh-intel: Should this be `sandbox-id` now? :)

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+			print_version();
+			exit(EXIT_SUCCESS);
+		default:
+			print_usage();
+			exit(EXIT_FAILURE);
+		}
+	}
+
+	if (!params->netns_path) {
+		printf("Missing network namespace path\n");
```
jodh-intel: Error, so `ERROR:` prefix and print to `stderr` (same comment for tests below).

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+ * Entering the provided network namespace since the point of this binary is
+ * to monitor any network change happening inside a specific network namespace.
+ */
+int enter_netns(const char *netns_path)
+{
+	int fd;
+
+	fd = open(netns_path, O_RDONLY);
+	if (fd == -1) {
+		printf("Failed opening network ns %s: %s\n",
```
jodh-intel: `ERROR:` and `stderr`.

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+	return -1;
+}
+
+/* 
+ * Enter the network namespace.
+ *
+ * Entering the provided network namespace since the point of this binary is
+ * to monitor any network change happening inside a specific network namespace.
+ */
+int enter_netns(const char *netns_path)
```
jodh-intel: Validate param.

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+{
+	int sock;
+	struct sockaddr_nl sa;
+
+	memset(&sa, 0, sizeof(sa));
+
+	sa.nl_family = AF_NETLINK;
+	sa.nl_pid = getpid();
+	sa.nl_groups = RTMGRP_LINK |
+		RTMGRP_IPV4_IFADDR |
```
jodh-intel: hard-coded IPv4. Is this intentional?

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+		RTMGRP_IPV4_ROUTE;
+
+	sock = socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
+	if (sock == -1) {
+		printf("Failed creating netlink socket: %s\n",
```
jodh-intel: `ERROR:` and `stderr` (here and everywhere ;)

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+}
+
+/*
+ * Get network interface name from index.
+ * 
+ * It's the caller responsibility to free the memory.
+ */
+char* iface_idx_to_name(int idx)
+{
+	int buf_len = 32 * sizeof(char);
```
jodh-intel: What is `32`?

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+	char *buf = (char *) malloc(buf_len);
```
jodh-intel: `calloc()`.

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+ */
+void print_version(void)
+{
+	printf("%s v0.1\n", PROGRAM_NAME);
+}
+
+/*
+ * Free internal fields of the watcher_params structure.
+ */
+void free_watcher_params(struct watcher_params *params)
```
jodh-intel: All params in all functions need to be checked.

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+
+	struct iface nif = {
+		.idx      = INIT_IF_IDX,
+		.hw_addr  = NULL,
+		.name     = NULL,
+		.mtu      = NULL,
+		.ip_addrs = NULL,
+	};
+
+	iface_list[idx] = nif;
```
jodh-intel: This looks wrong - you're appending a local variable to a global array?

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+		free(ipa_curr);
+		ipa_curr = NULL;
+	}
+}
+
+/*
+ * Free internal fields of the iface structure.
+ */
+void free_iface(struct iface *nif) {
+	int i;
```
jodh-intel: Unused.

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+
+	return sock;
+}
+
+/*
+ * Execute the runtime call from a child process.
+ */
+int fork_runtime_call(char *params[])
+{
+	int ret;
```
jodh-intel: Unused.

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
...
+	}
+}
+
+/*
+ * Fill a new interface from the information given by ifaddrs.
+ */
+int create_iface_from_ifaddrs(struct ifaddrs *ifa, struct iface *nif)
+{
+	int sock;
+	int family;
```
jodh-intel: Unused.

virtcontainers/netns_watcher/netns_watcher.c
```diff
@@ -0,0 +1,1251 @@
+/*
```
jodh-intel: How about giving it a prefix like `kata` or `vc`? :)

I assume it's intentional that this isn't actually mentioned in the `Makefile` yet too?

jodh-intel:
Since this PR introduces C code, we're also going to have to add C code static analysis checks specifically for this file. Ideally, we'd also register the project with https://scan.coverity.com/, but fwics that only works with Travis [which we're trying to get rid of](https://github.com/kata-containers/kata-containers/issues/22).

jodh-intel:
Could you update the commit message to explain what this tool does when it detects a change to a network ns?

sboeuf:
Depends on implementation and decisions made on #287 
Once #287 will get merged, the development on this PR could resume.

katabuilder:
PSS Measurement:
Qemu: 162111 KB
Proxy: 8749 KB
Shim: 10807 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  2001904 KB

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: done

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: done

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: done

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: done

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: done

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: you're right, done!

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: done

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: done

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: done

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: I have replaced all warnings and errors with `fprintf(stderr...`

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: done

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: done

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: done

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: done

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: Yes, as a first step, I want to make sure we handle properly ipv4 only.

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: Lol, I have added `fprintf(stderr,...)` everywhere.

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: done

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: Nevermind, I have just read the doc https://linux.die.net/man/3/if_indextoname and it clearly says that we should allocate a buffer of __at least__ `IF_NAMESIZE` (being 16). I think we can stick to that for now.

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: done

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: done

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: done

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: Yes let's call this `kata-netmon` which stands for "Kata network monitor". And yes, the Makefile will come later. I need to finalize the program first.

virtcontainers/netns_watcher/netns_watcher.c
sboeuf: Well, this is not simply to make sure every field of `iface_list[idx]` get reset to the proper values. But I can do something different.

jodh-intel:
btw, I think you'll need to add a commit to enable Coverity Scan as this is the first piece of C code in the runtime repo. Basically crib what you can from:

- https://github.com/intel/cc-oci-runtime/blob/master/.travis.yml

.. and do:

- https://scan.coverity.com/travis_ci

jodh-intel:
... for static analysis, we could start by setting up https://github.com/marketplace/lgtm which is free for OSS projects and has full github integration fwics.

grahamwhaley:
I can't think of any reason not to give [lgtm]( https://github.com/marketplace/lgtm) a tryout at least...

jodh-intel:
Hi @sboeuf - please could you push your latest branch as your comments don't appear to match up with reality for us :)

jodh-intel:
Sounds good - thanks for the update @sboeuf.

sboeuf:
I want to keep this PR around since we might move to the C implementation later. But for now, we want to prototype this through the new PR https://github.com/kata-containers/runtime/pull/534

virtcontainers/netmon/netmon.c
```diff
@@ -0,0 +1,1301 @@
...
+
+	free_iface(&nif);
+
+	return 0;
+}
+
+/*
+ * Add route.
+ */
+int add_route(const struct nlmsghdr *nh, const char *sandbox_id,
```
bergwolf: For this kind of change, I would insist that either we make change to network config via the sandbox network hotplug API, or we put this binary in the cli directory. I do not think we should make virtcontainers call the kata-runtime binary directly.

sboeuf: See my comment [here](https://github.com/kata-containers/runtime/pull/534#discussion_r206663151), let me know what you think about it !

sboeuf:
Closing this PR as it's very unlikely that we'll use this C code. The code will still be available on my Github fork on the branch `netns_watcher`.

PR #196, opened 2018-04-10
cli: implement spec command

### LINKED ISSUE #188: Implement the spec command
The runtime does not implement the `runc spec` command that generates a basic `config.json`.

caoruidong:
Add spec command that  that generates a basic config.json for kata.

fixes #188

### DIFF
--- a/cli/spec.go
+++ b/cli/spec.go
@@ -1,0 +1,105 @@
+// Copyright (c) 2014,2015,2016,2017 Docker, Inc.
+// Copyright (c) 2018 Huawei Corporation.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/opencontainers/runc/libcontainer/specconv"
+	"github.com/urfave/cli"
+)
+
+var specCLICommand = cli.Command{
+	Name:      "spec",
+	Usage:     "create a new specification file",
+	ArgsUsage: "",
+	Description: `The spec command creates the new specification file named "` + specConfig + `" for
+the bundle.
+
+The spec generated is just a starter file. Editing of the spec is required to
+achieve desired results. For example, the newly generated spec includes an args
+parameter that is initially set to call the "sh" command when the container is
+started. Calling "sh" may work for an ubuntu container or busybox, but will not
+work for containers that do not include the "sh" program.
+
+EXAMPLE:
+  To run docker's hello-world container one needs to set the args parameter
+in the spec to call hello. This can be done using the sed command or a text
+editor. The following commands create a bundle for hello-world, change the
+default args parameter in the spec from "sh" to "/hello", then run the hello
+command in a new hello-world container named container1:
+
+    mkdir hello
+    cd hello
+    docker pull hello-world
+    docker export $(docker create hello-world) > hello-world.tar
+    mkdir rootfs
+    tar -C rootfs -xf hello-world.tar
+    kata-runtime spec
+    sed -i 's;"sh";"/hello";' ` + specConfig + `
+    kata-runtime run container1
+
+In the run command above, "container1" is the name for the instance of the
+container that you are starting. The name you provide for the container instance
+must be unique on your host.
+
+An alternative for generating a customized spec config is to use "oci-runtime-tool", the
+sub-command "oci-runtime-tool generate" has lots of options that can be used to do any
+customizations as you want, see runtime-tools (https://github.com/opencontainers/runtime-tools)
+to get more information.
+
+When starting a container through kata-runtime, kata-runtime needs root privilege. If not
+already running as root, you can use sudo to give kata-runtime root privilege. For
+example: "sudo kata-runtime start container1" will give kata-runtime root privilege to start the
+container on your host.
+
+Alternatively, you can start a rootless container, which has the ability to run
+without root privileges. For this to work, the specification file needs to be
+adjusted accordingly. You can pass the parameter --rootless to this command to
+generate a proper rootless spec file.`,
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:  "bundle, b",
+			Value: "",
+			Usage: "path to the root of the bundle directory",
+		},
+	},
+	Action: func(context *cli.Context) error {
+		spec := specconv.Example()
+
+		checkNoFile := func(name string) error {
+			_, err := os.Stat(name)
+			if err == nil {
+				return fmt.Errorf("File %s exists. Remove it first", name)
+			}
+			if !os.IsNotExist(err) {
+				return err
+			}
+			return nil
+		}
+		bundle := context.String("bundle")
+		if bundle != "" {
+			if err := os.Chdir(bundle); err != nil {
+				return err
+			}
+		}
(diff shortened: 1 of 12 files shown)


caoruidong:
Sorry, forget to update dep. Wait

cli/spec.go
```diff
@@ -0,0 +1,108 @@
+package main
```
devimc: missing license

cli/spec.go
```diff
@@ -0,0 +1,108 @@
...
+			}
+		}
+		if err := checkNoFile(specConfig); err != nil {
+			return err
+		}
+		data, err := json.MarshalIndent(spec, "", "\t")
+		if err != nil {
+			return err
+		}
+		if err := ioutil.WriteFile(specConfig, data, 0666); err != nil {
```
devimc: pure evil  ehh :smile:

cli/spec.go
```diff
@@ -0,0 +1,108 @@
...
+			Name:  "bundle, b",
+			Value: "",
+			Usage: "path to the root of the bundle directory",
+		},
+		cli.BoolFlag{
+			Name:  "rootless",
+			Usage: "generate a configuration for a rootless container",
+		},
+	},
+	Action: func(context *cli.Context) error {
```
devimc: please add an unit test for this

cli/spec.go
caoruidong: Change mode to 640?

cli/spec.go
```diff
@@ -0,0 +1,108 @@
...
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name:  "bundle, b",
+			Value: "",
+			Usage: "path to the root of the bundle directory",
+		},
+		cli.BoolFlag{
+			Name:  "rootless",
+			Usage: "generate a configuration for a rootless container",
+		},
```
WeiZhang555: I'm not sure if we can support "rootless" container in kata in future, as I know this is a runc feature that allows you to run a container without root privilege.

This flag needs to be removed if it's not supported yet.

cli/spec.go
```diff
@@ -0,0 +1,108 @@
...
+	},
+	Action: func(context *cli.Context) error {
+		spec := specconv.Example()
+
+		rootless := context.Bool("rootless")
+		if rootless {
+			specconv.ToRootless(spec)
+		}
```
WeiZhang555: This part needs to be removed if `rootless` container isn't supported yet

WeiZhang555:
I'm wondering if these vendor files added by `dep` tool? Please update the vendor with `dep` tool, one reference link is: 
https://github.com/kata-containers/runtime/pull/141

bergwolf:
Could you add a test case the makes sure we can run a kata container using the default spec? There is [an example in runv] (https://github.com/hyperhq/runv/blob/master/tests/go-integration/test_data/Makefile) for creating a busybox bundle that might help you.

caoruidong:
@bergwolf Sure. Thanks. I'm working on this.
@WeiZhang555 Yes.  I will update with dep together with unit test.

cli/spec.go
devimc: runc uses 0664

WeiZhang555: I think runc is using 0666: https://github.com/opencontainers/runc/blob/master/spec.go#L110

But I don't know why, I think 0640 could be enough.

devimc: that's weird

```
$ docker-runc spec
$ ls -l config.json
-rw-rw-r-- 1 X X 597 Apr 11 10:33 config.json
```

sboeuf: Yes, let's do better than `runc` here, `0640` will be enough.

cli/spec.go
caoruidong: Done

cli/spec.go
caoruidong: Done

cli/spec.go
caoruidong: Done

caoruidong:
@WeiZhang555 Vendor updated.
@bergwolf After testing, the default spec can work with busybox. Do you mean that add a test case in spec_test.go to explicitly call kata-runtime run with default spec and busybox bundle ?

bergwolf:
@caoruidong I don't think we can do it in `spec_test.go`. I was thinking some place in https://github.com/kata-containers/tests. It does not block this PR though.

caoruidong:
@bergwolf So I think test_data/Makefile should be moved to https://github.com/kata-containers/tests.

bergwolf:
@caoruidong Feel free to use it when adding new tests. I just pointed to the runv tests for your reference.

sboeuf:
@devimc PTAL.

caoruidong:
@jodh-intel @WeiZhang555 PTAL

WeiZhang555:
LGTM after rebase.
Nice patch!

caoruidong:
Rebased!

caoruidong:
Can this be merged?

jodh-intel:
Done on https://github.com/kata-containers/documentation/pull/48/commits/2f2956e38c6668e70973fbc9b694e6651ae4a70c.

jodh-intel:
Tests fine in stand-alone mode fwiw (using the strategy below for `kata-runtime`):

- https://github.com/clearcontainers/runtime/wiki/Running-standalone

grahamwhaley:
The only thing stopping the merge is the red codecov reports - if we are happy with that @jodh-intel , then let's merge...

cli/spec.go
```diff
@@ -0,0 +1,114 @@
+// Copyright (c) 2014,2015,2016,2017 Docker, Inc.
+// Copyright (c) 2018 Huawei Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
grahamwhaley: Ah, I thought the standard for this project (the whole of kata) was the SPDX style copyright like:
https://github.com/kata-containers/runtime/blob/master/cli/fatal.go#L3
but, I note we have some non-SPDX copyright in the code base already, such as:
https://github.com/kata-containers/runtime/blob/master/cli/config.go#L3

@egernst @sboeuf @jodh-intel , thoughts?
And, maybe we can add a static check to catch any future deviations (whichever way we go) to the CI - maybe.

Other than that, this copyright looks correct to me afaict - the original code coming from https://github.com/opencontainers/runc/blob/master/spec.go I believe. That original file does not have an explicit copyright (it relies on the project wide LICENSE file I believe), so I presume we looked at the git logs to get the date range in the above - which I believe is fine and correct.

Maybe we can add a URL reference to that original file just below the copyright to make the historical origins clear?

grahamwhaley: btw, if we do all agree on SPDX (which we did I think at project inception, but we probably did not document well enough, and have not always caught transgressions), then I'm happy to Issue/PR to update/fix them all etc.

jodh-intel: Good spot @grahamwhaley! Yep, I agree - we should ensure all our files have the SPDX header where possible.

I've raised https://github.com/kata-containers/tests/issues/240 to add a CI check to enforce this for new / modified files, but would be great to have the existing codebase "SPDX-ified" :smile:

caoruidong:
@grahamwhaley It's green last time... I'll update aggin. 
@jodh-intel SPDX-ing together.

jodh-intel:
yep - let's do this! ;)

PR #211, opened 2018-04-13
Assign address to a pci bridge while appending it

amshinde:
This will allow to determine the PCI address of the device attached to the bridge within the guest.

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -353,6 +353,10 @@ func (q *qemu) createSandbox(sandboxConfig SandboxConfig) error {
 		},
 	}
 
+	// Add bridges before any other devices. This way we make sure that
+	// bridge gets the first available PCI address i.e bridgePCIStartAddr
+	devices = q.arch.appendBridges(devices, q.state.Bridges)
+
 	devices = q.arch.append9PVolumes(devices, sandboxConfig.Volumes)
 	devices = q.arch.appendConsole(devices, q.getSandboxConsole(sandboxConfig.ID))
 
@@ -361,11 +365,6 @@ func (q *qemu) createSandbox(sandboxConfig SandboxConfig) error {
 		if err != nil {
 			return err
 		}
-	}
-
-	devices = q.arch.appendBridges(devices, q.state.Bridges)
-	if err != nil {
-		return err
 	}
 
 	var ioThread *govmmQemu.IOThread
--- a/virtcontainers/qemu_amd64.go
+++ b/virtcontainers/qemu_amd64.go
@@ -8,6 +8,7 @@ package virtcontainers
 import (
 	"fmt"
 	"os"
+	"strconv"
 
 	govmmQemu "github.com/intel/govmm/qemu"
 )
@@ -204,6 +205,8 @@ func (q *qemuAmd64) appendBridges(devices []govmmQemu.Device, bridges []Bridge) 
 			t = govmmQemu.PCIEBridge
 		}
 
+		b.Addr = bridgePCIStartAddr + idx
+
 		devices = append(devices,
 			govmmQemu.BridgeDevice{
 				Type: t,
@@ -212,6 +215,7 @@ func (q *qemuAmd64) appendBridges(devices []govmmQemu.Device, bridges []Bridge) 
 				// Each bridge is required to be assigned a unique chassis id > 0
 				Chassis: (idx + 1),
 				SHPC:    true,
+				Addr:    strconv.FormatInt(int64(b.Addr), 10),
 			},
 		)
 	}
--- a/virtcontainers/qemu_arch_base.go
+++ b/virtcontainers/qemu_arch_base.go
@@ -9,6 +9,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"os"
+	"strconv"
 
 	govmmQemu "github.com/intel/govmm/qemu"
 )
@@ -99,6 +100,12 @@ const (
 	maxDevIDSize            = 31
 	defaultMsize9p          = 8192
 )
+
+// This is the PCI start address assigned to the first bridge that
+// is added on the qemu command line. In case of x86_64, the first two PCI
+// addresses (0 and 1) are used by the platform while in case of ARM, address
+// 0 is reserved.
+const bridgePCIStartAddr = 2
 
 const (
 	// VirtioBlock means use virtio-blk for hotplugging drives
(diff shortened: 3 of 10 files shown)


virtcontainers/qemu_arch_base.go
```diff
@@ -110,6 +111,8 @@ const (
 	maxDevIDSize            = 31
 )
 
+const bridgePCIStartAddr = 5
```
sboeuf: Why 5 ? I thought Qemu was reserving only the slot 0 and 1. Could you add a comment explaining this ?

amshinde: @sboeuf  Slot 0 and 1 are reserved in case of bridges. I am choosing address 5 here on the main pci bus. There are a few devices that we add on the main bus currently, eg
```
sudo docker run -it --runtime=kata-runtime  centos/tools
[root@e9f831b1e6e7 /]# lspci
00:00.0 Host bridge: Intel Corporation 440FX - 82441FX PMC [Natoma] (rev 02)
00:01.0 ISA bridge: Intel Corporation 82371SB PIIX3 ISA [Natoma/Triton II]
00:01.1 IDE interface: Intel Corporation 82371SB PIIX3 IDE [Natoma/Triton II]
00:01.3 Bridge: Intel Corporation 82371AB/EB/MB PIIX4 ACPI (rev 03)
00:02.0 Communication controller: Red Hat, Inc Virtio console
00:03.0 PCI bridge: Red Hat, Inc. QEMU PCI-PCI bridge
00:04.0 SCSI storage controller: Red Hat, Inc Virtio SCSI
00:05.0 Unclassified device [0002]: Red Hat, Inc Virtio filesystem
00:06.0 Ethernet controller: Red Hat, Inc Virtio network device
```
Here pci bridge is added at 00:03.0, I just chose to add bridges starting at 00.04.0, with the other devices occupying the other slots.

amshinde: @sboeuf We dont care about the location of other devices, we just care about the bridge address. Bridges are cold-plugged, the first available slot in the current sequence is 0x03,  I could have just chosen that. By simply choosing 0x5,  I am adding bridges from that address, leaving the other addresses for other devices, qemu will find the first available slot for those devices.

sboeuf: I totally get it, I was wondering if other cold plugged devices (do we have some ?), could be cold plugged before the bridge. If that's the case, then you cannot guarantee that nothing else will appear at address 5.
Following this explanation, if you're sure that the bridge is gonna be the first thing being added to the main bridge, then I think it is better to take the first PCI address that we know available (and explain that previous addresses 0 and 1 are indeed used internally by Qemu, but this is a known thing).

amshinde: I added a buffer to have some flexibility. i can change this from 0x5 to 0x3 if it makes things any clearer.

amshinde: @sboeuf Changes pushed following our conversation. I have added in comments and moved the bridge append early on. ptal.

virtcontainers/qemu.go
```diff
@@ -364,6 +364,13 @@ func (q *qemu) createSandbox(sandboxConfig SandboxConfig) error {
 		},
 	}
 
+	// Add bridges before any other devices. This way we make sure that
+	// bridge gets the first available PCI adress i.e bridgePCIStartAddr
```
jodh-intel: typo: "address".

virtcontainers/qemu.go
```diff
@@ -364,6 +364,13 @@ func (q *qemu) createSandbox(sandboxConfig SandboxConfig) error {
...
+	devices = q.arch.appendBridges(devices, q.state.Bridges)
+	if err != nil {
```
jodh-intel: Huh? This check looks bogus as `err` wasn't set by `appendBridges()`.

virtcontainers/qemu_amd64.go
```diff
@@ -215,6 +216,8 @@ func (q *qemuAmd64) appendBridges(devices []govmmQemu.Device, bridges []Bridge)
 			t = govmmQemu.PCIEBridge
 		}
 
+		b.Addr = bridgePCIStartAddr + idx
```
jodh-intel: What happens on ARM?

virtcontainers/qemu_arch_base.go
```diff
@@ -110,6 +111,12 @@ const (
 	maxDevIDSize            = 31
 )
 
+// This is the PCI start address assigned to the first bridge that
+// is added on the qemu command line. In case of x86_64, the first two PCI
+// addresses (0 and 1) are used by the platform while in case of ARM, address
+// 0 is reserved.
```
jodh-intel: Good comment, but if the behaviour is different for x86_64 vs ARM, why is the (x86_64-specific) constant in the generic base file?

virtcontainers/qemu_arch_base.go
```diff
@@ -331,6 +338,8 @@ func (q *qemuArchBase) appendBridges(devices []govmmQemu.Device, bridges []Bridg
 			t = govmmQemu.PCIEBridge
 		}
 
+		b.Addr = bridgePCIStartAddr + idx
```
jodh-intel: Isn't this a duplicate of what is already in `qemu_amd64.go`? If so, can't this be removed?

virtcontainers/qemu.go
amshinde: yeah you are right, this was part of a later change I am working on.

virtcontainers/qemu_arch_base.go
amshinde: The code to add a bridge was different, and the address needs to be incremented while adding bridges in both the cases.

virtcontainers/qemu_arch_base.go
amshinde: We just need to make sure that the address does not conflict. So I chose greater of the two addresses.

virtcontainers/qemu_amd64.go
amshinde: The code in qemu_arch_base will be used.

amshinde:
@jodh-intel Comments addresses, ptal.

amshinde:
@devimc @sboeuf Rebased.

devimc:
@amshinde `TestQemuArchBaseAppendBridges` failed

amshinde:
@sboeuf k8s CI seems to be failing. Is this something we have seen before?

sboeuf:
@amshinde no the failure is about the kata-log-parser. We have something in the pipe for that, in the meantime, please re-trigger the jenkins job.

PR #213, opened 2018-04-14
virtcontainers: Improve container searching

### LINKED ISSUE #212: virtcontainers: Improve container searching
Currently, the more we are running some containers on the system, the more time it takes our runtime to start new ones. The reason behind is that we always ask the list of all sandboxes in order to parse it from the CLI. This parsing resulting in finding if a container exists or not, and in case it does exist, we need its associated sandbox ID.

An easy way to simplify the current complexity O(n²) into an O(1), would be to use a hash table listing all the containers as a key and their associated sandbox as a value.

We should keep track of the containers and their associated pods through a hash table stored on the host filesystem, in order to support the short lifetime of the CLI process.

sboeuf:
The more we are running some containers on the system, the more
time it takes our runtime to start new ones. The reason behind is
that we always ask the list of all sandboxes in order to parse it
from the CLI. This parsing resulting in finding if a container
exists or not, and in case it does exist, we need its associated
sandbox ID.

An easy way to simplify the current complexity O(n²) into an O(1),
is to use a hash table listing all the containers as a key and their
associated sandbox as a value.

That's what this patch does, it introduces a hash table being stored
on the filesystem through a JSON file. This way, we can keep track of
the containers and their associated sandboxes, even in case of the
short lifetime of the kata-runtime CLI.

Fixes #212

### DIFF
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -639,3 +639,25 @@ func ProcessListContainer(sandboxID, containerID string, options ProcessListOpti
 
 	return c.processList(options)
 }
+
+// ContainerSandboxList returns the list of sandboxes IDs for all
+// the sandboxes including the provided container ID.
+// It also provides the information about the container ID being part
+// of the containers map.
+func ContainerSandboxList(containerID string) ([]string, bool, error) {
+	if containerID == "" {
+		return nil, false, errNeedContainerID
+	}
+
+	ctrsMap, err := fetchContainersMap()
+	if err != nil {
+		return nil, false, err
+	}
+
+	sandboxList, exist := ctrsMap[containerID]
+	if !exist {
+		return nil, false, nil
+	}
+
+	return sandboxList, true, nil
+}
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -537,6 +537,13 @@ func createContainer(sandbox *Sandbox, contConfig ContainerConfig) (c *Container
 		return
 	}
 
+	// Update containers map if stateful flag.
+	if !c.sandbox.config.Stateless {
+		if err = addToContainersMap(contConfig.ID, sandbox.id); err != nil {
+			return
+		}
+	}
+
 	return c, nil
 }
 
@@ -551,7 +558,18 @@ func (c *Container) delete() error {
 		return err
 	}
 
-	return c.sandbox.storage.deleteContainerResources(c.sandboxID, c.id, nil)
+	if err := c.sandbox.storage.deleteContainerResources(c.sandboxID, c.id, nil); err != nil {
+		return err
+	}
+
+	// Update containers map if stateful flag.
+	if !c.sandbox.config.Stateless {
+		if err := delFromContainersMap(c.id, c.sandbox.id); err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 // checkSandboxRunning validates the container state.
--- a/virtcontainers/filesystem.go
+++ b/virtcontainers/filesystem.go
@@ -48,6 +48,9 @@ const (
 
 	// devicesFileType represents a device file type
 	devicesFileType
+
+	// containersMapFileType represents a containers map file type
+	containersMapFileType
 )
 
 // configFile is the file name used for every JSON sandbox configuration.
@@ -76,6 +79,9 @@ const mountsFile = "mounts.json"
 // devicesFile is the file name storing a container's devices.
 const devicesFile = "devices.json"
 
+// containersMapFile is the file name storing a map of containers/sandboxes.
+const containersMapFile = "containers_map.json"
+
 // dirMode is the permission bits used for creating a directory
 const dirMode = os.FileMode(0750) | os.ModeDir
 
@@ -101,6 +107,11 @@ type resourceStorage interface {
 	// for the actual resource and the URI base.
 	containerURI(sandboxID, containerID string, resource sandboxResource) (string, string, error)
 	sandboxURI(sandboxID string, resource sandboxResource) (string, string, error)
+	globalSandboxURI(resource sandboxResource) (string, string, error)
+
+	// Global resources
+	fetchContainersMap(ctrsMap *map[string][]string) error
+	storeContainersMap(ctrsMap map[string][]string) error
 
 	// Sandbox resources
 	storeSandboxResource(sandboxID string, resource sandboxResource, data interface{}) error
@@ -427,6 +438,33 @@ func (fs *filesystem) sandboxURI(sandboxID string, resource sandboxResource) (st
 	return fs.resourceURI(true, sandboxID, "", resource)
 }
 
+func (fs *filesystem) globalSandboxURI(resource sandboxResource) (string, string, error) {
+	var dirPath string
+
+	switch resource {
+	case lockFileType, containersMapFileType:
+		dirPath = runStoragePath
+		break
+	default:
+		return "", "", errInvalidResource
+	}
+
+	var filename string
+
+	switch resource {
+	case lockFileType:
+		filename = lockFileName
+		break
+	case containersMapFileType:
+		filename = containersMapFile
+		break
+	default:
+		return "", "", errInvalidResource
+	}
+
+	return filepath.Join(dirPath, filename), dirPath, nil
+}
+
 // commonResourceChecks performs basic checks common to both setting and
 // getting a sandboxResource.
 func (fs *filesystem) commonResourceChecks(sandboxSpecific bool, sandboxID, containerID string, resource sandboxResource) error {
@@ -617,6 +655,35 @@ func (fs *filesystem) fetchAgentState(sandboxID string, state interface{}) error
 	return fs.fetchResource(true, sandboxID, "", agentFileType, state)
 }
 
+func (fs *filesystem) fetchContainersMap(ctrsMap *map[string][]string) error {
+	resource := containersMapFileType
+
+	filePath, _, err := fs.globalSandboxURI(resource)
+	if err != nil {
+		return err
+	}
+
+	if _, err := os.Stat(filePath); os.IsNotExist(err) {
+		f, err := os.Create(filePath)
+		if err != nil {
+			return err
+		}
+		f.Close()
+
+		m := make(map[string][]string)
+		if err := fs.storeContainersMap(m); err != nil {
+			return err
+		}
+	}
+
+	fileData, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		return err
+	}
+
+	return json.Unmarshal(fileData, ctrsMap)
+}
+
 func (fs *filesystem) storeSandboxNetwork(sandboxID string, networkNS NetworkNamespace) error {
 	return fs.storeSandboxResource(sandboxID, networkFileType, networkNS)
 }
@@ -637,6 +704,17 @@ func (fs *filesystem) storeAgentState(sandboxID string, state interface{}) error
 	}
 
 	return fs.storeFile(agentFile, state)
+}
+
+func (fs *filesystem) storeContainersMap(ctrsMap map[string][]string) error {
+	resource := containersMapFileType
+
+	filePath, _, err := fs.globalSandboxURI(resource)
+	if err != nil {
+		return err
+	}
+
+	return fs.storeFile(filePath, ctrsMap)
 }
 
 func (fs *filesystem) deleteSandboxResources(sandboxID string, resources []sandboxResource) error {
--- a/cli/create_test.go
+++ b/cli/create_test.go
@@ -305,14 +305,14 @@ func TestCreateInvalidArgs(t *testing.T) {
 		return sandbox, nil
 	}
 
-	testingImpl.ListSandboxFunc = func() ([]vc.SandboxStatus, error) {
+	testingImpl.ContainerSandboxListFunc = func(containerID string) ([]string, bool, error) {
 		// No pre-existing sandboxes
-		return []vc.SandboxStatus{}, nil
+		return []string{}, false, nil
 	}
 
 	defer func() {
 		testingImpl.CreateSandboxFunc = nil
-		testingImpl.ListSandboxFunc = nil
+		testingImpl.ContainerSandboxListFunc = nil
 	}()
 
 	tmpdir, err := ioutil.TempDir("", "")
@@ -355,13 +355,13 @@ func TestCreateInvalidArgs(t *testing.T) {
 func TestCreateInvalidConfigJSON(t *testing.T) {
 	assert := assert.New(t)
 
-	testingImpl.ListSandboxFunc = func() ([]vc.SandboxStatus, error) {
+	testingImpl.ContainerSandboxListFunc = func(containerID string) ([]string, bool, error) {
 		// No pre-existing sandboxes
-		return []vc.SandboxStatus{}, nil
+		return []string{}, false, nil
 	}
 
 	defer func() {
-		testingImpl.ListSandboxFunc = nil
+		testingImpl.ContainerSandboxListFunc = nil
 	}()
 
 	tmpdir, err := ioutil.TempDir("", "")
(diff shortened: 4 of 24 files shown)


sboeuf:
@grahamwhaley thank you for confirming this with your tests :+1:

sboeuf:
@jodh-intel thanks for the review but I am currently working on this PR to be more compliant. This version was only done as a POC

jodh-intel:
ack - thanks @sboeuf!

virtcontainers/api.go
```diff
@@ -650,3 +650,7 @@ func ProcessListContainer(sandboxID, containerID string, options ProcessListOpti
 
 	return c.processList(options)
 }
+
+func ContainersMap() (map[string]string, error) {
```
bergwolf: IIUC, the purpose is to find out a container's sandboxID. Can you simply the API to just do the job? e.g.,
```func CantainerSandbox(containerID string) (sandboxID string, err error)```
So that it does not make internal data structure part of the API.

File: virtcontainers/filesystem.go [hunk 2]
```diff
@@ -112,6 +118,11 @@ type resourceStorage interface {
 	// for the actual resource and the URI base.
 	containerURI(sandboxID, containerID string, resource sandboxResource) (string, string, error)
 	sandboxURI(sandboxID string, resource sandboxResource) (string, string, error)
+	sandboxesURI(resource sandboxResource) (string, string, error)
```
bergwolf: The interface naming is really confusing. How about `globalSandboxURI`?

virtcontainers/filesystem.go
```diff
@@ -438,6 +449,55 @@ func (fs *filesystem) sandboxURI(sandboxID string, resource sandboxResource) (st
...
 }
 
+func (fs *filesystem) sandboxesURI(resource sandboxResource) (string, string, error) {
+	var dirPath string
+
+	switch resource {
+	case configFileType:
+		dirPath = configStoragePath
+		break
+	case stateFileType, networkFileType, processFileType, lockFileType, mountsFileType, devicesFileType, hypervisorFileType, agentFileType, containersMapFileType:
```
bergwolf: Are we going to handle all these kind of types at all?

virtcontainers/api.go
sboeuf: Well the answer is that we don't want virtcontainers to run under the same assumptions as the CLI. In this case that a container ID is unique across sandboxes, meaning there is a single sandbox ID corresponding to the container ID provided. That being said, the whole logic handling a map assumes so... I guess this could all be implemented into the CLI instead, but it would mean that we would have to store this somewhere in the filesystem. Where would it be ?
The nice thing about virtcontainers is that it was handling this for us completely.

virtcontainers/filesystem.go
sboeuf: Yes sounds good.

virtcontainers/filesystem.go
sboeuf: Nope, as I said to @jodh-intel, this was only a POC, I have to rework the code.

virtcontainers/api.go
bergwolf: Yup, the use of a map does not look accurate even inside virtcontainers.

I would vote for implementing it in the cli and we can just store the sandboxID of a container in its own directory. Thus no global maps nor global on disk locks.

We can store it in the root storage of kata cli (default to `/var/run/kata-containers`), such as `/var/run/kata-containers/containers/<container_id>/state-or-whatever.json`.

sboeuf: Well I have thought about it and I am gonna implement it in virtcontainers in a generic way, by using a hash table of containers pointing to a list of pods. Now, the function will return the list of pods related to a container ID, which is okay and compliant with virtcontainers. This way we keep the storage being handled by virtcontainers. From the CLI perspective, we will get a list of pods which should never contain more than one element.

bergwolf: Well, in that case please make it optional so that we can opt out in the CRI use cases. We will definitely disable it in frakti. We might also want to disable it for cri-o and cri-containerd as well to remove the unnecessary global locking and extra file system access, once we decouple the spec validation of cri containers from docker containers. In the cri shim case, we only need global uniqueness of sandboxID and sandbox-wide uniqueness of containerID, which can both be validated in virtcontainers. Then the new API is only necessary for docker.

bergwolf: What's more, we always use containerID as sandboxID in the docker use case. The uniqueness of containerID == the uniqueness of sandboxID. So I do not think we need the API at all. Just validate proper uniqueness of sandboxID and containerID in virtcontainers.

virtcontainers/filesystem.go
```diff
@@ -628,6 +688,41 @@ func (fs *filesystem) fetchAgentState(sandboxID string, state interface{}) error
 	return fs.fetchResource(true, sandboxID, "", agentFileType, state)
 }
 
+func (fs *filesystem) fetchContainersMap(ctrsMap *map[string]string) error {
+	fs.Logger().Info("### fetchContainersMap 1 ####")
```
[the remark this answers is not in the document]
sboeuf: I have removed them

virtcontainers/sandboxes.go
```diff
@@ -0,0 +1,132 @@
...
+	if err := fs.fetchContainersMap(&ctrsMap); err != nil {
+		return nil, err
+	}
+
+	return ctrsMap, nil
+}
+
+// addToContainersMap adds a pair containerID/sandboxID to the map and stores
+// the result.
+func addToContainersMap(ctrID, sandboxID string) error {
```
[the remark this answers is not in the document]
sboeuf: Check added.

virtcontainers/sandboxes.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	ctrsMap[ctrID] = sandboxID
+
+	return fs.storeContainersMap(ctrsMap)
+}
+
+// delFromContainersMap removes a pair containerID/sandboxID from the map and
+// stores the result.
+func delFromContainersMap(ctrID, sandboxID string) error {
```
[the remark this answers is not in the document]
sboeuf: ditto

virtcontainers/sandboxes.go
```diff
@@ -0,0 +1,132 @@
+// Copyright (c) 2018 Intel Corporation
```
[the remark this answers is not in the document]
sboeuf: done

virtcontainers/filesystem.go
```diff
@@ -438,6 +449,55 @@ func (fs *filesystem) sandboxURI(sandboxID string, resource sandboxResource) (st
...
+	default:
+		return "", "", errInvalidResource
+	}
+
+	var filename string
+
+	switch resource {
+	case configFileType:
+		filename = configFile
+		break
```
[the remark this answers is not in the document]
sboeuf: I have simplified this function

virtcontainers/api.go
sboeuf: @bergwolf I have introduced a flag in the SandboxConfig. It'll be up to Frakti to define that it needs a stateless configuration.

virtcontainers/filesystem.go
sboeuf: done

virtcontainers/filesystem.go
sboeuf: done

sboeuf:
@jodh-intel @bergwolf PTAL

virtcontainers/sandboxes.go
```diff
@@ -0,0 +1,163 @@
...
```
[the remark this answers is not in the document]
sboeuf: done

sboeuf:
@bergwolf I broke this down into 4 different commits ! I hope this helps ;)

virtcontainers/api.go
```diff
@@ -639,3 +639,25 @@ func ProcessListContainer(sandboxID, containerID string, options ProcessListOpti
 
 	return c.processList(options)
 }
+
+// ContainerSandboxList returns the list of sandboxes IDs for all
+// the sandboxes including the provided container ID.
+// It also provides the information about the container ID being part
+// of the containers map.
+func ContainerSandboxList(containerID string) ([]string, bool, error) {
```
bergwolf: TBH, I am still not a fan of this API and the global container-sandbox mapping file. It introduces a global on disk lock and extra json marshaling/unmarshaling for every container creation/deletion. And this extra json marshaling/unmarshaling gets slower as the number of containers increases on the same node. Of course, the extra stateless flag can help frakti/hyperd avoid the situation. But kata cli will still be affected.

I think we should draw wider attention to the change or discuss it in the arch meeting. /cc @sameo @WeiZhang555 @laijs @gnawux @jodh-intel @amshinde @devimc  WDYT? Another option is to implement it in the client side, which does not require global lock nor global mapping file, but it has a drawback that cli need to maintain some container-related information.

sboeuf: Backed by @grahamwhaley test results, this change introduces more improvements than it introduces latency.
Also, implementing this at the CLI level does not mean we don't need any locking, the CLI could be called concurrently in order to start several containers on different pods. And in this case you still need a lock to make sure two different `kata-runtime` processes are not trying to write to this file at the same time.

bergwolf: Surely this is much faster than current implementation... I was saying it is slower and more complex than being implemented in the cli. And its performance decreases linearly with the number of containers on the host.

> implementing this at the CLI level does not mean we don't need any locking

The containerID uniqueness can be ensured by virtcontainers. Then kata cli only need to record the containerID->sandboxID mapping, which can be done on a per container basis (e.g., using a file like `runtime-root/containerID/sandboxID`). So there is no global mapping file, thus no global disk file locking either.

sboeuf: Any implementation will obviously be slower as the number of containers increases.

> The containerID uniqueness can be ensured by virtcontainers

I don't want this since I don't expect this enforcement from virtcontainers. I still want virtcontainers to be able to run 3 different pods with each having some containers with the same IDs. I would expect that Frakti could leverage that for instance. The uniqueness of containerID is really an OCI/CLI requirement only.

> Then kata cli only need to record the containerID->sandboxID mapping, which can be done on a per container basis (e.g., using a file like runtime-root/containerID/sandboxID). So there is no global mapping file, thus no global disk file locking either.

So you're proposing that we're going to create a new file for each new container, which puts more management complexity into the CLI, and that's what I don't really like. But maybe other can give some input here ?

bergwolf: > Any implementation will obviously be slower as the number of containers increases.

If you do not maintain a global mapping file, you do not suffer from this.

>  I still want virtcontainers to be able to run 3 different pods with each having some containers with the same IDs.

I expect the same from virtcontainers, -- sandboxID is globally unique, and containerID is unique inside one sandbox.

However for kata cli, we can differentiated docker containers from CRI containers with container annotations.
- For docker containers, containerID uniqueness == sandboxID uniqueness because we always use containerID as sandboxID for docker containers.
- For CRI containers, the uniqueness is ensured by virtcontainers.

So in summary we do not have to check containerID in CLI.

sboeuf: @grahamwhaley @jodh-intel @egernst @amshinde WDYT ? The proposal from @bergwolf would work but I have a concern about letting the CLI handling some file creation/removal, this would kind of duplicate the file management between the CLI and virtcontainers.

sboeuf:
@sameo @WeiZhang555 @laijs @gnawux @jodh-intel @amshinde @devimc @grahamwhaley @egernst
It'd be nice to have your input on this, in order to choose which way we want to go.

Maybe as a tradeoff, and since this PR is already greatly improving our perfs, we could use it as an intermediate step. Leaving the room for more discussion and more improvement into a follow up PR.

virtcontainers/api.go
sameo: @bergwolf @sboeuf What about supporting several entries per `runtime-root/containerID/` from virtcontainers instead? If there are several of them, then we fallback to the current implementation. Multiple `sanboxID` per `runtime-root/containerID/` should be a very rare occurrence. And since we can safely assume that there will not be concurrent write access to any file under `runtime-root/containerID/` there would not be any need for file locking. Concurrency should only happen at the `runtime-root/containerID/sandboxID` read level.

sboeuf: No need to fallback, implementing the same solution inside virtcontainers would work very well. There is only one case that is bothering me, how can we delete `runtime-root/containerID` directory safely ? Because at the time we'll do the delete, we never know that another pod could create the same containerID.

laijs:
Please ensure every runtime process touches/access the less sandboxes/containers as possible.
When the runtime process is crashed/killed, the fewer sandboxes/containers are affected the better.
(e.g. the runtime process is crashed when updating the map)

Adding a simple symlink located in the container OCI state dir and pointed to the sandbox dir can fix the problem.

virtcontainers/api.go
amshinde: I think you can handle this. If you delete a delete a sandbox directory and the check the upper {cid} directory is empty at that point, simply delete it. If another pod has created a directory between the time you delete and check, you are good, you do not delete {cid}.

sboeuf:
@bergwolf @egernst @grahamwhaley @sameo 
Ok I think I have spent enough time thinking about this, and I have taken the decision to implement this into the CLI. The reasons:
- The CLI implementation is going to be very simple since we can assume the uniqueness of the container ID, leading to a lock free solution.
- The VC implementation cannot be lock free. At some point, and because we try to support generic use case from VC, we will have to cover a concurrent case (deletion of the `containerID` directory while another sandbox with similar containerID could be created).
- The VC implementation would need to be generic but we will never use it for the generic case (several sandboxes with same container IDs).

grahamwhaley:
That theory sounds sound to me - some good points there.
I wonder, should we document in VC that you will need to supply your own locking under certain circumstances?

sboeuf:
@grahamwhaley well I don't see the need to document anything here since I am gonna implement this into the CLI. This means virtcontainers will not be aware about this.
Did I miss your point ?

sboeuf:
Here you go, #256 implements from the CLI side !

virtcontainers/filesystem.go
```diff
@@ -617,6 +655,35 @@ func (fs *filesystem) fetchAgentState(sandboxID string, state interface{}) error
...
+		}
+		f.Close()
+
+		m := make(map[string][]string)
+		if err := fs.storeContainersMap(m); err != nil {
+			return err
+		}
+	}
+
+	fileData, err := ioutil.ReadFile(filePath)
```
jodh-intel: This code implicitly only handles the current on-disk format of the containers map. We need to either:

- Add a statement in the code explaining that.

- Add something like a `version` field to the on-disk JSON format and have the code deal with any "older" versions it finds.

virtcontainers/container.go
```diff
@@ -551,7 +558,18 @@ func (c *Container) delete() error {
...
 	}
 
-	return c.sandbox.storage.deleteContainerResources(c.sandboxID, c.id, nil)
+	if err := c.sandbox.storage.deleteContainerResources(c.sandboxID, c.id, nil); err != nil {
+		return err
+	}
+
+	// Update containers map if stateful flag.
+	if !c.sandbox.config.Stateless {
+		if err := delFromContainersMap(c.id, c.sandbox.id); err != nil {
```
jodh-intel: The ordering looks wrong here as we seem to have this:

- `create()`
  - create container on-disk state.
  - add container to map.

- `delete()`
  - remove container on-disk state.
  - remove container from map.

Whereas I'd expect:

- `delete()`
  - remove container from map.
  - remove container on-disk state.

jodh-intel:
I like the idea, but I'm not convinced this is the way to go. I can see the point of adding the `Stateless` option but rather than that, how about we consider changing https://github.com/kata-containers/documentation/blob/master/design/kata-api-design.md#metadata-storage-plugin slightly so that we add a `AddToMap()` and `RemoveFromMap()` interface? That way the caller could just provide a nop interface if they don't want virtcontainers to create a map.

grahamwhaley:
From a high level view I like that idea @jodh-intel , as that makes the developer choice a deliberate and conscious one, and avoids any possible 'bad default' if nobody implements any locking anywhere when they might need it.

bergwolf:
If we decide to go with VC implementation, please change the implementation to a lock free and global map file free version as @sameo and @sboeuf suggested in https://github.com/kata-containers/runtime/pull/213/files#r183570992 and  https://github.com/kata-containers/runtime/pull/256#issuecomment-384393993

sboeuf:
#256 has been merged, we can close this PR now.

Issue #215, opened 2018-04-16
Remove redundant OCI "config" data from sandbox config and container config


WeiZhang555:
# Description of problem

Now under `/var/lib/virtcontainers/sandboxes/<sandbox-id>/`， there is a `config.json` and a `<container-id>/config.json`, these two files both contain an annotation named "com.github.katacontainers.runtime.virtcontainers.pkg.oci.config" whose value is equal to container's OCI spec.
This is totally unnecessary because we already store the OCI bundle path into the annotation, such as
"Annotations": {
        "com.github.katacontainers.runtime.virtcontainers.pkg.oci.bundle_path": "/run/docker/libcontainerd/21d0970bcf1dda614abec8ea7363e010d81d0ad08dc5a909ee0e993cb2da2422"
}

We can easily get config data from `bundlePath`/config.json, and save the data space from persist data file.

# Expected result

Remove unnecessary config json data from sandbox and container annotations.

# Got result

`config.json`:

```
{
    "AgentConfig": {
        "GRPCSocket": ""
    },
    "AgentType": "kata",
    "Annotations": {
        "com.github.containers.virtcontainers.pkg.oci.bundle_path": "/run/docker/libcontainerd/6e8e5f156f42c531deedf23a6256392068a053869990f032567ce4414c343346",
[...]

PR #216, opened 2018-04-16
Remove ConfigJSONKey annotation

WeiZhang555:
Currently we already saved "BundlePath" in sandbox and config's config, with
`BundlePath`, we can easily get config of json format from "BundlePath"/config.json,
So ConfigJSONKey is a duplicate and unnecessary field and we can remove it for
saving more space from persist data.

Fixes: #215

Signed-off-by: Zhang Wei <zhangwei555@huawei.com>


WeiZhang555:
CI failed.
Adding a [WIP] for fixing the CI

sboeuf:
Thanks @WeiZhang555 I'll take a look

virtcontainers/kata_agent.go
```diff
@@ -962,3 +963,22 @@ func (k *kataAgent) sendReq(request interface{}) (interface{}, error) {
...
 	}
 }
+
+// ParseConfigJSON unmarshals the config.json file.
+// TODO: this is duplicate with virtcontainers/pkg/oci/utilsgo:ParseConfigJSON
+// but we can't use that function because there will be cycle dependency problem
+// better find a way to eliminate duplicate codes @weizhang555
+func ParseConfigJSON(bundlePath string) (*specs.Spec, error) {
+	configPath := filepath.Join(bundlePath, "config.json")
+	configByte, err := ioutil.ReadFile(configPath)
```
bergwolf: With this, we are pushing the cli dependence on `config.json` file to virtcontainers. If we are certain that the content of OCI annotations is exactly the same as `config.json`, and assuming the purpose of the PR is to reduce persistent data, can we still let the cli read `config.json` and pass its contents to virtcontainers, and skip persisting the OCI spec in virtcontainers?

WeiZhang555: Then it will need to filter out "configJSON" content overtime we try to persist the data, it could be ugly codes and everywhere.

Are you worrying about the operation performance? I think it will influence a little on performance because it will need to open files more times, but the influence won't be much in my opinion. Though I don't have a test data. 😄

bergwolf: @WeiZhang555 I'm thinking about the following things:
1. We do not change the image bundle's config.json in hyperd for a container one. I think we can deal with it if required.
2. This requires `config.json` to be exactly as what's in the annotations. Are we certain about it?
3. If we ever change the OCI spec in virtcontainers, do we update the `config.json` file?

WeiZhang555: I didn't fully get your concerns. But in theory, I think it's better virtcontainers do never change `config.json` file because it's an input from upper level component or user, and we can add custom or additional modifications into virtcontainer's persist data but not original bundle config.json file.
Is this what you want to discuss?

bergwolf: Yes, that's it. We do change the OCI spec in certain cases (e.g., #138) and we do need to persist the new OCI spec in such case to allow future fetchSandbox() caller to get the same spec view of the container. I agree with you that we should **NOT** change the contents of `config.json`. So it means we need to find a way to persist optional modification to the spec and merge with `config.json` when reading it. If you do not want to handle it in this PR, we need to file an issue to make sure it gets implemented later.

WeiZhang555: @bergwolf I didn't find why we need to change OCI spec in #138, I think we can support `-v /dev/block-dev:xxx` with standard OCI spec. 
Can you point me the dedicated discussion about this? Or some more description?

bergwolf: @WeiZhang555 e.g. https://github.com/kata-containers/runtime/pull/138/files#diff-7ba6ce1174401126aaba6a1cff25434dR559

WeiZhang555: @bergwolf This is a good point!  I didn't realize we have such logic.

But even in previous code, after we modified the spec, the new spec data isn't persisted or saved back to annotation, so the good point is this PR didn't influence any of the function.

bergwolf: @WeiZhang555 I think we need to fix it before releasing. It is quite easy to fix before this PR. If we merge the PR, it won't be straightforward to fix any more.

WeiZhang555: Do you mean we need to persist the modified spec?

I mean the modified spec is calculated everytime you send a request to kata-agent, so in current logic, there's nothing need to be fixed.

bergwolf: No, it's only calculated when we create a new container (look for callers of `replaceOCIMountsForStorages`). Later API calls only read the OCI spec from persistent data and that is not the same one we have sent to the agent.

WeiZhang555: Here I'm still confused on what to fix.

Here when we modified the OCI spec, we want to send a temporary modified spec to kata-agent, but we don't want to persist the modified ones. So bundle/config.json will not be modified in any cases, and reading from bundle/config.json should be right logic as this PR is doing. Right?

So what should I fix here?

bergwolf: >  but we don't want to persist the modified ones

That's what I wanted to fix. If we decide not to persist the modified spec, that's a design decision and we need to be really careful handling it because we are not recording what's sent to the agent and other virtcontainers API calls are getting different OCI spec than the one sent to the agent.

cli/tags
```diff
@@ -0,0 +1,1172 @@
+!_TAG_FILE_FORMAT	2
```
bergwolf: btw, this file seems to be a mistake.

virtcontainers/kata_agent.go
bergwolf: Fair enough. `createContainer()` in kata_agent.go seems to be the only user of the OCI spec in virtcontainers. I might just be over worried.

Another thing is container hooks. If we read the spec so late, how do we handle pre-start hooks?

WeiZhang555: @bergwolf for container hooks, it will be stored inside SandboxConfig from OCI spec, so when the sandbox runs the hooks, it doesn't need to read OCI spec again.

https://github.com/kata-containers/runtime/blob/master/virtcontainers/pkg/oci/utils.go#L494

bergwolf: oops, sorry I missed it. We are reading the OCI spec file twice, one in cli.Create and the other in virtcontainers.createContainer.

sboeuf:
-1

WeiZhang555:
@sboeuf I understand your concerns. 

> . Your PR introduces more accesses to the filesystem and more Unmarshalling functions than the code before, 

The only concern for me is it truly introduced more accesses to the filesystem, but it didn't introduce more Unmarshalling functions, because in original code, the spec data is stored in annotation as string, you still need to unmarshal it to specs.Spec in every time you want to use it.

If you check the sandbox and container's config in virtcontainers, you can see the spec data is stored in 3 part, that means the content in config.json is duplicated and we save extra 3 **exactly same** data in 3 different positions. That's why I want to remove this because it's not really a necessary part.

I think I will find a better way to reduce the filesystem access, maybe at least merge three part of config.json into one.

sboeuf:
@WeiZhang555 Thanks. I am always down for some code optimizations :)

PR #218, opened 2018-04-16
api: add fetchsandbox and release sandbox

bergwolf:
`FetchSandbox()` allows a caller to retrieve sandbox with its ID. `sandbox.Release()` allows a caller to actively drop the runtime-agent connection. A new config option `LongLiveConn` is added to kata agent's config. When set, the connection to agent is kept across multiple calls to the agent and caller is expected to call `sandbox.Release()` to explicitly close the connection.


bergwolf:
@kata-containers/runtime PTAL

bergwolf:
Jenkins ubuntu 16-04 failed on kata-containers/tests/issues/238, a ci configuration issue, unrelated to the PR.

virtcontainers/agent.go
```diff
@@ -144,6 +144,9 @@ type agent interface {
 	// supported by the agent.
 	capabilities() capabilities
 
+	// disconnect will disconnect the connection to the agent
+	disconnect()
```
jodh-intel: I don't understand why this cannot be:

```
disconnect() error
```

bergwolf: I was thinking that we cannot do more recovery if disconnect() ever fails. I can add it back if you want though.

bergwolf:
CI green. ping @sboeuf @jodh-intel @WeiZhang555 @laijs @amshinde

virtcontainers/agent.go
amshinde: @bergwolf Will be useful to return the error to atleast log it even if it we cannot recover.

virtcontainers/kata_agent.go
```diff
@@ -73,16 +74,17 @@ func (s *kataVSOCK) String() string {
 // KataAgentState is the structure describing the data stored from this
 // agent implementation.
 type KataAgentState struct {
-	ProxyPid     int
-	ProxyBuiltIn bool
```
amshinde: I see that  you moved the ProxyBuiltIn from KataAgentState to kataAgent. Can you provide an explanation in the commit message, maybe move this to a separate commit.

virtcontainers/interfaces.go
```diff
@@ -20,17 +20,23 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// VC is the Virtcontainers interface
-type VC interface {
+// VCManager is the Virtcontainers sandbox management interface
+type VCManager interface {
```
sboeuf: Why do we need this new interface ? Is there a benefit ? I am asking cause otherwise I would prefer keep things simple when not needed.

virtcontainers/kata_agent_test.go
```diff
@@ -112,9 +112,7 @@ func TestKataAgentDisconnect(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if err := k.disconnect(); err != nil {
-		t.Fatal(err)
-	}
+	k.disconnect()
```
sboeuf: Why did you remove the error check here ?

virtcontainers/kata_agent.go
```diff
@@ -32,8 +32,10 @@ import (
 	vcAnnotations "github.com/kata-containers/runtime/virtcontainers/pkg/annotations"
 	ns "github.com/kata-containers/runtime/virtcontainers/pkg/nsenter"
 	"github.com/kata-containers/runtime/virtcontainers/pkg/uuid"
+
```
sboeuf: No need for a blank line here.

virtcontainers/pkg/vcmock/types.go
```diff
@@ -48,6 +48,7 @@ type VCMock struct {
 	CreateSandboxFunc func(sandboxConfig vc.SandboxConfig) (vc.VCSandbox, error)
 	DeleteSandboxFunc func(sandboxID string) (vc.VCSandbox, error)
 	ListSandboxFunc   func() ([]vc.SandboxStatus, error)
+	FetchSandboxFunc  func(sandboxID string) (vc.VCSandbox, error)
```
sboeuf: This should be part of the previous commit.

virtcontainers/pkg/vcmock/mock_test.go
```diff
@@ -577,3 +577,31 @@ func TestVCMockProcessListContainer(t *testing.T) {
 	assert.Error(err)
 	assert.True(IsMockError(err))
 }
+
+func TestVCMockFetchSandbox(t *testing.T) {
```
sboeuf: This should be part of the previous commit.

virtcontainers/interfaces.go
bergwolf: It is not strictly needed but it make the interface cleaner and we do want a clean interface IMHO.

bergwolf: The separation also makes it clear that the rest of the APIs in `VC` interface will be wrappers of `VCSandbox`.

sboeuf: Well I don't understand the clear separation here. Why do we have `StartSandbox()` into the `VCManager` and `StopSandbox()` into the `VC` interface then ?

bergwolf: Because we didn't define VCSandbox.Stop() :(

I'll drop this commit to move it out of the API change's way.

bergwolf:
Jenkins failed on network issues and the rebuild button landed on a 504 page.

grahamwhaley:
Rebuild worked for me - I've kicked one off for you (is in the queue right now it seems)

grahamwhaley:
that was on 16.04 btw - if that works then we can kick 17.10 as well if needed etc.

bergwolf:
Thanks @grahamwhaley !

bergwolf:
cc @jodh-intel @WeiZhang555 @laijs @amshinde @miaoyq

laijs:
LGTM
with :s/p *Sandbox/s *Sandbox/g

virtcontainers/sandbox.go
```diff
@@ -946,6 +946,28 @@ func (s *Sandbox) DeleteContainer(containerID string) (VCContainer, error) {
 	return c, nil
 }
 
+// StatusContainer gets the status of a container
+func (s *Sandbox) StatusContainer(containerID string) (ContainerStatus, error) {
```
sboeuf: This function should be smarter than this. It should be inspired from what is done in `virtcontainers/api.go`:
```go
func statusContainer(sandbox *Sandbox, containerID string) (ContainerStatus, error) {
```
The reason is that we need the status to be updated if virtcontainers realizes the shim has terminated. Otherwise, this assumes the caller will monitor the shim process in order to call `StopContainer()` when the shim terminates, but I don't think we want to rely on that.

virtcontainers/sandbox.go
```diff
@@ -535,6 +535,31 @@ func (s *Sandbox) Release() error {
 	return s.agent.disconnect()
 }
 
+// Status gets the status of the sandbox
+func (s *Sandbox) Status() SandboxStatus {
+	var contStatusList []ContainerStatus
+	for _, c := range s.containers {
```
sboeuf: The loop getting all the container statuses should leverage the function:
```go
func (s *Sandbox) StatusContainer(containerID string) (ContainerStatus, error)
```
from the previous commit. This would ensure that we also get some updated status for each container that needs some update.
Also this would change slightly the current function prototype, by adding an error return:
```go
func (s *Sandbox) Status() (SandboxStatus, error) {
```

virtcontainers/sandbox.go
bergwolf: @sboeuf I'm afraid the shim pid is not enough for representing container status because there can be no shim at all in the kata_builtin_shim case, which is the main usage of frakti along with this API. I'd like to handle it properly in a separate PR since it's an existing issue `StatusContainer()` as well in api.go. WDYT?

virtcontainers/sandbox.go
bergwolf: For the same reason as `StatusContainer()`, I want to fix it properly in a separate PR.

virtcontainers/sandbox.go
sboeuf: @bergwolf fair enough since this won't break the existing API. In the meantime (until you submit this follow up PR), could you add some warning comments on top of this function to explain that it should not be used as it might return the wrong state.

virtcontainers/sandbox.go
sboeuf: Same here, please add a temporary comment ;)

virtcontainers/sandbox.go
bergwolf: @sboeuf how about tracking it in a separate issue (kata-containers/runtime/issues/253) instead? An outstanding issue works better at tracking than comments in the code. Besides, currently the CI system is broken. I don't want to wait another day just for CI to pass. I'm having other PRs piling up on top of this one.

bergwolf: updated. Thanks @sboeuf !

PR #220, opened 2018-04-16
Handle device nodes and regular files in /dev

### LINKED ISSUE #219: We should not be bind-mounting any files under /dev, /sys and /proc from the host inside the container
These files belong to the host OS. Essentially need to revert this commit : 08909b2

amshinde:
This reverts commit 08909b2213a23ab00a442a505f4b1a146e34b41b first.

It removes /dev from the system ignored mounts retaining /sys and /proc.
Regular files in /dev are treated separately, they are shared with container in the guest using 9p. Device nodes in /dev are ignored. One caveat is we handle block device files early on, these are passed to the guest using virtio-blk.

Fixes #219

### DIFF
--- a/virtcontainers/mount.go
+++ b/virtcontainers/mount.go
@@ -18,6 +18,18 @@ import (
 
 var rootfsDir = "rootfs"
 
+var systemMountPrefixes = []string{"/proc", "/dev", "/sys"}
+
+func isSystemMount(m string) bool {
+	for _, p := range systemMountPrefixes {
+		if m == p || strings.HasPrefix(m, p+"/") {
+			return true
+		}
+	}
+
+	return false
+}
+
 func major(dev uint64) int {
 	return int((dev >> 8) & 0xfff)
 }
--- a/virtcontainers/mount.go
+++ b/virtcontainers/mount.go
@@ -18,7 +18,7 @@ import (
 
 var rootfsDir = "rootfs"
 
-var systemMountPrefixes = []string{"/proc", "/dev", "/sys"}
+var systemMountPrefixes = []string{"/proc", "/sys"}
 
 func isSystemMount(m string) bool {
 	for _, p := range systemMountPrefixes {
--- a/virtcontainers/mount.go
+++ b/virtcontainers/mount.go
@@ -30,6 +30,34 @@ func isSystemMount(m string) bool {
 	return false
 }
 
+func isHostDevice(m string) bool {
+	if m == "/dev" {
+		return true
+	}
+
+	if strings.HasPrefix(m, "/dev/") {
+		// Check if regular file
+		s, err := os.Stat(m)
+
+		// This should not happen. In case file does not exist let the
+		// error be handled by the agent, simply return false here.
+		if err != nil {
+			return false
+		}
+
+		if s.Mode().IsRegular() {
+			return false
+		}
+
+		// This is not a regular file in /dev. It is either a
+		// device file, directory or any other special file which is
+		// specific to the host system.
+		return true
+	}
+
+	return false
+}
+
 func major(dev uint64) int {
 	return int((dev >> 8) & 0xfff)
 }
--- a/virtcontainers/mount_test.go
+++ b/virtcontainers/mount_test.go
@@ -18,6 +18,30 @@ import (
 	"testing"
 )
 
+func TestIsSystemMount(t *testing.T) {
+	tests := []struct {
+		mnt      string
+		expected bool
+	}{
+		{"/sys", true},
+		{"/sys/", true},
+		{"/sys//", true},
+		{"/sys/fs", true},
+		{"/sys/fs/", true},
+		{"/sys/fs/cgroup", true},
+		{"/sysfoo", false},
+		{"/home", false},
+		{"/dev/block/", true},
+	}
+
+	for _, test := range tests {
+		result := isSystemMount(test.mnt)
+		if result != test.expected {
+			t.Fatalf("Expected result for path %s : %v, got %v", test.mnt, test.expected, result)
+		}
+	}
+}
+
 func TestMajorMinorNumber(t *testing.T) {
 	devices := []string{"/dev/zero", "/dev/net/tun"}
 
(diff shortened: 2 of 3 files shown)


amshinde:
cc @sboeuf @egernst @bergwolf

virtcontainers/mount.go
```diff
@@ -29,6 +29,18 @@ import (
 
 var rootfsDir = "rootfs"
 
+var systemMountPrefixes = []string{"/proc", "/dev", "/sys"}
```
sboeuf: We should only constraint this list to the devices.

amshinde: Why just /dev? We cannot share /proc and /sys between the guest and host as well.

sboeuf: Yes sure, but those are already handled by the fact that we won't try to bind mount them.
I guess either way, that's fine. But I want to see the CI results first

amshinde: So this is the error message I am seeing 
```
Apr 16 21:02:22 jenkins-azure-vm-ubuntu-16-04-templatee71490 crio[68573]: time="2018-04-16 21:02:22.686454249Z" level=error msg="Container creation error: rpc error: code = Internal desc = Could not run process: container_linux.go:348: starting container process caused "process_linux.go:402: container init caused \"rootfs_linux.go:58: mounting \\\"/var/lib/kubelet/pods/1f253791-41b9-11e8-b40a-000d3afd4c4f/containers/kubedns/fc4f90c8\\\" to rootfs \\\"/run/kata-containers/shared/containers/31d498c2c53faf8fe556eb7ebc30ce91ccedb9a423e71205808b88a425a9c217/rootfs\\\" at \\\"/dev/termination-log\\\" caused \\\"stat /var/lib/kubelet/pods/1f253791-41b9-11e8-b40a-000d3afd4c4f/containers/kubedns/fc4f90c8: no such file or directory\\\"\""
```

bergwolf: k8s requires the termination log and it by default resides under `/dev`.

amshinde: I do not like log files living under /dev. I think a way to handle this is not to ignore all files under /dev,  but ignore device files and pass regular files to the container as 9p shared mounts.
@bergwolf @sboeuf What do you think?

sboeuf: I agree with @bergwolf that we want to support it. That being said @bergwolf , how do you think we should handle the support of something like `-v /dev/gpus/:/dev/gpus/` where the directory `/dev/gpus/` contains a list of graphics devices ? Do you think we should parse the whole directory and add one by one every device (`card0`, `card1`, `card2`, ...) to the container ? And maybe also rely on a blacklist of paths that should not be passed from the host to the guest ?

amshinde:
@sboeuf I wanted to do things in order, apply the revert first rather than apply this as a fresh new commit just to keep the context in history.  Regarding why this is needed, I feel that we need to consider /proc and /sys in the ignored mounts.

sboeuf:
Yeah fair enough.

sboeuf:
LGTM

virtcontainers/mount.go
```diff
@@ -29,6 +29,47 @@ import (
...
+		s, err := os.Stat(m)
+
+		// This should not happen. In case file does not exist let the
+		// error be handled by the agent, simply return false here.
+		if err != nil {
+			return false
+		}
+
+		mode := s.Mode()
+		if mode.IsRegular() {
```
bergwolf: nits: please just use `s.Mode().IsRegular()`.

virtcontainers/mount.go
```diff
@@ -29,6 +29,47 @@ import (
...
+	}
+
+	if strings.HasPrefix(m, "/dev/") {
+		// Check if regular file
+		s, err := os.Stat(m)
+
+		// This should not happen. In case file does not exist let the
+		// error be handled by the agent, simply return false here.
+		if err != nil {
+			return false
```
devimc: can you log the error?

WeiZhang555:
One thing I want to mention, this will make https://github.com/kata-containers/runtime/pull/138 not work anymore.
And  #138 is written and upload by @amshinde (also author of this PR). 
So do you want to bring #138 back later? Actually I like feature from #138

virtcontainers/mount.go
caoruidong: Is it ok to modify a revert commit?

sboeuf: I agree the PR message should be modified

amshinde: @sboeuf @caoruidong Will change the PR message.

amshinde: Done

amshinde:
I am seeing CI failing:
```
INFO: Checking for SPDX license headers
ERROR: Required license identifier ('SPDX-License-Identifier: Apache-2.0') missing from following files:

cli/config/configuration.toml.in
```
@jodh-intel Should we be excluding this file from the SPDX check?

virtcontainers/mount.go
```diff
@@ -29,6 +29,46 @@ import (
...
+
+	return false
+}
+
+func isHostDevice(m string) bool {
+	if m == "/dev" {
+		return true
+	}
+
+	if strings.HasPrefix(m, "/dev/") {
```
WeiZhang555: I think we need to evaluate symbol link first, in case the source is a symlink and linked to a device

WeiZhang555:
@amshinde Oh, you're right, it's here: https://github.com/kata-containers/runtime/pull/220/files#diff-8a56f5a1fa5971cc43e650ffa24f1e2bL322

I missed it, sorry for didn't read the code carefully.

jodh-intel:
@amshinde - I've raised a PR to add the SPDX header to that file (https://github.com/kata-containers/runtime/pull/235).

virtcontainers/mount_test.go
```diff
@@ -29,6 +29,64 @@ import (
...
+	}{
+		{"/sys", true},
+		{"/sys/", true},
+		{"/sys//", true},
+		{"/sys/fs", true},
+		{"/sys/fs/", true},
+		{"/sys/fs/cgroup", true},
+		{"/sysfoo", false},
+		{"/home", false},
+		{"/dev/block/", false},
```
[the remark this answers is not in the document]
amshinde: done

amshinde:
@sboeuf can you merge

Issue #221, opened 2018-04-16
When adding "cpuset_cpus": "0" and "cpuset_mems": "0"  in the linux resources of the container.json, kata-runtime fails to launch a container.


chavafg:
When adding "cpuset_cpus": "0" and "cpuset_mems": "0"  in the linux resources of the container.json, kata-runtime fails to launch a container.

From kata-runtime log:

Apr 16 17:12:29 jenkins-azure-vm-ubuntu-16-04-template7a43d0 kata-runtime[86554]: time="2018-04-16T17:12:29.082328309Z" level=error msg="rpc error: code = Internal desc = Could not run process: container_linux.go:348: starting container process caused \"process_linux.go:402: container init caused \\\"process_linux.go:367: setting cgroup config for procHooks process caused \\\\\\\"cannot set limits on the cpuset cgroup, as the container has not joined it\\\\\\\"\\\"\"" command=create name=kata-runtime pid=86554 source=runtime

PR #222, opened 2018-04-16
virtcontainers: Disable cpuset and cpumem

chavafg:
Disable cpuset and cpumem constraints as this is not properly
supported yet.
    
If we add "cpuset_cpus" and "cpuset_mems" to the container.json,
kata-runtime failed to start, so we need to disable them.

Fixes: #221.


sboeuf:
@bergwolf PTAL !

bergwolf:
lgtm

Issue #225, opened 2018-04-17
SPDX: many cli files not using SPDX license header


grahamwhaley:
Many (almost all) of the go files in the cli subdir use the fairly full Apache header, whereas the project standard/default is to use the shortened SPDX style header.

PR #226, opened 2018-04-17
SPDX: update cli and arch files to use SPDX

grahamwhaley:
Many cli and arch files were using the 'older style' fairly full
Apache license text. The project standard is the shorter SPDX style.
Convert them over.

Fixes: #225

### DIFF
--- a/arch/amd64-options.mk
+++ b/arch/amd64-options.mk
@@ -1,16 +1,7 @@
 # Copyright (c) 2018 Intel Corporation
 #
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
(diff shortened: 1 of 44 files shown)


bergwolf:
@jodh-intel PTAL and merge if it's OK to you. Let's ignore the codecov/project error since it is just license change.

grahamwhaley:
Oh, yeah, I meant to mention the codecov. I don't really understand how it went down!
@jodh-intel, @sboeuf - do we know how reproducible the codecov results are (you would think very...?) - maybe we need to set a small 'acceptable change' level on them, if we can.

jodh-intel:
Looking at codecov is on my todo list, but alas it isn't near to top ;)

Let's merge this thing!

Issue #227, opened 2018-04-17
virtcontainers: move to using SPDX headers


grahamwhaley:
It looks like the whole virtcontainers codebase was imported with the older style Apache license headers in place, and not converted over to the project standard of SPDX.
With the new https://github.com/kata-containers/tests/pull/242 checking tool, any file touched in future will need to be updated. Better if we do a once-and-for-all sweep on the 80(+?) files that need an update maybe.

PR #231, opened 2018-04-18
license: SPDX: update all vc files to use SPDX style

grahamwhaley:
When imported, the vc files carried in the 'full style' apache
license text, but the standard for kata is to use SPDX style.
Update the relevant files to SPDX.

Fixes: #227

### DIFF
--- a/virtcontainers/agent.go
+++ b/virtcontainers/agent.go
@@ -1,17 +1,6 @@
-//
 // Copyright (c) 2016 Intel Corporation
 //
-// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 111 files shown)


grahamwhaley:
We'd like to land this ASAP, as due to PR https://github.com/kata-containers/tests/pull/242, all pending PRs that have touched a vc file with the old copyright will now fail the CI.
Even with this PR, those pending PRs will need rebasing and pushing, but that is easier than forcing the developers to now update the licenses only in the files they have touched etc.

Related, we should also try to land #226 at the same time.

PR #244, opened 2018-04-23
cli: Backtrace on SIGUSR1

### LINKED ISSUE #241: Dump stacks when received SIGUSR1 in Agent/Runtime/Shim/Proxy?
When we debugging some kind of deadlock or so issues in Docker/Containerd, sending SIGUSR1 to dockerd or containerd, and collect their goroutine stacks for further analysis are very helpful.

Should we implement the similar feature also in Agent/Runtime/Shim/Proxy in kata? 

Thanks guys :-)

jodh-intel:
Rework the signal handling code to backtrace on receipt of `SIGUSR1`
but continue to run.

Added some basic tests for the signal handling code.

Fixes #241.


jodh-intel:
Ping @bergwolf, @sboeuf.

cli/signals.go
```diff
@@ -54,13 +57,27 @@ func backtrace() {
 }
 
 func fatalSignal(sig syscall.Signal) bool {
-	return sigFatal[sig]
+	_, exists := handledSignalsMap[sig]
```
sboeuf: Why are you checking the same map `handledSignalsMap` from both `fatalSignal()` and `nonFatalSignal()`. Am I missing something ?

jodh-intel: That map contains all the signals that are handled. But `SIGUSR1` is handled differently - the bool map value is `false` meaning "don't die on receipt of this signal". The two functions you mention are similar but not that `nonFatalSignal()` is only dealing with `SIGUSR1` and hence has the `!handledSignalsMap[sig]` test at the end.

jodh-intel:
Updated branch to only backtrace on `SIGUSR1` when debug enabled at @sboeuf's suggestion.

cli/signals.go
sboeuf: Oh yes, I missed this slight difference !

cli/main.go
```diff
@@ -168,6 +170,9 @@ func setupSignalHandler() {
 			if fatalSignal(nativeSignal) {
 				kataLog.WithField("signal", sig).Error("received fatal signal")
 				die()
+			} else if debug && nonFatalSignal(nativeSignal) {
```
sboeuf: Just wondering if `nonFatalSignal()` is appropriate naming here, when the whole thing is only about handling a specific signal to generate a backtrace.

jodh-intel: I chose that name as I thought we might want to extend the list of signals at some future point. As the code is trying to hide the implementation details from `main()`, what would you suggest? We could call it something like `signalIsUSR1()` but that sort of breaks the info hiding ;)

sboeuf: Okay fair enough ;)

cli/signals.go
```diff
@@ -54,15 +57,28 @@ func backtrace() {
...
+	return handledSignalsMap[sig]
+}
+
+func nonFatalSignal(sig syscall.Signal) bool {
+	_, exists := handledSignalsMap[sig]
+	if !exists {
+		return false
+	}
+
+	return !handledSignalsMap[sig]
```
devimc: you already looked  for the signal, I think you could improve the performance doing this

```go
	fatalSignal, exists := handledSignalsMap[sig]
	if !exists {
		return false
	}

	return !faltalSignal
```

jodh-intel: Good call! Branch updated.

PR #252, opened 2018-04-24
API: support sandbox monitor operation

### LINKED ISSUE #251: API: support sandbox monitor operation
As described in [the API design](kata-containers/documentation/blob/master/design/kata-api-design.md), we need to support `Sandbox.Monitor()` operation that returns an error channel for the caller to watch on sandbox fetal errors.

bergwolf:
As described in [the API design](kata-containers/documentation/blob/master/design/kata-api-design.md), we need to support `Sandbox.Monitor()` operation that returns an error channel for the caller to watch on sandbox fetal errors.

Fixes #251

### DIFF
--- a/virtcontainers/agent.go
+++ b/virtcontainers/agent.go
@@ -133,6 +133,9 @@ type agent interface {
 	// supported by the agent.
 	capabilities() capabilities
 
+	// check will check the agent liveness
+	check() error
+
 	// disconnect will disconnect the connection to the agent
 	disconnect() error
 
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -15,6 +15,7 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"time"
 
 	kataclient "github.com/kata-containers/agent/protocols/client"
 	"github.com/kata-containers/agent/protocols/grpc"
@@ -934,6 +935,11 @@ func (k *kataAgent) disconnect() error {
 	return nil
 }
 
+func (k *kataAgent) check() error {
+	_, err := k.sendReq(&grpc.CheckRequest{})
+	return err
+}
+
 func (k *kataAgent) sendReq(request interface{}) (interface{}, error) {
 	if err := k.connect(); err != nil {
 		return nil, err
@@ -943,6 +949,11 @@ func (k *kataAgent) sendReq(request interface{}) (interface{}, error) {
 	}
 
 	switch req := request.(type) {
+	case *grpc.CheckRequest:
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		_, err := k.client.Check(ctx, req)
+		return nil, err
 	case *grpc.ExecProcessRequest:
 		_, err := k.client.ExecProcess(context.Background(), req)
 		return nil, err
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -17,6 +17,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/gogo/protobuf/proto"
 	kataclient "github.com/kata-containers/agent/protocols/client"
 	"github.com/kata-containers/agent/protocols/grpc"
 	vcAnnotations "github.com/kata-containers/runtime/virtcontainers/pkg/annotations"
@@ -74,6 +75,7 @@ type kataAgent struct {
 	shim         shim
 	proxy        proxy
 	client       *kataclient.AgentClient
+	reqHandlers  map[string]reqFunc
 	state        KataAgentState
 	keepConn     bool
 	proxyBuiltIn bool
@@ -917,6 +919,7 @@ func (k *kataAgent) connect() error {
 		return err
 	}
 
+	k.installReqFunc(client)
 	k.client = client
 
 	return nil
@@ -930,7 +933,9 @@ func (k *kataAgent) disconnect() error {
 	if err := k.client.Close(); err != nil && err != golangGrpc.ErrClientConnClosing {
 		return err
 	}
+
 	k.client = nil
+	k.reqHandlers = nil
 
 	return nil
 }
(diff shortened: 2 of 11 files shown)


bergwolf:
@sboeuf @amshinde @jodh-intel @laijs PTAL.

sboeuf:
@bergwolf I'll take a look at this tomorrow, could you please fix the CI.

virtcontainers/sandbox.go
```diff
@@ -561,6 +566,23 @@ func (s *Sandbox) Status() SandboxStatus {
 	}
 }
 
+// Monitor returns a error channel for watcher to watch at
+func (s *Sandbox) Monitor() (chan error, error) {
+	if s.state.State != StateRunning {
+		return nil, fmt.Errorf("Sandbox not running")
```
devimc: Sandbox **is** not running, might be?

virtcontainers/monitor.go
```diff
@@ -0,0 +1,106 @@
...
+	defer m.Unlock()
+
+	watcher := make(chan error, 1)
+	m.watchers = append(m.watchers, watcher)
+
+	if !m.running {
+		m.running = true
+
+		// create and start agent watcher
+		go func() {
```
devimc: I wonder if you should add this routine to a wait group ?

virtcontainers/sandbox.go
```diff
@@ -561,6 +566,23 @@ func (s *Sandbox) Status() SandboxStatus {
...
+	}
+
+	if s.monitor == nil {
+		s.Lock()
+		if s.monitor == nil {
```
sboeuf: Second time you check if `s.monitor` is nil, I think that's a bit too much :smiley:

virtcontainers/agent.go
```diff
@@ -146,27 +146,27 @@ type agent interface {
...
 
 	// createContainer will tell the agent to create a container related to a Sandbox.
 	createContainer(sandbox *Sandbox, c *Container) (*Process, error)
 
 	// startContainer will tell the agent to start a container related to a Sandbox.
-	startContainer(sandbox Sandbox, c *Container) error
+	startContainer(sandbox *Sandbox, c *Container) error
 
 	// stopContainer will tell the agent to stop a container related to a Sandbox.
-	stopContainer(sandbox Sandbox, c Container) error
```
sboeuf: The rationale behind using pointers versus copies was to make sure the sandbox could not get modified by some functions that obviously didn't need it.
Now I understand your concern about memory being overused because we're generating a lot of copies and I think I pretty okay with changing things into pointers. But please submit this into a separate PR as it is an important change and this way we can discuss properly about it. I don't want this to be hidden through the monitor PR.

virtcontainers/kata_agent.go
```diff
@@ -940,6 +945,50 @@ func (k *kataAgent) check() error {
 	return err
 }
 
+type reqFunc func(context.Context, interface{}, ...golangGrpc.CallOption) (interface{}, error)
+
+func (k *kataAgent) installReqFunc(c *kataclient.AgentClient) {
+	k.reqHandlers = make(map[string]reqFunc)
```
sboeuf: Yes, in this case, I think the map is the only way to avoid the gocyclo complexity going over 15.
Any reason you didn't use a static map here ?

virtcontainers/sandbox.go
```diff
@@ -561,6 +566,23 @@ func (s *Sandbox) Status() SandboxStatus {
...
+			s.monitor = newMonitor(s)
+		}
+		s.Unlock()
+	}
+
+	return s.monitor.newWatcher()
```
sboeuf: I am a bit concerned about the way this is handled here. The caller can basically call a million times into `Monitor()`, this will result into a million go routines being spawned here. If think we should try to handle things better, by returning the existing channel from `newWatcher()` if it already exists. This means we would always have only one channel, because I don't see the usage for having several channels referring to the same thing. Maybe I am missing something.

virtcontainers/agent.go
bergwolf: Without it, I got CI errors about copying mutex.

virtcontainers/monitor.go
bergwolf: Good point! I'll let stop() wait on it.

virtcontainers/sandbox.go
bergwolf: good point!

virtcontainers/sandbox.go
bergwolf: It ensures we only create and set the monitor once. Otherwise concurrent callers might race in and we end up setting monitor twice.

virtcontainers/sandbox.go
bergwolf: No, the monitor makes sure there is only one watcher goroutine per monitor. We just add more watch channels to it.

virtcontainers/kata_agent.go
bergwolf: Because the agent client is dynamically allocated. I also wanted to use a static map but it turns out there is no static grpc request handler.

virtcontainers/sandbox.go
bergwolf: Done

virtcontainers/monitor.go
bergwolf: Done.

virtcontainers/agent.go
sboeuf: Ok, but please submit it as a separate PR.

virtcontainers/sandbox.go
sboeuf: You can simply do the lock and test the monitor value. This is really cheap here.

virtcontainers/sandbox.go
sboeuf: Oh I see. So we'll have one go routine with an infinite number of channels. Sounds good.

virtcontainers/agent.go
bergwolf: Then this PR cannot pass CI.

virtcontainers/sandbox.go
bergwolf: Fair enough, I'll change it.

virtcontainers/agent.go
sboeuf: That's fine, we'll merge the new PR, so that you can rebase this one.

bergwolf: ok, here it is #263

sboeuf:
@bergwolf please rebase (and make sure it removes the last commit that we've already merged in a previous PR)

virtcontainers/sandbox.go
bergwolf: Done.

bergwolf:
PR rebased.

egernst:
@sboeuf  - PTAL and merge when you're happy with the changes.

Issue #254, opened 2018-04-24
change go version to 1.9.2 in versions.yaml 


chavafg:
Our https://github.com/kata-containers/runtime/blob/master/versions.yaml says that we support go1.10 to build our project, but we have static and unit tests issues, so we need to change the version to 1.9.2.

By doing this, we will be able to remove the hardcoded version from the jenkins scripts and instead install golang using `.ci/install_go.sh` from the tests repository. And when moving to go1.10 using a PR, the CI will test that the static checks and unit tests pass correctly.

PR #255, opened 2018-04-24
versions: change newest supported go version

chavafg:
change from go1.10 to 1.9.2.
Our static checks and unit tests fail when using
go 1.10. Since we use go 1.9.2 to test in our CI,
reflect this version in versions.yaml

By doing this, we will be able to remove the hardcoded version
from the jenkins scripts and instead install golang using
`.ci/install_go.sh` from the tests repository. And when moving
to go1.10 using a PR, the CI will test that the static checks
and unit tests pass correctly.

Fixes: #254.

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -139,7 +139,7 @@ languages:
     notes: "'version' is the default minimum version used by this project."
     version: "1.8.3"
     meta:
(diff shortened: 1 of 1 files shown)


jcvenegas:
lgtm

PR #256, opened 2018-04-24
cli: Optimize container research

### LINKED ISSUE #212: virtcontainers: Improve container searching

sboeuf:
This commit will allow for better performance regarding the time spent
to retrieve the sandbox ID related to a container ID.

The way it works is by relying on a specific mapping between container
IDs and sanbox IDs, meaning it allows to retrieve directly the sandbox
ID related to a container ID from the CLI. This lowers complexity from
O(n²) to O(1), because we don't need to call into ListPod() which was
parsing all the pods and all the containers on the system everytime
the CLI need to retrieve this mapping.

This commit also updates the whole unit tests as a consequence. This
is involving most of them since they were all relying on ListPod()
before.

Fixes #212

### DIFF
--- a/cli/create.go
+++ b/cli/create.go
@@ -239,6 +239,10 @@ func createSandbox(ociSpec oci.CompatOCISpec, runtimeConfig oci.RuntimeConfig,
 		return vc.Process{}, fmt.Errorf("BUG: Container list from sandbox is wrong, expecting only one container, found %d containers", len(containers))
 	}
 
+	if err := addContainerIDMapping(containerID, sandbox.ID()); err != nil {
+		return vc.Process{}, err
+	}
+
 	return containers[0].Process(), nil
 }
 
@@ -257,6 +261,10 @@ func createContainer(ociSpec oci.CompatOCISpec, containerID, bundlePath,
 
 	_, c, err := vci.CreateContainer(sandboxID, contConfig)
 	if err != nil {
+		return vc.Process{}, err
+	}
+
+	if err := addContainerIDMapping(containerID, sandboxID); err != nil {
 		return vc.Process{}, err
 	}
 
--- a/cli/oci.go
+++ b/cli/oci.go
@@ -9,6 +9,7 @@ import (
 	"bufio"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"os"
 	"path/filepath"
@@ -24,10 +25,11 @@ import (
 
 // Contants related to cgroup memory directory
 const (
-	cgroupsTasksFile = "tasks"
-	cgroupsProcsFile = "cgroup.procs"
-	cgroupsDirMode   = os.FileMode(0750)
-	cgroupsFileMode  = os.FileMode(0640)
+	cgroupsTasksFile   = "tasks"
+	cgroupsProcsFile   = "cgroup.procs"
+	cgroupsDirMode     = os.FileMode(0750)
+	cgroupsFileMode    = os.FileMode(0640)
+	ctrsMappingDirMode = os.FileMode(0750)
 
 	// Filesystem type corresponding to CGROUP_SUPER_MAGIC as listed
 	// here: http://man7.org/linux/man-pages/man2/statfs.2.html
@@ -40,6 +42,8 @@ var cgroupsDirPath string
 
 var procMountInfo = "/proc/self/mountinfo"
 
+var ctrsMapTreePath = "/var/run/kata-containers/containers-mapping"
+
 // getContainerInfo returns the container status and its sandbox ID.
 func getContainerInfo(containerID string) (vc.ContainerStatus, string, error) {
 	// container ID MUST be provided.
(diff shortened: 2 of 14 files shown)


sboeuf:
This PR is the same as https://github.com/kata-containers/runtime/pull/213, but it implements the container<==>sandbox mapping from the CLI.
@egernst @jodh-intel @bergwolf @grahamwhaley @amshinde PTAL

cli/oci.go
```diff
@@ -40,30 +42,32 @@ var cgroupsDirPath string
 
 var procMountInfo = "/proc/self/mountinfo"
 
+var ctrsMapTreePath = "/var/run/kata-containers/containers-mapping"
```
bergwolf: Please put it in the runtime root storage directory instead of a hard coded one.

sboeuf: As agreed, this should be done in a follow up PR introducing the support of `kata-runtime --root`

jodh-intel:
Although this is now mergeable, can we finish the discussion we're having on #213 as both these PRs are trying to solve the same problem?

Marking as dnm for now...

jodh-intel:
@sboeuf - please can you change the commit message to explain *how* you've optimised the search ;)

sboeuf:
@jodh-intel commit message updated.

cli/oci.go
```diff
@@ -339,3 +343,63 @@ func getCgroupsDirPath(mountInfoFile string) (string, error) {
...
+// ID directory. If there are several files, we could not determine which
+// file name corresponds to the sandbox ID associated, and this would throw
+// an error.
+func fetchContainerIDMapping(containerID string) (string, error) {
+	if containerID == "" {
+		return "", fmt.Errorf("Missing container ID")
+	}
+
+	dirPath := filepath.Join(ctrsMapTreePath, containerID)
+	if _, err := os.Stat(dirPath); os.IsNotExist(err) {
```
amshinde: You can get rid of this check. The ReadDir call below will return an error if the directory does not exist.

cli/create.go
```diff
@@ -239,6 +239,10 @@ func createSandbox(ociSpec oci.CompatOCISpec, runtimeConfig oci.RuntimeConfig,
 		return vc.Process{}, fmt.Errorf("BUG: Container list from sandbox is wrong, expecting only one container, found %d containers", len(containers))
 	}
 
+	if err := addContainerIDMapping(containerID, sandbox.ID()); err != nil {
```
amshinde: I think we talked that this could be done in virtcontainers itself. Any reason this is being done in the CLI?

sboeuf: Well I was about to say that we were still needing a lock, but now I am confused. I think this could work:
- `fetchSandboxIDs(containerID)` returns a list of what is inside the directory `/.../containerID/`. No lock needed since the call into `readdir()` is atomic.
- `addSandboxID(containerID, sandboxID)` would add a new sandbox using `os,MkdirAll()`, meaning the path `/.../containerID/sandboxID` would be created, no matter if the parent directory `/.../containerID` exist or not.
- `delSandboxID(container, sandboxID)` would remove the path `/.../containerID/sandboxID` and after, it would call into `os.Rmdir()` `/.../containerID` without checking the error. Indeed, the call to rmdir might fail if something is still in the directory, and also if a `addSandboxID()` has just been called.

sboeuf: Ok so from our offline discussion, `os.MkdirAll()` is obviously not atomic, and that's the reason why we would need some locking. But given the fact that this case would never happen since we will not use it for real, we will never hit it. Let me ask others about what they think.

sboeuf:
@bergwolf @jodh-intel @egernst There is one alternative solution which consists in not being completely bug free from a virtcontainers perspective (a potential container directory creation could happen at the same time another sandboxID is removed, and only in case the container directory becomes empty), but assuming that the corner case has a very low probability in theory, and given that it won't happen at all for us since only the CLI will use it.
I don't know if that's something we can consider in order to make sure we let this implementation inside virtcontainers, and this way we don't spread the storage.
So to summarize, we have 0% chance of hitting this issue, but from virtcontainers perspective, this is not completely safe. We might want to add a retry for the `addSandboxID()` in order to reduce further the chance of failure.

amshinde:
@sboeuf Like I mentioned in your other PR for CLI, I think this could well be implemented inside virtcontainers itself, ie virtcontainers handling the creation/deletion of  sandox entries inside a container directory.

There is a slight chance of a race when the sandbox directory entry is created during container creation since this involves creating/checking the container directory followed by creating the sandbox directory itself(/run/virtcontainers/{cid}/{sandbox_id}). The container directory may be deleted by another process between the two steps. Assuming that we run into this issue, we can retry creating the sandbox entry a few times. (Assuming processes context switch between the two steps exactly !)

We may run into this issue based on the supposition that containers may end up having the same container-id. Now, our cli and docker itself makes sure that container-ids are unique. Since CRIO relies on OCI, container-ids would be unique in CRIO case as well.
@bergwolf I think this is true for frakti as well?

So I think we could just go with the simpler approach and avoid all the locking.

bergwolf:
@amshinde frakti would rely on virtcontainers to ensure the uniqueness of sandboxID globally, and the uniqueness of containerID within a sandbox. For kata cli, it verifies the uniqueness of contaienrID itself but it is really unnecessary.

@sboeuf I'm fine with implementing it inside virtcontainers. I just want to make sure that we use a lockless solution and do not rely on a global map file.

sboeuf:
@bergwolf ok then I'll wait for @jodh-intel @grahamwhaley and @egernst input on that.

sboeuf:
@egernst @bergwolf could you please merge it ?

cli/oci.go
bergwolf: Yes, indeed.

sboeuf:
@bergwolf comment addressed.

PR #259, opened 2018-04-25
add sandbox process operation relay API support

bergwolf:
As described in [the API design](kata-containers/documentation/blob/master/design/kata-api-design.md), we need to support `sandbox.WaitProcess()`, `sandbox.WinsizeProcess()`, `sandbox.SignalProcess()` and `sandbox.IOStream()` operations.

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -8,6 +8,7 @@ package virtcontainers
 import (
 	"encoding/hex"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"syscall"
@@ -756,6 +757,16 @@ func (c *Container) winsizeProcess(processID string, height, width uint32) error
 	}
 
 	return c.sandbox.agent.winsizeProcess(c, processID, height, width)
+}
+
+func (c *Container) ioStream(processID string) (io.WriteCloser, io.Reader, io.Reader, error) {
+	if c.state.State != StateReady && c.state.State != StateRunning {
+		return nil, nil, nil, fmt.Errorf("Container not ready or running, impossible to signal the container")
+	}
+
+	stream := newIOStream(c.sandbox, c, processID)
+
+	return stream.stdin(), stream.stdout(), stream.stderr(), nil
 }
 
 func (c *Container) processList(options ProcessListOptions) (ProcessList, error) {
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -663,7 +663,7 @@ func (c *Container) stop() error {
 	// return an error, but instead try to kill it forcefully.
 	if err := waitForShim(c.process.Pid); err != nil {
 		// Force the container to be killed.
-		if err := c.sandbox.agent.killContainer(c.sandbox, *c, syscall.SIGKILL, true); err != nil {
+		if err := c.kill(syscall.SIGKILL, true); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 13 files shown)


sboeuf:
Just adding do-not-merge for now, since we need to get #252 in first.

sboeuf:
https://github.com/kata-containers/runtime/pull/252 got merged, please rebase.

bergwolf:
PR rebased. PTAL.

bergwolf:
CI green. @sboeuf @egernst @amshinde PTAL.

virtcontainers/hyperstart_agent.go
```diff
@@ -806,3 +806,33 @@ func (h *hyper) check() error {
 	// cc-agent does not support check
 	return nil
 }
+
+func (h *hyper) waitProcess(c *Container, processID string) (int32, error) {
+	// cc-agent does not support wait process
```
egernst: s/cc-agent/hyperstart-agent?

virtcontainers/hyperstart_agent.go
```diff
@@ -806,3 +806,33 @@ func (h *hyper) check() error {
...
+	return nil
+}
+
+func (h *hyper) readProcessStdout(c *Container, processID string, data []byte) (int, error) {
+	// cc-agent does not support stdout read request
+	return 0, nil
+}
+
+func (h *hyper) readProcessStderr(c *Container, processID string, data []byte) (int, error) {
+	// cc-agent does not support stderr read request
```
egernst: s/cc-agent/hyperstart-agent?
 same for all the other comments above...

virtcontainers/container.go
```diff
@@ -652,7 +652,7 @@ func (c *Container) stop() error {
 	// return an error, but instead try to kill it forcefully.
 	if err := waitForShim(c.process.Pid); err != nil {
 		// Force the container to be killed.
-		if err := c.sandbox.agent.killContainer(c.sandbox, *c, syscall.SIGKILL, true); err != nil {
+		if err := c.kill(syscall.SIGKILL, true); err != nil {
```
amshinde: Why not just call into c.signalProcess here. `c.kill` api is not really required.

bergwolf: It's a helper function for cases like this when we want to send signals to the container process itself. And it hides the detail that we use `c.process.Token` as process ID of the container process, useful because it's also called by `KillContainer()` in api.go.

virtcontainers/hyperstart_agent.go
bergwolf: Done.

virtcontainers/hyperstart_agent.go
bergwolf: Done.

egernst:
@amshinde - PTAL at the update and merge if you're okay with the explanation provided?

virtcontainers/iostream.go
```diff
@@ -0,0 +1,91 @@
...
+
+package virtcontainers
+
+import (
+	"errors"
+	"io"
+)
+
+type iostream struct {
+	sandbox   *Sandbox
```
WeiZhang555: I think an iostream is only bounded to a container, so we don't need the sandbox here, and sandbox can be fetched from container.sandbox if needed.
We have lots of redundant struct fields in many places, maybe we can rework them later.

WeiZhang555:
LGTM

laijs:
LGTM

Issue #262, opened 2018-04-26
virtcontainers: always pass sandbox as pointers


bergwolf:
We have a lot of copies of sandbox around and it waste time and memory to do so. Instead, we can always pass it as pointers and it also ensures we have a consistent view of sandbox everywhere.

PR #263, opened 2018-04-26
virtcontainers: always pass sandbox as a pointer

bergwolf:
Currently we sometimes pass it as a pointer and other times not. As
a result, the view of sandbox across virtcontainers may not be the same
and it costs extra memory copy each time we pass it by value. Fix it
by ensuring sandbox is always passed by pointers.

Fixes: #262

Signed-off-by: Peng Tao <bergwolf@gmail.com>

### DIFF
--- a/virtcontainers/cc_proxy.go
+++ b/virtcontainers/cc_proxy.go
@@ -13,7 +13,7 @@ type ccProxy struct {
 }
 
 // start is the proxy start implementation for ccProxy.
-func (p *ccProxy) start(sandbox Sandbox, params proxyParams) (int, string, error) {
+func (p *ccProxy) start(sandbox *Sandbox, params proxyParams) (int, string, error) {
 	config, err := newProxyConfig(sandbox.config)
 	if err != nil {
 		return -1, "", err
@@ -38,6 +38,6 @@ func (p *ccProxy) start(sandbox Sandbox, params proxyParams) (int, string, error
 	return cmd.Process.Pid, proxyURL, nil
 }
 
-func (p *ccProxy) stop(sandbox Sandbox, pid int) error {
+func (p *ccProxy) stop(sandbox *Sandbox, pid int) error {
 	return nil
 }
--- a/virtcontainers/agent.go
+++ b/virtcontainers/agent.go
@@ -143,27 +143,27 @@ type agent interface {
 	exec(sandbox *Sandbox, c Container, cmd Cmd) (*Process, error)
 
 	// startSandbox will tell the agent to start all containers related to the Sandbox.
-	startSandbox(sandbox Sandbox) error
+	startSandbox(sandbox *Sandbox) error
 
 	// stopSandbox will tell the agent to stop all containers related to the Sandbox.
-	stopSandbox(sandbox Sandbox) error
+	stopSandbox(sandbox *Sandbox) error
 
 	// createContainer will tell the agent to create a container related to a Sandbox.
 	createContainer(sandbox *Sandbox, c *Container) (*Process, error)
 
 	// startContainer will tell the agent to start a container related to a Sandbox.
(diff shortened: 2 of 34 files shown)


virtcontainers/cc_proxy.go
```diff
@@ -13,7 +13,7 @@ type ccProxy struct {
 }
 
 // start is the proxy start implementation for ccProxy.
-func (p *ccProxy) start(sandbox Sandbox, params proxyParams) (int, string, error) {
+func (p *ccProxy) start(sandbox *Sandbox, params proxyParams) (int, string, error) {
```
jodh-intel: Don't we now need check on all these params now they are pointers?:

```
var errNeedSandbox = errors.New("Sandbox cannot be empty")

    :

func (p *ccProxy) start(sandbox *Sandbox, params proxyParams) (int, string, error) {
    if sandbox == nil {
        return -1, "", errNeedSandbox
    }

    :
}

```

bergwolf: These are all internal functions. I think we should trust internal callers because otherwise we do not trust ourselves. That said, even if we do not trust internal callers, the check should be added as a separate PR since it is an existing issue for places where sandbox is already passed as pointer.

bergwolf:
Jenkins failed on kata-containers/tests#264

jcvenegas:
Please push again, now the test is skipped.

bergwolf:
CI green. @sboeuf @egernst  @jodh-intel @laijs PTAL.

virtcontainers/cc_proxy.go
bergwolf: @jodh-intel Do you want to check every parameter in every function? Since that's what _not trust anything_ means to me. From my point of view, I think we should check all parameters in all of the virtcontainers APIs but that's all. We are already doing this and it ensures that we fail fast.

OTOH, if we do not trust ourselves and check parameter in every function, we waste CPU time and memory on checking the same thing over and over again. Please create an issue for it if you think that's really necessary. I don't mind having such a discussion but it does not need to block this PR since it's a different/existing issue.

bergwolf: A few examples of where `*Sandbox` is not checked before being dereferenced in master:
- https://github.com/kata-containers/runtime/blob/master/virtcontainers/kata_agent.go#L132
- https://github.com/kata-containers/runtime/blob/master/virtcontainers/kata_agent.go#L183
- https://github.com/kata-containers/runtime/blob/master/virtcontainers/shim.go#L175
- https://github.com/kata-containers/runtime/blob/master/virtcontainers/container.go#L404

There a more of them for sure. And if we want to take precaution in all these functions, it should also apply to other parameters as well, not just the sandbox pointers.

bergwolf:
@jodh-intel can we merge this?

sboeuf:
@bergwolf I know you need this PR to be merged in order to move forward with https://github.com/kata-containers/runtime/pull/252.
But we should definitely follow @jodh-intel comments and add some checks to the codebase. The introduction of pointers has to be backed up by some better testing of the sandbox pointers.

Also, I know, as you mentioned it, that we might not check it properly in functions where it is already a pointer, that's why I propose that you can open an issue tracking this as a global thing to address through the entire codebase, marking this as high priority.

In the meantime, I am okay to merge this one, but I'd like @egernst and @amshinde feedbacks on this.

BTW, you need to rebase it before we can merge it.

bergwolf:
@jodh-intel @egernst @sboeuf PR rebased and #280 is created to follow up the sanity check discussion.

grahamwhaley:
@egernst @sboeuf @bergwolf  - this is GREEN now - if we are not wanting to merge right now, please put a DNM on it - or somebody is going to press that button....

egernst:
Too late :)

Issue #257, opened 2018-04-25
--root= does nothing


sboeuf:
From @jodh-intel:
_The runtime supports the --root= global option but, although it is mentioned in a few strings, this option does nothing._

_Further, the build is setting the value of defaultRootDirectory to $(PKGRUNDIR) which is also incorrect given that container data is stored by virtcontainers, not the runtime._

sboeuf:
Related to #22

jodh-intel:
If we want this option to be truly honoured, we'll clearly need to add a new virtcontainers API to set it's `configStoragePath` / `runStoragePath` (`fileystem.go`).

PR #200, opened 2018-04-11
runtime: rename pod to sandbox

bergwolf:
As agreed in [the kata containers API design](https://github.com/kata-containers/documentation/blob/master/design/kata-api-design.md), we need to rename pod notion to sandbox. The patch is a bit big but the actual change is done through the script:
```
sed -i -e 's/pod/sandbox/g' -e 's/Pod/Sandbox/g' -e 's/POD/SB/g'
```

Fixes: #199

Signed-off-by: Peng Tao <bergwolf@gmail.com>

PR #269, opened 2018-04-27
socket: Enforce socket length

### LINKED ISSUE #268: enforce socket length
Only some parts of the virtcontainers code enforce unix domain socket length.

jodh-intel:
A Unix domain socket is limited to 107 usable bytes on Linux. However,
not all code creating socket paths was checking for this limits.

Created a new `buildSocketPath()` function (with tests) to encapsulate
the logic and updated all code creating sockets to use it.

Fixes #268.

### DIFF
--- a/virtcontainers/hyperstart_agent.go
+++ b/virtcontainers/hyperstart_agent.go
@@ -739,12 +739,17 @@ func (h *hyper) register() error {
 	}
 	defer h.disconnect()
 
+	console, err := h.sandbox.hypervisor.getSandboxConsole(h.sandbox.id)
+	if err != nil {
+		return err
+	}
+
 	registerVMOptions := &proxyClient.RegisterVMOptions{
-		Console:      h.sandbox.hypervisor.getSandboxConsole(h.sandbox.id),
+		Console:      console,
 		NumIOStreams: 0,
 	}
 
-	_, err := h.client.RegisterVM(h.sandbox.id, h.sockets[0].HostPath,
+	_, err = h.client.RegisterVM(h.sandbox.id, h.sockets[0].HostPath,
 		h.sockets[1].HostPath, registerVMOptions)
 	return err
 }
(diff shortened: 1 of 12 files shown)


jodh-intel:
This PR raises a few issues:

- We're already almost at the `107`-byte path limit.
  - #200 made the problem worse by increasing the `storagePathSuffix` path by 5 bytes (`len("sandboxes")-len("pods")`)
  - We should consider changing `storagePathSuffix` from `"/virtcontainers/sandboxes"` to `"/kata/pods"`, saving `15` bytes.
- The changes to `qmpSocketPath()` are frankly gross but I've attempted to re-implement what was there before. The logic is indicative of the lack of remaining bytes available for the files we need to create.
- We should have found this bug earlier (aka we need more unit tests in virtcontainers).

Related: #257.

virtcontainers/utils.go
```diff
@@ -7,15 +7,22 @@ package virtcontainers
...
 )
 
 const cpBinaryName = "cp"
 
 const fileMode0755 = os.FileMode(0755)
 
+// Effective maximum Unix domain socket length
+//
+// See unix(7).
+const maxSocketPathLen = 107
```
grahamwhaley: I wish there was some go system level const for this eh?
Found a maybe relevant ticket https://github.com/golang/go/issues/6895 - and, looks like @jessfraz was around this area back in 2015 maybe - so maybe can give some insight/input.

jodh-intel:
The tests are failing, but atleast the code is working! :smile:

```
--- FAIL: TestCreateSandboxKataAgentSuccessful (0.00s)
	api_test.go:334: path too long (got 109, max 107): /tmp/virtcontainers-tmp-708439037/virtcontainers/sandboxes/run/7f49d00d-1995-4156-8c79-5f5ab24ce138/kata.sock

```

grahamwhaley:
Heh heh @jodh-intel 
So, looking at that path, the most obvious fix that springs to my mind is s/virtcontainers/vc/ - which would claim us back 24 chars or so I reckon...

grahamwhaley:
@jodh-intel - if the tests are failing due to the new check, does that mean we cannot merge this until we have fixed some other code (to reduce the path length)? If so, which I think, then I'm going to mark this as DNM to make it clear we are waiting for something else. Please do revert that if I've mis-understood...

sboeuf:
Leaving the decision to @jodh-intel to merge it, but completely fine with me ;)

WeiZhang555:
Codes LGTM. But need rebase :-)

jodh-intel:
Branch rebased. I've raised #290 to reduce the path lengths.

PR #270, opened 2018-04-27
cli: implement update command

### LINKED ISSUE #189: Implement the update command
*Originally raised as https://github.com/clearcontainers/runtime/issues/380*.

devimc:
Update command is used to update container's resources at run time.
All constraints are applied inside the VM to each container cgroup.
By now only CPU constraints are fully supported, vCPU are hot added
or removed depending of the new constraint.
    
fixes #189


devimc:
@kata-containers/runtime please take a look

virtcontainers/api.go
```diff
@@ -600,3 +601,34 @@ func ProcessListContainer(sandboxID, containerID string, options ProcessListOpti
...
+	}
+	defer unlockSandbox(lockFile)
+
+	p, err := fetchSandbox(sandboxID)
+	if err != nil {
+		return err
+	}
+
+	// Fetch the container.
+	c, err := p.findContainer(containerID)
```
bergwolf: Please make UpdateContainer part of the VCSandbox interface and implement `UpdateContainer()` here as a wrapper of it. You can move `p.findContainer()` and `c.update()` to `sandbox.UpdateContainer()`.

cli/update.go
```diff
@@ -0,0 +1,258 @@
...
+    "quota": 0,
+    "period": 0,
+    "realtimeRuntime": 0,
+    "realtimePeriod": 0,
+    "cpus": "",
+    "mems": ""
+  },
+  "blockIO": {
+    "weight": 0
+  }
```
bergwolf: missing pids?

devimc: I don't think so, it's base on runC implementation

virtcontainers/api.go
devimc: will do :)

cli/update.go
bergwolf: It's a bug in runc? `--pids-limit` and `--l3-cache-schema` were added after the initial `runc update` implementation and it seems the developer forgot to add them to the resources file.

bergwolf: ref opencontainers/runc/issues/1793

bergwolf: We do read pid limit at https://github.com/kata-containers/runtime/pull/270/files#diff-1cc31db47bce8b7af5fda9d984c048e6R162. So I think we should at least add pids here. l3-cache-schema may be left out since it is not part of `specs.LinuxResources` anyway.

devimc: good catch!

devimc:
@bergwolf changes applied, thanks

virtcontainers/hyperstart_agent.go
```diff
@@ -605,6 +606,11 @@ func (h *hyper) processListContainer(sandbox *Sandbox, c Container, options Proc
 	return h.processListOneContainer(sandbox.id, c.id, options)
 }
 
+func (h *hyper) updateContainer(sandbox *Sandbox, c Container, resources specs.LinuxResources) error {
+	// cc-agent does not support update
```
WeiZhang555: I think this should be hyperstart_agent ?

virtcontainers/interfaces.go
```diff
@@ -34,6 +35,7 @@ type VC interface {
 	StatusContainer(sandboxID, containerID string) (ContainerStatus, error)
 	StopContainer(sandboxID, containerID string) (VCContainer, error)
 	ProcessListContainer(sandboxID, containerID string, options ProcessListOptions) (ProcessList, error)
+	UpdateContainer(sandboxID, containerID string, resources specs.LinuxResources) error
```
WeiZhang555: Why do we need this API? I think `sandbox.UpdateContainer` is enough for us.

virtcontainers/hyperstart_agent.go
devimc: I think cc-agent is correct, see https://github.com/kata-containers/runtime/blob/master/virtcontainers/hyperstart_agent.go#L800-L808

virtcontainers/interfaces.go
devimc: see cli/update_test.go

devimc:
@egernst changes applied, thanks

virtcontainers/hyperstart_agent.go
WeiZhang555: Hmmm, so weird, this file is written for hyperstart agent but the comment is saying cc-agent.

I'm confused a bit, is this correct? @bergwolf

virtcontainers/interfaces.go
WeiZhang555: It's OK for now. I think we have lots of exported functions just used for test, hope we can find a way to remove them and keep the test cases at the same time later.
It's another topic though. This part looks good for now.

virtcontainers/hyperstart_agent.go
WeiZhang555: Oh, I believed this is a typo now. Check: https://github.com/kata-containers/runtime/pull/259/files#diff-d4339baeea8e93444b3ddc3e97b53c87R806

It's already been fixed in #259 now

WeiZhang555:
Codes LGTM. Only a typo left s/cc-agent/hyperstart-agent/

So anyone please feel free to merge this after the typo is fixed.

devimc:
@WeiZhang555 changes applied, thanks

egernst:
@devimc please rebase and lets move forward to merge after...

PR #275, opened 2018-04-28
virtcontainers: Properly remove the container when shim gets killed

### LINKED ISSUE #274: Support proper removal of the container when the shim gets killed
Here is an interesting case I have been debugging. I was trying to understand why a `kubeadm reset` was not working for the case of Kata. In this case, the only pod started with Kata is the `kube-dns` pod. For some reasons, when this pod is stopped and removed, first its containers receive some signals. 2 of them receive some `SIGTERM` signals, which seems the way to properly stop them, but the third container receives a `SIGCONT`. Obviously, nothing happens in this case, but apparently CRI-O considers this should be the end of the container and after a few seconds, it kills the container process (being the shim in Kata case). Because it is using a SIGKILL, the signal does not get transfered to the agent because the shim itself is killed right away.
After this happened, CRI-O calls into `kata-runtime state`, we detect the shim is not running anymore and we try to stop the container. The code will eventually call into `agent.RemoveContainer()`, but this will fail and return an error because inside the agent, the container is still running !
Now here are the different approach we could consider:
- Enable a watcher in the shim, as a child process of the shim itself. The discussion had been already opened before, but the problem is that starting a child is almost like duplicating the footprint of the process. This is a real issue given the fact that we have a shim written in Go.
The solution here is to write a shim in C, and the problem of duplicating will become negligible. But we're talking mid term here since the shim will unlikely be changed in one or two weeks.
- From the agent code, force a container to be killed if this container is still running when the `RemoveContainer()` function gets called. The issue with this solution is that we're handling this case only from the Kata agent, where this issue should probably be handled from virtcontainers so that it would be reused across agent implementations.
- From virtcontainers, from `(c *Container) stop() error` in `container.go`, after the shim has been waited, we could always throw a SIGKILL without checking for the error returned. Most of the cases will end up returning an error because the shim itself not being there actually represents the container inside the VM has terminated already. And in case the shim has been killed without the possibility to forward the signal, the SIGKILL will work and will allow the following call to `agent.stopContainer()` to proceed on the removal of the container inside the agent. This is the easiest solution for now.
- Now, an evolution of the previous solution adding a call to `killContainer(SIGKILL)` would be to make such a call only if the container state is running. But this would involve providing a new agent protocol function to the API, something like `ContainerStatus()`. This would be called from `container.go` to check if the shim being stopped is matching the status of the container inside the VM. In case it does, nothing needs to be done, but in case it does not, then we should force a call into `killContainer(SIGKILL)`. This is an evolution of the previous solution preventing from calling dumbly into `killContainer(SIGKILL)` everytime we remove a container, but instead only when it would be needed.

@egernst @bergwolf @jodh-intel @amshinde @grahamwhaley please let me know what you think make more sense here. I can implement the easiest solution for virtcontainers in one line of code, and I don't see drawbacks to it.

sboeuf:
Here is an interesting case I have been debugging. I was trying to
understand why a "kubeadm reset" was not working for kata-runtime
compared to runc. In this case, the only pod started with Kata is
the kube-dns pod. For some reasons, when this pod is stopped and
removed, its containers receive some signals, 2 of them being SIGTERM
signals, which seems the way to properly stop them, but the third
container receives a SIGCONT. Obviously, nothing happens in this
case, but apparently CRI-O considers this should be the end of the
container and after a few seconds, it kills the container process
(being the shim in Kata case). Because it is using a SIGKILL, the
signal does not get forwarded to the agent because the shim itself
is killed right away. After this happened, CRI-O calls into
"kata-runtime state", we detect the shim is not running anymore
and we try to stop the container. The code will eventually call
into agent.RemoveContainer(), but this will fail and return an
error because inside the agent, the container is still running.

The approach to solve this issue here is to send a SIGKILL signal
to the container after the shim has been waited for. This call does
not check for the error returned because most of the cases, regular
use cases, will end up returning an error because the shim itself
not being there actually represents the container inside the VM has
already terminated.
And in case the shim has been killed without the possibility to
forward the signal (like described in first paragraph), the SIGKILL
will work and will allow the following call to agent.stopContainer()
to proceed to the removal of the container inside the agent.

Fixes #274

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -665,6 +665,16 @@ func (c *Container) stop() error {
 		}
 	}
 
+	// Force the container to be killed. For most of the cases, this
+	// should not matter and it should return an error that will be
+	// ignored.
+	// But for the specific case where the shim has been SIGKILL'ed,
+	// the container is still running inside the VM. And this is why
+	// this signal will ensure the container will get killed to match
+	// the state of the shim. This will allow the following call to
+	// stopContainer() to succeed in such particular case.
+	c.sandbox.agent.killContainer(*(c.sandbox), *c, syscall.SIGKILL, true)
+
 	if err := c.sandbox.agent.stopContainer(*(c.sandbox), *c); err != nil {
 		return err
 	}


sboeuf:
@egernst @jodh-intel @amshinde @bergwolf PTAL

sboeuf:
@chavafg if you try this out, you should be able to properly reset a k8s cluster with `kubeadm reset`. Given that you have your agent patched with https://github.com/kata-containers/agent/pull/225.

grahamwhaley:
@sboeuf Did you figure out exactly why we see the SIGCONT - and what actually happens in the `runc` case? I didn't see an explanation of if that flow is something different, and for good reason, and if we are exactly emulating that flow or just working around something?
This old k8s Issue maybe looks relevant? https://github.com/kubernetes/kubernetes/issues/33050

I just want to make sure we understand what we are working around and why, and not making a fix that may not hold if other k8s things change in the future etc.

sboeuf:
@grahamwhaley I 100% agree we should also investigate why the SIGCONT does not work. But the issue highlighted because the SIGCONT didn't terminate the container is also a valid issue, solved by this PR and explained by #274.
A separate issue for SIGCONT seems appropriate in this case, don't you think ?

grahamwhaley:
Agreed @sboeuf - if we can open an Issue to note we should figure out the oddity of workflow with the `kube-dns` pod and signals, that would be great.

sboeuf:
@grahamwhaley I have just opened https://github.com/kata-containers/runtime/issues/276 to track this down !

chavafg:
I am late on this, but yes it works, now the jenkins jobs don't have the errors when running the `kubeadm reset`. Thanks @sboeuf

sboeuf:
@chavafg thanks for confirmation!

Issue #283, opened 2018-05-02
make install does not (re)build


stefanha:
# Description of problem

"make install" fails on a clean working directory:

```
$ make install
install: cannot stat ‘data/kata-collect-data.sh’: No such file or directory
```

This happens because install and install-scripts do not depend on the
runtime.  Make doesn't know it needs to build the runtime before it can
be installed.

# Expected result

"make install" builds the runtime and then installs the files successfully.

# Actual result

install: cannot stat ‘data/kata-collect-data.sh’: No such file or directory

PR #284, opened 2018-05-02
Makefile: add missing dependencies to install targets

stefanha:
"make install" fails on a clean working directory:

  $ make install
  install: cannot stat ‘data/kata-collect-data.sh’: No such file or directory

This happens because install and install-scripts do not depend on the
runtime.  Make doesn't know it needs to build the runtime before it can
be installed.

Add the missing dependencies to the install targets so that "make
install" works on a clean working directory and rebuilds when source
files have been modified.

Note that SCRIPTS contains the generated kata-collect-data.sh script.
That file needs to be generated before it can be installed, so make
SCRIPTS a dependency of install-scripts.

Fixes: #283
Signed-off-by: Stefan Hajnoczi <stefanha@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -377,11 +377,11 @@ check-go-static:
 coverage:
 	$(QUIET_TEST).ci/go-test.sh html-coverage
 
-install: default install-scripts
+install: default runtime install-scripts
 	$(QUIET_INST)install -D $(TARGET) $(DESTTARGET)
 	$(QUIET_INST)install -D $(CONFIG) $(DESTCONFIG)
 
-install-scripts:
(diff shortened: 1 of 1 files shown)


sboeuf:
@stefanha thanks for the PR. I have restarted the fedora build, we got a glitch here. I'll merge when the CI turn green.

bergwolf:
CI failed crio redis-server test:
```
time="2018-05-02T16:44:27Z" level=fatal msg="execing command in container failed: unable to upgrade connection: 404 page not found"
```

bergwolf:
lgtm

PR #214, opened 2018-04-16
[WIP] virtcontainers: support network hotplug

miaoyq:
At present, only implement the `add network`, the function entry is:
```
func addNetworkToPod(podID,  ifName string) error
```
The function accepts two parameters: pod ID and net interface name. Other network information will be obtained by virtcontainers according to `ifName`.

This is only the initial completion of the basic code flow, but has not yet been functionally tested.
/cc @sboeuf 

Signed-off-by: Yanqiang Miao <miao.yanqiang@zte.com.cn>

PR #287, opened 2018-05-07
api: add sandbox hotplug network 

### LINKED ISSUE #113: network configuration hotplug support
Continuation of discussion from https://github.com/containers/virtcontainers/issues/241

cc/ @mcastelino @amshinde @sboeuf @bergwolf @laijs @guangxuli @miaoyq

caoruidong:
Fixes #113
Refactor generate interface and route, add network hotplug interface


caoruidong:
This pr is mainly based on #214. It is a draft and not fully tested.There is still a lot of work to do.
1. Make a pr to govmm
2. Better to write a new function to tell the type of interface. But it may call vhostUserSocketPath one more time.
3. Add unit tests
4. What is the scope of network hotplug? sandbox.AddNetwork(), ListNetwork, UpdateNetwork and UpdateRoute? Which interface do we use to pull out a nic?

caoruidong:
Seems that "switch devType { }" appears many time. Should I call the new interface (s *Sandbox) HotplugAddDevice or hypervisor.hotplugDevice ?

virtcontainers/kata_agent.go
```diff
@@ -453,36 +471,11 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
 	if err != nil {
 		return err
 	}
-	for _, ifc := range interfaces {
-		// send update interface request
```
caoruidong: Why not send add interface request here?

virtcontainers/sandbox.go
```diff
@@ -921,6 +925,166 @@ func (s *Sandbox) removeNetwork() error {
...
+				return err
+			}
+		}
+	}
+	return fmt.Errorf("update failed:interface %s not found", ifName)
+}
+
+//UpdateRoute updates the sandbox route table (e.g. for portmapping support).
+func (s *Sandbox) UpdateRoute(routes []*grpc.Route) error {
+	if err := s.agent.plugRoutes(routes); err != nil {
```
caoruidong: Will agent's route table inconsistent with that in sandbox storage?

caoruidong:
@bergwolf @amshinde PTAL too.

egernst:
@sboeuf  - can you PTAL?

bergwolf:
Generally looks good. Can you please add cli side change so that these APIs are actually used and tested?

virtcontainers/agent.go
```diff
@@ -194,4 +195,13 @@ type agent interface {
 	// This function should be called after hot adding vCPUs or Memory.
 	// cpus specifies the number of CPUs that were added and the agent should online
 	onlineCPUMem(cpus uint32) error
+
+	// hotPlugNetwork will tell the agent to add a new nic for an existed Sandbox.
```
devimc: s/hotPlugNetwork/plugRoutes

virtcontainers/agent.go
```diff
@@ -194,4 +195,13 @@ type agent interface {
...
+	plugRoutes(routes []*grpc.Route) error
+
+	// hotPlugNetwork will tell the agent to add a new nic for an existed Sandbox.
+	hotPlugNetwork(sandbox *Sandbox, endpoint Endpoint) error
+
+	// hotPlugNetwork will tell the agent to del a new nic for an existed Sandbox.
```
devimc: s/hotPullNetwork/hotPlugNetwork

caoruidong: Thx. Fixed

caoruidong:
@bergwolf OK, I'm working on it.

katabuilder:
PSS Measurement:
Qemu: 143935 KB
Proxy: 6723 KB
Shim: 11105 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  2011588 KB

katabuilder:
PSS Measurement:
Qemu: 144111 KB
Proxy: 4714 KB
Shim: 12932 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  2014752 KB

virtcontainers/agent.go
```diff
@@ -197,4 +198,13 @@ type agent interface {
 
 	// statsContainer will tell the agent to get stats from a container related to a Sandbox
 	statsContainer(sandbox *Sandbox, c Container) (*ContainerStats, error)
+
+	// plugRoutes will tell the agent to update route table for an existed Sandbox.
+	plugRoutes(routes []*grpc.Route) error
```
sboeuf: Please replace with:
```go
updateRoutes(routes []netlink.Route) error
```
The name will be more appropriate, and the parameter will not be tied to our gRPC protocol.

virtcontainers/api.go
```diff
@@ -653,3 +653,23 @@ func StatsContainer(sandboxID, containerID string) (ContainerStats, error) {
...
 	return s.StatsContainer(containerID)
 }
+
+// AddNetwork is the virtcontainers entry point to add a new nic.
+func AddNetwork(sandboxID, ifName string) error {
+	if sandboxID == "" {
+		return errNeedSandboxID
+	}
+
+	lockFile, err := rLockSandbox(sandboxID)
```
sboeuf: This should be a RW lock as I expect this function to store the newly added interface at some point.

virtcontainers/agent.go
```diff
@@ -197,4 +198,13 @@ type agent interface {
...
+
+	// hotPlugNetwork will tell the agent to add a new nic for an existed Sandbox.
+	hotPlugNetwork(sandbox *Sandbox, endpoint Endpoint) error
```
sboeuf: Please replace with:
```go
addNetwork(sandbox *Sandbox, endpoint Endpoint) error
```

virtcontainers/agent.go
```diff
@@ -197,4 +198,13 @@ type agent interface {
...
+
+	// hotPullNetwork will tell the agent to del a new nic for an existed Sandbox.
+	hotPullNetwork(sandbox *Sandbox, endpoint Endpoint) error
```
sboeuf: Please replace with:
```go
delNetwork(sandbox *Sandbox, endpoint Endpoint) error
```

virtcontainers/api.go
```diff
@@ -653,3 +653,23 @@ func StatsContainer(sandboxID, containerID string) (ContainerStats, error) {
...
+	}
+	defer unlockSandbox(lockFile)
+
+	s, err := fetchSandbox(sandboxID)
+	if err != nil {
+		return err
+	}
+
+	return s.AddNetwork(ifName)
+}
```
sboeuf: I guess you will complete this by providing the implementation for `UpdateRoutes()`, `ListNetwork()`, and `UpdateNetwork()` ?

cli/interface.go
```diff
@@ -0,0 +1,86 @@
...
+	},
+	Action: func(context *cli.Context) error {
+		return cli.ShowSubcommandHelp(context)
+	},
+}
+
+var infAddCommand = cli.Command{
+	Name:      "add",
+	Usage:     "add a new nic into a container",
+	ArgsUsage: "<container-id> <interface-name>",
```
sboeuf: We need to discuss how we see this working. Do we expect the caller to provide the whole interface (that's the declarative model we've been talking about), or do we simply provide an interface name and we expect virtcontainers to scan the network namespace to get all the information about this interface ?

virtcontainers/network.go
```diff
@@ -154,6 +154,8 @@ type Endpoint interface {
 	SetProperties(NetworkInfo)
 	Attach(hypervisor) error
 	Detach() error
+	HotAttach(hypervisor) error
+	HotDetach(hypervisor) error
```
sboeuf: I don't think we need such a thing, or maybe we do as a temporary thing, but eventually, we will get rid of the network cold plug, and this will be hotplugged by default.

cli/interface.go
```diff
@@ -0,0 +1,86 @@
...
+
+		tw.Flush()
+		return nil
+	},
+}
+
+var infUpdateCommand = cli.Command{
+	Name:      "update",
+	Usage:     "update configuration of interface",
+	ArgsUsage: `update <container-id> <interface-name>`,
```
sboeuf: Same question here, do we expect to provide a simple interface name or a full description ?

caoruidong:
@sboeuf Thanks a lot for your detailed review, I'll update the pr. I am confused by the CLI arguments too. Because the API design doc just lists the name of API, not the args.  Maybe we can discuss the CLI commands in a new issue.

sboeuf:
@caoruidong I think https://github.com/kata-containers/runtime/issues/113 is appropriate to discuss the expectations regarding the parameters of the API/CLI.

katabuilder:
PSS Measurement:
Qemu: 161465 KB
Proxy: 6636 KB
Shim: 10742 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1996252 KB

virtcontainers/sandbox.go
```diff
@@ -925,6 +929,172 @@ func (s *Sandbox) removeNetwork() error {
 	return nil
 }
 
+// AddNetwork adds new nic to the sandbox.
+func (s *Sandbox) AddNetwork(ifName string) error {
```
bergwolf: I would prefer the API to be declarative so that the interface is configured explicitly at the API level, e.g., to let caller to set bridge, ip, mac, mtu, gw etc through the API, instead of letting virtcontainers discover what's there through netns. This also aligns with `UpdateRoute()`. @caoruidong @sboeuf @WeiZhang555 WDYT?

vendor/github.com/intel/govmm/qemu/qmp.go
```diff
@@ -727,6 +727,47 @@ func (q *QMP) ExecuteBlockdevDel(ctx context.Context, blockdevID string) error {
 	return q.executeCommand(ctx, "x-blockdev-del", args, nil)
 }
 
+func (q *QMP) ExecuteNetdevAdd(ctx context.Context, netdevID, ifname, downscript, script string) error {
```
WeiZhang555: This part should firstly upstream to github.com/intel/govmm, and re-vendor back to kata-containers after it's merged.
And it's better to put vendor commit in seperate commit.

virtcontainers/agent.go
caoruidong: Done. But seems that we will lost grpc.Route.Device info cause it comes from endpoint.Name()

virtcontainers/network.go
caoruidong: Now Attach() calls hypervisor.addDevice() which just manipulates qemu launch command args. So at this time we may need a HotAttach() interface. I agree with you eventually they will unified. Maybe it's better to refactor in another later PR？

virtcontainers/sandbox.go
caoruidong: Agree. I will change my code.

vendor/github.com/intel/govmm/qemu/qmp.go
caoruidong: Yes. If the QMP part is correct and tests are passed, I will make a PR to govmm, then re-vendor and rebase this PR.

sboeuf:
@caoruidong what is the status on this PR ?
Have you removed a bunch of code here ? Please make sure you have proper commit message so that it is clear what is done in each of them.

caoruidong:
@sboeuf Yes, I’m rewriting this PR. Sorry to mix it with another branch in my repo. I’ll clean up commits later.

sboeuf:
@caoruidong no problem ;)

amshinde:
Looks like this PR needs to be updated. @caoruidong Let us know when this is ready for review.

caoruidong:
@amshinde Sure.

caoruidong:
Meet a problem. When update interfaces, we may delete eth0. I want to delete the device added by qemu command like "-device driver=virtio-net-pci,netdev=network-0,mac=02:42:ac:11:00:02,mq=on,vectors=18". But I find it has no id. So I can't send QMP command device_del. Do we need to delete this device？

bergwolf:
@caoruidong With network hotplug, I think we should stop appending the default nic to qemu command line when creating a new VM, but hotplug it after the VM is up and running. And then you have the required id of it.

@sboeuf @WeiZhang555 WDYT?

amshinde:
I agree with @bergwolf. The default NIC needs to be hotplugged as well rather than passing it on the qemu command line.

sboeuf:
@bergwolf yes I agree with that, we don't want an hybrid solution since we can hotplug everything.

WeiZhang555:
Agree with @bergwolf

caoruidong:
@bergwolf OK，I have changed the code.
@sboeuf I’m not familiar with Macvtap mode and fail to create a macvtap interface. I’m implementing the bridge mode first.

caoruidong:
Need some help. See https://github.com/kata-containers/agent/issues/280

amshinde:
@caoruidong I havent looked at the PR yet, but saw your comment about macvtap.
Currently macvtap is passed using the same tun/tap configuration interface to qemu as 
`-net tap,fd=3` where `3` is the file descriptor for the macvtap opened in r/w mode.

So, hotplugging macvtap should be the same qemu option as tap+bridge.

Also, currently we support 2 other interfaces along with macvtap:
- Physical interfaces that are passed using VFIO. We already support hotplugging vfio devices in Kata, so you just need to use existing code for implementing this.
- `vhost-user` interfaces. I havent yet looked if qemu has support for this upstream. Last I had seen patches for this some months back.

caoruidong:
@amshinde Thanks for your advice. I have looked at https://github.com/intel/govmm. Seems that it has no interface to add fd or send QMP instruction "getfd". Maybe need a bigger PR to govmm?

virtcontainers/agent.go
caoruidong: I revert this change. Cause input is in the grpc.Route format. Create a new type to decouple it with grpc?

amshinde:
@caoruidong Yes, fd support will need to be added in. I'll try to see what are the commands that need to be  implemented for this.

sboeuf:
@caoruidong what is the status on this PR ? Is MACVTAP the only thing that is missing here ? Or do you plan to add more things ?

BTW, I think you could start submitting to `govmm` in order to get what is needed merged first.

virtcontainers/sandbox.go
```diff
@@ -926,6 +929,102 @@ func (s *Sandbox) removeNetwork() error {
...
+func (s *Sandbox) UpdateRoute(routes []*grpc.Route) error {
+	if err := s.agent.updateRoutes(routes); err != nil {
+		return err
+	}
+	s.networkNS.Routes = routes
+	return s.storage.storeSandboxNetwork(s.id, s.networkNS)
+}
+
+// AddNetwork adds new nic to the sandbox.
+func (s *Sandbox) AddNetwork(inf grpc.Interface) error {
```
amshinde: @caoruidong You will need to call AddNetwork after entering the network namespace with DoNetNs()

caoruidong: Some people worry about overhead about veth pair in #113 .  I will update this PR if most people agree on the veth pair way.

caoruidong:
@sboeuf Now it can hot plug interface or route correctly to a sandbox. Sadly there is another problem. The nic in the sandbox is not connected to host because I'm a little confuse with the network model as in #113 . So at least two things to do. 
I will start to make a PR to govmm and update this PR with the veth pair way.

caoruidong:
@amshinde Another problem: QEMU can use an existing tap device by assigning ifname in “netdev_add" QMP. If I create a tap by "ip tuntap add dev tap0 mode tap" and pass it to QMP, it works well. When I use tap created by https://github.com/kata-containers/runtime/blob/b2bec3362bc6af9c2412d483acfbf5b8d6975df6/virtcontainers/network.go#L924
QMP return {\"error\": {\"class\": \"GenericError\", \"desc\": \"could not configure /dev/net/tun (tap0): Invalid argument\"}} . QEMU doesn't treat it as a tap. So createLink() and "ip tuntap add" are not equivalent. If I want to add a tap created by createLink() to the sandbox, I must pass the fd?

virtcontainers/network.go
```diff
@@ -737,6 +803,23 @@ func xconnectVMNetwork(netPair *NetworkInterfacePair, connect bool) error {
 	}
 }
 
+func xsetupVMNetwork(endpoint *VirtualEndpoint) error {
```
sboeuf: Why do we need to define this new function here ?
What's the problem that you cannot solve with `xconnectVMNetwork()` ?

virtcontainers/network.go
```diff
@@ -244,6 +247,39 @@ func (endpoint *VirtualEndpoint) Detach(netNsCreated bool, netNsPath string) err
 	})
 }
 
+// HotAttach for the virtual endpoint uses hot plug device
+func (endpoint *VirtualEndpoint) HotAttach(h hypervisor, netNsPath string) error {
+	networkLogger().Info("Hot attaching virtual endpoint")
+	if err := xsetupVMNetwork(endpoint); err != nil {
```
sboeuf: It'd be better if we could use `xconnectVMNetwork(&(endpoint.NetPair), true)` here instead. Unless there's a real blocker, could you tell more about it ?

virtcontainers/sandbox.go
sboeuf: Yes thanks @caoruidong, let's have the "standard" network hotplug first and move to some optimizations/improvements later (we'll make this configurable I think).

virtcontainers/sandbox.go
```diff
@@ -926,6 +929,102 @@ func (s *Sandbox) removeNetwork() error {
...
+		Iface: NetlinkIface{
+			LinkAttrs: netlink.LinkAttrs{
+				Name:         inf.Name,
+				HardwareAddr: hw,
+				MTU:          int(inf.Mtu),
+			},
+			Type: "",
+		},
+		Addrs: addrs,
+	}
```
sboeuf: Maybe you could introduce a function translating a `grpc.Interface` structure into a `NetworkInfo` and vice-versa. This would be nice in case it needs to be reused by some other part of the code.

virtcontainers/network.go
```diff
@@ -413,6 +469,8 @@ type NetworkNamespace struct {
 	NetNsPath    string
 	NetNsCreated bool
 	Endpoints    []Endpoint
+	Interfaces   []grpc.Interface
+	Routes       []*grpc.Route
```
sboeuf: Why do you need to add those fields ? `NetworkNamespace` structure already have the list of `Endpoints`, and each Endpoint has a getter function called `Properties()` returning the `NetworkInfo` related to the sandbox.

virtcontainers/kata_agent.go
```diff
@@ -1175,6 +1175,15 @@ func (k *kataAgent) closeProcessStdin(c *Container, ProcessID string) error {
...
 
+func (k *kataAgent) updateNetwork(sandbox *Sandbox, endpoint Endpoint) error {
+	if endpoint == nil {
+		return fmt.Errorf("endpoint can not be nil")
+	}
+
+	interfaces := []*grpc.Interface{k.generateInterface(endpoint)}
+	return k.updateInterfaces(interfaces)
+}
+
```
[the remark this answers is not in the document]
caoruidong: This function is refactored in the first commit. https://github.com/caoruidong/runtime/blob/a8c43583e614475aa240fff053864946e7b15a1c/virtcontainers/kata_agent.go#L438-L456

virtcontainers/network.go
caoruidong: In https://github.com/kata-containers/runtime/issues/113#issuecomment-402918673. 
xconnectVMNetwork() assumes that a device named endporint.netPair.VirtIface.Name already exists in the shim net ns. 
https://github.com/kata-containers/runtime/blob/ff7b4f6ed7f622217f456a9215f5e7c6215ff916/virtcontainers/network.go#L917-L939
This is OK if we insert a nic named eth0. Because docker creates a veth pair and rename one of it to eth0 and put it in the net ns. But if we insert another nic eth1, what should we do? We are not responsible for the veth creation.

virtcontainers/network.go
caoruidong: I see Endpoints store most of these info. But not all.
1. Need to refactor https://github.com/kata-containers/runtime/blob/ff7b4f6ed7f622217f456a9215f5e7c6215ff916/virtcontainers/network.go#L1136 to pass grpc.Interface.Device as tapName. This is easy. 
2. Endpoint includes netlink.Route which does not have a Device field. So we will lost grpc.Route.Device info. Though this field is unused exactly now (just used in agent logs).

virtcontainers/interfaces.go
```diff
@@ -69,6 +73,10 @@ type VCSandbox interface {
 	SignalProcess(containerID, processID string, signal syscall.Signal, all bool) error
 	WinsizeProcess(containerID, processID string, height, width uint32) error
 	IOStream(containerID, processID string) (io.WriteCloser, io.Reader, io.Reader, error)
+	AddNetwork(inf grpc.Interface) error
+	ListNetwork() ([]grpc.Interface, []*grpc.Route, error)
+	UpdateNetwork(interfaces []grpc.Interface) error
+	UpdateRoute(routes []*grpc.Route) error
```
WeiZhang555: I guess this should align with VC interfaces:
```
+	ListNetwork() ([]grpc.Interface, []*grpc.Route, error)
+	UpdateNetwork(interfaces []grpc.Interface, routes []*grpc.Route) error
```

virtcontainers/kata_agent.go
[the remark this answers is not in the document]
WeiZhang555: The `sandbox` isn't used

virtcontainers/interfaces.go
caoruidong: Refer to https://github.com/kata-containers/documentation/blob/master/design/kata-api-design.md#sandbox-hotplug-api. Though it is a little wired?

virtcontainers/agent.go
```diff
@@ -206,4 +207,10 @@ type agent interface {
 
 	// resumeContainer will resume a paused container
 	resumeContainer(sandbox *Sandbox, c Container) error
+
+	// updateRoutes will tell the agent to update route table for an existed Sandbox.
+	updateRoutes(routes []*grpc.Route) error
+
+	// updateNetwork will tell the agent to update a new nic for an existed Sandbox.
+	updateNetwork(sandbox *Sandbox, endpoint Endpoint) error
```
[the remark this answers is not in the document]
caoruidong: @sboeuf I don't know what I've done... Fixed.
@WeiZhang555 Since updateInterface and updateRoute are split in agent grpc code, maybe it's better to rename updateNetwork to updateInterface instead of merging them?

virtcontainers/api.go
```diff
@@ -696,3 +697,56 @@ func PauseContainer(sandboxID, containerID string) error {
...
+	s, err := fetchSandbox(sandboxID)
+	if err != nil {
+		return
+	}
+
+	oldInterfaces := s.networkNS.Interfaces
+	oldRoutes := s.networkNS.Routes
+	defer func() {
+		// Roll back
+		if err != nil {
```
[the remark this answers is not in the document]
caoruidong: Done

virtcontainers/kata_agent.go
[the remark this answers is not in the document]
caoruidong: Fixed.

virtcontainers/network.go
```diff
@@ -295,6 +331,16 @@ func (endpoint *VhostUserEndpoint) Detach(netNsCreated bool, netNsPath string) e
 	return nil
 }
 
+// HotAttach for vhostuser endpoint not supported yet
+func (endpoint *VhostUserEndpoint) HotAttach(h hypervisor, netNsPath string) error {
+	return fmt.Errorf("VhostUserEndpoint don't support Hot attach")
```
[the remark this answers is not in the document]
caoruidong: They are all fixed.

virtcontainers/sandbox.go
caoruidong: Done. But the function is a little wired. Should we delete the log part?

virtcontainers/sandbox.go
```diff
@@ -931,6 +934,122 @@ func (s *Sandbox) removeNetwork() error {
...
+}
+
+// ListNetwork lists all nics and their configurations in the sandbox.
+func (s *Sandbox) ListNetwork() ([]grpc.Interface, []*grpc.Route, error) {
+	return s.networkNS.Interfaces, s.networkNS.Routes, nil
+}
+
+// UpdateNetwork updates the configuration of an existing network interface.
+func (s *Sandbox) UpdateNetwork(interfaces []grpc.Interface) error {
+	if err := s.deleteNetworks(); err != nil {
```
bergwolf: The semantics of `UpdateNetwork` is a bit weird. It will unplug all existing NICs and re-plug them together with any additional NICs. E.g, if a VM has one NIC and we want to add another one. Why do we have to remove the existing one first? I am not sure that is the best way of doing network hotplug.

Can we have `AddNetwork` and `DeleteNetwork` at least since that would cover the most common use cases and does not wipe out all existing NICs during the process?

virtcontainers/sandbox.go
```diff
@@ -931,6 +934,122 @@ func (s *Sandbox) removeNetwork() error {
...
+
+// deleteNetworks deletes all existing network interfaces.
+func (s *Sandbox) deleteNetworks() error {
+	for i := len(s.networkNS.Endpoints) - 1; i >= 0; i-- {
+		endpoint := s.networkNS.Endpoints[i]
+		if err := endpoint.HotDetach(s.hypervisor, s.networkNS.NetNsCreated, s.networkNS.NetNsPath); err != nil {
+			return err
+		}
+		s.networkNS.Endpoints = s.networkNS.Endpoints[:i]
+		if err := s.storage.storeSandboxNetwork(s.id, s.networkNS); err != nil {
```
bergwolf: Each add network and each delete network would save the network config once. From metadata persistency point of view, can we simplify it to just do whatever configuration the API needs, and then save the persistent data once before returning? This would save many IO operations and speed up network operations.

virtcontainers/sandbox.go
WeiZhang555: Deleting existing ones is definitely not good,  we can have a better implementation.

virtcontainers/network.go
WeiZhang555: I agree with @sboeuf , I think it's weird to introduce a new `grpc.Interface` , and  I guess it's better to rework a little of `Endpoint` to involve all the necessary infos

virtcontainers/sandbox.go
```diff
@@ -931,6 +934,122 @@ func (s *Sandbox) removeNetwork() error {
...
+			return err
+		}
+	}
+	s.networkNS.Interfaces = interfaces
+	return s.storage.storeSandboxNetwork(s.id, s.networkNS)
+}
+
+// deleteNetworks deletes all existing network interfaces.
+func (s *Sandbox) deleteNetworks() error {
+	for i := len(s.networkNS.Endpoints) - 1; i >= 0; i-- {
```
WeiZhang555: Can this be simplified as this?
```
for I, endpoints := range s.networkNS.Endpoints {
}
```

virtcontainers/sandbox.go
sboeuf: @WeiZhang555 @bergwolf I agree with this, we don't want to remove an existing interface since the point of `UpdateNetwork()` is to update something that already exists.

virtcontainers/interfaces.go
WeiZhang555: I agree that we should modify the design docs instead

virtcontainers/sandbox.go
```diff
@@ -931,6 +934,122 @@ func (s *Sandbox) removeNetwork() error {
...
+	if err := s.deleteNetworks(); err != nil {
+		return err
+	}
+	for _, inf := range interfaces {
+		if err := s.AddNetwork(inf); err != nil {
+			return err
+		}
+	}
+	s.networkNS.Interfaces = interfaces
+	return s.storage.storeSandboxNetwork(s.id, s.networkNS)
```
WeiZhang555: There's a trick need to be taken care of very carefully.

Here's a list of interfaces, there will be a situation that some interfaces are added successfully and some are not, we need to store the added ones even there are errors happening. So this piece of code should be 
```
func (s *Sandbox) UpdateNetwork(interfaces []grpc.Interface) error {                 
    if err := s.deleteNetworks(); err != nil {                                       
        return err                                                                   
    }                                                                                
                                                                                     
    var err error                                                                    
    for _, inf := range interfaces {                                                 
        if err = s.AddNetwork(inf); err != nil {                                     
            break                                                                    
        }                                                                            
        s.networkNS.Interfaces = append(s.networkNS.Interfaces, inf)                 
    }                                                                                
    if errStore := s.storage.storeSandboxNetwork(s.id, s.networkNS); errSte != nil && err == nil {
        return errStore                                                          
    }                                                                                
                                                                                     
    return err                                                                       
}                    
```

virtcontainers/kata_agent.go
```diff
@@ -342,81 +343,126 @@ func (k *kataAgent) generateInterfacesAndRoutes(networkNS NetworkNamespace) ([]*
...
+			Routes: &grpc.Routes{
+				Routes: routes,
+			},
+		}
+
+		resultingRoutes, err := k.sendReq(routesReq)
+		if err != nil {
+			k.Logger().WithFields(logrus.Fields{
+				"routes-requested": fmt.Sprintf("%+v", routes),
+				"resulting-routes": fmt.Sprintf("%+v", resultingRoutes),
```
WeiZhang555: the `resultingRoutes` need to be returned back to function caller, it's not only for logging.
The  `resultingRoutes` contains the final route rules applied inside Guest, which should be saved in `NetworkNamespace.Routes`.

The logic here is, client send all final "routes" rules kata-agent need to apply, kata-agent will apply these rules, some may fail, but whatever the kata-agent will return all the rules exising inside Guest, and client should update its storage to these rules.

So it's something like "state determination"

virtcontainers/network.go
WeiZhang555: I think we can modify function "bridgeNetworkPair", if the veth pair doesn't exists, we skip the bridge creation and don't insert the veth and tap in the bridge(no bridge to insert), then we can reuse `xconnectVMNetwork()` again.

And in future, CNM and CNI will likely call the same CMD interface "kata-runtime update-network", so I think they should share same logic. You can't just split `kata-runtime update-network` functions with current CNM supporting.

virtcontainers/sandbox.go
caoruidong: OK. I will rewrite it

virtcontainers/kata_agent.go
egernst: Agreed with @WeiZhang555's feedback here.

Stepping back, I don't know what the use case for update routes is compared to just using generate routes, since if we are changing any routes we are going to write all of them.

egernst: ie, for state determination: our state is "failed"

egernst:
@WeiZhang555 @caoruidong @sboeuf @amshinde  - let's try to talk in real time on this patch tonight.  Seb/Archana and myself had a good session walking through this and wanted to see if we can sync.  A UML that I *think* you can view based on our discussion:

https://www.planttext.com/plantuml/txt/hLJDRjim3BxhANXB1zZTtOUW1RP33WNQeZiFYaN7H6J9bUXaVVjHSk8uQJAWm8vHI7v_f1yYwy13PvMgBQ7ZShdLh7y_FgygU3G6LeunDBf6TNmlbpiAN7l7cXo6M84wVoYNBp-VVbKlJyzbwxdKnfGEUU_3zbEXtWXaDVH6Cqv8h_W-O6JbF2C4shIik4N8jOi85zLGUuEm9sl18HfW3sif7ZQTbADHyhlLEta1fWxzmD_tcXXqZBuc6MIabrlWbY8qFa22Ej8rk2D1a16wxr47k2Ci8FUdqlUkJ__LlkijCdvJw0marYfpBPUZfHLW20NGn5WybnguuhYboA_24UDSVSWle0EA56UiS2OdWXRUoJ3nU05h7IaMAY8pkGreSSCtMSkX-KBQFRmbmScnA1gKM5PBaqpMZT0yaHrX9KCq1GXumXkiWQb1NMq8sFaTcdjOyQoZaTYpOvcVawKQcsNy1MlyqsFDMUXnhinguCCFi7L-dshpLWICWysQHw45H98uFZ96UXPBZXuSWiOfCPNyFlABGzD8sgSGRoxyZJMUXNUHVuOHPIEFAxcFEtre_YhkMVCvSbwyqnOeMRehw3ymcV4VJZrqDozzTWhprJwxMpFoa0yI2GsMfGs-csyjECcNcdIDbcgzjhEqRbpJ9EeVldvMTHdSQGkLUX18-M3-1G00

egernst:
@caoruidong - did you intend to do a squash?

caoruidong:
@egernst Yes. I rewrite some code. Now wait for rebase and re-vendor govmm. Then I will split these into about 4 commits.

amshinde:
@caoruidong I have a wip PR for waiting for the network interface for configuring it:
https://github.com/kata-containers/agent/pull/306

I have not tested it end-to-end, will need changes on the runtime side for the PCI Address to be added in grpc.Interface. For the expected format of the PCI address, you can refer to: 
https://github.com/kata-containers/runtime/blob/99954d50257760929b6d72bfa628dadb6ea96394/virtcontainers/qemu.go#L680

Basically, I think we need to add an additional field to `Endpoint` to store the PCI address after the `Endpoint` has been hotplugged.

virtcontainers/qemu.go
```diff
@@ -784,6 +784,45 @@ func (q *qemu) hotplugVFIODevice(device deviceDrivers.VFIODevice, op operation)
...
+
+	qmp, err := q.qmpSetup()
+	if err != nil {
+		return err
+	}
+	q.qmpMonitorCh.qmp = qmp
+	devID := "virtio-" + drive.NetPair.ID
+
+	if op == addDevice {
+		if err := q.qmpMonitorCh.qmp.ExecuteNetdevAdd(q.qmpMonitorCh.ctx, "tap", drive.NetPair.Name, drive.NetPair.TAPIface.Name, "no", "no", defaultQueues); err != nil {
```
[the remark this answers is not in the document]
caoruidong: Yes. For macvtap, seems it only receives fds

caoruidong:
@amshinde Thx. Do you mean we make the change in this PR? OK.

virtcontainers/qemu.go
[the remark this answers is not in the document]
ningbo9: FYI: https://github.com/ningbo9/runtime/blob/e9ce39790bc9eeb36add1ab0472600fd1d91b378/vendor/github.com/intel/govmm/qemu/qmp.go#L868

virtcontainers/sandbox.go
```diff
@@ -934,6 +938,105 @@ func (s *Sandbox) removeNetwork() error {
...
+	}
+
+	// Update the sandbox storage
+	s.networkNS.Endpoints = append(s.networkNS.Endpoints, endpoint)
+	if err := s.storage.storeSandboxNetwork(s.id, s.networkNS); err != nil {
+		return nil, err
+	}
+
+	// Add network for vm
+	return s.agent.updateInterface(inf)
```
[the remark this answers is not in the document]
caoruidong: I think it will has a default route. And we have changed the CLI. Now it's caller's responsibility to call update routes after adding a interface.

virtcontainers/qemu.go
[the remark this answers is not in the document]
caoruidong: Thank you so much. I will try to drive this as in https://github.com/intel/govmm/pull/24#issuecomment-407822886

amshinde:
@caoruidong Yes, it should be part of this PR.

virtcontainers/kata_agent.go
WeiZhang555: @egernst Yes, I agree.

jodh-intel:
Hi @caoruidong - please can you add details of your govmm PR.

caoruidong:
@jodh-intel OK. I'll update the PR later. And try to reduce the cyclomatic complexity to make CI happy.

jodh-intel:
Thanks @caoruidong! :)

sboeuf:
Hey @caoruidong, sorry I didn't follow up on this ! Have you any update here ? Any blockers that you need some help with ?

caoruidong:
@sboeuf Yes. I have updated this PR following the discussion in #113. For cli command like "add interface" still using json format to describe an interface.
```
{
        "device":"tap1",
        "name":"eth0",
        "IPAddresses":[{"address":"192.168.0.2","mask":"24"}],
        "mtu":1500,
        "hwAddr":"00:e0:4c:97:ad:66"
}
```
1. It supports bridged networking model now. Should macvtap model be supported in this PR? I think it should. And how about physical and vhost-user endpoints？
2. Now "list interfaces" and "list routes" just return nil because agent dose not have these APIs. Do we need a PR for agent?

cli/network.go
```diff
@@ -0,0 +1,214 @@
...
+var addIfaceCommand = cli.Command{
+	Name:      "add-iface",
+	Usage:     "add an interface to a container",
+	ArgsUsage: `add-iface <container-id> file or - for stdin`,
+	Flags:     []cli.Flag{},
+	Action: func(context *cli.Context) error {
+		if context.Args().Present() == false {
+			return fmt.Errorf("missing container ID, should provide one")
+		}
+		containerID := context.Args().First()
```
jodh-intel: All these new `cli.Command`'s will require *two* calls to `setExternalLoggers()` to ensure logging is correct (see #528).

cli/network.go
```diff
@@ -0,0 +1,214 @@
...
+		return err
+	},
+}
+
+var delIfaceCommand = cli.Command{
+	Name:      "del-iface",
+	Usage:     "delete an interface from a container",
+	ArgsUsage: `del-iface <container-id> file or - for stdin`,
+	Flags:     []cli.Flag{},
+	Action: func(context *cli.Context) error {
```
jodh-intel: This anonymous function is 99% identical to the one for `addIfaceCommand` - I'd create a generic `toggleIfaceCommand(add bool)` which will call `vci.AddInterface()` if `add == true`, else `vci.RemoveInterface()`.

Also the function in `updateRoutesCommand()` is extremely similar too if that could also be dealt with in a common way?

cli/network.go
```diff
@@ -0,0 +1,214 @@
...
+				return err
+			}
+			defer f.Close()
+		}
+		if err = json.NewDecoder(f).Decode(&inf); err != nil {
+			return err
+		}
+
+		resultingInf, err := vci.AddInterface(sandboxID, inf)
+		fmt.Println(resultingInf)
```
jodh-intel: Log too?

virtcontainers/api.go
```diff
@@ -760,3 +761,100 @@ func PauseContainer(sandboxID, containerID string) error {
...
+
+	s, err := fetchSandbox(sandboxID)
+	if err != nil {
+		return nil, err
+	}
+	return s.AddInterface(inf)
+}
+
+// RemoveInterface is the virtcontainers remove interface entry point.
+func RemoveInterface(sandboxID string, inf *grpc.Interface) (*grpc.Interface, error) {
```
jodh-intel: Again, this is almost identical to `AddInterface()` so you could make them both call a common function like the following to avoid duplication:

```
func toggleInterface(add bool, sandboxID string, inf *grpc.Interface) (*grpc.Interface, error)
```

cli/network.go
```diff
@@ -0,0 +1,214 @@
...
+	"fmt"
+	"os"
+
+	"github.com/kata-containers/agent/protocols/grpc"
+	vc "github.com/kata-containers/runtime/virtcontainers"
+	"github.com/urfave/cli"
+)
+
+var networkCLICommand = cli.Command{
+	Name:  "network",
```
devimc: I don't see a `network` command in `runc`, I think this is `kata` specific and it should be named `kata-network`

cli/main.go
```diff
@@ -110,6 +110,7 @@ var runtimeCommands = []cli.Command{
 	execCLICommand,
 	killCLICommand,
 	listCLICommand,
+	networkCLICommand,
```
devimc: `// Kata Containers specific extensions` ?

cli/network.go
```diff
@@ -0,0 +1,214 @@
...
+				return err
+			}
+			defer f.Close()
+		}
+		if err = json.NewDecoder(f).Decode(&inf); err != nil {
+			return err
+		}
+
+		resultingInf, err := vci.RemoveInterface(sandboxID, inf)
+		fmt.Println(resultingInf)
```
devimc: Log

cli/network.go
```diff
@@ -0,0 +1,214 @@
...
+			return err
+		}
+		containerID = status.ID
+		// container MUST be running
+		if status.State.State != vc.StateRunning {
+			return fmt.Errorf("container %s is not running", containerID)
+		}
+
+		interfaces, err := vci.ListInterfaces(sandboxID)
+		fmt.Println(interfaces)
```
devimc: Log

cli/network.go
```diff
@@ -0,0 +1,214 @@
...
+			if err != nil {
+				return err
+			}
+			defer f.Close()
+		}
+		if err = json.NewDecoder(f).Decode(&routes); err != nil {
+			return err
+		}
+		resultingRoutes, err := vci.UpdateRoutes(sandboxID, routes)
+		fmt.Println(resultingRoutes)
```
devimc: Log again :smile:

cli/network.go
```diff
@@ -0,0 +1,214 @@
...
+			return err
+		}
+		containerID = status.ID
+		// container MUST be running
+		if status.State.State != vc.StateRunning {
+			return fmt.Errorf("container %s is not running", containerID)
+		}
+
+		routes, err := vci.ListRoutes(sandboxID)
+		fmt.Println(routes)
```
devimc: one more

virtcontainers/kata_agent.go
```diff
@@ -441,6 +441,64 @@ func (k *kataAgent) generateInterfacesAndRoutes(networkNS NetworkNamespace) ([]*
...
+		if ok {
+			return resultRoutes.Routes, err
+		}
+		return nil, err
+	}
+	return nil, nil
+}
+
+func (k *kataAgent) listInterfaces() ([]*grpc.Interface, error) {
+	return nil, nil
```
devimc: This is a `TODO` ?

virtcontainers/kata_agent.go
```diff
@@ -441,6 +441,64 @@ func (k *kataAgent) generateInterfacesAndRoutes(networkNS NetworkNamespace) ([]*
...
+}
+
+func (k *kataAgent) listRoutes() ([]*grpc.Route, error) {
+	return nil, nil
```
devimc: `TODO` ?

virtcontainers/qemu.go
```diff
@@ -755,6 +755,45 @@ func (q *qemu) hotplugVFIODevice(device deviceDrivers.VFIODevice, op operation)
...
+			return err
+		}
+		addr, bridge, err := q.addDeviceToBridge(drive.NetPair.ID)
+		if err != nil {
+			return err
+		}
+		drive.PCIAddr = fmt.Sprintf("%02x", bridge.Addr) + "/" + addr
```
devimc: I think this is the same 

```go
fmt.Sprintf("%02x/%s", bridge.Addr, addr)
```

virtcontainers/sandbox.go
```diff
@@ -967,6 +971,106 @@ func (s *Sandbox) removeNetwork() error {
...
+// ListRoutes lists all routes and their configurations in the sandbox.
+func (s *Sandbox) ListRoutes() ([]*grpc.Route, error) {
+	return s.agent.listRoutes()
+}
+
+// deleteNetworks deletes all existing network interfaces.
+func (s *Sandbox) deleteNetworks() error {
+	for i, endpoint := range s.networkNS.Endpoints {
+		if err := endpoint.HotDetach(s.hypervisor, s.networkNS.NetNsCreated, s.networkNS.NetNsPath); err != nil {
+			s.networkNS.Endpoints = s.networkNS.Endpoints[i:]
```
devimc: `panic: runtime error: slice bounds out of range`

you can't iterate over `s.networkNS.Endpoints` and update it at the same time

see https://play.golang.org/p/2yj9Ahrzsuu

caoruidong: Just update slice and end iteration when an endpoint detaching returns an error not every time. Is this correct?

cli/network.go
caoruidong: Fixed

cli/network.go
caoruidong: Combine these three functions to one.

virtcontainers/api.go
caoruidong: Updated

cli/network.go
caoruidong: Yes. Fixed.

virtcontainers/qemu.go
caoruidong: Fixed

virtcontainers/kata_agent.go
caoruidong: Addressed.

virtcontainers/sandbox.go
jodh-intel: You've got to copy the slice and update that copy to remove all the endpoints that were successfully deleted to avoid the issue @devimc has pointed out.

caoruidong: OK. But I think it's safe. In other codes
https://github.com/kata-containers/runtime/blob/fc0142ec8e403b784a7e240b04d4b09b7b552426/virtcontainers/sandbox.go#L1108-L1113

cli/network.go
caoruidong: Done

cli/network.go
caoruidong: Log added.

virtcontainers/sandbox.go
caoruidong: Good point. I will update this. In the end, I find that updating s.networkNS.Endpoints seems useless...

sboeuf:
@caoruidong 

>@sboeuf Yes. I have updated this PR following the discussion in #113. Cli commands like "add interface" still use json format to describe an interface.

This sounds good to me, that's what we agreed on.

>It supports bridged networking model now. Should macvtap model be supported in this PR? I think it should.

I agree MACVTAP should be supported through this PR as this is the model we use by default in Kata Containers right now, and I don't think we want to see any drop in performance because of hotplug.

>And how about physical and vhost-user endpoints？

Those ones should be handled as a follow up PR (but this is too much for this one). I think @amshinde has a pretty good idea on how to support the hotplug for every type of `Endpoint`.

>Now "list interfaces" and "list routes" just return nil because agent dose not have these APIs. Do we need a PR for agent?

We definitely need a PR on the agent. Please ask if you want someone else to do this ;)

caoruidong:
@sboeuf Thanks. I'll move on and make a PR on agent.

grahamwhaley:
Just to check status then @caoruidong @sboeuf (as this is a >150 comment thread with 9+ commentors on it...) - is this PR now blocked waiting for a new agent PR, or?? thx

caoruidong:
Yes, waiting for https://github.com/kata-containers/agent/pull/316. Sorry

sboeuf:
@caoruidong  kata-containers/agent#316 just got merged ;)

virtcontainers/qemu.go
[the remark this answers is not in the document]
caoruidong: Your code is really helpful. But one problem. If hotplug a macvtap, can host ping it successfully? Have you tried this?

sboeuf:
@caoruidong ok what's the new status on this PR ? I think there is no more external dependency that needs to be resolved anymore (all agent and govmm changes have been merged), right ?

caoruidong:
@sboeuf Yes. I have update all this PR. One thing to do is reducing the cyclomatic complexity.
There is another problem, if I use bridged network model and set up network correctly, I can ping the nic from host. But if I use macvtap, I cannot ping. Do you and @amshinde have any ideas?

sboeuf:
@caoruidong I have been playing with your PR as I wanted to understand what was missing here. I have realized that because we said we should move to a fully hotplugged model, we are far from having this in place here because there are so many things involved and implicit with the way the network is handled, which makes it very hard to decouple.

Having the network fully hotplugged means that we cannot make any assumption that some underlying network will be there when Kata will be called, and this means Kata should not have to scan the network. I realize this task has to go to the CLI level, and with this also comes the rework of the OCI hooks that should have been handled from the OCI CLI level from a long time ago. Because we have this very strong assumption with the OCI hooks, and the way we expect to find the network, we cannot be generic here...
I strongly believe that we should do this refactoring/redesigning work asap, but at the same time I want to go step by step with this PR as I would like to see it merged. And also I care about not breaking Kata, so here is my proposal if @caoruidong @bergwolf @egernst @amshinde @WeiZhang555 @mcastelino  all agree with it.

1. We merge this PR once we can hotplug some extra networks by using CLI commands or directly the Kata API, but keeping the first default network as being coldplugged, for the sake of not breaking Kata, and also not to introduce too many changes.
2. At this point, we know the hotplug mechanism works, but we want to get the first network also hotplugged and this will mean:
    a. Move the OCI hooks handling to the CLI implementation
    b. Scan the content of the network namespace from the CLI after the network has been created
    c. Call into Kata hotplug API to add the network needed by the sandbox
Globally, I am proposing to move a lot of the OCI specific assumptions to the CLI layer since it's OCI specific. This means Kata API (virtcontainers) becomes lighter and more straightforward about the network, by simply doing the plumbing between interfaces of the network namespace and interfaces of the VM.

Please let me know what you all think about this !

caoruidong:
I think that we can keep the code plug way as default. If one day we are clear with the hot plug usage scenario, we can change to that way carefully.

WeiZhang555:
> 1. We merge this PR once we can hotplug some extra networks by using CLI commands or directly the Kata API, but keeping the first default network as being coldplugged, for the sake of not breaking Kata, and also not to introduce too many changes.

This is so good to me.

> 2. At this point, we know the hotplug mechanism works, but we want to get the first network also hotplugged and this will mean:
> a. Move the OCI hooks handling to the CLI implementation
> b. Scan the content of the network namespace from the CLI after the network has been created
> c. Call into Kata hotplug API to add the network needed by the sandbox
> Globally, I am proposing to move a lot of the OCI specific assumptions to the CLI layer since it's OCI specific. This means Kata API (virtcontainers) becomes lighter and more straightforward about the network, by simply doing the plumbing between interfaces of the network namespace and interfaces of the VM.

Not quite understand why we need to move OCI specific assumptions to CLI and how. I think  we can always "scan and hotplug interface" to the VM after virtcontainers finishes execution of OCI Prestart hooks. By that we can go on supporting docker libnetwork.
But we can discuss more details later as this is Step 2. Overall it looks good.

bergwolf:
@WeiZhang555 
> Not quite understand why we need to move OCI specific assumptions to CLI and how. I think we can always "scan and hotplug interface" to the VM after virtcontainers finishes execution of OCI Prestart hooks. By that we can go on supporting docker libnetwork.

The sandbox creation process becomes:
1. create an empty sandbox/VM
2. run prestart hooks
3. scan and hotplug interfaces
4. create containers

I'm fine with implementing it either in virtcontainers or in the CLI. With the hotplug APIs, it might feel better to implement it in the CLI.

amshinde:
> We merge this PR once we can hotplug some extra networks by using CLI commands or directly the Kata API, but keeping the first default network as being coldplugged, for the sake of not breaking Kata, and also not to introduce too many changes.

Yes, I am fine with this too. Lets keep the coldplugging code as is and we can revisit it once we have reworked the OCI hooks. 

> a. Move the OCI hooks handling to the CLI implementation
> b. Scan the content of the network namespace from the CLI after the network has been created
> c. Call into Kata hotplug API to add the network needed by the sandbox

Yes, we have discussed this in the past about moving the OCI hooks to the CLI. Scanning the net namespace from the CLI instead of virtcontainers sounds like a better approach to me.

sboeuf:
@bergwolf @amshinde @WeiZhang555 thanks for your feedback!

@caoruidong I have played with your PR all day, using the code I have developed for the network monitor. I need to cleanup a few thing on my branch and I'll provide you some fix that I needed on top of your PR to get things working. I also want to give a few comments about it. I'll do this later tonight or tomorrow, just wanted to let you know I'm actively looking into this :)

virtcontainers/network.go
```diff
@@ -1256,7 +1314,7 @@ func deleteNetNS(netNSPath string) error {
 	return nil
 }
 
-func createVirtualNetworkEndpoint(idx int, ifName string, interworkingModel NetInterworkingModel) (*VirtualEndpoint, error) {
+func createVirtualNetworkEndpoint(idx int, ifName, tapName, hwAddr string, interworkingModel NetInterworkingModel) (*VirtualEndpoint, error) {
```
sboeuf: We don't need the `tapName` and `hwAddr` here since they have to be generated for veth. Please don't modify this.

virtcontainers/network.go
```diff
@@ -1288,6 +1346,14 @@ func createVirtualNetworkEndpoint(idx int, ifName string, interworkingModel NetI
 		endpoint.NetPair.VirtIface.Name = ifName
 	}
 
+	if tapName != "" {
+		endpoint.NetPair.TAPIface.Name = tapName
+	}
+
+	if hwAddr != "" {
+		endpoint.NetPair.TAPIface.HardAddr = hwAddr
+	}
```
sboeuf: Because we don't need those variable, please remove both if statements here.

sboeuf:
@caoruidong oh one more thing, could you fix the code coverage ? `-2.4%` is too much...

bergwolf:
@WeiZhang555 Yes, I agree that we should rely on the netns monitor to do the (scan/hotplug) work.

sboeuf:
@WeiZhang555 @bergwolf 
I'd prefer to keep the network monitor solution a very specific one, only in case we want to detect network changes. But this monitoring process increases the footprint of the whole system, and might not be reliable for now in case of crash/restart. What I would suggest is that we can configure from `configuration.toml` if we want the network monitoring process or not. Based on this, we would either start the monitoring process before the OCI hooks from the CLI, or we would simply scan the netns from the CLI.

Does that make sense ?

Just to be clear, I'm trying to not push too much complexity with our current architecture, and that's why I'm trying to make sure we don't consider the network monitoring process as mandatory.

sboeuf:
@caoruidong I have rebased and repushed so that we can see the code coverage as it was getting confused.

@WeiZhang555 @devimc you both had requested some changes, could you please take another look and let us know if the PR is okay now?

caoruidong:
@sboeuf Codecov is still not working well. Maybe I need to update it again.

sboeuf:
@caoruidong give it some time, now that the pr has been rebased on master, it should be fine I think ;)

sboeuf:
@amshinde @chavafg do you know why codecov does not work here ? is it down ?

chavafg:
Not sure, 
I found this on the `codecov.io` site report [1].
```
Missing base report.
Unable to compare commits because the base of the pull request did not upload a coverage report.

Changes found in between 83008d4...cd514b6 (pseudo...base) which prevent comparing this pull request.
```
Maybe a new rebase is needed?

[1] https://codecov.io/gh/kata-containers/runtime/pull/287

amshinde:
@sboeuf No idea. Maybe as @chavafg suggests, a rebase may help.

sboeuf:
@amshinde @chavafg I did the rebase this morning, but didn't seem to fix the issue...

bergwolf:
@sboeuf I think we should separate the daemon part of netns monitor from the actual implementation that does the scan/hotplug job, -- e.g., by making scan/hotplug a standalone package. Then both the netns monitor process, and the CLI can use the standalone package to kick off network scan/hotplug. The only difference is that netns is a long running daemon and CLI only does it in one shot when/after executing the prestart hooks. When there are no prestart hooks, CLI does not even bother calling scan/hotplug -- which would allow us to create a working sandbox w/o any external NICs.

sboeuf:
@bergwolf yes definitely. I am currently working on reworking the whole OCI hook. This will be the first step. And then, we'll move more code to CLI if needed. But all of this sounds good to me :)

sboeuf:
@caoruidong please try to fix the build for this as I'd like to get it merged by tomorrow (unless @WeiZhang555 @bergwolf or @devimc has any objections).

WeiZhang555:
LGTM

> I'd prefer to keep the network monitor solution a very specific one, only in case we want to detect network changes. But this monitoring process increases the footprint of the whole system, and might not be reliable for now in case of crash/restart. What I would suggest is that we can configure from configuration.toml if we want the network monitoring process or not. Based on this, we would either start the monitoring process before the OCI hooks from the CLI, or we would simply scan the netns from the CLI.
> Does that make sense ?

I think this could make sense :-) @sboeuf

sboeuf:
@caoruidong thank you very much!
I'll review and merge if everything's good :)

sboeuf:
Merging as `-0.06%` coverage is clearly negligible here.

CaesarLXL:
This feature is currently not valid because the pull request 534 is not fully implemented.

Issue #291, opened 2018-05-09
Root dir espcially shared dirs remaining when container has been removed


jshachm:
`/var/run/kata-containers/shared/sandboxes` remaining a lot of sandbox dirs even if sandbox has already been removed.
Same thing happens to a single container.

devimc:
fixed https://github.com/kata-containers/runtime/pull/1139

PR #292, opened 2018-05-09
virtcontainers : fix shared dir resource remaining

jshachm:
Before this patch shared dir will reamin when sandox
has already removed.

Do clean up shared dirs after all mounts are umounted

Fix: #291

Signed-off-by: Haomin <caihaomin@huawei.com>

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -510,11 +510,19 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
 func (c *Container) unmountHostMounts() error {
 	for _, m := range c.mounts {
 		if m.HostPath != "" {
+			logger := c.Logger().WithField("host-path", m.HostPath)
 			if err := syscall.Unmount(m.HostPath, 0); err != nil {
-				c.Logger().WithFields(logrus.Fields{
-					"host-path": m.HostPath,
-					"error":     err,
-				}).Warn("Could not umount")
+				// Unable to unmount paths could be a really big problem here
+				// we need to make sure cause 'less damage' if things are
+				// really broken. For further, we need to give admins more of
+				// a chance to diagnose the problem. As the rules of `fail fast`,
+				// here we return an error as soon as we get it.
+				logger.WithError(err).Warn("Could not umount")
+				return err
+			} else if err := os.RemoveAll(m.HostPath); err != nil {
+				// since the mounts related to the shared dir is umounted
+				// we need to remove the host path to avoid resource remaining
+				logger.WithError(err).Warn("Could not be removed")
 				return err
 			}
 		}
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -549,6 +549,10 @@ func (k *kataAgent) stopSandbox(sandbox *Sandbox) error {
 	return k.proxy.stop(sandbox, k.state.ProxyPid)
 }
 
+func (k *kataAgent) cleanupSandbox(sandbox *Sandbox) error {
+	return os.RemoveAll(filepath.Join(kataHostSharedDir, sandbox.id))
+}
+
 func (k *kataAgent) replaceOCIMountSource(spec *specs.Spec, guestMounts []Mount) error {
 	ociMounts := spec.Mounts
 
@@ -951,7 +955,14 @@ func (k *kataAgent) stopContainer(sandbox *Sandbox, c Container) error {
 		return err
 	}
 
-	return bindUnmountContainerRootfs(kataHostSharedDir, sandbox.id, c.id)
+	if err := bindUnmountContainerRootfs(kataHostSharedDir, sandbox.id, c.id); err != nil {
+		return err
+	}
+
+	// since rootfs is umounted it's safe to remove the dir now
+	rootPathParent := filepath.Join(kataHostSharedDir, sandbox.id, c.id)
+
+	return os.RemoveAll(rootPathParent)
 }
 
 func (k *kataAgent) signalProcess(c *Container, processID string, signal syscall.Signal, all bool) error {
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -1182,6 +1182,13 @@ func (s *Sandbox) stop() error {
 		return err
 	}
 
+	// vm is stopped remove the sandbox shared dir
+	if err := s.agent.cleanupSandbox(s); err != nil {
+		// cleanup resource failed shouldn't block destroy sandbox
+		// just raise a warning
+		s.Logger().WithError(err).Warnf("cleanup sandbox failed")
+	}
+
 	return s.setSandboxState(StateStopped)
 }
 
--- a/virtcontainers/agent.go
+++ b/virtcontainers/agent.go
@@ -152,6 +152,9 @@ type agent interface {
 	// stopSandbox will tell the agent to stop all containers related to the Sandbox.
 	stopSandbox(sandbox *Sandbox) error
 
+	// cleanup will clean the resources for sandbox
+	cleanupSandbox(sandbox *Sandbox) error
+
 	// createContainer will tell the agent to create a container related to a Sandbox.
 	createContainer(sandbox *Sandbox, c *Container) (*Process, error)
 
--- a/virtcontainers/hyperstart_agent.go
+++ b/virtcontainers/hyperstart_agent.go
@@ -854,3 +854,7 @@ func (h *hyper) resumeContainer(sandbox *Sandbox, c Container) error {
 	// hyperstart-agent does not support resume container
 	return nil
 }
+
+func (h *hyper) cleanupSandbox(sandbox *Sandbox) error {
+	return nil
+}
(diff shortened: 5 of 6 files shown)


virtcontainers/container.go
```diff
@@ -405,6 +405,14 @@ func (c *Container) unmountHostMounts() error {
 					"error":     err,
 				}).Warn("Could not umount")
 				return err
+			} else if err := os.RemoveAll(m.HostPath); err != nil {
+				// since the mounts related to the shared dir is umounted
+				// we need to remove the host path to avoid resource remaining
+				c.Logger().WithFields(logrus.Fields{
```
jodh-intel: This log call essentially simulates a call to `WithError()` and although it is simpler than having to call `WithError()`, I think it's safer to explicitly call `WithError()` to protect against it changing its behaviour at some future point.

Since this is based on the log call code above, I'd do something like:

```
logger := c.Logger().WithField("host-path", m.HostPath)

if err := syscall.Unmount(m.HostPath, 0); err != nil {
    logger.WithError(err).Warn("could not unmount")
} else if err := os.RemoveAll(m.HostPath); err != nil {
    logger.WithError(err).Warn("could not remove path")
}
```

virtcontainers/container.go
```diff
@@ -405,6 +405,14 @@ func (c *Container) unmountHostMounts() error {
 					"error":     err,
 				}).Warn("Could not umount")
 				return err
+			} else if err := os.RemoveAll(m.HostPath); err != nil {
```
jodh-intel: I think you should be able to create a unit test for this in `container_test.go` where you set `c.mounts[0]` to contain a double bind-mounted path. That way, the `Unmount()` will succeed but the `RemoveAll()` will fail (as there is still one more `Unmount()` call to make).

virtcontainers/container.go
jshachm: Going to change it now~

virtcontainers/sandbox.go
```diff
@@ -1162,6 +1162,12 @@ func (s *Sandbox) stop() error {
 		return err
 	}
 
+	// vm is stopped remove the sandbox shared dir
+	sandboxSharedDir := filepath.Join(kataHostSharedDir, s.id)
```
sboeuf: This cannot be done here. `kataHostSharedDir` is specific to the `kata_agent.go` implementation, and should never be used by the generic sandbox implementation.

jshachm: Sry for missing this. This should be done more generic.

jshachm:
Need to find a better way to remove sandbox work dirs and solve the case that shared dirs are mounted to different destinations. So do-not-merge by now

virtcontainers/sandbox.go
amshinde: @jshachm You can introduce an agent interface to cleanup agent specific artifacts after the VM has stopped.

jshachm:
@amshinde Good suggestions. Working on it and maybe it's good for #351

virtcontainers/sandbox.go
sboeuf: But wait, any cleanup specific to an agent implementation should be part of either `stopContainer()` if it's related to a container, or `stopSandbox()` if it's related to the sandbox itself.
I don't see the value (and actually this would imply duplication) of introducing a new method to the agent interface.

amshinde: @sboeuf The kata host directory can only be deleted after the VM has stopped since it is shared with the VM through 9p. So this cannot be part of `stopContainer` or `stopSandbox`.

sboeuf: Oh yes good point, then in this case, a `agent.cleanup()` function as you suggested sounds like a good idea.

katabuilder:
PSS Measurement:
Qemu: 159779 KB
Proxy: 8730 KB
Shim: 10810 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1996996 KB

jshachm:
@sboeuf @amshinde    get `cleanupSandbox` done.

Maybe need more unit test as @jodh-intel suggested. Will pull those later.

amshinde:
thanks @jshachm!
lgtm

virtcontainers/kata_agent.go
```diff
@@ -528,6 +528,14 @@ func (k *kataAgent) stopSandbox(sandbox *Sandbox) error {
 	return k.proxy.stop(sandbox, k.state.ProxyPid)
 }
 
+func (k *kataAgent) cleanupSandbox(sandbox *Sandbox) error {
+	sandboxSharedDir := filepath.Join(kataHostSharedDir, sandbox.id)
+	if err := os.RemoveAll(sandboxSharedDir); err != nil {
```
sboeuf: Please simplify the function this way:
```go
func (k *kataAgent) cleanupSandbox(sandbox *Sandbox) error {
        return os.RemoveAll(filepath.Join(kataHostSharedDir, sandbox.id))
}
```

virtcontainers/sandbox.go
```diff
@@ -1181,6 +1181,12 @@ func (s *Sandbox) stop() error {
 		return err
 	}
 
+	// vm is stopped remove the sandbox shared dir
+	if err := s.agent.cleanupSandbox(s); err != nil {
+		// cleanup resource failed shouldn't block destroy sandbox
+		// just raise a warning
+		s.Logger().WithError(err).Warnf("cleanup sandbox failed")
```
sboeuf: Looks fair !

virtcontainers/sandbox.go
```diff
@@ -1181,6 +1181,12 @@ func (s *Sandbox) stop() error {
...
+	}
```
sboeuf: Please add a blank line to let the code breathe :)

virtcontainers/virtcontainers_test.go
```diff
@@ -41,7 +41,7 @@ var testQemuPath = ""
 var testHyperstartCtlSocket = ""
 var testHyperstartTtySocket = ""
 
-// cleanUp Removes any stale sandbox/container state that can affect
+// cleanup Removes any stale sandbox/container state that can affect
```
sboeuf: This should not be part of this PR.

jshachm:
@sboeuf all changes done~

sboeuf:
Thanks I'll take a look !

sboeuf:
@jshachm LGTM
Could you please repush to retrigger the CI now that we should not have glitches anymore.

virtcontainers/container.go
```diff
@@ -510,11 +510,14 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
...
 	for _, m := range c.mounts {
 		if m.HostPath != "" {
+			logger := c.Logger().WithField("host-path", m.HostPath)
 			if err := syscall.Unmount(m.HostPath, 0); err != nil {
-				c.Logger().WithFields(logrus.Fields{
-					"host-path": m.HostPath,
-					"error":     err,
-				}).Warn("Could not umount")
+				logger.WithError(err).Warn("Could not umount")
+				return err
```
grahamwhaley: Not a change by your patch, but directly related - it seems if we fail one unmount then we quit the loop. Would it maybe be better for us to continue to try and unmount the remaining mounts in `c.mounts`, and clean up as much as we can (even though we know we will ultimately fail this function)?
Same for the following `RemoveAll` as well.

jodh-intel:
The test failures I think are https://github.com/kata-containers/tests/issues/427 (see https://github.com/kata-containers/runtime/pull/385#issuecomment-397922893).

virtcontainers/container.go
jshachm: @grahamwhaley  great catch and suggestions~

When I coded these part, it also took me a lot of time to think whether `continue to next` is better than `break from here`.After thinking and discussion with teammates,  and some "sad" experience from our production environment, I think `umount` failure is a very bad situation maybe caused by `mount point leak` and other bad thing... So we need to break right here and raise a warning to upper platform like `k8s`. 

In other word, If a remaining has already happened, maybe it better to alarm the `SRE` department as quick as possible to get involved in this problem.

What do you think about this, glad to hear your suggestion in this situation~~~ ^_^

jshachm:
@grahamwhaley thx for the remind of `commit msg`. I will update it and make it more clear once I get to my workstation next morning... And for CI thing, with the help of @chavafg I finally get my button back~ 

@jodh-intel As you mentioned , CI will not be happy until we change the scripts of ping CNI?

virtcontainers/container.go
grahamwhaley: Hi @jshachm Yeah, it is a tricky call if we quit or carry on. I can see your point that 'fail fast' and try not to make things worse is perfectly valid. Maybe that is the safer way which will:
- possibly cause 'less damage' if things are really broken
- give the admins more of a chance to diagnose the problem

So, I'm happy to leave as is. If we do, can we add a big comment saying why we quit on the first failure ;-)
Probably best if we get some other input from maybe @egernst and @gnawux maybe, and others if anybody has a strong opinion on this?

sboeuf:
@jshachm if you rebase your PR on top of master (make sure you fetch), you should get the CI test passing since the CNI issue got fixed.

amshinde:
@jshachm Yes, CI has been fixed now. Please rebase on top of master.

PR #296, opened 2018-05-09
virtcontainers/qemu: reduce memory footprint

### LINKED ISSUE #295: kata containers memory footprint
The memory footprint of kata containers is pretty high even inside the guest. Comparing to what runv+hyperstart has for a 256MB busybox container:
```
/ # free -m
             total       used       free     shared    buffers     cached
Mem:           235         33        202         18          0         19
-/+ buffers/cache:         13        222
Swap:            0          0          0
```

kata container only leaves less than 120MB free memory for the container app:
```
/ # free -m
             total       used       free     shared    buffers     cached
Mem:           195         82        113         26          0         26
-/+ buffers/cache:         55        139
Swap:            0          0          0
```

There are two places we need to look into:
1. total memory: both cases are specifying 256 MB in qemu arguments but kata container guest only sees around 200MB of total memory, as also confirmed by agent log `system-memory=\"200572 kB`.
2. used memory: likely this is caused by C vs. go. IWO, the go runtime costs around 50MB more than a plain C agent. I guess it's a price we'll have to pay with having a go agent. I wonder if there is some way to still reduce the agent memory footprint?

@sboeuf @WeiZhang555 @amshinde @devimc @egernst @laijs any thought?

devimc:
There is a relation between the maximum number of vCPUs and the
memory footprint, if QEMU maxcpus option and kernel nr_cpus
cmdline argument are big, then memory footprint is big, this
issue only occurs if CPU hotplug support is enabled in the kernel,
might be because of kernel needs to allocate resources to watch all
sockets waiting for a CPU to be connected (ACPI event).

For example

```
+---------------+-------------------------+
|               | Memory Footprint (KB)   |
+---------------+-------------------------+
| NR_CPUS=240   | 186501                  |
+---------------+-------------------------+
| NR_CPUS=8     | 110684                  |
+---------------+-------------------------+
```

In order to do not affect CPU hotplug and allow to users to have containers
with the same number of physical CPUs, this patch tries to mitigate the
big memory footprint by using the actual number of physical CPUs as the
maximum number of vCPUs for each container.

Before this patch a container with 256MB of RAM

```
              total        used        free      shared  buff/cache   available
Mem:           195M         40M        113M         26M         41M        112M
Swap:            0B          0B          0B
```

With this patch

```
              total        used        free      shared  buff/cache   available
Mem:           236M         11M        188M         26M         36M        186M
Swap:            0B          0B          0B
```

fixes #295

### DIFF
--- a/cli/config.go
+++ b/cli/config.go
@@ -74,6 +74,7 @@ type hypervisor struct {
 	KernelParams          string `toml:"kernel_params"`
 	MachineType           string `toml:"machine_type"`
 	DefaultVCPUs          int32  `toml:"default_vcpus"`
+	DefaultMaxVCPUs       uint32 `toml:"default_maxvcpus"`
 	DefaultMemSz          uint32 `toml:"default_memory"`
 	DefaultBridges        uint32 `toml:"default_bridges"`
 	Msize9p               uint32 `toml:"msize_9p"`
@@ -202,6 +203,25 @@ func (h hypervisor) defaultVCPUs() uint32 {
 	return uint32(h.DefaultVCPUs)
 }
 
+func (h hypervisor) defaultMaxVCPUs() uint32 {
+	numcpus := uint32(goruntime.NumCPU())
+	maxvcpus := vc.MaxQemuVCPUs()
+	reqVCPUs := h.DefaultMaxVCPUs
+
+	//don't exceed the number of physical CPUs. If a default is not provided, use the
+	// numbers of physical CPUs
+	if reqVCPUs >= numcpus || reqVCPUs == 0 {
+		reqVCPUs = numcpus
+	}
+
+	// Don't exceed the maximum number of vCPUs supported by hypervisor
+	if reqVCPUs > maxvcpus {
+		return maxvcpus
+	}
+
+	return reqVCPUs
+}
+
 func (h hypervisor) defaultMemSz() uint32 {
 	if h.DefaultMemSz < 8 {
 		return defaultMemSize // MiB
@@ -313,6 +333,7 @@ func newQemuHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 		KernelParams:          vc.DeserializeParams(strings.Fields(kernelParams)),
 		HypervisorMachineType: machineType,
 		DefaultVCPUs:          h.defaultVCPUs(),
+		DefaultMaxVCPUs:       h.defaultMaxVCPUs(),
 		DefaultMemSz:          h.defaultMemSz(),
 		DefaultBridges:        h.defaultBridges(),
 		DisableBlockDeviceUse: h.DisableBlockDeviceUse,
@@ -418,6 +439,7 @@ func loadConfiguration(configPath string, ignoreLogging bool) (resolvedConfigPat
 		MachineAccelerators:   defaultMachineAccelerators,
 		HypervisorMachineType: defaultMachineType,
 		DefaultVCPUs:          defaultVCPUCount,
+		DefaultMaxVCPUs:       defaultMaxVCPUCount,
 		DefaultMemSz:          defaultMemSize,
 		DefaultBridges:        defaultBridgesCount,
 		MemPrealloc:           defaultEnableMemPrealloc,
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -57,15 +57,11 @@ type ContainerStatus struct {
 
 // ContainerResources describes container resources
 type ContainerResources struct {
-	// CPUQuota specifies the total amount of time in microseconds
-	// The number of microseconds per CPUPeriod that the container is guaranteed CPU access
-	CPUQuota int64
+	// VCPUs are the number of vCPUs that are being used by the container
+	VCPUs uint32
 
-	// CPUPeriod specifies the CPU CFS scheduler period of time in microseconds
-	CPUPeriod uint64
-
-	// CPUShares specifies container's weight vs. other containers
-	CPUShares uint64
+	// Mem is the memory that is being used by the container
+	Mem uint32
 }
 
 // ContainerConfig describes one container runtime configuration.
@@ -804,8 +800,7 @@ func (c *Container) update(resources specs.LinuxResources) error {
 	}
 
 	newResources := ContainerResources{
-		CPUPeriod: *resources.CPU.Period,
-		CPUQuota:  *resources.CPU.Quota,
+		VCPUs: uint32(utils.ConstraintsToVCPUs(*resources.CPU.Quota, *resources.CPU.Period)),
 	}
 
 	if err := c.updateResources(currentConfig.Resources, newResources); err != nil {
@@ -866,7 +861,7 @@ func (c *Container) hotplugDrive() error {
 		Index:  driveIndex,
 	}
 
-	if err := c.sandbox.hypervisor.hotplugAddDevice(&drive, blockDev); err != nil {
+	if _, err := c.sandbox.hypervisor.hotplugAddDevice(&drive, blockDev); err != nil {
 		return err
 	}
 
@@ -903,7 +898,7 @@ func (c *Container) removeDrive() (err error) {
 		l := c.Logger().WithField("device-id", devID)
 		l.Info("Unplugging block device")
 
-		if err := c.sandbox.hypervisor.hotplugRemoveDevice(drive, blockDev); err != nil {
+		if _, err := c.sandbox.hypervisor.hotplugRemoveDevice(drive, blockDev); err != nil {
 			l.WithError(err).Info("Failed to unplug block device")
 			return err
 		}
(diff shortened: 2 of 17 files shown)


bergwolf:
Nice findings! I wasn't aware the maxcpus limit can have so much impact on memory footprint! Thanks @devimc ! And it

LGTM!

egernst:
LGTM, though I see the CI failures.  

--- FAIL: TestQemuKernelParameters (0.00s)
	qemu_test.go:50: Got: panic=1 initcall_debug nr_cpus=4 foo=foo bar=bar, Expecting: panic=1 initcall_debug foo=foo bar=bar

Perhaps the test needs updating?

gnawux:
Awesome

Never noticed the impaction of `maxcpus`

WeiZhang555:
This is a great improvement! 
One question, should we make this configurable ? Then the service provider can choose to make the VM smaller via a smaller `maxcpus`

ChristophSGR:
Von meinem iPhone gesendet

Am 10.05.2018 um 15:50 schrieb zhangwei_cs <notifications@github.com>:

This is a great improvement!
One question, should we make this configurable ? Then the service provider can choose to make the VM smaller via a smaller maxcpus

egernst:
@WeiZhang555 @ChristophSGR   - Yes I think we should.  @devimc is OOO today -- I'm sure he can take a look tomorrow (unless someone wants to add a patch to this PR in the meantime!)

gnawux:
I thought about this ("make this configurable") as well, but I'd like to accept this patch and have another patch on configurable things if @devimc do not have a slot on this.

egernst:
@gnawux - sure this is fine.  I think we shoudl get this in ASAP for sure.  I want to wait until we can get the tests update (seems like a legit CI failure)

gnawux:
@egernst agree

virtcontainers/qemu_amd64.go
```diff
@@ -73,7 +74,7 @@ var supportedQemuMachines = []govmmQemu.Machine{
 
 // returns the maximum number of vCPUs supported
 func maxQemuVCPUs() uint32 {
-	return uint32(240)
+	return uint32(runtime.NumCPU())
```
WeiZhang555: There will be some problems if the physical machine has more than 240 physical CPUs, right?
I think we should choose `MAX(runtime.NumCPU(), 240)` here

WeiZhang555:
Sounds fair enough to merge this first, after fixing the CI problem.

devimc:
CI is not happy because of Azure VM has 4 CPUs and CRIO tries to create a container with 5 vCPUs

```
# time="2018-05-11 13:06:20.495616418Z" level=error msg="Container creation error: Unable to hotplug 2 CPUs, currently this SB has 3 CPUs and the maximum amount of CPUs is 4
```

hence I need to make `maxvcpus` configurable in this patch

devimc:
please take a look, all these changes are needed to reduce memory footprint and make CRIO happy

cli/config.go
```diff
@@ -202,6 +203,28 @@ func (h hypervisor) defaultVCPUs() uint32 {
...
+
+	if h.DefaultMaxVCPUs <= 0 {
+		// Don't exceed the maximum number of vCPUs supported by QEMU
+		if numcpus > int(maxvcpus) {
+			return maxvcpus
+		}
+		return uint32(numcpus)
+	}
+
+	if h.DefaultMaxVCPUs > int32(numcpus) {
```
gnawux: do we need this comparison?

virtcontainers/qemu.go
```diff
@@ -800,19 +802,28 @@ func (q *qemu) hotplugCPUs(vcpus uint32, op operation) error {
...
+// try to hot add an amount of vCPUs, returns the number of vCPUs added
+func (q *qemu) hotplugAddCPUs(amount uint32) (uint32, error) {
 	currentVCPUs := q.qemuConfig.SMP.CPUs + uint32(len(q.state.HotpluggedVCPUs))
 
-	// Don't exceed the maximum amount of vCPUs
+	// Don't fail if the number of max vCPUs is exceeded, log a warning and hot add the vCPUs needed
+	// to reach out max vCPUs
 	if currentVCPUs+amount > q.config.DefaultMaxVCPUs {
-		return fmt.Errorf("Unable to hotplug %d CPUs, currently this SB has %d CPUs and the maximum amount of CPUs is %d",
+		q.Logger().Warnf("Cannot to hotplug %d CPUs, currently this SB has %d CPUs and the maximum amount of CPUs is %d",
```
egernst: nit s/to//

cli/config.go
```diff
@@ -74,6 +74,7 @@ type hypervisor struct {
 	KernelParams          string `toml:"kernel_params"`
 	MachineType           string `toml:"machine_type"`
 	DefaultVCPUs          int32  `toml:"default_vcpus"`
+	DefaultMaxVCPUs       int32  `toml:"default_maxvcpus"`
```
egernst: Why signed int?  Is there a scenario where you'd have a negative number number of CPUs?

cli/config.go
```diff
@@ -202,6 +203,28 @@ func (h hypervisor) defaultVCPUs() uint32 {
 	return uint32(h.DefaultVCPUs)
 }
 
+func (h hypervisor) defaultMaxVCPUs() uint32 {
+	numcpus := goruntime.NumCPU()
+	maxvcpus := vc.MaxQemuVCPUs()
+
+	if h.DefaultMaxVCPUs <= 0 {
```
egernst: can the default be negative?

cli/config.go
devimc: this is part of configuration file, if this value is <= 0 then the actual number of vCPUs will be used

cli/config.go
devimc: yes, same as `DefaultVCPUs`

virtcontainers/qemu.go
devimc: good catch! thanks

cli/config.go
devimc: yes, in case of user value is greater than the actual number of physical CPUs, we need to check this values doesn't exceed the maximum number of vCPUs supported by QEMU/KVM.
Adding a comment in the code.
Thanks

virtcontainers/container.go
```diff
@@ -938,14 +933,38 @@ func (c *Container) addResources() error {
...
-		if err := c.sandbox.hypervisor.hotplugAddDevice(uint32(vCPUs), cpuDev); err != nil {
+		data, err := c.sandbox.hypervisor.hotplugAddDevice(vCPUs, cpuDev)
+		if err != nil {
 			return err
 		}
 
-		return c.sandbox.agent.onlineCPUMem(uint32(vCPUs))
+		vcpusAdded, ok := data.(uint32)
+		if !ok {
+			return fmt.Errorf("Could not get the number of vCPUs added")
```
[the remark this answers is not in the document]
devimc: I'll include `data` in the error, thanks

cli/config.go
devimc: ohhh wait.. docker and kubernetes don't allow you to create containers with a number of CPUs greater than the actual number of physical CPUs, hence we don't need this condition, thanks

```
$ docker run --cpus 10 -ti centos bash
docker: Error response from daemon: Range of CPUs is from 0.01 to 8.00, as there are only 8 CPUs available.
```

virtcontainers/container.go
```diff
@@ -957,22 +976,45 @@ func (c *Container) removeResources() error {
 		return nil
 	}
 
-	vCPUs := utils.ConstraintsToVCPUs(c.config.Resources.CPUQuota, c.config.Resources.CPUPeriod)
+	// fetch current configuration
```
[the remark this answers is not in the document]
devimc: you're right, updating commit, thanks

cli/config.go
egernst: my question is why negative needs to be supported?  Seems == 0 should be enough here?

cli/config.go
```diff
@@ -202,6 +203,21 @@ func (h hypervisor) defaultVCPUs() uint32 {
...
+
+	if h.DefaultMaxVCPUs <= 0 || h.DefaultMaxVCPUs > int32(numcpus) {
+		// Don't exceed the maximum number of vCPUs supported by QEMU
+		if numcpus > int(maxvcpus) {
+			return maxvcpus
+		}
+		return uint32(numcpus)
+	}
+
+	return uint32(h.DefaultMaxVCPUs)
```
egernst: This logic is still a bit odd to me.  What happens if default-provided is not greater than numcpus, but it is still greater than  is greater than  maxvcpus?  
e.g., max-from-hypervisor is 100, number of physical cores is 120, but our configured default is 110?

Also I think you should s/QEMU/hypervisor?

devimc:
@egernst @jodh-intel @gnawux changes applied, thanks

cli/config.go
devimc: > e.g., max-from-hypervisor is 100, number of physical cores is 120, but our configured default is 110?

to avoid issues with KVM, the maximum number of CPUs supported by KVM is used, in this example `100`

> Also I think you should s/QEMU/hypervisor?

sure, I can do it, but in another PR, wdyt?

cli/config.go
devimc: yep, it should be enough, do you prefer to have an `uint32` here?

devimc: I used an `int32` to support negative values and be similar to `default_vcpus`

devimc:
@egernst  changes applied

cli/config.go
```diff
@@ -202,6 +203,25 @@ func (h hypervisor) defaultVCPUs() uint32 {
 	return uint32(h.DefaultVCPUs)
 }
 
+func (h hypervisor) defaultMaxVCPUs() uint32 {
+	numcpus := goruntime.NumCPU()
+	maxvcpus := vc.MaxQemuVCPUs()
+
+	// Don't exceed the maximum number of vCPUs supported by hypervisor
+	if h.DefaultMaxVCPUs >= maxvcpus {
+		return maxvcpus
```
egernst: what if the default is also greater than numcpus?  I think what you want is:
```
//don't exceed the number of physical CPUs. If a default is not provided, use the
// numbers of physical CPUs
if h.DefaultMaxVCPUS >= numcpus || h.DefaultMaxVCPUS == 0 {
      h.DefaultMaxVCPUs := numcpus
}

// Don't exceed the maximum number of vCPUs supported by hypervisor
if (h.DefaultMaxVCPUS > maxvcpus )  {
   return maxvcpus
}
  return h.DefaultMacVCPUS

devimc: actually is  the same, but your code looks better, updating pr, thanks

PR #298, opened 2018-05-11
Share pid namespace

### LINKED ISSUE #343: Containers in a pod should be able to share PID namespace.
Related to https://github.com/kata-containers/agent/issues/251

amshinde:
namespace: Check if pid namespaces need to be shared

k8s provides a configuration for sharing PID namespace
among containers. In case of crio and cri plugin, an infra
container is started first. All following containers are
supposed to share the pid namespace of this container.

In case a non-empty pid namespace path is provided for a container,
we check for the above condition while creating a container
and pass this out to the kata agent in the CreatContainer
request as SandboxPidNs flag. We clear out the PID namespaces
in the configuration passed to the kata agent.

Fixes #343


virtcontainers/kata_agent.go
```diff
@@ -769,16 +768,31 @@ func (k *kataAgent) createContainer(sandbox *Sandbox, c *Container) (p *Process,
...
 	grpcSpec.Root.Path = rootPath
 
+	sharedPidNs := false
+
+	// Sandbox pid will be set once the sandbox container is created.
+	// Check if pid namespace path is the same as the sandbox/infra container
+	if sandbox.state.Pid > 0 {
+		for _, ns := range grpcSpec.Linux.Namespaces {
+			if ns.Type == string(specs.PIDNamespace) {
+				if ns.Path == fmt.Sprintf("/proc/%d/ns/pid") {
```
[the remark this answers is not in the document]
amshinde: fixed

virtcontainers/sandbox.go
```diff
@@ -467,6 +470,8 @@ type Sandbox struct {
 	annotationsLock *sync.RWMutex
 
 	wg *sync.WaitGroup
+
+	//sandboxContainer *Container
```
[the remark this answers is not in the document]
amshinde: fixed.

amshinde:
@jodh-intel Yes, PR in the agent needs to land first.

jodh-intel:
Adding dnm as this is blocked on https://github.com/kata-containers/agent/pull/233.

amshinde:
Now that the agent PR is merged, I have revendored the agent protocol changes.
@sboeuf @devimc @bergwolf @jodh-intel PTAL

katabuilder:
PSS Measurement:
Qemu: 142766 KB
Proxy: 8110 KB
Shim: 10867 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  2012332 KB

virtcontainers/sandbox.go
```diff
@@ -1274,6 +1277,20 @@ func (s *Sandbox) decrementSandboxBlockIndex() error {
 	return nil
 }
 
+// setSandboxPid sets the Pid of the the shim process belonging to the
+// sandbox container as the Pid of the sandbox.
+func (s *Sandbox) setSandboxPid(pid int) error {
+	s.state.Pid = pid
+
+	// update on-disk state
+	err := s.storage.storeSandboxResource(s.id, stateFileType, s.state)
```
sboeuf: Can be simplified with:
```go
return s.storage.storeSandboxResource(s.id, stateFileType, s.state)
```

virtcontainers/sandbox.go
```diff
@@ -70,6 +70,9 @@ type State struct {
 
 	// PCI slot at which the block device backing the container rootfs is attached.
 	RootfsPCIAddr string `json:"rootfsPCIAddr"`
+
+	// Pid is the process id of the container process
```
sboeuf: Could you elaborate a bit here. The comment describing `Pid` is not verbose enough IMO.

virtcontainers/kata_agent.go
```diff
@@ -777,16 +776,43 @@ func (k *kataAgent) createContainer(sandbox *Sandbox, c *Container) (p *Process,
...
+				pidNsIndex = i
+				break
+			}
+		}
+	}
+
+	// Remove the pid namespace.
+	if pidNsIndex >= 0 {
+		grpcSpec.Linux.Namespaces = append(grpcSpec.Linux.Namespaces[:pidNsIndex], grpcSpec.Linux.Namespaces[pidNsIndex+1:]...)
+
```
sboeuf: You can remove this blank line.

virtcontainers/kata_agent.go
```diff
@@ -777,16 +776,43 @@ func (k *kataAgent) createContainer(sandbox *Sandbox, c *Container) (p *Process,
...
 
+	sharedPidNs := false
+	pidNsIndex := -1
+
+	// Sandbox pid will be set once the sandbox container is created.
+	// Check if pid namespace path is the same as the sandbox/infra container
+	if sandbox.state.Pid > 0 {
+		for i, ns := range grpcSpec.Linux.Namespaces {
+			if ns.Type == string(specs.PIDNamespace) {
+				if ns.Path == fmt.Sprintf("/proc/%d/ns/pid", sandbox.state.Pid) {
```
sboeuf: Are we sure this is the only form a namespace path could have ? What if a symlink is provided ? Is it possible to find the root namespace to make sure we don't miss some corner cases ?

virtcontainers/kata_agent.go
```diff
@@ -777,16 +776,43 @@ func (k *kataAgent) createContainer(sandbox *Sandbox, c *Container) (p *Process,
 	// We need to give the OCI spec our absolute rootfs path in the guest.
 	grpcSpec.Root.Path = rootPath
 
+	sharedPidNs := false
+	pidNsIndex := -1
+
+	// Sandbox pid will be set once the sandbox container is created.
+	// Check if pid namespace path is the same as the sandbox/infra container
+	if sandbox.state.Pid > 0 {
+		for i, ns := range grpcSpec.Linux.Namespaces {
```
sboeuf: Ok I understand that given the current architecture, this is the only thing we can do. But I think we should handle the detection at the `container.go` level, meaning more generically so that it can be reused across agent implementation and provided directly through a new boolean field of the `Container` structure.
Now, this is currently not feasible because this means we would introduce some OCI assumptions into `container.go`. The way to avoid that is to parse the namespaces provided by the OCI spec from `virtcontainers/pkg/oci/utils.go`, and fill the `ContainerConfig` with those information.
This way, we could analyze the namespaces expected for the container and determine if we have to share the PID namespace, or any other namespace from `container.go`.
@amshinde WDYT ?

amshinde: @sboeuf Sounds reasonable to check for shared namespaces in a generic way for reuse by the agents. I'll rework this.

virtcontainers/kata_agent.go
amshinde: yeah, I'll check for symlinks too

katabuilder:
PSS Measurement:
Qemu: 142137 KB
Proxy: 6744 KB
Shim: 8803 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  2011216 KB

virtcontainers/kata_agent.go
```diff
@@ -843,6 +848,43 @@ func (k *kataAgent) handleBlockVolumes(c *Container) []*grpc.Storage {
 	return volumeStorages
 }
 
+func (k *kataAgent) handlePidNamespace(grpcSpec *grpc.Spec, sandbox *Sandbox) (bool, error) {
+	sharedPidNs := false
+
+	for i, ns := range grpcSpec.Linux.Namespaces {
+		if ns.Type == string(specs.PIDNamespace) {
```
sboeuf: I would remove 3 indentation levels by doing:
```go
if ns.Type != string(specs.PIDNamespace) ||
        ns.Path == "" ||
        sandbox.state.Pid <= 0 {
        continue
}
```

virtcontainers/kata_agent.go
```diff
@@ -843,6 +848,43 @@ func (k *kataAgent) handleBlockVolumes(c *Container) []*grpc.Storage {
...
+						if ln != pidNsPath {
+							return sharedPidNs, fmt.Errorf("Pid namespace path other than sandbox")
+						}
+						sharedPidNs = true
+					}
+				}
+			}
+
+			// Remove pid namespace.
+			grpcSpec.Linux.Namespaces = append(grpcSpec.Linux.Namespaces[:i], grpcSpec.Linux.Namespaces[i+1:]...)
```
sboeuf: I think it would be more appropriate to handle the spec modification from `constraintGRPCSpec(grpcSpec)`, by providing `sharedPidNs` flag. This way, based on the value of `sharedPidNs`, the spec will be updated accordingly. Otherwise, we're modifying the spec in too many places IMO.

virtcontainers/kata_agent.go
```diff
@@ -843,6 +848,43 @@ func (k *kataAgent) handleBlockVolumes(c *Container) []*grpc.Storage {
...
+					pidNsPath := fmt.Sprintf("/proc/%d/ns/pid", sandbox.state.Pid)
+
+					//  Check if pid namespace path is the same as the sandbox
+					if ns.Path == pidNsPath {
+						sharedPidNs = true
+					} else {
+						// Check for symlinks
+						ln, err := filepath.EvalSymlinks(ns.Path)
+						if err != nil {
+							return sharedPidNs, err
```
jodh-intel: This is very deeply indented. I'd suggest reworking to invert the logi and continue if conditions are not `true`:

```
for i, ns := range grpcSpec.Linux.Namespaces {
    if ns.Type != string(specs.PIDNamespace) {
        continue
    }

    if ns.Path == "" {
        continue
    }

    ...

}
```

virtcontainers/kata_agent.go
```diff
@@ -843,6 +848,43 @@ func (k *kataAgent) handleBlockVolumes(c *Container) []*grpc.Storage {
...
+		if ns.Type == string(specs.PIDNamespace) {
+			if ns.Path != "" {
+				if sandbox.state.Pid > 0 {
+					pidNsPath := fmt.Sprintf("/proc/%d/ns/pid", sandbox.state.Pid)
+
+					//  Check if pid namespace path is the same as the sandbox
+					if ns.Path == pidNsPath {
+						sharedPidNs = true
+					} else {
+						// Check for symlinks
```
jodh-intel: I'd remove this comment as it's obvious from the function call what it's doing.

virtcontainers/kata_agent.go
```diff
@@ -843,6 +848,43 @@ func (k *kataAgent) handleBlockVolumes(c *Container) []*grpc.Storage {
...
+						ln, err := filepath.EvalSymlinks(ns.Path)
+						if err != nil {
+							return sharedPidNs, err
+						}
+
+						// We have arbitrary pid namespace path here
+						if ln != pidNsPath {
+							return sharedPidNs, fmt.Errorf("Pid namespace path other than sandbox")
```
jodh-intel: Log `ln` and `pidNsPath`?

virtcontainers/kata_agent.go
jodh-intel: Whenever I see this standard idiom, I always have to do a double-take as it just looks wrong to me :)

amshinde: The reason I did it this way was to keep all the logic related to PID namespaces in a single function. If I move it to `constraintGRPCSpec` , it still means that we have to loop over all the namespaces yet again.

sboeuf: Yes good point, then if we go with this solution, you should add some comments about the function itself, describing the fact that it modifies the grpcSpec structure.

virtcontainers/kata_agent.go
amshinde: done

virtcontainers/kata_agent.go
amshinde: done

virtcontainers/kata_agent.go
amshinde: @jodh-intel I  have added that

virtcontainers/kata_agent.go
amshinde: done

amshinde:
@jodh-intel @sboeuf PTAL

sboeuf:
Thanks for the changes @amshinde 
LGTM

jodh-intel:
lgtm

katabuilder:
PSS Measurement:
Qemu: 162205 KB
Proxy: 6771 KB
Shim: 8923 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1997856 KB

katabuilder:
PSS Measurement:
Qemu: 159478 KB
Proxy: 6683 KB
Shim: 8693 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1996004 KB

PR #245, opened 2018-04-23
[WIP]virtcontainers: pre add storage

WeiZhang555:
Signed-off-by: Zhang Wei <zhangwei555@huawei.com>

PR #301, opened 2018-05-13
virtcontainers: support pre-add storage for frakti

### LINKED ISSUE #50: virtcontainers: Implement Storage hotplug API
To help optimize the container boot process, virtcontainers needs to support hotplugging of storage and network devices.  We need to define and implement an API for attaching NICs and storage endpoints to a Pod/Sandbox, such as:
```go
func AddStorage(podID string) error {
        ...
}


AND/OR

func (p VCPod) AddStorage() error {
        ...
}
```

let's use this issue to define the API for storage  hotplug.  This should be closed once we have the design documented and implementation merged.

WeiZhang555:
Commit1: virtcontainers: same device can only attach once

Fix #50

This commit enhanced device manager in several aspects:

* allocate device ID in global device mapper to guarantee every device
has a uniq device ID and there won't be any ID collision.

* block device can only be attached once, first reason is that it
doesn't make sense to attach one block twice, second is for pre-add
storage for frakti, if one block is identified as hotplugged before, we
can skip and don't need to do it again. To identify if two block device
points to same one, we check the HostPath of device information, they
are regarded same when their HostPath is same.

Signed-off-by: Zhang Wei <zhangwei555@huawei.com>


WeiZhang555:
work in progress and replace https://github.com/kata-containers/runtime/pull/245

virtcontainers/container.go
```diff
@@ -300,6 +300,38 @@ func (c *Container) createContainersDirs() error {
...
+		if err != nil {
+			return nil, err
+		}
+
+		// Attach this block device, all other devices passed in the config have been attached at this point
+		if err := b.Attach(c.sandbox); err != nil {
+			return nil, err
+		}
+
+		// TODO: remove dependency of package drivers
```
jodh-intel: If this TODO is going to be retained, please could you change it to reference a github issue so we don't forget to do this.

Same comment for the other TODO's.

virtcontainers/device/drivers/block.go
```diff
@@ -88,53 +78,68 @@ func (device *BlockDevice) Attach(devReceiver api.DeviceReceiver) (err error) {
...
-			deviceLogger().WithError(err).Error("Failed to unplug block device")
-			return err
+	device.DevManager.Lock()
+	defer device.DevManager.Unlock()
+
+	// make sure this is the last device we are referencing
+	attached := device.DevManager.AttachedDevices(device)
+	switch len(attached) {
+	case 0:
+		// this should never happen, report error if it occurs
```
jodh-intel: Could you add a `return fmt.Errorf("BUG: ...")` or similar?

virtcontainers/device/manager/manager.go
```diff
@@ -21,59 +26,125 @@ const (
...
+	switch d := dev.(type) {
+	case *drivers.BlockDevice:
+		d.DevManager = dm
+		dm.devices[d.DeviceID()] = d
+		return d
+	case *drivers.VFIODevice:
+		d.DevManager = dm
+		dm.devices[d.DeviceID()] = d
+		return d
+	case *drivers.GenericDevice:
```
jodh-intel: As these three cases are identical, you could add explicit `fallthrough`'s. Or something like:

```
switch d := dev.(type) {
	case *drivers.BlockDevice:
	case *drivers.VFIODevice:
	case *drivers.GenericDevice:
	default:
		return nil
}

d.DevManager = dm
dm.devices[d.DeviceID()] = d
return d
```

virtcontainers/hyperstart_agent.go
```diff
@@ -231,6 +232,31 @@ func fsMapFromMounts(mounts []Mount) []*hyperstart.FsmapDescriptor {
 	return fsmap
 }
 
+func fsMapFromDevices(devices []api.Device) ([]*hyperstart.FsmapDescriptor, error) {
+	var fsmap []*hyperstart.FsmapDescriptor
+	for _, device := range devices {
+		// skip if it's not block device
```
jodh-intel: Nit: this comment is redundant I think as the test is very clear.

virtcontainers/kata_agent.go
```diff
@@ -625,13 +625,17 @@ func constraintGRPCSpec(grpcSpec *grpc.Spec) {
...
 func (k *kataAgent) appendDevices(deviceList []*grpc.Device, devices []api.Device) []*grpc.Device {
 	for _, device := range devices {
-		d, ok := device.(*drivers.BlockDevice)
-		if !ok {
+		if device.DeviceType() != config.DeviceBlock {
+			continue
+		}
+		d, ok := device.GetDeviceDrive().(*config.BlockDrive)
+		if !ok || d == nil {
+			k.Logger().Error("malformed block drive")
```
jodh-intel: I'd add details of the invalid device. Something like:

```
k.Logger().WithField("device", device).Error("malformed block drive")
```

Same comment to error below.

virtcontainers/sandbox.go
```diff
@@ -1334,11 +1349,43 @@ func togglePauseSandbox(sandboxID string, pause bool) (*Sandbox, error) {
...
+							}).Error("failed to remove vfio device for rolling back")
+					}
+				}
+			}
+		}()
+
+		// adding a group of VFIO devices
+		for _, dev := range vfioDevices {
+			if err = s.hypervisor.hotplugAddDevice(dev, vfioDev); err != nil {
+				s.Logger().WithError(err).
```
jodh-intel: You could define a new logger variable here to avoid the `WithFields()` duplication:

```
logger := s.Logger().WithFields(...)

defer func () {
    :
    logger.WithError(rollbackErr).Error("failed to remove vfio device for rolling back")
   :
}()

:

logger.WithError(err).Error("failed to hotplug VFIO device")
```

Same comment for code below.

virtcontainers/container.go
WeiZhang555: Sure

virtcontainers/device/manager/manager.go
WeiZhang555: `fallthrough` isn't supported in type switch in my test. But latter one looks good, will modify, thank!

virtcontainers/sandbox.go
WeiZhang555: This is great, thank!

WeiZhang555:
Add commit 2:
```
    virtcontainers: support pre add devices for container
    
    Fixes: #50
    
    This commit adds support for adding container devices before container created,
    this feature is mainly used by frakti.
    
    The idea is first we create a sandbox, before container created, we can call the
    `sandbox.PreAddDevices()` to create and attach devices to hypervisor from container
    config, then the devices and storage are stored in disk as `devices.json` and
    `mounts.json`.
    Next time when we create the container, the container can fetch mounts and devices
    from disk and restore them, so next time when it tries to attach them to hypervisor,
    the device manager can know they have been attached before and stopped them from
    being attached twice.
```

WeiZhang555:
ping @bergwolf Please check commit 2 with the new sandbox interface, see if this is what you want.

Note: new interface isn't fully tested, I am adding new test cases to check if it works. 

And by the way, did anyone update the gometalinter? I see lots of new warnings for gometalinter.

virtcontainers/device/manager/manager.go
WeiZhang555: Oops, sorry, just found that the latter one doesn't work either.
I need to use `d` in type switch, if I put the logic at below, it can't visit d anymore.
So have to keep the original way.

virtcontainers/sandbox.go
WeiZhang555: Oops, this is in for loop, and I need to display the detailed information for each dev, so I'm afraid a logger template won't be good enough.

virtcontainers/device/drivers/block.go
WeiZhang555: Addressed: and find a wrong test case via fixing this!

virtcontainers/interfaces.go
```diff
@@ -65,6 +67,7 @@ type VCSandbox interface {
 	SignalProcess(containerID, processID string, signal syscall.Signal, all bool) error
 	WinsizeProcess(containerID, processID string, height, width uint32) error
 	IOStream(containerID, processID string) (io.WriteCloser, io.Reader, io.Reader, error)
+	PreAddDevices(containerID ContainerConfig) ([]api.Device, error)
```
bergwolf: @WeiZhang555 Can you please decouple devices from container here? The original idea is to make storage and devices sandbox level resource, and AddStorage() would be able to add storage to an empty sandbox that has no containers. The AddStorage API need to do the following two things:
1. add storage resource info to sandbox, and save it in persistent storage
2. hotplug the storage devices if necessary

And it does not need to return a Device slice. The storage resources are saved in sandbox and VCSandbox.CreateContainer() can reference existing storage/devices with e.g. the device id.

some pseudo code of `PreAddDevices()`
```
func (s *Sandbox) PreAddStorage(devices []DeviceConfig) error {
        for _, d := range devices {
                s.hypervisor.hotplugDevice(d)
                s.storages = append(s.devices, d)
        }
        s.filesystem.saveDeviceInfo()
}
```

And in VCSandbox.CreateContainer(), we allow containerConfig.Mount to reference an existing device with its ID, and update corresponding container.Mount.BlockDevice to match the referenced device.

WeiZhang555: I agree with you in every point. The difficult point here is even after I did lots of reworks, the devices are still coupled to container in some point.
First one is device is created based on input of `config.DeviceInfo` which contains a `containerPath` in it so it has to bind to a container unless we modify the `DeviceInfo` construct.
Second is every device is store in `sandbox/containerDir/devices.json` and `sandbox/containerDir/mount.json`, we might need to move `sandbox/containerDir/devices.json` to `sandbox/devices.json`. 

I agree with you this is the right way, but I don't think I can finish it before 1.0. That's why I choose another way, keep devices.json under the `containerDir/` as before, and hotplug the devices for container in `PreAddStorage` function, then while we are creating the container, it can find the `device.json` under disk and skip the device hotplug process.

Do you want me to modify to the new way you described? It's a right approach, but there's a big risk that I can't finish it in several days. @bergwolf

bergwolf: @WeiZhang555 Please change it so that we do not need to maintain the PreAddStorage API in this form post 1.0.  And frakti won't be able to use it anyway. frakti wants to be able to add storage/device to an empty sandbox. If we cannot make it 1.0, let do post release like 1.1.0.

virtcontainers/container.go
WeiZhang555: Addressed in https://github.com/kata-containers/runtime/issues/306 and self-assigned.
Will handle this soon

WeiZhang555:
https://github.com/kata-containers/runtime/pull/301#discussion_r188158749

Discussed with @bergwolf , so I will continue to work on this towards the right way, and try to merge this after 1.0.  So sorry that it's impossible to catch release 1.0 now :-(

Ping @egernst to let you know

virtcontainers/device/config/config.go
```diff
@@ -90,11 +90,49 @@ type DeviceInfo struct {
...
+	DeviceInfo *DeviceInfo
 	SocketPath string
 	ID         string
 }
 
+// BlockDrive represents a block storage drive which may be used in case the storage
+// driver has an underlying block storage device.
+type BlockDrive struct {
+
+	// Path to the disk-image/device which will be used with this drive
```
devimc: comment should start with `File`, isn't it?

virtcontainers/device/config/config.go
```diff
@@ -90,11 +90,49 @@ type DeviceInfo struct {
...
+	Index int
+
+	// PCIAddr is the PCI address used to identify the slot at which the drive is attached.
+	PCIAddr string
+
+	// SCSI Address of the block device, in case the device is attached using SCSI driver
+	// SCSI address is in the format SCSI-Id:LUN
+	SCSIAddr string
+
+	// Path at which the device appears inside the VM, outside of the container mount namespace
```
devimc: and here with `VirtPath`

virtcontainers/device/config/config.go
```diff
@@ -90,11 +90,49 @@ type DeviceInfo struct {
...
+	VirtPath string
+
+	// Device path inside the container
```
devimc: and `ContainerPath`

virtcontainers/sandbox.go
```diff
@@ -633,6 +633,80 @@ func (s *Sandbox) IOStream(containerID, processID string) (io.WriteCloser, io.Re
...
+	c, err = newContainer(s, contConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = c.createContainersDirs(); err != nil {
+		return nil, err
+	}
+
+	// TODO: hotplug rootfs
```
devimc: is there an issue for this?

virtcontainers/sandbox.go
```diff
@@ -1357,17 +1478,49 @@ func (s *Sandbox) HotplugAddDevice(device api.Device, devType config.DeviceType)
...
 			return fmt.Errorf("device type mismatch, expect device type to be %s", devType)
 		}
-		return s.hypervisor.hotplugRemoveDevice(*vfioDevice, vfioDev)
+		removedDev := []*config.VFIODrive{}
+		var err error
+		defer func() {
+			// if err happens,roll back and remove added device!
+			if err != nil {
+				for _, dev := range removedDev {
+					if rollbackErr := s.hypervisor.hotplugAddDevice(dev, vfioDev); rollbackErr != nil {
```
devimc: comment says `// if err happens,roll back and remove added device!`, but here you are adding devices ?

virtcontainers/sandbox.go
```diff
@@ -633,6 +633,80 @@ func (s *Sandbox) IOStream(containerID, processID string) (io.WriteCloser, io.Re
...
+// it will be mainly used by frakti to boost startup time by doing this in parallel
+func (s *Sandbox) PreAddDevices(contConfig ContainerConfig) ([]api.Device, error) {
+	var (
+		c       *Container
+		devices []api.Device
+		err     error
+	)
+
+	defer func() {
+		if err != nil {
```
devimc: to reduce 1 indentation level and improve code readability, you should check `if err == nil` and return in all rolling back functions, what do you think?

virtcontainers/interfaces.go
sboeuf: I am gonna try to summarize my understanding here. We want an API that allows to hotplug blindly some storage to the VM without having to relate this to any container semantic.
Given the fact that the caller knows what it's doing, we could have a simple `AddStorage()` invoking some hypervisor function in order to hotplug what is needed as a first step.
As a second step, the caller (frakti most likely) will call later into `CreateContainer()`, providing the appropriate volume description expected for this container. If it wants the container to use the storage previously hotplugged, it will have to describe it through an appropriate `Mount` structure. We could add a specific field `NeedToPlug = false` to this structure to specify nothing needs to be done on the host, but a bind mount, or something else has to happen inside the guest.
Am I missing something or this could simply work ?

sboeuf: @bergwolf @WeiZhang555 ^^^

virtcontainers/device/drivers/block.go
```diff
@@ -88,53 +78,68 @@ func (device *BlockDevice) Attach(devReceiver api.DeviceReceiver) (err error) {
...
-
 	driveName, err := utils.GetVirtDriveName(index)
 	if err != nil {
 		return err
 	}
 
 	customOptions := device.DeviceInfo.DriverOptions
 	if customOptions != nil && customOptions["block-driver"] == "virtio-blk" {
-		device.VirtPath = filepath.Join("/dev", driveName)
-		device.PCIAddr = drive.PCIAddr
```
amshinde: Why  is the PCIAddr assignment deleted here? kata-agent relies on the PCIAddr for determining the virtio-block device name inside the guest. The assignment to Virtpath is not used by kata-agent, but was retained for the clear-container agent until that becomes obsolete.

virtcontainers/device/drivers/vfio.go
```diff
@@ -57,32 +60,29 @@ func (device *VFIODevice) Attach(devReceiver api.DeviceReceiver) error {
 
 	// Pass all devices in iommu group
 	for _, deviceFile := range deviceFiles {
-
 		//Get bdf of device eg 0000:00:1c.0
 		deviceBDF, err := getBDF(deviceFile.Name())
 		if err != nil {
 			return err
 		}
-
```
amshinde: you are not checking here if the device has been already attached here? We should avoid attaching VFIO devices more than once as well.

virtcontainers/device/manager/manager.go
```diff
@@ -21,59 +26,125 @@ const (
...
 		deviceLogger().WithField("device", path).Info("Device has not been passed to the container")
-		return drivers.NewGenericDevice(devInfo), nil
+		return drivers.NewGenericDevice(dm, &devInfo), nil
 	}
 }
 
-// NewDevices creates bundles of devices based on array of DeviceInfo
-func (dm *deviceManager) NewDevices(devInfos []config.DeviceInfo) ([]api.Device, error) {
+// NewDevice creates bundles of devices based on array of DeviceInfo
+func (dm *deviceManager) NewDevice(devInfo config.DeviceInfo) (api.Device, error) {
```
amshinde: Why not just retain NewDevices? that way you can lock and unlock just once.

virtcontainers/device/manager/manager.go
```diff
@@ -21,59 +26,125 @@ const (
...
+		dm.devices[dev.DeviceID()] = dev
+	}
+	return dev, err
+}
+
+// DeviceAttached returns a device list of all hotplugged devices with same hostpath
+// with input paramater device
+func (dm *deviceManager) AttachedDevices(device api.Device) []api.Device {
 	var devices []api.Device
+	devInfo := device.GetDeviceInfo()
```
amshinde: do you need a read lock here?

virtcontainers/device/drivers/vfio.go
```diff
@@ -82,13 +93,16 @@ func (device *VFIODevice) Attach(devReceiver api.DeviceReceiver) error {
...
+	device.DeviceInfo.Hotplugged = true
 	return nil
 }
 
 // Detach is standard interface of api.Device, it's used to remove device from some
 // DeviceReceiver
 func (device *VFIODevice) Detach(devReceiver api.DeviceReceiver) error {
+	device.DevManager.Lock()
+	defer device.DevManager.Unlock()
+	device.DeviceInfo.Hotplugged = true
```
amshinde: Shouldnt this be false?

virtcontainers/device/config/config.go
WeiZhang555: Correct. I'll fix it, thanks!

virtcontainers/sandbox.go
WeiZhang555: A lots of more work need to be done for this, so I added a TODO tag here.

virtcontainers/sandbox.go
WeiZhang555: The comment is wrong, thanks! It's a typical copy-and-paste error :wink:

virtcontainers/sandbox.go
WeiZhang555: I see your point, nice suggestion. Will fix it

virtcontainers/container.go
```diff
@@ -300,6 +300,38 @@ func (c *Container) createContainersDirs() error {
...
+	}
+
+	// Check if mount is a block device file. If it is, the block device will be attached to the host
+	// instead of passing this as a shared mount.
+	if c.checkBlockDeviceSupport() && stat.Mode&unix.S_IFBLK == unix.S_IFBLK {
+		b, err := c.sandbox.devManager.NewDevice(config.DeviceInfo{
+			HostPath:      m.Source,
+			ContainerPath: m.Destination,
+			DevType:       "b",
+			Major:         int64(stat.Rdev / 256),
```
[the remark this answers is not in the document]
WeiZhang555: Nice suggestion, thanks

virtcontainers/interfaces.go
bergwolf: @sboeuf You description is correct. One thing to note is that `NeedToPlug` is not necessary in device config and virtcontainers can decide if hotplug is needed depending on the device types.

virtcontainers/device/api/interface.go
```diff
@@ -51,11 +53,25 @@ type VhostUserDevice interface {
...
+	// Caller could cast the return value to device specific struct
+	// e.g. Block device returns *config.BlockDrive and
+	// vfio device returns *config.VFIODrive
+	GetDeviceDrive() interface{}
 }
 
 // DeviceManager can be used to create a new device, this can be used as single
 // device management object.
 type DeviceManager interface {
-	NewDevices(devInfos []config.DeviceInfo) ([]Device, error)
```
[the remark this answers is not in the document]
WeiZhang555: I don't see any necessity to create many devices in one API, I was hoping to break this API into smaller granularity and let upper function call to decide whether to operate on many or one.

virtcontainers/device/config/config.go
```diff
@@ -90,11 +90,49 @@ type DeviceInfo struct {
...
+	ContainerPath string
+}
+
+// VFIODrive represents a VFIO drive used for hotplugging
```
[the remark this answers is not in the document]
WeiZhang555: Any good suggestion?

I am intending to abstract vfio device's ID and BDF to a standard struct but I'm not good at naming it:

https://github.com/kata-containers/runtime/blob/813c8c3b56f12f8d130400a43fd6db6bd39e8c03/virtcontainers/device/drivers/vfio.go#L62-L79

virtcontainers/device/drivers/vfio.go
WeiZhang555: Oh, right. A critical mistake. Will modify

virtcontainers/device/drivers/block.go
WeiZhang555: It's not removed, just moved to struct `BlockDrive`:

 https://github.com/kata-containers/runtime/pull/301/files#diff-50f996b3aaf4e1ec24964719e24d75c5R115

virtcontainers/device/drivers/vfio.go
WeiZhang555: Will modify.

katabuilder:
PSS Measurement:
Qemu: 160245 KB
Proxy: 6817 KB
Shim: 11075 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1997004 KB

katabuilder:
PSS Measurement:
Qemu: 161536 KB
Proxy: 6787 KB
Shim: 11033 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1996756 KB

katabuilder:
PSS Measurement:
Qemu: 161875 KB
Proxy: 6630 KB
Shim: 13049 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1997624 KB

katabuilder:
PSS Measurement:
Qemu: 166184 KB
Proxy: 4718 KB
Shim: 10989 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  2003888 KB

virtcontainers/device/config/config.go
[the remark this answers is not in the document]
sboeuf: Maybe a simple `VFIODev` because VFIO can be related to any type of device, not only drives for storage.

WeiZhang555:
The CI still keeps failing, I think it should be a CI problem
```
panic: ERROR: docker daemon is not installed, not running, or not accessible to current user: Client:
 Version:      18.03.1-ce
 API version:  1.37
 Go version:   go1.9.5
 Git commit:   9ee9f40
 Built:        Thu Apr 26 07:19:31 2018
 OS/Arch:      linux/amd64
 Experimental: false
 Orchestrator: swarm
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.37/version: dial unix /var/run/docker.sock: connect: permission denied (error exit status 1)

goroutine 1 [running, locked to thread]:
github.com/kata-containers/runtime/cli.init.2()
    /tmp/jenkins/workspace/kata-containers-runtime-fedora-27-PR/go/src/github.com/kata-containers/runtime/cli/main_test.go:92 +0xe2c
FAIL    github.com/kata-containers/runtime/cli    0.097s
make: *** [Makefile:386: go-test] Error 1
Performing Post build task...
Could not match :Build was aborted  : False
Match found for :Build step 'Execute shell' marked build as failure : True
Logical operation result is TRUE
Running script  : #!/bin/bash
```

Maybe that's because we are not giving enough permission to run the test, so the test program can't visit "/etc/docker/daemon.sock"? 
I think this is a jenkins mis-config problem, who can take a look at this? @jodh-intel ?

grahamwhaley:
@chavafg and @mnaser I think will pick this up in a few hours. We had a CI failure/issue a couple of days ago, and had to rebuild the Jenkins CI on some new services (thanks @mnaser !) - but it looks like we still have a couple of teething issues to iron out.
This one does look like maybe a docker group perms issue for some of that.

jodh-intel:
Yes - that error is coming from the unit test setup code that needs to talk to docker:

- https://github.com/kata-containers/runtime/blob/master/cli/main_test.go#L92

chavafg:
I've checked the init files in the openstack plugin configuration, and we use 
`sudo gpasswd -a ${USER} docker` 
to add the jenkins user into the docker group, but I am not sure that init file is being run as jenkins, maybe that is the issue. 
For the moment I have changed to:
`sudo gpasswd -a jenkins docker`

Restarting the jobs.

WeiZhang555:
@chavafg Thanks! The issue is fixed. Other failure should be real PR problem, I will fix the codes.
Thanks again for fixing this so quickly!

virtcontainers/device/config/config.go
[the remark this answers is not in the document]
amshinde: Yeah, I would go with `VFIODev` too.

amshinde:
@WeiZhang555 Just rebase on master, CI failures should get fixed :)

amshinde:
@WeiZhang555 Can you split this PR into several commits?  Will make this 800+ line PR easier to digest and re-review.

WeiZhang555:
@amshinde I'll try but it's hard to split, one change links to too many other changes, and I hope to get every commit work, don't want to introduce a single broken commit.
And, this is only first step, actually I have more commits to append. I think I failed to make this PR small enough for reviewing :crying_cat_face:

virtcontainers/container.go
```diff
@@ -1076,18 +1111,30 @@ func (c *Container) removeDrive() (err error) {
...
+		//TODO: roll back?
+		return err
+	}
 	return nil
 }
 
 func (c *Container) detachDevices() error {
-	for _, device := range c.devices {
-		if err := device.Detach(c.sandbox); err != nil {
+	for _, dev := range c.devices {
```
jodh-intel: This function is identical to `attachDevices()` except for the different API call. It might be worth creating a `toggleAttachDevices(attach bool)` function that both `attachDevices()` and `detachDevices()` call to avoid the duplication and ensure they remain consistent.

virtcontainers/container.go
```diff
@@ -1076,18 +1111,30 @@ func (c *Container) removeDrive() (err error) {
...
+		if err := c.sandbox.devManager.DetachDevice(dev.ID, c.sandbox); err != nil {
+			if err == manager.ErrDeviceNotAttached {
+				// skip if device is already attached before
```
jodh-intel: Great to see that we can do this skip here. However, I'd be tempted to add a log call (at debug level but maybe info) so that it's clear when this scenario is being hit.

virtcontainers/device/config/config.go
```diff
@@ -87,12 +87,48 @@ type DeviceInfo struct {
...
 }
 
+// BlockDrive represents a block storage drive which may be used in case the storage
+// driver has an underlying block storage device.
+type BlockDrive struct {
+	// File is the path to the disk-image/device which will be used with this drive
+	File string
+
+	// Format of the drive
+	Format string
```
jodh-intel: What format is `Format`? :)

- Could this be a type instead?
- Could you give some examples in a comment maybe?

virtcontainers/container.go
WeiZhang555: Worth a try, I'll modify as per your suggestion.
Thank @jodh-intel for helping review!

virtcontainers/container.go
WeiZhang555: Sounds good

virtcontainers/device/config/config.go
WeiZhang555: This part is picked from original code, it's the specified block format when pass to qemu, I think "raw" is the only supported format currently, not sure if there's any plan to add more format.
I'll add some comments to address this.

WeiZhang555:
The PR is split into 5 commits, hope this can help make review easier.
@sboeuf @amshinde @bergwolf @jodh-intel

sboeuf:
@WeiZhang555 Thanks for doing this rework, I'll take a look later !

amshinde:
Thanks @WeiZhang555! Will take a look.

virtcontainers/sandbox.go
```diff
@@ -1389,18 +1419,49 @@ func (s *Sandbox) HotplugAddDevice(device api.Device, devType config.DeviceType)
...
 		}
-		_, err := s.hypervisor.hotplugRemoveDevice(*vfioDevice, vfioDev)
-		return err
+		removedDev := []*config.VFIODrive{}
+		var err error
+		defer func() {
+			// if err happens,roll back and add the removed devices back!
+			if err != nil {
+				for _, dev := range removedDev {
+					if _, rollbackErr := s.hypervisor.hotplugAddDevice(dev, vfioDev); rollbackErr != nil {
```
sboeuf: Do we care about rolling back such a thing ? I mean the real scenario here is that removing will not work for some reasons, and it will return an error. What we expect after this is a `delete` to work, and I don't think trying to re-add the device will help with that.

virtcontainers/sandbox.go
```diff
@@ -1364,12 +1363,43 @@ func togglePauseSandbox(sandboxID string, pause bool) (*Sandbox, error) {
...
 		}
-		_, err := s.hypervisor.hotplugAddDevice(*vfioDevice, vfioDev)
-		return err
+		addedDev := []*config.VFIODrive{}
+		var err error
+		defer func() {
+			// if err happens,roll back and remove added device!
+			if err != nil {
+				for _, dev := range addedDev {
+					if _, rollbackErr := s.hypervisor.hotplugRemoveDevice(dev, vfioDev); rollbackErr != nil {
```
sboeuf: Do we care about rolling back such a thing ? I mean the real scenario here is that adding a device will not work for some reasons, and it will return an error. What we expect after this is a `delete` to work, and I don't think trying to remove the device right now or later during the `delete` will help with that.

virtcontainers/device/config/config.go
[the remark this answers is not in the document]
sboeuf: Could you do the change regarding `VFIODrive` into `VFIODev` directly into this commit ? This way you fix it directly where it has been introduced.

virtcontainers/device/drivers/vfio.go
```diff
@@ -57,44 +56,62 @@ func (device *VFIODevice) Attach(devReceiver api.DeviceReceiver) error {
...
+		"device-type":  "vfio-passthrough",
+	}).Info("Device group attached")
+	device.DeviceInfo.Hotplugged = true
 	return nil
 }
 
 // Detach is standard interface of api.Device, it's used to remove device from some
 // DeviceReceiver
 func (device *VFIODevice) Detach(devReceiver api.DeviceReceiver) error {
+	device.DeviceInfo.Hotplugged = false
```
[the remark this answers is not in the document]
WeiZhang555: OK

virtcontainers/container.go
```diff
@@ -1022,7 +1022,7 @@ func (c *Container) hotplugDrive() error {
 
 	// Add drive with id as container id
 	devID := utils.MakeNameID("drive", c.id, maxDevIDSize)
-	drive := drivers.Drive{
+	drive := config.BlockDrive{
```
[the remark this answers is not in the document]
WeiZhang555: Yes, I think so.

virtcontainers/sandbox.go
WeiZhang555: I think you're right, will remove the rollback logic. Thanks!

virtcontainers/device/api/interface.go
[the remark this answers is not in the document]
WeiZhang555: Actually there is a reason:

I modified the container.go: https://github.com/kata-containers/runtime/pull/301/files#diff-8a56f5a1fa5971cc43e650ffa24f1e2bR613
```
+		for _, info := range contConfig.DeviceInfos {
 +			dev, err := sandbox.devManager.NewDevice(info)
 +			if err != nil {
 +				return &Container{}, err
 +			}

 +			c.devices = append(c.devices, ContainerDevice{
 +				ID:            dev.DeviceID(),
 +				ContainerPath: info.ContainerPath,
 +			})
  		}
  	}
```

The `ContainerPath` is in each `contConfig.DeviceInfos` and I need to do a map with each device created by `NewDevice` and `info.ContainerPath`,  if I use `NewDevices()` then there's no way to find which `info` links to which device, that's why I moved `NewDevices()` to a `NewDevice()`.  @amshinde @sboeuf

WeiZhang555:
@sboeuf @amshinde comments are addressed, thank you! Please take another look at it.

virtcontainers/container.go
```diff
@@ -1076,18 +1111,30 @@ func (c *Container) removeDrive() (err error) {
...
 func (c *Container) attachDevices() error {
-	for _, device := range c.devices {
-		if err := device.Attach(c.sandbox); err != nil {
+	for _, dev := range c.devices {
+		if err := c.sandbox.devManager.AttachDevice(dev.ID, c.sandbox); err != nil {
+			if err == manager.ErrDeviceAttached {
+				// skip if device is already attached before
+				continue
+			}
 			return err
```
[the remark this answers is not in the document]
WeiZhang555: According to comments: https://github.com/kata-containers/runtime/pull/301#discussion_r200522277 from @sboeuf , "runtime delete" will handle the rollback correctly

virtcontainers/device/config/config.go
bergwolf: Possible future format values are `rbd`, `iscsi`, `nbd` etc. We can add a slice of `key`:`value` mappings to allow to pass extra options. That is not required for this PR though. We can add it later.

virtcontainers/container.go
```diff
@@ -566,6 +562,38 @@ func newContainer(sandbox *Sandbox, contConfig ContainerConfig) (*Container, err
...
+		}
+
+		// Check if mount is a block device file. If it is, the block device will be attached to the host
+		// instead of passing this as a shared mount.
+		if c.checkBlockDeviceSupport() && stat.Mode&unix.S_IFBLK == unix.S_IFBLK {
+			b, err := c.sandbox.devManager.NewDevice(config.DeviceInfo{
+				HostPath:      m.Source,
+				ContainerPath: m.Destination,
+				DevType:       "b",
+				Major:         int64(unix.Major(stat.Rdev)),
```
bergwolf: Nits: it seems that `Major` and `Minor` is never used and can be cleaned up entirely. From VM point of view, there is no need to ask for host device major and minor. So it's better to just remove them from the device API. Not required for this PR though.

virtcontainers/device/drivers/vfio.go
```diff
@@ -57,44 +60,71 @@ func (device *VFIODevice) Attach(devReceiver api.DeviceReceiver) error {
...
 		}
+		device.vfioDevs = append(device.vfioDevs, vfio)
+	}
 
-		deviceLogger().WithFields(logrus.Fields{
-			"device-group": device.DeviceInfo.HostPath,
-			"device-type":  "vfio-passthrough",
-		}).Info("Device group attached")
+	// hotplug a VFIO device is actually hotplugging a group of iommu devices
+	if err := devReceiver.HotplugAddDevice(device, config.DeviceVFIO); err != nil {
```
bergwolf: Why is this preferred than the original hotplug one-by-one interface? The interface is different than other device hotplug interfaces so I assume there is a valid reason for it. Please document it a bit in the above comments.

virtcontainers/container.go
WeiZhang555: I think `Major` and `Minor` is better option for identifying a device than a path, because we can have  lots of alias, besides the `Major` and `Minor` is from OCI spec but HostPath isn't

virtcontainers/device/drivers/vfio.go
WeiZhang555: That's because `HotplugAddDevice` accepts parameter of "device", in VFIO scenario, it's actually not a single device but all devices in same iommu group, and I need to save all the devices inside `device`. 
That's why I  modified the interface, and I think the comments from above "// Pass all devices in iommu group" already explained this well.

bergwolf: Yes, I understand that. I was asking why `devReceiver.HotplugAddDevice` must take a slice of VFIO devices rather than just one device (like the other hotplug handlers) and then we can do iteration in the loop here like the original code does. Because in the end, you still call `s.hypervisor.hotplugAddDevice` in a loop, just at a different place.

virtcontainers/container.go
bergwolf: Yes, I get that. I was thinking to push the burden of converting from <Major, Minor> to a HostPath to the client , because they do not make sense for VMs, and for remove block devices, there is no host device major/minor.  Once that is done, we can remove `Major`/`Minor` from the API.

amshinde:
@WeiZhang555 One minor comment.
LGTM otherwise.

virtcontainers/container.go
WeiZhang555: We can discuss more later in seperate PR :-)

virtcontainers/device/drivers/vfio.go
WeiZhang555: That will look weird, you see the `HotplugAddDevice` takes a `device` parameter, if I do the iteration here, we need to save one vfioDev inside `device`, and in next iteration, we need to modify the device struct and replace the original vfio dev with another dev, that will be annoying.

Besides that, for unplug part, if we don't save slice(all vfio devices in same IOMMU group), we need to do another "scan" again to get all vfio devices, and then another for loop to unplug them one by one. In my opinion the "scan" part can be avoided by saving all vfio dev into a slice in `device`, which can make the unplug logic very simple.

virtcontainers/device/drivers/vfio.go
```diff
@@ -57,44 +60,71 @@ func (device *VFIODevice) Attach(devReceiver api.DeviceReceiver) error {
...
+	// hotplug a VFIO device is actually hotplugging a group of iommu devices
+	if err := devReceiver.HotplugRemoveDevice(device, config.DeviceVFIO); err != nil {
+		deviceLogger().WithError(err).Error("Failed to remove device")
+		return err
+	}
+
+	deviceLogger().WithFields(logrus.Fields{
+		"device-group": device.DeviceInfo.HostPath,
+		"device-type":  "vfio-passthrough",
+	}).Info("Device group attached")
```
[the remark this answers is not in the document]
WeiZhang555: Nice catch, fixed @amshinde

bergwolf:
CI failed on kata-containers/tests#513 and kata-containers/tests#506

WeiZhang555:
```
time="2018-07-23T07:59:23Z" level=error msg="open /sys/module/kvm_intel/parameters/nested: no such file or directory" name=kata-runtime pid=7299 source=runtime
open /sys/module/kvm_intel/parameters/nested: no such file or directory
```

I think the CI machine isn't set up properly, the nested virtualization should be open for testing in VM

sboeuf:
Thanks @WeiZhang555 for addressing all the comments.
This LGTM, and will be ready to be merged once you rebase it and get the CI passing ;)

grahamwhaley:
CI's failing. The F27 at least looks like it might be a network glitch, so I will re-kick that one and we'll see how it goes:
```
package manager update test check apt-get update 
  should not fail
  /tmp/jenkins/workspace/kata-containers-runtime-fedora-27-PR/go/src/github.com/kata-containers/tests/integration/docker/package_manager_test.go:38
Running command '/usr/bin/docker [docker run --runtime kata-runtime --rm --name CMm81EKhiFzL03HMUgOn71pPRuU4Fh debian apt-get -y update]'
Killing process timeout reached '60' seconds
Running command '/usr/bin/docker [docker ps -a -f name=CMm81EKhiFzL03HMUgOn71pPRuU4Fh --format {{.Status}}]'
[docker ps -a -f name=CMm81EKhiFzL03HMUgOn71pPRuU4Fh --format {{.Status}}]
Timeout: 60 seconds
Exit Code: 0
Stdout: Up 57 seconds

Stderr: 

• Failure [60.063 seconds]
package manager update test
/tmp/jenkins/workspace/kata-containers-runtime-fedora-27-PR/go/src/github.com/kata-containers/tests/integration/docker/package_manager_test.go:15
  check apt-get update
  /tmp/jenkins/workspace/kata-containers-runtime-fedora-27-PR/go/src/github.com/kata-containers/tests/integration/docker/package_manager_test.go:37
    should not fail [It]
    /tmp/jenkins/workspace/kata-containers-runtime-fedora-27-PR/go/src/github.com/kata-containers/tests/integration/docker/package_manager_test.go:38

    Expected
        <int>: -1
    to be zero-valued
```

WeiZhang555:
I think it turns out to be some random test case failure.

@bergwolf, do you want to take a look at this?

virtcontainers/device/api/interface.go
```diff
@@ -27,35 +27,43 @@ func DeviceLogger() *logrus.Entry {
...
+	// e.g. Block device returns *config.BlockDrive and
+	// vfio device returns []*config.VFIODev
+	GetDeviceInfo() interface{}
+	// IsAttached checks if the device is attached
+	IsAttached() bool
 }
 
 // DeviceManager can be used to create a new device, this can be used as single
 // device management object.
 type DeviceManager interface {
```
bergwolf: @WeiZhang555 I'm wondering how this can be consumed by frakti. Currently sandbox has its private device manager and that will manage all device hotplug internally. If frakti wants to use this API, it will have to create a device manager on its own -- so that it can call functions in the DeviceManager interface. But that seems to conflict with the internal device manager in the sandbox structure. Would you please add device manager as an argument to `CreateSandbox`/`RunSandbox` and `FetchSandbox`, and use that to set sandbox.devManager if it's non-nil?

Another way is to make `sandbox.HotplugAddDevice`/`sandbox.HotplugRemoveDevice` top level API and use device manager underneath. But since you have already chosen a different path, there is no need to go back there. Thanks for the hard work!

WeiZhang555: @bergwolf 

> Another way is to make sandbox.HotplugAddDevice/sandbox.HotplugRemoveDevice top level API and use device manager underneath. 

This is what I intended to do in seperate PR , and because this PR is too huge, my initial plan was only keeping the rework part in this one.

What do you think of the `sandbox.HotplugAddDevice/sandbox.HotplugRemoveDevice` ?

WeiZhang555: And `sandbox.HotplugAddDevice` will be simply wrapper of `DeviceManager.HotplugAddDevice`

bergwolf: Sounds good to me. Thanks!

bergwolf:
LGTM! Thanks @WeiZhang555 !

WeiZhang555:
I'll merge this as we have 3 LGTM now, there will be a new Sandbox API for Frakti support in following PR, I'll do it ASAP.
Thanks everyone for helping review!

sboeuf:
@WeiZhang555 I see that you merged this PR. I agree it was ready from a code review perspective, but usually, we don't merge when the CI is not passing... Zuul CI is not needed for now as it might not be completely stable, but `jenkins-ci-ubuntu-16-04` should definitely be passing before we merge anything. Otherwise, we might break the master branch.

Issue #61, opened 2018-03-15
Add support for ephemeral volumes 


harche:
Hi,

As of now all volumes are created on the host and passed to VM via 9pfs. But k8s allows you to create [ephemeral volumes](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) Also, these volumes can be backed by a ramdisk.  Ephemeral volumes, as the name indicates, live and die with the pod. There is no reason to use 9pfs for this type of volume. 

Kata needs to support these volumes by creating tempfs based volume inside of the VM. 

The possible approach that I can think of,
1. Detect if the volume getting attached is backed by a ramdisk (temfs)
2. When VM boots, instruct init to create tempfs inside of VM
3. Use the ramdisk created inside VM in step 2 with the containers of the pod. 

Any thoughts?

bergwolf:
The kata agent protocol already supports it. Please see https://github.com/kata-containers/agent/blob/master/protocols/grpc/agent.proto#L194

We can implement a new type of storage driver (e.g. `tmpfs`) that instruct the kata agent to setup a tmpfs mount point at `pb.Storage.Mountpoint` and reference it via the container oci spec.

These need to be implemented in both the kata agent and kata runtime though.

harche:
@bergwolf Sounds good. I will try to submit PR for supporting ephemeral volumes.

egernst:
@harche - checking to see if you've made any progress here.

harche:
@egernst Sorry I was away for medical reasons. Just back today. I will start working on it.

harche:
Docker and Kubernetes take different approaches to attach emphemeral volumes (backed by tmpfs) to the container. 

When an empheral volume is attached using kubernetes (by setting empty.medium to `"Memory"` in the yaml as described [here](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir)), the corresponding docker container's config.v2.json looks this, 
```json
"MountPoints": {
		"/cache": {
			"Source": "/var/lib/kubelet/pods/366c3a75-4869-11e8-b479-507b9ddd5ce4/volumes/kubernetes.io~empty-dir/cache-volume",
			"Destination": "/cache",
			"RW": true,
			"Name": "",
			"Driver": "",
			"Type": "bind",
			"Propagation": "rprivate",
			"Spec": {
				"Type": "bind",
[...]

PR #307, opened 2018-05-14
virtcontainers: Add support for ephemeral volumes

harche:
Ephemeral volumes should not be passed at 9pfs mounts.
They should be created inside the VM.

This patch disables ephemeral volumes from getting
mounted as 9pfs from the host and instead a corresponding
tmpfs is created inside the VM.

Fixes : #61

Signed-off-by: Harshal Patil <harshal.patil@in.ibm.com>

### DIFF
--- a/cli/create.go
+++ b/cli/create.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 
 	vc "github.com/kata-containers/runtime/virtcontainers"
+	"github.com/kata-containers/runtime/virtcontainers/device/config"
 	"github.com/kata-containers/runtime/virtcontainers/pkg/oci"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
@@ -103,7 +104,6 @@ func create(containerID, bundlePath, console, pidFilePath string, detach bool,
 	disableOutput := noNeedForOutput(detach, ociSpec.Process.Terminal)
 
 	var process vc.Process
-
 	switch containerType {
 	case vc.PodSandbox:
 		process, err = createSandbox(ociSpec, runtimeConfig, containerID, bundlePath, console, disableOutput)
@@ -259,6 +259,19 @@ func createContainer(ociSpec oci.CompatOCISpec, containerID, bundlePath,
 		return vc.Process{}, err
 	}
 
+	// Add the ephemeral device if ephemeral volume
+	// has to be attached to the container. For the given pod
+	// ephemeral volume is created only once backed by tmpfs
+	// inside the VM. For successive containers of the same
+	// pod the already existing volume is reused.
+	for _, mnt := range contConfig.Mounts {
+		if IsEphemeralStorage(mnt.Source) {
+			deviceInfo := config.DeviceInfo{}
+			deviceInfo.DevType = "e"
+			deviceInfo.ContainerPath = mnt.Source
+			contConfig.DeviceInfos = append(contConfig.DeviceInfos, deviceInfo)
+		}
+	}
 	_, c, err := vci.CreateContainer(sandboxID, contConfig)
 	if err != nil {
 		return vc.Process{}, err
--- a/cli/utils.go
+++ b/cli/utils.go
@@ -15,7 +15,10 @@ import (
 	"strings"
 )
 
-const unknown = "<<unknown>>"
+const (
+	unknown     = "<<unknown>>"
+	k8sEmptyDir = "kubernetes.io~empty-dir"
+)
 
 // variables to allow tests to modify the values
 var (
@@ -41,6 +44,26 @@ func getFileContents(file string) (string, error) {
 	}
 
 	return string(bytes), nil
+}
+
+// IsEphemeralStorage returns true if the given path
+// to the storage belongs to kubernetes ephemeral storage
+//
+// This method depends on a specific path used by k8s
+// to detect if it's of type ephemeral. As of now,
+// this is a very k8s specific solution that works
+// but in future there should be a better way for this
+// method to determine if the path is for ephemeral
+// volume type
+func IsEphemeralStorage(path string) bool {
+	splitSourceSlice := strings.Split(path, "/")
+	if len(splitSourceSlice) > 1 {
+		storageType := splitSourceSlice[len(splitSourceSlice)-2]
+		if storageType == k8sEmptyDir {
+			return true
+		}
+	}
+	return false
 }
 
 func getKernelVersion() (string, error) {
--- a/cli/utils_test.go
+++ b/cli/utils_test.go
@@ -38,6 +38,20 @@ func TestFileExists(t *testing.T) {
 		fmt.Sprintf("File %q should exist", file))
 }
 
+func TestIsEphemeralStorage(t *testing.T) {
+	sampleEphePath := "/var/lib/kubelet/pods/366c3a75-4869-11e8-b479-507b9ddd5ce4/volumes/kubernetes.io~empty-dir/cache-volume"
+	isEphe := IsEphemeralStorage(sampleEphePath)
+	if !isEphe {
+		t.Fatalf("Unable to correctly determine volume type")
+	}
+
+	sampleEphePath = "/var/lib/kubelet/pods/366c3a75-4869-11e8-b479-507b9ddd5ce4/volumes/cache-volume"
+	isEphe = IsEphemeralStorage(sampleEphePath)
+	if isEphe {
+		t.Fatalf("Unable to correctly determine volume type")
+	}
+}
+
 func TestGetFileContents(t *testing.T) {
 	type testData struct {
 		contents string
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -418,6 +418,7 @@ func (c *Container) createContainersDirs() error {
 // available when we will need to unmount those mounts.
 func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) ([]Mount, error) {
 	var sharedDirMounts []Mount
+
 	for idx, m := range c.mounts {
 		if isSystemMount(m.Destination) || m.Type != "bind" {
 			continue
@@ -433,6 +434,17 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
 		var stat unix.Stat_t
 		if err := unix.Stat(m.Source, &stat); err != nil {
 			return nil, err
+		}
+
+		// If there exists an ephemeral device corresponding
+		// to the given mount point we will skip the shared mounts
+		//
+		// We are not going to bind mount host ephemeral directory
+		// provided by k8s, instead at later stage we are going to
+		// create a tmpfs inside a VM and use it to bind mount to the
+		// containers of the pod.
+		if utils.IsEphemeralDevice(c.config.DeviceInfos, m.Source) {
+			continue
 		}
 
 		// Check if mount is a block device file. If it is, the block device will be attached to the host
@@ -476,7 +488,6 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
 		if err := bindMount(m.Source, mountDest, false); err != nil {
 			return nil, err
 		}
-
 		// Save HostPath mount value into the mount list of the container.
 		c.mounts[idx].HostPath = mountDest
 
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -19,6 +19,7 @@ import (
 	kataclient "github.com/kata-containers/agent/protocols/client"
 	"github.com/kata-containers/agent/protocols/grpc"
 	"github.com/kata-containers/runtime/virtcontainers/device/api"
+	"github.com/kata-containers/runtime/virtcontainers/device/config"
 	"github.com/kata-containers/runtime/virtcontainers/device/drivers"
 	vcAnnotations "github.com/kata-containers/runtime/virtcontainers/pkg/annotations"
 	ns "github.com/kata-containers/runtime/virtcontainers/pkg/nsenter"
@@ -51,6 +52,7 @@ var (
 	sharedDir9pOptions    = []string{"trans=virtio,version=9p2000.L", "nodev"}
 	shmDir                = "shm"
 	kataEphemeralDevType  = "ephemeral"
+	ephemeralPath         = filepath.Join(kataGuestSandboxDir, kataEphemeralDevType)
 )
 
 // KataAgentConfig is a structure storing information needed
@@ -775,11 +777,37 @@ func (k *kataAgent) createContainer(sandbox *Sandbox, c *Container) (p *Process,
 		return nil, err
 	}
 
+	var devInfos []config.DeviceInfo
+	for _, devInfo := range c.config.DeviceInfos {
+		if devInfo.DevType == "e" {
+			filename := filepath.Join(ephemeralPath, filepath.Base(devInfo.ContainerPath))
+			epheStorage := &grpc.Storage{
+				Driver:     kataEphemeralDevType,
+				Source:     "tmpfs",
+				Fstype:     "tmpfs",
+				MountPoint: filename,
+			}
+			ctrStorages = append(ctrStorages, epheStorage)
+
+		} else {
+			devInfos = append(devInfos, devInfo)
+		}
+	}
+
 	// Handle container mounts
 	newMounts, err := c.mountSharedDirMounts(kataHostSharedDir, kataGuestSharedDir)
 	if err != nil {
 		return nil, err
 	}
+
+	// Modify the mount source for ephemeral volume
+	for idx, mnt := range ociSpec.Mounts {
+		if utils.IsEphemeralDevice(c.config.DeviceInfos, mnt.Source) {
+			ociSpec.Mounts[idx].Source = filepath.Join(ephemeralPath, filepath.Base(mnt.Source))
+		}
+	}
+
+	c.config.DeviceInfos = devInfos
 
 	// We replace all OCI mount sources that match our container mount
 	// with the right source path (The guest one).
--- a/virtcontainers/utils/utils.go
+++ b/virtcontainers/utils/utils.go
@@ -12,6 +12,8 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+
+	"github.com/kata-containers/runtime/virtcontainers/device/config"
 )
 
 const cpBinaryName = "cp"
@@ -65,6 +67,18 @@ func ReverseString(s string) string {
 	}
 
 	return string(r)
+}
+
+// IsEphemeralDevice returns true if there exists an ephemeral
+// device in the configuration who's ContainerPath matches with
+// the mount source
+func IsEphemeralDevice(deviceInfos []config.DeviceInfo, source string) bool {
+	for _, devInfo := range deviceInfos {
+		if devInfo.DevType == "e" && devInfo.ContainerPath == source {
+			return true
+		}
+	}
+	return false
 }
 
 // CleanupFds closed bundles of open fds in batch
--- a/virtcontainers/device/config/config.go
+++ b/virtcontainers/device/config/config.go
@@ -60,6 +60,8 @@ type DeviceInfo struct {
 	// p - FIFO
 	// b - block(buffered) special file
 	// More info in mknod(1).
+	// also,
+	// e - ephemeral volume
 	DevType string
 
 	// Major, minor numbers for device.
--- a/virtcontainers/filesystem.go
+++ b/virtcontainers/filesystem.go
@@ -319,7 +319,6 @@ func (fs *filesystem) fetchDeviceFile(fileData []byte, devices *[]api.Device) er
 			}
 			tempDevices = append(tempDevices, &device)
 			l.Infof("Generic device unmarshalled [%v]", device)
-
 		default:
 			return fmt.Errorf("Unknown device type, could not unmarshal")
 		}


harche:
There will be the corresponding PR on agent project.

harche:
Corresponding PR in `agent`, https://github.com/kata-containers/agent/pull/236

virtcontainers/container.go
```diff
@@ -362,12 +363,28 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
 		filename := fmt.Sprintf("%s-%s-%s", c.id, hex.EncodeToString(randBytes), filepath.Base(m.Destination))
 		mountDest := filepath.Join(hostSharedDir, c.sandbox.id, filename)
 
-		if err := bindMount(m.Source, mountDest, false); err != nil {
-			return nil, err
+		var ephiVol = false
+		splitSourceSlice := strings.Split(m.Source, "/")
+		if len(splitSourceSlice) > 2 {
```
egernst: Can you clarify the expected format of the original Source str in a comment?

egernst:
@WeiZhang555 @amshinde  - PTAL

WeiZhang555:
@harche Sorry I didn't participate the discussion before. Need to confirm the cases first, I see in the issue #61 , you mentioned the volume is backed by tmpfs
```
# mount |grep 366c3a75
tmpfs on /var/lib/kubelet/pods/366c3a75-4869-11e8-b479-507b9ddd5ce4/volumes/kubernetes.io~empty-dir/cache-volume type tmpfs (rw,relatime)
tmpfs on /var/lib/kubelet/pods/366c3a75-4869-11e8-b479-507b9ddd5ce4/volumes/kubernetes.io~secret/default-token-2w6d4 type tmpfs (rw,relatime)
```

So in this PR, you only handled `kubernetes.io~empty-dir` but let the `kubernetes.io~secret/` go, I guess it's because we need to write data to "kubernetes.io~secret"from host so we should keep it `bind-mounted`. 
The question here is, are we sure there's no scenario we will write data to `kubernetes.io~empty-dir` from host? Because if this happens it will fail after we merge this.
Another concern is, is this code too "K8S" specific?

virtcontainers/container.go
```diff
@@ -362,12 +363,28 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
...
+			k8sStorageType := splitSourceSlice[len(splitSourceSlice)-2]
+			if k8sStorageType == "kubernetes.io~empty-dir" {
```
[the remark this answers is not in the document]
WeiZhang555: Yep, this is also what I am worrying about now, it's a bit too "K8S" specific.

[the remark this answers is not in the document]
bergwolf: Can we move the detection to kata cli to keep the knowledge out of virtcontainers? We can add a tmpfs device type for it in container DeviceInfos and kata cli is responsible for detecting ephemeral volumes and construct proper deviceinfo there.

virtcontainers/container.go
harche: @egernst A typical ephemeral storage path looks like, `/var/lib/kubelet/pods/366c3a75-4869-11e8-b479-507b9ddd5ce4/volumes/kubernetes.io~empty-dir/cache-volume`

I will update the code with this sample path in code comments.

virtcontainers/container.go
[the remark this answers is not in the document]
harche: @WeiZhang555 any solution to handle ephemeral volumes will be very specific to the k8s. Ephemeral volumes are k8s concept after all. But we are interested in bringing them in kata because it helps to increase the isolation with the host. If you have a scenario where you need to share the data between say, k8s init and regular containers of the pod then those containers can share the data amongst themselves without explicitly putting it on the host filesystem. 

@jodh-intel ephemeral volumes are define in k8s yaml at pod level like this, 

```
apiVersion: v1
kind: Pod
metadata:
  name: myapp-pod
  annotations:
  labels:
    app: myapp
spec:
  containers:
  # Application Container
  - name: myapp-container
    volumeMounts:
    - mountPath: /cache
      name: cache-volume
    imagePullPolicy: IfNotPresent 
    image: localhost:5000/app_image:latest
    command: ['sh']
  initContainers:
  - name: data-container
    volumeMounts:
    - mountPath: /cache
      name: cache-volume
    imagePullPolicy: IfNotPresent 
    image: localhost:5000/data_image:latest
    command: ['sh', '-c', 'cp /test.gpg /cache/']
  volumes:
  - name: cache-volume
    emptyDir:
      medium: "Memory"
```

So far the SanboxConfig.Volumes (https://github.com/kata-containers/runtime/blob/master/virtcontainers/sandbox.go#L444) doens't register this volume type. 

This is the reason I resorted to parsing the OCI mounts of the invidual container mounts to infer if there are any k8s ephemeral volumes.

katabuilder:
PSS Measurement:
Qemu: 142172 KB
Proxy: 6730 KB
Shim: 13117 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  2010968 KB

virtcontainers/utils/utils.go
```diff
@@ -67,6 +68,19 @@ func ReverseString(s string) string {
 	return string(r)
 }
 
+// From the given path to the storage determine
+// if it's a kubernetes ephemeral storage
+func IsEphemeralStroage(path string) bool {
```
bergwolf: The check is too k8s specific to be in the virtcontainers library. Please move it to kata cli and use a new device type to describe the ephemeral storage in container config.

virtcontainers/kata_agent.go
```diff
@@ -743,6 +744,18 @@ func (k *kataAgent) createContainer(sandbox *Sandbox, c *Container) (p *Process,
 		return nil, err
 	}
 
+	for _, m := range c.mounts {
+		if utils.IsEphemeralStroage(m.Source) {
+			filename := filepath.Join(ephemeralPath, filepath.Base(m.Source))
+			teststorage := &grpc.Storage{
```
bergwolf: `teststorage` is not a good variable name :)

harche: :) sorry. Leftover of an aftermath. I will clean it and update the patch.

amshinde:
@harche Are ephemeral volumes meant to be shared, that is created once for the sandbox and bind-mounted by the rest of the containers?

harche:
@amshinde Yes.

virtcontainers/utils/utils.go
harche: @bergwolf Need some clarification so just to make sure we are on the same page. 

When you say use a new DeviceType, do you mean the one declared here - https://github.com/kata-containers/runtime/blob/master/virtcontainers/device/config/config.go#L19 ?

Also, you suggested to use container config to describe the storage. Staying within the cli package I could imagine the new device can be added at this point, https://github.com/kata-containers/runtime/blob/master/cli/create.go#L261 (by appending to contConfig.DeviceInfos). Because after that point the code jumps into virtcontainer library.

Coming back to the DeviceType, it's used in the driver implementions in virtcontainer library https://github.com/kata-containers/runtime/tree/master/virtcontainers/device/drivers 

All the uses of the DeviceType including it's declaration is in virtcontainers only. So this isn't helping us taking the code away from virtcontainers and putting it in cli package. 

To me this seems like a overkill to implement all those methods in those drivers when we aren't even going to use any real device from the host. 

Maybe I am missing something, would you like to elaborate more on what you have in mind here?

bergwolf: Yes, your understand is correct. I asked for it because this is the device/storage model of virtcontainers. We want all devices/storage to follow the same flow. ephemeral is just another storage. I don't think we should make it a special case and handle it differently.

Besides, `IsEphemeralStroage` is totally bound to k8s. You cannot create an ephemeral storage unless the source path has strict k8s empty dir pattern. That is not flexible at all. We want a generic API that one can create ephemeral storage with any source path. In fact, for ephemeral storage, the source path should not matter at all.

WeiZhang555:
I don't feel very good about this "improvement" though it can be useful in K8S scenarios.
All the judgement of "ephemeral volumes" is based on the file path which is not a really good decision, this is extremely customized code and isn't generic at all, even after move the code from virtcontainers/ to cli/.
I'm leaning to -1 on this, unless we find a better way to detect `ephemeral volume`

virtcontainers/utils/utils.go
bergwolf: > The current code in this PR which uses pb.Storage.Mountpoint is implemented as per our discussion here, #61 (comment)

That comment only describes the fact that the tmpfs storage is already supported by the agent grpc protocol. You have done a great job to implement it in kata agent. But the runtime PR does not satisfy the virtcontainer API we would like to merge.

At the runtime API level, we want an explicit API to let callers specify a tmpfs storage and its mountpoint. It can be a better workaround to check `/proc/mounts` for tmpfs storage, but again, the detection needs to be done at the kata cli level and the virtcontainers API need to be explicit about using a tmpfs storage, instead of trying to do the auto-detection internally.

IOW, what you did is an API change and we need to be explicit about using a tmpfs storage.

WeiZhang555: @harche 

>A typical characteristic of an ephemeral storage would be,
>1. A directory on the host (may or may not be backed by tmpfs)
>2. This directory is bind mounted to all the containers of the same pod so that they can share the data.

I wonder if this is a trusted detection point, I think ephemeral storage has another difference with normal tmpfs storage that a normal tmpfs storage **DOES** allow you to share contents between host and guest but ephemeral storage *DOESN'T* .

So if user wants to share host contents with all containers in pod and this volume happens to base on tmpfs(I think this is also valid use case), the case could be broken.

sboeuf: Trying to catch up here, and trying to summarize, we don't have a reliable way to detect this specific use case, right ?
The reason being: a volume backed by `tmpfs` could be either a directory with some data the host want to share with the container, or an ephemeral volume that is an empty directory.

What about checking both the tmpfs and the content of the source path provided ?

Now that being said, I realize that ephemeral volume support in Kata Containers is more an improvement thing since it's already supported if two containers link to the same ephemeral volume on the host via 9p, right @harche ?
Following this, do we really need this "improvement" since it seems to me that we'll have to rely on strong assumptions to detect the case ?

harche:
Edited subject to `work in progress` since we are yet to arrive at the consensus.

virtcontainers/utils/utils.go
harche: We use ephemeral volumes regularly for sharing the data between containers of the pod (especially between init and regular containers). Some of the applications are sensitive to the security and this is where we would love to use kata in our kubernetes environment for the added isolation it provides. With upcoming techonologies like Intel's [TME and MKTME](https://software.intel.com/en-us/blogs/2017/12/22/intel-releases-new-technology-specification-for-memory-encryption), [IBM Ultravisor](https://www.ibm.com/developerworks/library/l-support-protected-computing/) or [AMD memory encryption](http://amd-dev.wpengine.netdna-cdn.com/wordpress/media/2013/12/AMD_Memory_Encryption_Whitepaper_v7-Public.pdf) we can take the isolation provided by kata to the next level. 

Using those technologies you will be able to encrypt the memory pages of the guest, this is where this patch will enable kata to play a crucial role in providing the strongest isolation any runtime can provide for security sensitive kubernetes workloads. Since you will place the ephemeral volume inside the guest, which in turn has encrypted memory pages, even a compromised host won't be able to read the data containers are sharing between them. 

I strongly believe this work aligns with the kata's mission to provide the more secure environment for the container workloads.

@WeiZhang555 @bergwolf @sboeuf @amshinde

harche:
What happened here? it says merged?

grahamwhaley:
I think that 'merged' a little way down the thread was for a PR that referenced this one. This one now says 'closed', by you @harche .

harche:
I didn't close it.

harche:
We use ephemeral volumes regularly for sharing the data between containers of the pod (especially between init and regular containers). Some of the applications are sensitive to the security and this is where we would love to use kata in our kubernetes environment for the added isolation it provides. With upcoming techonologies like Intel's TME and MKTME, IBM Ultravisor or AMD memory encryption we can take the isolation provided by kata to the next level.

Using those technologies you will be able to encrypt the memory pages of the guest, this is where this patch will enable kata to play a crucial role in providing the strongest isolation any runtime can provide for security sensitive kubernetes workloads. Since you will place the ephemeral volume inside the guest, which in turn has encrypted memory pages, even a compromised host won't be able to read the data containers are sharing between them.

I strongly believe this work aligns with the kata's mission to provide the more secure environment for the container workloads.

@WeiZhang555 @bergwolf @sboeuf @amshinde @jodh-intel

harche:
@bergwolf I have moved ephemeral volume detection to `cli` package. Let me know what do you think.

virtcontainers/container.go
```diff
@@ -472,13 +472,29 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
 		// These mounts are created in the shared dir
 		filename := fmt.Sprintf("%s-%s-%s", c.id, hex.EncodeToString(randBytes), filepath.Base(m.Destination))
 		mountDest := filepath.Join(hostSharedDir, c.sandbox.id, filename)
+		sourcePath := filepath.Join(guestSharedDir, filename)
 
-		if err := bindMount(m.Source, mountDest, false); err != nil {
-			return nil, err
+		epheVol := false
+		for _, devInfo := range c.config.DeviceInfos {
+			if devInfo.DevType == "e" && devInfo.HostPath == m.Source {
```
bergwolf: Please define the new device type in device/config/config.go, and compare `m.Source` with `devInfo.ContainerPath` instead.

And for epheVol, you don't need to put it in the shared mounts. Let me explain how it works:

1. on kata cli side, add a device to container config, where devinfo.type is epheVol, and a Mount to container config, where `m.Source == devinfo.ContainerPath`
2. in kata runtime, skip epheVol by **NOT** adding it to `sharedMounts`
3. in kata_agent.go, create tmpfs storage where its mountpoint is devinfo.ContainerPath

This would follow the same storage/device devinfo semantics as other types of devices/storages.

virtcontainers/utils/utils.go
sboeuf: Thanks for the explanation @harche. This makes sense to support this case, but it's annoying that is something we have to detect based on the path name (as mentioned by @WeiZhang555)

virtcontainers/kata_agent.go
```diff
@@ -742,12 +745,38 @@ func (k *kataAgent) createContainer(sandbox *Sandbox, c *Container) (p *Process,
...
 
+	var devInfos []config.DeviceInfo
+	for _, devInfo := range c.config.DeviceInfos {
+		if devInfo.DevType == "e" {
+			filename := devInfo.ContainerPath
+			epheStorage := &grpc.Storage{
+				Driver:     "ephemeral",
+				Source:     "tmpfs",
+				Fstype:     "tmpfs",
+				MountPoint: filename,
```
bergwolf: IIUC you need "/ephemeralPath/" + filepath.Base(devInfo.ContainerPath) here, to match line 744 below?

katabuilder:
PSS Measurement:
Qemu: 159762 KB
Proxy: 6658 KB
Shim: 10893 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1999608 KB

virtcontainers/kata_agent.go
harche: Thanks for pointing that out. Turns out I was looping over a wrong slice. Instead of looping over `ociSpec.Mounts` I was looping over `newMounts`. The reason it worked during testing because I kept the source path intact, which meant it created a path like, `/var/lib/kubelet/pods/366c3a75-4869-11e8-b479-507b9ddd5ce4/volumes/kubernetes.io~empty-dir/cache-volume` inside a VM. 

I will update the PR with the fix.

cli/utils.go
```diff
@@ -43,6 +43,19 @@ func getFileContents(file string) (string, error) {
...
+func IsEphemeralStroage(path string) bool {
+	splitSourceSlice := strings.Split(path, "/")
+	if len(splitSourceSlice) > 1 {
+		storageType := splitSourceSlice[len(splitSourceSlice)-2]
+		if storageType == "kubernetes.io~empty-dir" {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
harche: I wish I had a better way here. :( 

In future, if k8s decides to change that path, kata might just end up provisioning the storage over 9pfs just like it does right now. 

I am trying to see if we can come up with some sort of standard for ephemeral storage that everyone can comply with (k8s, runtimes etc). Until then, I don't see any other way.

katabuilder:
PSS Measurement:
Qemu: 161852 KB
Proxy: 4690 KB
Shim: 10780 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1997368 KB

virtcontainers/device/drivers/ephemeral.go
```diff
@@ -0,0 +1,41 @@
...
+
+package drivers
+
+import (
+	"github.com/kata-containers/runtime/virtcontainers/device/api"
+	"github.com/kata-containers/runtime/virtcontainers/device/config"
+)
+
+// GenericDevice refers to a device that is neither a VFIO device or block device.
+type EphemeralDevice struct {
```
WeiZhang555: I don't see where this EphemeralDevice is been used, I think this file is unnecessary, right?

cli/create.go
```diff
@@ -259,6 +259,16 @@ func createContainer(ociSpec oci.CompatOCISpec, containerID, bundlePath,
...
 	}
 
+	// Add the ephemeral device if ephemeral volume
+	// has to be attached to the container
+	for _, mnt := range contConfig.Mounts {
+		if IsEphemeralStroage(mnt.Source) {
+			deviceInfo := config.DeviceInfo{}
+			deviceInfo.DevType = "e"
+			deviceInfo.ContainerPath = mnt.Source
+			contConfig.DeviceInfos = append(contConfig.DeviceInfos, deviceInfo)
```
WeiZhang555: I maybe missed something.

Why do you have to convert ephemeral volume to a fake device, and then convert it back to grpc.Storage in virtcontainers? A DevType of "e" is nonexistent so it should work too to add a fake mount type as "ephemeral" and set the contConfig.Mounts[x].Type to "ephemeral".

And deviceInfo is often used to insert some kinds of devices to hypervisor, so I don't think it's appropriate for carrying the ephemeral volume

virtcontainers/device/drivers/ephemeral.go
harche: You are correct, this has not been used and will be removed from final PR. I was just trying to model it like `BlockDevice` but unlike `BlockDevice` I didn't find a way to use it anywhere since we don't end up actually passing any device to the container.

cli/create.go
harche: My initial PR didn't use device. It initiated grpc.Storage directly from virtcontainers library. But @bergwolf suggested [here](https://github.com/kata-containers/runtime/pull/307#pullrequestreview-125180817) that we should move the detection to `cli` because `virtcontainers` wasn't the right place. 

```
Please define the new device type in device/config/config.go, and compare m.Source with devInfo.ContainerPath instead.

And for epheVol, you don't need to put it in the shared mounts. Let me explain how it works:

1.on kata cli side, add a device to container config, where devinfo.type is epheVol, and a Mount to container config, where m.Source == devinfo.ContainerPath
2.in kata runtime, skip epheVol by NOT adding it to sharedMounts
3.in kata_agent.go, create tmpfs storage where its mountpoint is devinfo.ContainerPath
This would follow the same storage/device devinfo semantics as other types of devices/storages.
```

WeiZhang555: Moving detection to `cli` package makes sense, there's no problem for this change.

@bergwolf I don't think we need a special device type, instead a special mount type "ephemeral" should work too in my opinion. WDYT?

bergwolf: @WeiZhang555 Yes, it can work too. The difference is that I wanted to stick to a storage model that a storage is a sandbox level resource and containers can share it in its mount slice. It's the same storage model we have in the kata-agent.

WeiZhang555: I see your point, I'm working on device manager refactoring for this device model. But I'm thinking that making the ephemeral volume as a fake device looks weird, because this device has a "fake" device type and can't be hot plugged or appended ...

bergwolf: ephemeral volume still can be shared. We do it either by creating a sandbox level storage resource, or by crafting mount.Source to point to the same one. And if we choose to share by letting API callers set the same mount.Source, we'd have to make a static link from mount.Source to `grpc.Storage.Mountpoint` so that kata-agent knows about the sharing as well. The static path conversion and meaningful ephemeral volume mount source are the main cons compared to a real sandbox level storage structure.

In fact, I think devinfo needs to be restructured so that it can describe any device or volume rather than just host local devices. Think of ceph rbd as an example. It is a device that needs hotplug, but there is no host device node for it. <-- I was hoping your device refactoring PR to take care of it but if not, we can do it separately.

harche: @WeiZhang555 @bergwolf So I presume we have to wait till @WeiZhang555 submits new PR to refactor device manager and then we will restructure this PR according to new refactored device manager? Just trying to understand what would be the next steps.

cli/utils.go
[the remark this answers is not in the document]
harche: /cc @saad-ali

cli/create.go
bergwolf: @harche I think the patch is in reasonable shape now. We can update the ephemeral storage later when the device refactor is done.

harche:
This particular talk highlights the importance of ephemeral volumes in storage volume attacks. https://youtu.be/GLwmJh-j3rs?t=876

cli/create.go
harche: @bergwolf Thanks.

cli/create.go
```diff
@@ -259,6 +259,16 @@ func createContainer(ociSpec oci.CompatOCISpec, containerID, bundlePath,
 		return vc.Process{}, err
 	}
 
+	// Add the ephemeral device if ephemeral volume
+	// has to be attached to the container
+	for _, mnt := range contConfig.Mounts {
+		if IsEphemeralStroage(mnt.Source) {
+			deviceInfo := config.DeviceInfo{}
+			deviceInfo.DevType = "e"
```
sboeuf: It would be more explicit to have something like `ephemeral` instead of `e`, it would read better.
Also, whatever value we choose, we need to reference this from a constant or a global variable to be reused across all virtcontainers.

cli/create.go
```diff
@@ -259,6 +259,16 @@ func createContainer(ociSpec oci.CompatOCISpec, containerID, bundlePath,
 		return vc.Process{}, err
 	}
 
+	// Add the ephemeral device if ephemeral volume
+	// has to be attached to the container
```
sboeuf: Maybe we could be a bit more verbose here because we could think that adding this ephemeral storage might fail if does not exist yet, but the agent actually takes care of creating the volume inside the VM if it does not exist yet, and does nothing if it already exists.

cli/create.go
harche: @sboeuf I agree that it should use a full word instead of just a letter. However, all other devices use deviceInfo.DevType which is declared [here](https://github.com/kata-containers/runtime/blob/master/virtcontainers/device/config/config.go#L63) single letter. So just for ephemeral if I use a word that would deviate from the standard other device types are following. 

I spoke about this with @bergwolf . We agreed that in future we might have to change that for all devices, but that would be out of scope for this PR.

cli/utils.go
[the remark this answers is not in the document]
harche: @jodh-intel @WeiZhang555 @sboeuf I have created a topic in k8s community on the stability of ephemeral volumes they provision, https://groups.google.com/forum/#!topic/kubernetes-dev/4JtfGSSUO_A

Feel free to join the discussion.

cli/utils.go
```diff
@@ -43,6 +43,19 @@ func getFileContents(file string) (string, error) {
 	return string(bytes), nil
 }
 
+// From the given path to the storage determine
+// if it's a kubernetes ephemeral storage
+func IsEphemeralStroage(path string) bool {
+	splitSourceSlice := strings.Split(path, "/")
+	if len(splitSourceSlice) > 1 {
+		storageType := splitSourceSlice[len(splitSourceSlice)-2]
+		if storageType == "kubernetes.io~empty-dir" {
```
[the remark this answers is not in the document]
harche: I have created a topic on k8s community to discuss just that, https://groups.google.com/forum/#!topic/kubernetes-dev/4JtfGSSUO_A

cli/utils.go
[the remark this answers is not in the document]
sboeuf: Nice, thx @harche.
In the meantime, just make sure our code has enough comment to explain this is not a perfect/stable solution.

cli/create.go
sboeuf: Yes, fair enough !

cli/utils.go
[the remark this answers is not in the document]
WeiZhang555: Thanks for submitting topic in K8S community! I think it will make me feel safer once get K8S community's confirm :+1:

virtcontainers/device/drivers/ephemeral.go
WeiZhang555: I'll mark this as "need change" in case we forget and merge it in accidently.

cli/utils.go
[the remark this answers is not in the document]
harche: @sboeuf sure. 

Meanwhile, Looking at Michelle Au's (Google) [comment](https://kubernetes.slack.com/messages/C09QZFCE5/) and Tim Hockin's (Google) [comment](https://groups.google.com/d/msg/kubernetes-dev/4JtfGSSUO_A/q3PfCLJsAQAJ), the general tone there resonates some of the concerns raised here. 

However, Michelle mentioned in her reply on slack,
`to answer your question, emprydir path on the kubelet should be stable` 

Quoting's Tim's reply on mailing list,
`I would probably NAK any change to the volume implementations that changed the path, but I don't think it is guaranteed.`

Although for the longer term, the we need to come up with a solution that doesn't depend on these assurances (or lack of them). There are a couple of approaches, 

1. As Michelle suggested, `we need to find a way to generically invoke volume plugins in the runtime`

2. Or In my opinion, we should decouple storage drivers from kata. Storage drivers should be these pluggable entities which anyone can write for the storage they are dealing with. 

@egernst @WeiZhang555 @sboeuf @bergwolf @jodh-intel 
Let me know your thoughts on how should we proceed in short term as well as in long term.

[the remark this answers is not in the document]
WeiZhang555: Since this is still a reasonable improvement especially on security aspect, I'm fine to merge it as a short term enhancement.
But it's better to find a solid method doing this later.
LGTM for now to merge this.

cli/create.go
WeiZhang555: device manager rework PR is almost done: https://github.com/kata-containers/runtime/pull/301
But I met some CI problem(probably related to CI itself not the PR)

harche: @WeiZhang555 Should I modify this PR to work with the changes in  #301?

WeiZhang555: CI has some problem, so I'm not 100% sure about correctness of #301 , so maybe you can wait.

Or you can go on with your work, I'll rebase on your changes, it's fine for me.

harche:
@WeiZhang555 @sboeuf @bergwolf I have updated the PR

harche:
Removing [WIP] from the title.

cli/utils_test.go
```diff
@@ -38,6 +38,20 @@ func TestFileExists(t *testing.T) {
 		fmt.Sprintf("File %q should exist", file))
 }
 
+func TestIsEphemeralStroage(t *testing.T) {
```
sboeuf: s/`TestIsEphemeralStroage`/`TestIsEphemeralStorage`/

cli/utils.go
```diff
@@ -43,6 +43,26 @@ func getFileContents(file string) (string, error) {
...
+// to detect if it's of type ephemeral. As of now,
+// this is a very k8s specific solution that works
+// but in future there should be a better way for this
+// method to determine if the path is for ephemeral
+// volume type
+func IsEphemeralStorage(path string) bool {
+	splitSourceSlice := strings.Split(path, "/")
+	if len(splitSourceSlice) > 1 {
+		storageType := splitSourceSlice[len(splitSourceSlice)-2]
+		if storageType == "kubernetes.io~empty-dir" {
```
sboeuf: Please define a global variable for `kubernetes.io~empty-dir`

virtcontainers/container.go
```diff
@@ -418,6 +418,7 @@ func (c *Container) createContainersDirs() error {
 // available when we will need to unmount those mounts.
 func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) ([]Mount, error) {
 	var sharedDirMounts []Mount
+MntLoop:
```
sboeuf: I don't like this kind of syntax, it's not really Go friendly.

virtcontainers/container.go
```diff
@@ -435,6 +436,16 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
 			return nil, err
 		}
 
+		for _, devInfo := range c.config.DeviceInfos {
+			if devInfo.DevType == "e" && devInfo.ContainerPath == m.Source {
+				// We are not going to bind mount host ephemeral directory
+				// provided by k8s, instead at later stage we are going to
+				// create a tmpfs inside a VM and use it to bind mount to the
+				// containers of the pod.
+				continue MntLoop
```
sboeuf: In order to avoid the `goto` here, please use something like this:
```go
                loopOver := false
                for _, devInfo := range c.config.DeviceInfos {
			if devInfo.DevType == "e" && devInfo.ContainerPath == m.Source {
				// We are not going to bind mount host ephemeral directory
				// provided by k8s, instead at later stage we are going to
				// create a tmpfs inside a VM and use it to bind mount to the
				// containers of the pod.
				loopOver = true
                                break
			}
		}

                if loopOver {
                            continue
                }
```

sboeuf:
@WeiZhang555 please give it another review !

virtcontainers/kata_agent.go
```diff
@@ -775,12 +777,40 @@ func (k *kataAgent) createContainer(sandbox *Sandbox, c *Container) (p *Process,
 		return nil, err
 	}
 
+	var devInfos []config.DeviceInfo
+	for _, devInfo := range c.config.DeviceInfos {
+		if devInfo.DevType == "e" {
+			filename := filepath.Join(ephemeralPath, filepath.Base(devInfo.ContainerPath))
```
amshinde: If the Base guaranteed to be unique here? If not, I would append this with a randomly generated string to make sure there are no conflicts.

amshinde:
LGTM overall.
Thanks @harche! 

You will need to refactor the code introduced in createContainer and mountSharedDirMounts in separate functions to reduce the cyclomatic complexity. Static checks are failing right now. You will then need to rebase on master to fix a CI issue we had recently.

virtcontainers/kata_agent.go
harche: Yes, the Base is guaranteed to be unique. In k8s ephemeral volume is defined by, 
```
  volumes:
  - name: cache-volume
    emptyDir:
      medium: "Memory"
```
In the pod yaml. 

The Base comes from `name` of the volume which is unique across the pod (which in the example above would be `cache-volume`)

harche:
@amshinde code in `mountSharedDirMounts` already has the cyclomatic complexity of 15 (without my changes). Any new condition will bump it to 16 and above (15 is the cut off right now). I need to have at least one if condition there to decide if that mount source should be skipped or not. 

For `createContainer` too cyclomatic complexity is at 15 without my changes, so this is too on the edge.

WeiZhang555:
The "ephemeral device" design is still weird to me, because it's a fake device, there's no real device type associated with it and it's not related to device, it's only about "Mount".

@bergwolf I know you suggested this design, I am thinking to just use "Mount" info instead of adding a "Device" type.
For example, from client side, it modify the ephemeral mount to new Mount:
```
Mount {
"Source": "/var/lib/kubelet/pods/366c3a75-4869-11e8-b479-507b9ddd5ce4/volumes/kubernetes.io~empty-dir/cache-volume",
"Destination": "/container/ephemeral",
"Type": "ephemeral",
}
```

Then it will be translated into a grpcStorage: 
```
grpc.Storage{
 				Driver:     kataEphemeralDevType,
 				Source:     "tmpfs",
 				Fstype:     "tmpfs",
 				MountPoint: filepath.Join(ephemeralPath, filepath.Base(mount.Source)),
 			}
```

and a modified OCI spec with modified Mount filed to kata-agent as:
```
Mount {
"Source": filepath.Join(ephemeralPath, filepath.Base(mount.Source)),,
"Destination": "/container/ephemeral",
"Type": "bind",
}
```

So we can omit a fake device type, instead we introduce a new fake storage type "ephemeral" similar to "tmpfs" or "proc". 
My suggestion isn't about saving code lines, it's just because I think the fake device type looks weird and easily confuse users.

What do you think? I didn't try but I think the new design could work. @bergwolf @harche

sboeuf:
I like the proposal @WeiZhang555. I think using `Storage` in this case is a better option than `Device`.

bergwolf:
@WeiZhang555 It feels weird because `Storage` was renamed to `Device` :)

The main reason I suggested the device description instead of relying only on container Mounts is that there is no sandbox level storage structure that can describe a storage without a real device. There is much in common between `storage` and `device` and there is also subtle difference. If we want to exclude fake devices from the device infrastructure, we need a storage structure to allow proper tracking of those storages w/o real devices at the **sandbox level**.

IWO, we either treat storage the same as device and allow fake devices, or we define storage and device separately and clearly.

For ephemeral storage, your proposal does work and it is actually what @harche did in the first place. But still, it misses a sandbox level description. If we ever want to check for ephemeral storage sharing, we have to iterate through all containers' config in the sandbox.

If we choose to exclude storage w/o real devices from the device handling infrastructure, we need to at least file an issue to say so and that we will add the missing storage abstraction at sandbox level. And then this PR can move forward with your proposal and @harche's original implementation.

tallclair:
Sorry I'm late to the party, but I think I can provide a bit more context on the Kubernetes side. Long term, the CRI will need to change to really support this case (I've started those conversations, but AFAIK there hasn't been any progress). Without the full k8s support, I have a few concerns with pushing forward with this approach:

- Using tmpfs means that writes to the emptyDir volume will consume pod memory. How is this memory accounted? We don't yet have a notion of pod overhead, so I don't think there's a way to reserve memory for these volumes. An alternative approach is to use a block or "image device" (I'm don't know the correct terminology) that provides disk-backed storage that is opaque to the host.
- How does this interact non-emptyDir volume types that delegate to the emptydir volume handler? This includes: secrets, configmaps, projected, gitRepo, downwardAPI
- We'll eventually want to treat the container's writeable layer the same way (opaque to the host)

Considering all this, I think this should be configureable, and maybe treated as an experimental feature (have per-pod opt-in, as well as a runtime-level configuration to enable it). This also solves the issue of whether the filepath to the volume could ever change, since that could be part of the config.

That was a bit of a ramble, but hopefully it helps. I'm very excited about this feature, thanks for working on it!

harche:
For people following this issue, I have created a PR that doesn't use devices - https://github.com/kata-containers/runtime/pull/458

harche:
@tallclair Ephemeral volumes will consume the memory allocated for pod VM. But attaching the block storage to provision ephemeral volumes defeats the main motivation to have this feature in first place. I have explained in this, https://github.com/kata-containers/runtime/pull/307#issuecomment-393845024, in detail on why we need to make sure the ephemeral volume that we provisioned needs to remain in VM memory. 

TLDR; Upcoming memory encryption technologies allow us to encrypt VM pages and gaurd against a compromised host from reading that memory. We inject our ephemeral volume in that VM memory so that even a compromised host can't read what containers of that pod are sharing with each other.

sboeuf:
Replaced by https://github.com/kata-containers/runtime/pull/458, this can be closed.

tallclair:
@harche couldn't the attached block storage just be encrypted? That said, I think we're currently very far away from being able to protect VM pods running a compromised node, and it will only be possible by taking a holistic look at Kubernetes, possibly rethinking some fundamental concepts. I'm not sure it's worth designing individual features for this use case without the full picture, and I'm worried this is the security equivalent of premature optimization (premature hardening?).

PR #311, opened 2018-05-15
kata-env: Remove static agent version

### LINKED ISSUE #310: Remove agent version from kata-env output
The `kata-env` command currently shows the following agent details:

```
[Agent]
  Type = "kata"
  Version = "<<unknown>>"
```

The `Version` element was added for parity with the other elements whose details `kata-env` displays. However, that version is simply a static string because the runtime cannot determine the agent version *without creating a container and connecting to the agent*.

As such, let's just remove the string since:

- it never changes.
- it's potentially misleading.
- users *can* determine the agent version by running `kata-collect-data.sh`, which now shows details of the image, including something like:

```
agent:
  url: "https://github.com/kata-containers/agent"
  name: "kata-agent"
  version: "0.2.0-e444fe2bbbc57fdd47dad00f369935009b094817"
  agent-is-init-daemon: "no"

```

jodh-intel:
Remove the agent version from the output of `kata-env`. The value was
always a static string (`<<unknown>>`) because the runtime cannot
determine the agent version without creating a container.

Note that agent details, including the version, *are* displayed when
the user runs `kata-collect-data.sh`.

Fixes #310.

### DIFF
--- a/cli/kata-env.go
+++ b/cli/kata-env.go
@@ -21,7 +21,7 @@ import (
 //
 // XXX: Increment for every change to the output format
 // (meaning any change to the EnvInfo type).
-const formatVersion = "1.0.11"
+const formatVersion = "1.0.12"
 
 // MetaInfo stores information on the format of the output itself
 type MetaInfo struct {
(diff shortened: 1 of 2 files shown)


WeiZhang555:
LGTM
once CI is green, I believe it will.

jodh-intel:
@sboeuf, @devimc, @amshinde - ptal.

Issue #293, opened 2018-05-09
Is qemu-system-x86_64 or qemu-lite-system-x86_64 in kata-containers configuration.toml 


gnodli:
# Description of problem
I follow [Developer-Guide.md](https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md)  

conditions:
- initrd image and hyperstart kernel
-default runtime is kata-runtime

when I use below command to start a container, which return an error
`  sudo docker run -ti busybox sh`

# Expected result
start a busybox container

# Actual result
output:
```
docker: Error response from daemon: OCI runtime create failed: /usr/share/defaults/kata-containers/configuration.toml: file /usr/bin/qemu-system-x86_64 does not exist: unknown.
```

---
Finally, I found the QEMU version should be qemu-lite, and the dir of `/usr/bin/` only have qemu-lite-system-x86_64。  Then, I change the qemu-system-x86_64 to qemu-lite-system-x86_64, it works.

jodh-intel:
Thanks for raising this. I'm slightly confused though - the error states that you do not have qemu installed.

The developer guide does state that you need to install `qemu-lite` here:

- https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#assumptions

That said, we do need to update the dev guide now that we also support vanilla qemu (see #119). I've raised https://github.com/kata-containers/documentation/issues/67 to deal with this.

gnodli:
Thank you for your answer. I did not install qemu, just install qemu-lite package according to above guide. That said, qemu-lite can also work , but it has been dropped.

jodh-intel:
Oh sorry, I see what you mean now. I'll try to update the dev guide today to avoid this confusion...

jodh-intel:
Hi @gnodli - are you happy for this issue to be closed when https://github.com/kata-containers/documentation/pull/68 lands?

gnodli:
Now I understand，thank you very much

urbandroid:
Configuration looks for qemu-lite but snap package has qemu bin and bam! you can't change conf file because snaps are read only that's why i love snap, reveals the lazy, pain in the ass software at early stage. Issue is still there but issue is closed what should i do? i know $sudo snap remove kata-containers --never-look-back

caoruidong:
cc @jcvenegas @jodh-intel

bergwolf:
I think we should provide qemu-lite snap package as well cc @devimc .

PR #312, opened 2018-05-16
build: default to using qemu-lite-x86_64

egernst:
Update the runtime to use qemu-lite by default. After a
build this will be observed as the default in configuration.toml.

Fixes: #293

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -25,6 +25,7 @@ format: |
      commit: "<commit>"
      version: "<version>"
      release: "<version>"
+     branch: "<git-branch>"
      meta:
        <key-1>: "<value-1>"
        <key-n>: "<value-n>"
@@ -45,7 +46,8 @@ format: |
   - A section MAY define a "meta" section to store additional
     information about a project or group.
 
-  - Each entry MUST specify ATLEAST one of "commit", "version" and "release".
+  - Each entry MUST specify ATLEAST one of "commit", "version", "release"
+    and "branch".
 
   - WARNING: Gotcha alert! Remember to double-quote all strings
     (except multi-line strings)! This avoids the possibility of a
@@ -57,6 +59,12 @@ assets:
 
   hypervisor:
     description: "Component used to create virtual machines"
+
+    qemu-lite:
+      description: "lightweight VMM that uses KVM"
+      url: "https://github.com/kata-containers/qemu"
+      branch: "qemu-lite-2.11.0"
+      commit: "6ba2bfbee9a80bfd03605c5eb2ca743c8b68389e"
 
     qemu:
       description: "VMM that uses KVM"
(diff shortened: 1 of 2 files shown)


egernst:
With this, our default build-from-source will be consistent with what we package and consistent with expectations in the developer guide.

chavafg:
Can you add something like: 
```
  qemu-lite:
      url: "https://github.com/kata-containers/qemu"
      version: "qemu-lite-2.11.0"
```
in https://github.com/kata-containers/runtime/blob/master/versions.yaml#L64 

So that I can grab this information to build qemu-lite in the tests repository?

I'll create the corresponding PR on `tests` to make the CI happy

chavafg:
@egernst seems like `qemu: ` line is not properly indented. Should be at the same level of `qemu-lite: ` line

egernst:
okay, fixed -- let's see what CI does now.

egernst:
@chavafg - Any ideas on the CI failures?  Would love to get the CI updated for qemu-lite and a new package rolled through this before EOW.

versions.yaml
```diff
@@ -58,6 +58,11 @@ assets:
   hypervisor:
     description: "Component used to create virtual machines"
 
+    qemu-lite:
+      description: "lite VMM that uses KVM"
```
jodh-intel: Nit: I'd say the following to avoid the self-referential definition:

> lightweight VMM that uses KVM

versions.yaml
```diff
@@ -58,6 +58,11 @@ assets:
...
+      url: "https://github.com/kata-containers/qemu"
+      version: "qemu-lite-2.11.0"
```
jodh-intel: This is actually a branch so you'll need to change the file to show something like this:

```diff
diff --git a/versions.yaml b/versions.yaml                                                                                                                                                    
index ab173d8..d10882b 100644
--- a/versions.yaml
+++ b/versions.yaml
@@ -25,6 +25,7 @@ format: |
      commit: "<commit>"
      version: "<version>"
      release: "<version>"
+     branch: "<git-branch>"
      meta:
        <key-1>: "<value-1>"
        <key-n>: "<value-n>"
@@ -45,7 +46,7 @@ format: |
   - A section MAY define a "meta" section to store additional
     information about a project or group.
 
-  - Each entry MUST specify ATLEAST one of "commit", "version" and "release".
+  - Each entry MUST specify ATLEAST one of "commit", "version", "release" and "branch".
 
   - WARNING: Gotcha alert! Remember to double-quote all strings
     (except multi-line strings)! This avoids the possibility of a
@@ -58,6 +59,12 @@ assets:
   hypervisor:
     description: "Component used to create virtual machines"
 
+    qemu-lite:
+      description: "lightweight VMM that uses KVM"
+      url: "https://github.com/kata-containers/qemu"
+      branch: "qemu-lite-2.11.0"
+      commit: "6ba2bfbee9a80bfd03605c5eb2ca743c8b68389e"
+
     qemu:
       description: "VMM that uses KVM"
       url: "https://github.com/qemu/qemu"
```

egernst:
@jodh-intel  -- changes applied.  Thanks for the diff!

jodh-intel:
lgtm

PR #319, opened 2018-05-18
config: Error if image+initrd specified

### LINKED ISSUE #318: Using a config file with both initrd= and image= defined will cause docker to hang
If you run a command like this:

```
$ sudo docker run -ti busybox sh
```

And it hangs and you also see this:

```
$ sudo journalctl -t kata-proxy |grep -m 1 -i cannot
May 17 17:27:25 artful kata-proxy[13710]: time="2018-05-17T17:27:25.503479797+01:00" level=info msg="[    0.667228] VFS: Cannot open root device \"pmem0p1\" or unknown-block(0,0): error -6\n" name=kata-proxy pid=13710 source=agent
```

... then the issue is that your `configuration.toml` has an `init=` **and** an `image=` option defined. That does not work, so please follow the instructions carefully in the Developer Guide here:

- https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#configure-to-use-initrd-or-rootfs-image

jodh-intel:
If you build and install the runtime, the config file will contain an
entry for both an `initrd=` and an `image=` entry.

The Developer Guide explains that the user must disable one but it is
easy to forget.

Modified the runtime to fail if both an image and an initrd are
specified. Also added a new test for this scenario.

Fixes #318.

### DIFF
--- a/cli/config.go
+++ b/cli/config.go
@@ -309,6 +309,11 @@ func newQemuHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 		return vc.HypervisorConfig{}, err
 	}
 
+	if image != "" && initrd != "" {
+		return vc.HypervisorConfig{},
+			errors.New("cannot specify an image and an initrd in configuration file")
+	}
+
 	firmware, err := h.firmware()
 	if err != nil {
 		return vc.HypervisorConfig{}, err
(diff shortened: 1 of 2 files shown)


jodh-intel:
Hi @bergwolf, @sboeuf, @egernst - ptal.

We've seen a few folks with this issue so this will make the problem immediately obvious.

devimc:
IMO runtime shouldn't fail if both are specified, we should take one and print a warning

devimc:
if you ask me which one is the best, I'll tell you `initrd` because it supports a bash bomb  :smile:

PR #348, opened 2018-05-30
unittest: tiny fix for incorrect parameters

### LINKED ISSUE #347: Testcase pull busybox every time whether it exists or not
#Testcase pull busybox every time whether it exists or not
the init  function in  cli/main_test.go,
```
       // Only hit the network if the image doesn't exist locally
	_, err = runCommand([]string{"docker", "image", "inspect", testDockerImage})
	if err == nil {
		fmt.Printf("INFO: docker image %v already exists locally\n", testDockerImage)
	} else {
		_, err = runCommand([]string{"docker", "pull", testDockerImage})
		if err != nil {
			panic(err)
		}
	}
```
runCommand([]string{"docker", "image", "inspect", testDockerImage}) will always fail, and it lead testcase pull busybox every time whether it exists or not.

keloyang:
runCommand([]string{"docker", "image", "inspect", testDockerImage}) will always fail, and it lead testcase pull busybox every time whether it exists or not.

Fixes #347

Signed-off-by: y00316549 <yangshukui@huawei.com>

### DIFF
--- a/cli/main_test.go
+++ b/cli/main_test.go
@@ -98,7 +98,7 @@ func init() {
 	fmt.Printf("INFO: ensuring required docker image (%v) is available\n", testDockerImage)
 
 	// Only hit the network if the image doesn't exist locally
-	_, err = runCommand([]string{"docker", "image", "inspect", testDockerImage})
+	_, err = runCommand([]string{"docker", "inspect", "--type=image", testDockerImage})
 	if err == nil {
 		fmt.Printf("INFO: docker image %v already exists locally\n", testDockerImage)
 	} else {


jodh-intel:
Hi @keloyang - I'm a little confused since the following two commands appear to be equivalent, atleast using docker 18.03.1-ce...?!

- `docker image inspect $image`
- `docker inspect $image`

WeiZhang555:
Guess you are trying to test it with an older version of docker? @keloyang

amshinde:
@keloyang Yes, curious about the docker version where this is failing.
If we decide to to go with docker inspect, a more specific check with `docker inspect --type=image busybox` should be used, rather than looking for any docker object.

keloyang:
@amshinde @jodh-intel @WeiZhang555 the version I used is 1.11.2, maybe it's better to use compatible command， WDYT ?

WeiZhang555:
I think it makes sense to use a more compatible command. As @amshinde said, `docker inspect --type=image busybox` can be a better improvement.
The change will be LGTM with @amshinde 's suggested modification.

amshinde:
@keloyang Agree on using a more compatible command.

keloyang:
ping @amshinde @WeiZhang555  added --type=image, thanks.

katabuilder:
PSS Measurement:
Qemu: 161983 KB
Proxy: 4636 KB
Shim: 8799 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1997500 KB

jodh-intel:
lgtm

jodh-intel:
pullapprove seems to be confused. We have the requisite number of acks and the CI is green, so merging...

Issue #297, opened 2018-05-10
bug: runtime should return fail when proxy get incorrect message


Ace-Tang:
# Description of problem

I use a incorrect kata-containers.img which has no kata-agent inside, and as a result, I can not start kata container successful. But the runtime not return fail.
I check proxy log
```
time="2018-05-10T10:02:54.143132026+08:00" level=info name=kata-proxy pid=7465 source=proxy version=0.0.1-baf8b9e191e0647ee6e6c92f15c3a71180b2862a
time="2018-05-10T10:02:54.549573982+08:00" level=info msg="[    0.270943] systemd[1]: Failed to load default target: No such file or directory\n" name=kata-proxy pid=7465 source=agent
```

it seems that proxy try to start something in vm but fail, I think proxy should return with error, or runtime should get error through other way like connect with agent in vm, since I am clear with the mechanism kata container , please correct me if I am wrong, thanks.

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #350, opened 2018-05-30
virtcontainers: add codes to process the case that kata-agent doesn't start in VM

woshijpf:
virtcontainers: add codes to process the case that kata-agent doesn't start in VM

If kata-agent process is not launched to run in VM, kata-runtime will exit because of timeout,
however kata-proxy and qemu process will still resident in host

Call Check grpc function to check whether grpc server is running in kata-agent process in VM,
if not and then do some rollback operation to stop kata-proxy and qemu processes

Fixes: #297

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -937,10 +937,18 @@ func (s *Sandbox) startVM() error {
 
 	s.Logger().Info("VM started")
 
+
+
 	// Once startVM is done, we want to guarantee
 	// that the sandbox is manageable. For that we need
 	// to start the sandbox inside the VM.
-	return s.agent.startSandbox(s)
+	err := s.agent.startSandbox(s)
+	defer func() {
+		if err != nil {
+			s.hypervisor.stopSandbox()
+		}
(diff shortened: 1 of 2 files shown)


caoruidong:
CI failed because title is too long

virtcontainers/sandbox.go
```diff
@@ -937,10 +937,18 @@ func (s *Sandbox) startVM() error {
 
 	s.Logger().Info("VM started")
 
+
+
```
caoruidong: Too many blank lines?

virtcontainers/sandbox.go
```diff
@@ -937,10 +937,18 @@ func (s *Sandbox) startVM() error {
...
 	// Once startVM is done, we want to guarantee
 	// that the sandbox is manageable. For that we need
 	// to start the sandbox inside the VM.
-	return s.agent.startSandbox(s)
+	err := s.agent.startSandbox(s)
+	defer func() {
```
caoruidong: Is defer necessary?

PR #351, opened 2018-05-30
virtcontainers: process the case that kata-agent doesn't start in VM

woshijpf:
virtcontainers: process the case that kata-agent doesn't start in VM

If kata-agent process is not launched to run in VM, kata-runtime will exit because of timeout,
however kata-proxy and qemu process will still resident in host

Call Check grpc function to check whether grpc server is running in kata-agent process in VM,
if not and then do some rollback operation to stop kata-proxy and qemu processes

Fixes: #297

### DIFF
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -39,6 +39,8 @@ func CreateSandbox(sandboxConfig SandboxConfig, factory Factory) (VCSandbox, err
 }
 
 func createSandboxFromConfig(sandboxConfig SandboxConfig, factory Factory) (*Sandbox, error) {
+	var err error
+
 	// Create the sandbox.
 	s, err := createSandbox(sandboxConfig, factory)
 	if err != nil {
@@ -46,22 +48,29 @@ func createSandboxFromConfig(sandboxConfig SandboxConfig, factory Factory) (*San
 	}
 
 	// Create the sandbox network
-	if err := s.createNetwork(); err != nil {
+	if err = s.createNetwork(); err != nil {
 		return nil, err
 	}
 
+	// network rollback
+	defer func() {
+		if err != nil && s.networkNS.NetNsCreated {
+			s.removeNetwork()
+		}
+	}()
+
 	// Start the VM
-	if err := s.startVM(); err != nil {
+	if err = s.startVM(); err != nil {
 		return nil, err
 	}
 
 	// Create Containers
-	if err := s.createContainers(); err != nil {
+	if err = s.createContainers(); err != nil {
 		return nil, err
 	}
 
 	// The sandbox is completely created now, we can store it.
-	if err := s.storeSandbox(); err != nil {
+	if err = s.storeSandbox(); err != nil {
 		return nil, err
 	}
 
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -64,6 +64,27 @@ func createSandboxFromConfig(sandboxConfig SandboxConfig, factory Factory) (*San
 		return nil, err
 	}
 
+	// rollback to stop VM if error occurs
+	defer func() {
+		if err != nil {
+			s.stopVM()
+		}
+	}()
+
+	// Once startVM is done, we want to guarantee
+	// that the sandbox is manageable. For that we need
+	// to start the sandbox inside the VM.
+	if err = s.agent.startSandbox(s); err != nil {
+		return nil, err
+	}
+
+	// rollback to stop sandbox in VM
+	defer func() {
+		if err != nil {
+			s.agent.stopSandbox(s)
+		}
+	}()
+
 	// Create Containers
 	if err = s.createContainers(); err != nil {
 		return nil, err
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -450,6 +450,8 @@ func (k *kataAgent) generateInterfacesAndRoutes(networkNS NetworkNamespace) ([]*
 }
 
 func (k *kataAgent) startProxy(sandbox *Sandbox) error {
+	var err error
+
 	if k.proxy == nil {
 		return errorMissingProxy
 	}
@@ -475,10 +477,18 @@ func (k *kataAgent) startProxy(sandbox *Sandbox) error {
 		return err
 	}
 
+	// If error occurs after kata-proxy process start,
+	// then rollback to kill kata-proxy process
+	defer func() {
+		if err != nil && pid > 0 {
+			k.proxy.stop(sandbox, pid)
+		}
+	}()
+
 	// Fill agent state with proxy information, and store them.
 	k.state.ProxyPid = pid
 	k.state.URL = uri
-	if err := sandbox.storage.storeAgentState(sandbox.id, k.state); err != nil {
+	if err = sandbox.storage.storeAgentState(sandbox.id, k.state); err != nil {
 		return err
 	}
 
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -512,6 +512,11 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
 		hostname = hostname[:maxHostnameLen]
 	}
 
+	// check grpc server is serving
+	if err = k.check(); err != nil {
+		return err
+	}
+
 	//
 	// Setup network interfaces and routes
 	//
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -971,10 +971,12 @@ func (s *Sandbox) startVM() error {
 
 	s.Logger().Info("VM started")
 
-	// Once startVM is done, we want to guarantee
-	// that the sandbox is manageable. For that we need
-	// to start the sandbox inside the VM.
-	return s.agent.startSandbox(s)
+	return nil
+}
+
+// stopVM: stop the sandbox's VM
+func (s *Sandbox) stopVM() error {
+	return s.hypervisor.stopSandbox()
 }
 
 func (s *Sandbox) addContainer(c *Container) error {


katabuilder:
PSS Measurement:
Qemu: 144708 KB
Proxy: 6742 KB
Shim: 8796 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  2013332 KB

katabuilder:
PSS Measurement:
Qemu: 144050 KB
Proxy: 6795 KB
Shim: 10930 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  2011712 KB

katabuilder:
PSS Measurement:
Qemu: 142416 KB
Proxy: 6825 KB
Shim: 10819 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  2011464 KB

virtcontainers/kata_agent.go
```diff
@@ -451,6 +451,21 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
...
 
+	//check grpc server is serving
+	if err := k.connect(); err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	if _, err := k.client.Check(ctx, &grpc.CheckRequest{}); err != nil {
+		k.Logger().Error("grpc server check failed")
```
jodh-intel: How about:

```
k.Logger().WithError(err).Error("grpc server check failed")
```

virtcontainers/sandbox.go
```diff
@@ -940,7 +940,13 @@ func (s *Sandbox) startVM() error {
 	// Once startVM is done, we want to guarantee
 	// that the sandbox is manageable. For that we need
 	// to start the sandbox inside the VM.
-	return s.agent.startSandbox(s)
+	err := s.agent.startSandbox(s)
+	defer func() {
```
devimc: is `defer` needed? why not just 

```go
if err != nil {
	s.hypervisor.stopSandbox()
}
```
:thinking:

devimc: or

```go
if err != nil {
	return s.hypervisor.stopSandbox()
}
```

in case of  `stopSandbox` fails

virtcontainers/kata_agent.go
```diff
@@ -451,6 +451,21 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
 		hostname = hostname[:maxHostnameLen]
 	}
 
+	//check grpc server is serving
+	if err := k.connect(); err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
devimc: why 5?  should that number be a constant?

virtcontainers/kata_agent.go
woshijpf: This log output is used to locate the problem from kata-runtime logfile when kata-agent is crashed in VM

jodh-intel: Sure, but your code is not logging the actual error, so I'm suggesting you keep your error message and also add the original error (which might include more detail).

virtcontainers/sandbox.go
woshijpf: Thanks! I agree with your last proposed code in case of `stopSandbox` fails

virtcontainers/kata_agent.go
woshijpf: Thanks! This is a great suggestion about the error log.Back to your first comment,  I agree with your opinion that  kata-agent isn't running is very small likelihood . But I have another question, do we have some measures to make sure that kata-agent will not fail in VM?

virtcontainers/sandbox.go
sboeuf: This is not semantically correct. We should not call `stopSandbox()` after `startSandbox()` returned some error. We would be running `stopSandbox()`, which assumes the sandbox has been properly started, when this is not true.

The valid case for such a thing could be to run `stopSandbox()` if further in the code, an error would occur, and we would want to rollback from there.

And if you want to do such thing, I think it should be part of its own commit for more clarity ;)

virtcontainers/kata_agent.go
```diff
@@ -451,6 +451,21 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
...
+	defer cancel()
+	if _, err := k.client.Check(ctx, &grpc.CheckRequest{}); err != nil {
+		k.Logger().Error("grpc server check failed")
+
+		//kill kata-proxy to release resources
+		k.proxy.stop(sandbox, k.state.ProxyPid)
```
sboeuf: The rollback of the proxy being started sounds like a good idea. But if we want this to be properly implemented, we should create a `defer` right after the proxy has been correctly started. You need to declare a global `err error` for the function, and if when the defer is executed, the `err != nil`, then we should stop the proxy. This will apply for any error case that could happen in this function after the proxy has been started.
You also want to handle the case where the agent fails to start, which is great, but a separate thing IMO, that's why it should be split into 2 different commits. First you introduce proper rollback for the proxy, and then you introduce the check for the agent availability.

virtcontainers/kata_agent.go
devimc: yep, just define a constant and add a comment, for example

```go
// timeout to check grpc server is serving
const  grpcCheckTimeout = 5*time.Second
....
func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
....
    ctx, cancel := context.WithTimeout(context.Background(), grpcCheckTimeout)

```

virtcontainers/sandbox.go
woshijpf: Maybe we have some misunderstanding about `s.agent.stopSandbox`() and `s.hypervisor.stopSandbox()`.

`s.agent.stopSandbox()` is used to stop the Sandbox and Containers which are running in VM, in other words, both sandbox and containers are started successfully in VM.

`s.hypervisor.stopSandbox()` is used to stop qemu process and release related resources, in other words, sandbox is not created successfully in VM.

In this place, if `s.agent.startSandbox()` return error , it indicates that sandbox is not created successfully in VM, so I want to call `s.hypervisor.stopSandbox()` to stop VM process.

katabuilder:
PSS Measurement:
Qemu: 163900 KB
Proxy: 6792 KB
Shim: 10786 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1996624 KB

devimc:
@woshijpf  do me a favour, run next command with this patch

```
docker run --runtime=kata-runtime -ti busybox bash
```

if you don't see a qemu process running  then this patch looks good to me

jodh-intel:
You have to look closely at the previous command to notice that it won't work as `bash` isn't included with `busybox`. Equivalent:

```
$ docker run --runtime=kata-runtime -ti does-not-exist
```

devimc:
@jodh-intel I think you mean 

```
docker run --runtime=kata-runtime -ti busybox does-not-exist
```

jodh-intel:
@devimc - I do! Thanks! :smile:

woshijpf:
@devimc I test this patch in my ubuntu 16.04 x86_64 laptop is good with next two cases:
1. kata-agent is running in VM
This case works well, container can run in VM

```bash
# need to run as root 
# docker run --runtime=kata-runtime -ti busybox sh
/ # ls
bin   dev   etc   home  proc  root  sys   tmp   usr   var
```

2. no kata-agent  in VM
First, we need to create a rootfs image without kata-agent, you can follow next steps:

```bash
$ cd <kata-containers>/osbuilder/rootfs-builder

# use euleros as example
$ ./rootfs.sh euleros

# remove kata-agent.service, so systemd can't launch the kata-agent process
$ cd ./rootfs-EulerOS/usr/lib/systemd/system/
$ rm kata-agent.service
$ rm rm kata-containers.target

# make rootfs-EulerOS into rootfs.img
$ cd <kata-containers>/osbuilder/image-builder

# we need to modify the rootfs make shell scipt
$ cp image_builder.sh image_builder-no-agent.sh
$ vim image_builder-no-agent.sh
 # comment follow lines
129 # [ "${AGENT_INIT}" == "yes" ] || [ -x "${ROOTFS}/usr/bin/${AGENT_BIN}" ] || \
130 #        die "/usr/bin/${AGENT_BIN} is not installed in ${ROOTFS}
131  #       use AGENT_BIN env variable to change the expected agent binary name"
132 # OK "Agent installed"

$ next, we can run  image_builder-no-agent.sh to make rootfs image
$ ./image_builder-no-agent.sh ../rootfs-builder/rootfs-EulerOS/

# finally, we get the Euleros rootfs.img without kata-agent
$ ls 
 ls
Dockerfile  image_builder-no-agent.sh  image_builder.sh  kata-containers.img   README.md

# we can copy kata-containers.img to anywhere, but you need to modify the "image = "  field in the /usr/share/defaults/kata-containers/configuration.toml file
```

Now, we can use the same command as previous case to test this patch
```bash
# need to run as root
# docker run --runtime=kata-runtime -ti busybox sh
docker: Error response from daemon: OCI runtime create failed: rpc error: code = DeadlineExceeded desc = context deadline exceeded: unknown.

# we get the error output as expected, and let's see if any kata-proxy and qemu processed is resident in host system
$ ps -auxww --sort=start_time

......
root     10857  0.0  0.0      0     0 ?        S    23:11   0:00 [kworker/0:1]
root     10858  0.0  0.0      0     0 ?        S    23:11   0:00 [kworker/1:3]
root     11118  0.0  0.0  44744  3752 pts/17   R+   23:14   0:00 ps -auxww --sort=start_time
```

we get the result as expected, so this patch works well.

devimc:
@woshijpf nice and thanks, and what about `docker run --runtime=kata-runtime -ti busybox does-not-exist` " ?   is qemu running?

woshijpf:
@devimc Yep，I just run the command `docker run --runtime=kata-runtime -ti busybox does-not-exist`, the result surprises me that qemu and kata-proxy processes both still reside in the host system, it is a bug need to be fixed.
Maybe we need to do some rollback if container doesn't run successfully.

virtcontainers/kata_agent.go
```diff
@@ -461,6 +464,19 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
...
 	}
 
+	//check grpc server is serving
+	if err = k.connect(); err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), grpcCheckTimeout)
+	defer cancel()
+	if _, err = k.client.Check(ctx, &grpc.CheckRequest{}); err != nil {
```
sboeuf: This looks good, but I was thinking that in order to avoid duplication of code, the check with timeout should be available through the agent interface itself. Thus, you could enhance the agent interface regarding `check()` function, and then rely on this new version to call it from here.

virtcontainers/kata_agent.go
```diff
@@ -461,6 +464,19 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
...
+		k.Logger().Error("grpc server check failed, get the error: ", err)
+		return err
+	}
+	k.disconnect()
```
sboeuf: You should check for `err` here.

virtcontainers/sandbox.go
```diff
@@ -944,7 +944,14 @@ func (s *Sandbox) startVM() error {
 	// Once startVM is done, we want to guarantee
 	// that the sandbox is manageable. For that we need
 	// to start the sandbox inside the VM.
-	return s.agent.startSandbox(s)
+	err := s.agent.startSandbox(s)
+	if err != nil {
+		//VM is started  but sandbox is not created successfully in VM
+		//so we need to rollback to stop VM qemu process
+		s.hypervisor.stopSandbox()
```
sboeuf: Sorry I didn't realize we were talking about `hypervisor.stopSandbox()`....
Could you isolate this into its own commit as this addresses the rollback of the VM.
Also, you should follow the same model you followed for the proxy rollback, meaning you defer the `hypervisor.stopSandbox()` after you properly started the VM. This way you will both cover the error case from `waitSandbox()` and `agent.startSandbox()` that could happen.

sboeuf:
@woshijpf there is a lot of work related to rollback, and I'd be more than happy if you want to look into it. We need to be careful about not having rollbacks that would collide with each others.

woshijpf:
@sboeuf Thanks for your careful review.I'm glad to deep into it to solve rollback related problems.

virtcontainers/kata_agent.go
```diff
@@ -429,14 +431,22 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
 
 	// Start the proxy here
 	pid, uri, err := k.proxy.start(sandbox, proxyParams)
+
+	// If error occurs after kata-proxy start, rollback to kill kata-proxy process
+	defer func() {
+		if err != nil && pid > 0{
+			k.proxy.stop(sandbox, pid)
+		}
+	}()
```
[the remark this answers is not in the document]
sboeuf: Yes, the defer is called all the time (when the function returns), and if you look further at the code in this function, there are numerous cases where we can return a non-nil error, meaning the proxy will get stopped.

virtcontainers/kata_agent.go
bergwolf: You can simply use k.check() here, and it handles grpc connect/disconnect properly.

jshachm: @bergwolf  k.check doesn‘t have `timeout`. Need to enhance it.

bergwolf: @jshachm Take a look at `installReqFunc()`. Check request has builtin a 5 second timeout, same as you defined in this PR.

jshachm: @bergwolf  sry for miss it~~  Yep,  with the builtin timeout, `k.check` will make a difference.

woshijpf: @bergwolf Thanks, it's a great suggestion,  it makes code more clean and simple

virtcontainers/kata_agent.go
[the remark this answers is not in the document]
woshijpf: @sboeuf Since I am not when the defer function will be called, I write a simple code to test it, finally I get the this conclusion that defer function isn't called all the time after function end, for example, if a return code is called before defer in the function body, then this defer function will not be called.
But in the place, I will move defer block after the check for `err` as you suggested, because it has no effect .
If `k.proxy.start()` return error that indicates the kata-proxy process doesn't start, so we don't need to do kata-proxy rollback operation.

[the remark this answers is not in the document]
sboeuf: Yes exactly, if the defer statement is reached, then it will be called when the function returns.
And yes, I have asked for the defer to be called after the error check of `k.proxy.start()` because we don't need any rollback if `err != nil` in this case.

woshijpf:
@sboeuf @jodh-intel @devimc Hi buddies~I am fixing the problem of qemu and kata-proxy processes reside in system while `createSandboxFromConfig()` is failed, but I meet a problem and I want to get help from your nice guys.
The problem is that how can I get to know whether qemu process  is running or not in system when `err := s.startVM() ` return error in `createSandboxFromConfig()` function, since the error return by `s.startVM()` may caused by `s.hypervisor.startSandbox()` or `s.hypervisor.waitSandbox` or `s.agent.startSandbox()` in `startVM()` function.Thanks!

katabuilder:
PSS Measurement:
Qemu: 164140 KB
Proxy: 6803 KB
Shim: 10818 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1996136 KB

katabuilder:
PSS Measurement:
Qemu: 162063 KB
Proxy: 4597 KB
Shim: 12829 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1997252 KB

woshijpf:
@sboeuf Thanks for your detailed reviews and suggestion, I learn a lot from your reviews, and I will finish this work as soon as possible.

woshijpf:
@sboeuf I'm sorry to do this work so late. Now, I follow your suggestion to refactor the code and split modified code into 4 commits which make the code more clearly.

virtcontainers/api.go
```diff
@@ -39,29 +39,53 @@ func CreateSandbox(sandboxConfig SandboxConfig) (VCSandbox, error) {
 }
 
 func createSandboxFromConfig(sandboxConfig SandboxConfig) (*Sandbox, error) {
+
```
jodh-intel: Could you remove the blank line before this variable (same comment for the function below)?

woshijpf: Thanks for your careful review, and I have removed the blank line

virtcontainers/api.go
```diff
@@ -39,29 +39,39 @@ func CreateSandbox(sandboxConfig SandboxConfig) (VCSandbox, error) {
...
 
 	// Create the sandbox network
-	if err := s.createNetwork(); err != nil {
+	if err = s.createNetwork(); err != nil {
 		return nil, err
 	}
 
+	// network rollback
+	defer func() {
+		if err != nil && s.networkNS.NetNsCreated{
```
sboeuf: `gofmt` will complain:
```go
if err != nil && s.networkNS.NetNsCreated {
```

virtcontainers/kata_agent.go
```diff
@@ -436,10 +439,18 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
 		return err
 	}
 
+	// If error occurs after kata-proxy process start,
+	// then rollback to kill kata-proxy process
+	defer func() {
+		if err != nil && pid > 0{
```
sboeuf: `gofmt` will complain:
```go
if err != nil && pid > 0 {
```

virtcontainers/sandbox.go
```diff
@@ -952,6 +952,11 @@ func (s *Sandbox) startVM() error {
 	return s.agent.startSandbox(s)
 }
 
+// stopVM: stop the sandbox's VM
+func (s *Sandbox)stopVM() error {
+	return s.hypervisor.stopSandbox()
```
sboeuf: We want to stop the sandbox with `s.agent.stopSandbox()` before we actually stop the VM. This way we mirror/rollback what has been done inside `startVM()`.

virtcontainers/api.go
```diff
@@ -55,7 +54,7 @@ func createSandboxFromConfig(sandboxConfig SandboxConfig) (*Sandbox, error) {
 
 	// network rollback
 	defer func() {
-		if err != nil && s.networkNS.NetNsCreated{
+		if err != nil && s.networkNS.NetNsCreated {
```
sboeuf: Please fix this in the commit where it was introduced.

virtcontainers/sandbox.go
sboeuf: Oh I've just seen your next commit and I can see you don't need `s.agent.stopSandbox()` since it actually removes the part `s.agent.startSandbox()` from `startVM()`.
This is fine, but please re-organize your code so that there is no inconsistent commit.

woshijpf: Yes, you are right! I don't consider the case that if `s.agent.startSandbox()` is executed successfully, but `s.createContainers()` fails in `createSandboxFromConfig()` function, my code doesn't process this case well, I think I need to add `s.agent.stopSandbox()` before `s.stopVM()` in the defer function  in `createSandboxFromConfig()` function.

virtcontainers/api.go
```diff
@@ -64,6 +65,27 @@ func createSandboxFromConfig(sandboxConfig SandboxConfig) (*Sandbox, error) {
...
+
+	// Once startVM is done, we want to guarantee
+	// that the sandbox is manageable. For that we need
+	// to start the sandbox inside the VM.
+	if err = s.agent.startSandbox(s); err != nil {
+		return nil, err
+	}
+
+	// set the sandboxRunning state to true
+	sandboxRunning = true
```
sboeuf: No need for this boolean if you define another defer function here:
```go
	defer func() {
		if err != nil {
			s.agent.stopSandbox(s)
		}
	}()
```
Because this defer function is defined after the first one, it will be called prior to it. That's a stack, not a FIFO, meaning it will do exactly what you need.
This way, no need for a complex defer function, just call a bunch of simple ones.

woshijpf:
@sboeuf Could you help me to review the code again, I modify the code as you suggested.
 Finally thanks for your patient and detailed reviews, I have learned a lot from you.

sboeuf:
@woshijpf I have just added one last comment. Thanks for your quick fixes !

virtcontainers/api.go
woshijpf: Yeap, your suggestion is more graceful!

sboeuf:
@woshijpf we just fixed the CI, could you rebase this PR on latest master, and it's good to be merged once the CI will pass.

sboeuf:
@woshijpf any updates on this ?

jshachm:
@sboeuf he is out of office for about a mouth. And will be back next week ,so updates will be promoted next week~

sboeuf:
@jshachm thx for the info !

egernst:
@jshachm @woshijpf  -- back yet?  I want to get this contribution merged.

woshijpf:
@egernst I'm sorry for leaving so long time, and I am going to rebase the code to pull a new request.

egernst:
no worries @woshijpf - hope it was a good break.  Hope the CI goes through smoothly.

PR #353, opened 2018-05-30
virtcontainers/api: use RW lock to update containers

### LINKED ISSUE #346: Potential race condition when updating container resources
The code path updating container resources eventually stores the modifications. This means we have to make sure the pod is RW locked when this function is called. https://github.com/kata-containers/runtime/blob/1.0.0/virtcontainers/api.go#L618 should be simply updated with a RW lock.

devimc:
When a container is updated, those modifications are stored, to
avoid race conditions with other operations, a RW lock should be used.

fixes #346

### DIFF
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -615,7 +615,7 @@ func UpdateContainer(sandboxID, containerID string, resources specs.LinuxResourc
 		return errNeedContainerID
 	}
 
-	lockFile, err := rLockSandbox(sandboxID)
+	lockFile, err := rwLockSandbox(sandboxID)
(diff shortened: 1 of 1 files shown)


katabuilder:
PSS Measurement:
Qemu: 166170 KB
Proxy: 10484 KB
Shim: 8823 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1996012 KB

jodh-intel:
Just for completeness, can you point to any failures that were the result of only using a read lock?

Ideally of course we'd have a specific test that thrashed this code and tried to provoke the error :)

PR #357, opened 2018-05-31
shm: Create shared /dev/shm

### LINKED ISSUE #356: Create a shared /dev/shm for containers in a sandbox
Containers in a pod share IPC namespace and should have a shared /dev/shm created within the guest. We should also first check the size of the shm mount setup on the host size for creating shm inside the guest.

Related to : https://github.com/kata-containers/kata-containers/issues/21

amshinde:
This commit checks the size of "/dev/shm" for the sandbox container
which is then used to create the shared memory inside the guest.
kata agent then uses this size to set up a sandbox level shm
storage. The containers then simply bind mount this sandbox level
shm.

With this, we will now be able to support docker --shm-size option
as well have a shared shm within containers in a pod, since they are
supposed to be in the same IPC namespace.

Fixed #356

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -42,12 +42,15 @@ var (
 	kataHostSharedDir     = "/run/kata-containers/shared/sandboxes/"
 	kataGuestSharedDir    = "/run/kata-containers/shared/containers/"
 	mountGuest9pTag       = "kataShared"
+	kataGuestSandboxDir   = "/run/kata-containers/sandbox/"
 	type9pFs              = "9p"
 	vsockSocketScheme     = "vsock"
 	kata9pDevType         = "9p"
 	kataBlkDevType        = "blk"
 	kataSCSIDevType       = "scsi"
 	sharedDir9pOptions    = []string{"trans=virtio,version=9p2000.L", "nodev"}
+	shmDir                = "shm"
+	kataEphemeralDevType  = "ephemeral"
 )
 
 // KataAgentConfig is a structure storing information needed
@@ -505,9 +508,27 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
 		Options:    sharedDir9pOptions,
 	}
 
+	storages := []*grpc.Storage{sharedVolume}
+
+	if sandbox.shmSize > 0 {
+		path := filepath.Join(kataGuestSandboxDir, shmDir)
+		shmSizeOption := fmt.Sprintf("size=%d", sandbox.shmSize)
+
+		shmStorage := &grpc.Storage{
+			Driver:     kataEphemeralDevType,
+			MountPoint: path,
+			Source:     "shm",
+			Fstype:     "tmpfs",
+			Options:    []string{"noexec", "nosuid", "nodev", "mode=1777", shmSizeOption},
+		}
+
+		storages = append(storages, shmStorage)
+	}
+
 	req := &grpc.CreateSandboxRequest{
-		Hostname: hostname,
-		Storages: []*grpc.Storage{sharedVolume},
+		Hostname:     hostname,
+		Storages:     storages,
+		SandboxPidns: false,
 	}
 
 	_, err = k.sendReq(req)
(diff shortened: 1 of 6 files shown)


amshinde:
@bergwolf PTAL. Let me know what you think about the approach. I am planning to add a new Storage in the kata agent with driver "shm" that will create a tmpfs based storage.

virtcontainers/kata_agent.go
```diff
@@ -611,14 +630,24 @@ func constraintGRPCSpec(grpcSpec *grpc.Spec) {
...
+
+		if sandbox.shmSize > 0 {
+			grpcSpec.Mounts[idx].Type = "bind"
+			grpcSpec.Mounts[idx].Source = filepath.Join(kataGuestSharedDir, shmDir)
+		} else {
+			sizeOption := fmt.Sprintf("size=%d", DefaultShmSize)
 			grpcSpec.Mounts[idx].Type = "tmpfs"
 			grpcSpec.Mounts[idx].Source = "shm"
-			grpcSpec.Mounts[idx].Options = []string{"noexec", "nosuid", "nodev", "mode=1777", "size=65536k"}
+			grpcSpec.Mounts[idx].Options = []string{"noexec", "nosuid", "nodev", "mode=1777", sizeOption}
```
jodh-intel: It might be worth adding a log call to show the requested size for debugging purposes.

virtcontainers/pkg/oci/utils.go
```diff
@@ -594,6 +602,29 @@ func ContainerConfig(ocispec CompatOCISpec, bundlePath, cid, console string, det
...
 
+func getShmSize(c vc.ContainerConfig) (uint64, error) {
+	var shmSize uint64
+
+	for _, m := range c.Mounts {
+		if m.Destination != "/dev/shm" {
+			continue
+		}
+
+		shmSize = vc.DefaultShmSize
```
jodh-intel: Why is this set inside the loop? Can't you just initialize `shmSize` to this value as currently if there is no host mount this function will return `0`?

bergwolf:
@amshinde Can you reuse kata-containers/agent/pull/236 and base your change on top of #307? shm is just a special case of tmpfs and tmpfs is handled as ephemeral storage in #307.

WeiZhang555:
@bergwolf Maybe we can create a new general storage type in kata-agent, for supporting the case that sharing a volume among containers in pod, but not share it with host?

bergwolf:
@WeiZhang555 Sharing a volume within the guest is already supported, by letting different containers bind mount the same volume source at the sandbox level. Each container can mount the referenced volume to its own namespace.

virtcontainers/pkg/oci/utils.go
amshinde: @jodh-intel I have added a `break` here. If no `shm` mount is found, then the `shmSize` will be zero, hence it is initialized to zero. In case host shm is to be used, this will be set to `DefaultShmSize`.

virtcontainers/kata_agent.go
amshinde: I have added log calls for the shm size.

amshinde:
@bergwolf I have changed this PR to use `ehemeral` storage now,, instead of introducing a new storage specific to just `shm`. PTAL.

virtcontainers/kata_agent.go
```diff
@@ -505,9 +507,27 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
...
+	storages := []*grpc.Storage{sharedVolume}
+
+	if sandbox.shmSize > 0 {
+		path := filepath.Join(kataGuestSharedDir, shmDir)
+		shmSizeOption := fmt.Sprintf("size=%d", sandbox.shmSize)
+
+		shmStorage := &grpc.Storage{
+			Driver:     kataEphemeralDevType,
+			MountPoint: path,
+			Source:     "shm",
```
amshinde: @bergwolf I just realised that the `source` in case of /dev/shm is set to `shm` and not `tmpfs`, various for ephemeral volumes it is `tmpfs` .
```
$ sudo docker run -it --runtime=runc  debian sh -c "mount | grep shm"
shm on /dev/shm type tmpfs (rw,nosuid,nodev,noexec,relatime,size=65536k)
```
This source field would need to be changed to `tmpfs` to use the ephemeral storage. I know shm is a special case of tmpfs, but I havent looked into the implications of doing this,  do you think it is ok to use `tmpfs` instead of `shm`.

I would prefer to introduce a new storage type for shm to be consistent with runc behaviour.

bergwolf: @amshinde for ephemeral storage, the source actually does not matter. You can set it as `shm` or `tmpfs` or whatever, and it will be shown as such inside the container. Do you mean that kata-agent would only accept `tmpfs` source for ephemeral storage? If so, we need to fix it in kata-agent side.

amshinde: @bergwolf Yes, you are right. I set this to shm and tested it.

bergwolf:
LTGM! Thanks @amshinde !

My only concern is that we are mounting all these volumes under `kataGuestSharedDir` that is actually the mount point of 9pfs. This might confuse the 9pfs, e.g., these internal volume mountpoints directories are also visible to host side via 9pfs.

jodh-intel:
/cc @grahamwhaley for the 9p comment.

lgtm although the CI doesn't look very happy.

katabuilder:
PSS Measurement:
Qemu: 159671 KB
Proxy: 4689 KB
Shim: 11070 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1996756 KB

katabuilder:
PSS Measurement:
Qemu: 161784 KB
Proxy: 4783 KB
Shim: 10822 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1995756 KB

katabuilder:
PSS Measurement:
Qemu: 161773 KB
Proxy: 6814 KB
Shim: 10823 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1996384 KB

katabuilder:
PSS Measurement:
Qemu: 159750 KB
Proxy: 6744 KB
Shim: 8975 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1997128 KB

amshinde:
@jodh-intel I am no longer using `kataGuestSharedDir` , so the shm mount will be not be  to host side via 9pfs. CI has been fixed as well.

jodh-intel:
Hi @amshinde - ok. codecov is complaining, but I think if you create a unit test for `getShmSize()` it'll be within thresholds.

katabuilder:
PSS Measurement:
Qemu: 162138 KB
Proxy: 6723 KB
Shim: 10733 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1998492 KB

amshinde:
@egernst @sboeuf Can you merge this one? Code coverage needed to be fixed which has been done now.

PR #385, opened 2018-06-11
network: Always bind back physical interfaces

### LINKED ISSUE #384: Bind back physical network interfaces to host driver when the sandbox is stopped.
We detach network network interfaces only if the network namespace has been created by the kata-runtime. While this applied to virtual endpoints, we need to call `Detach` on physical endpoints  even if the network namespace has not been created by kata-runtime to bind them back to host driver.

amshinde:
In case of physical network interfaces, we explicitly
pass through them to the VM. We need to bind them back to
the host driver when the sandbox is stopped, irrespective if
the network namespace has been created by virtcontainers or not.

Fixes #384

### DIFF
--- a/virtcontainers/cni.go
+++ b/virtcontainers/cni.go
@@ -153,10 +153,10 @@ func (n *cni) add(sandbox *Sandbox, config NetworkConfig, netNsPath string, netN
 	return networkNS, nil
 }
 
-// remove unbridges and deletes TAP interfaces. It also removes virtual network
-// interfaces and deletes the network namespace for the CNI network.
-func (n *cni) remove(sandbox *Sandbox, networkNS NetworkNamespace) error {
-	if err := removeNetworkCommon(networkNS); err != nil {
+// remove network endpoints in the network namespace. It also deletes the network
+// namespace in case the namespace has been created by us.
+func (n *cni) remove(sandbox *Sandbox, networkNS NetworkNamespace, netNsCreated bool) error {
+	if err := removeNetworkCommon(networkNS, netNsCreated); err != nil {
 		return err
 	}
 
@@ -164,5 +164,9 @@ func (n *cni) remove(sandbox *Sandbox, networkNS NetworkNamespace) error {
 		return err
 	}
 
-	return deleteNetNS(networkNS.NetNsPath, true)
+	if netNsCreated {
+		return deleteNetNS(networkNS.NetNsPath, true)
+	}
+
+	return nil
 }
--- a/virtcontainers/cni.go
+++ b/virtcontainers/cni.go
@@ -165,7 +165,7 @@ func (n *cni) remove(sandbox *Sandbox, networkNS NetworkNamespace, netNsCreated 
 	}
 
 	if netNsCreated {
-		return deleteNetNS(networkNS.NetNsPath, true)
+		return deleteNetNS(networkNS.NetNsPath)
 	}
 
 	return nil
--- a/virtcontainers/cnm.go
+++ b/virtcontainers/cnm.go
@@ -55,7 +55,7 @@ func (n *cnm) remove(sandbox *Sandbox, networkNS NetworkNamespace, netNsCreated 
 	}
 
 	if netNsCreated {
-		return deleteNetNS(networkNS.NetNsPath, true)
+		return deleteNetNS(networkNS.NetNsPath)
 	}
 
 	return nil
--- a/virtcontainers/cnm.go
+++ b/virtcontainers/cnm.go
@@ -47,12 +47,16 @@ func (n *cnm) add(sandbox *Sandbox, config NetworkConfig, netNsPath string, netN
 	return networkNS, nil
 }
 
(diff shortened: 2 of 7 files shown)


virtcontainers/cni.go
```diff
@@ -155,14 +155,18 @@ func (n *cni) add(sandbox *Sandbox, config NetworkConfig, netNsPath string, netN
 
 // remove unbridges and deletes TAP interfaces. It also removes virtual network
```
egernst: While we are in here, should we update the function description to be more accurate (ie, handle the non-virtual case as well).

virtcontainers/cnm.go
```diff
@@ -49,10 +49,14 @@ func (n *cnm) add(sandbox *Sandbox, config NetworkConfig, netNsPath string, netN
 
 // remove unbridges and deletes TAP interfaces. It also removes virtual network
```
egernst: same comment as CNI case...

katabuilder:
PSS Measurement:
Qemu: 163786 KB
Proxy: 4747 KB
Shim: 10936 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1996268 KB

virtcontainers/cni.go
amshinde: Fixed.

virtcontainers/cnm.go
amshinde: fixed.

katabuilder:
PSS Measurement:
Qemu: 158018 KB
Proxy: 6685 KB
Shim: 11009 KB

Memory inside container:
Total Memory: 2045972 KB
Free Memory:  1996872 KB

virtcontainers/cni.go
```diff
@@ -153,16 +153,20 @@ func (n *cni) add(sandbox *Sandbox, config NetworkConfig, netNsPath string, netN
...
+	if err := removeNetworkCommon(networkNS, netNsCreated); err != nil {
 		return err
 	}
 
 	if err := n.invokePluginsDelete(sandbox, networkNS); err != nil {
 		return err
 	}
 
-	return deleteNetNS(networkNS.NetNsPath, true)
+	if netNsCreated {
```
jodh-intel: Aside: It looks like all calls to `deleteNetNS()` now call it with `mounted=true` so this param (and the logic) can be dropped I think? Maybe you could do that cleanup on a separate commit on this PR?

sboeuf:
@amshinde thanks for the PR, looks very good, and I agree the logic related to the fact that we have created or not the network namespace should be only affecting the removal of this network namespace.

sboeuf:
@amshinde could you try to look into those CI issues ? They don't look like random issues, the patch might be the cause here.

amshinde:
@sboeuf I was seeing these unit tests failing in the last CI build for this PR:
```
=== RUN   TestStartStopSandboxHyperstartAgentSuccessfulWithCNINetwork
--- FAIL: TestStartStopSandboxHyperstartAgentSuccessfulWithCNINetwork (0.01s)
	cc_proxy_mock.go:80: [CCProxyMock] listening on /tmp/cc-proxy-test869276064/cc-proxy-test.sock
	api_test.go:1425: missing network name
	cc_proxy_mock.go:80: [CCProxyMock] Client connection already closed
	cc_proxy_mock.go:80: [CCProxyMock] Stopped
=== RUN   TestStartStopSandboxHyperstartAgentSuccessfulWithCNMNetwork
--- FAIL: TestStartStopSandboxHyperstartAgentSuccessfulWithCNMNetwork (0.00s)
	cc_proxy_mock.go:80: [CCProxyMock] listening on /tmp/cc-proxy-test430518911/cc-proxy-test.sock
	api_test.go:1463: sandbox 7f49d00d-1995-4156-8c79-5f5ab24ce138 exists
	cc_proxy_mock.go:80: [CCProxyMock] Client connection already closed
	cc_proxy_mock.go:80: [CCProxyMock] Stopped
```

This was introduced by this change in the CNI plugins repository:
https://github.com/containernetworking/plugins/commit/1f6e6ef198812a9a1257c43e314f9b1c92467767

This introduces a change requiring network name to be provided by the plugins, but the loopback plugin that we use in our tests has been been updated yet to include this as seen here :
https://github.com/containernetworking/plugins/commit/1f6e6ef198812a9a1257c43e314f9b1c92467767#diff-5a9f4821dfcce9f8ad9b7f8e616eeb7b

Since we we use the latest master in our virtcontainers setup script, our CI is failing with the error "missing network name".  So I introduced a change to use plugin binaries with the same version that we vendor in our runtime. Once this has been fixed upstream, we can update the vendoring and the setup script.

jodh-intel:
Tests issue raised for the CNI version issue: https://github.com/kata-containers/tests/issues/42.

amshinde:
@sboeuf @jodh-intel I have opened an issue upstream  to address the failure on latest master for CNI plugins:
https://github.com/containernetworking/plugins/issues/167

We can pull in the latest CNI plugins once the above issue is solved.

Meanwhile, can this PR be merged?

Issue #397, opened 2018-06-13
ppc64le: Kata-container fails to launch on POWER8 when SMT is ON. 


nitkon:
# Description of problem
SMT(Simultaneous multithreading) must be turned off on Power8 for KVM to work. Use this as a 
check before running kata-container on POWER systems.

PR #398, opened 2018-06-13
cli: kata-check if SMT is off on POWER8 systems

nitkon:
SMT must be turned off on Power8 for KVM to work. Put
this as a check for kata-runtime kata-check.

Fixes: #397

### DIFF
--- a/cli/kata-check_ppc64le.go
+++ b/cli/kata-check_ppc64le.go
@@ -7,6 +7,8 @@ package main
 
 import (
 	"fmt"
+	"os/exec"
+	"strings"
 
 	"github.com/sirupsen/logrus"
 )
@@ -15,6 +17,11 @@ const (
 	cpuFlagsTag        = genericCPUFlagsTag
 	archCPUVendorField = genericCPUVendorField
 	archCPUModelField  = genericCPUModelField
+)
+
+var (
+	ppc64CpuCmd     = "ppc64_cpu"
+	smtStatusOption = "--smt"
 )
 
 // archRequiredCPUFlags maps a CPU flag value to search for and a
@@ -50,6 +57,17 @@ func hostIsVMContainerCapable(details vmContainerCapableDetails) error {
 		return err
 	}
 
+	text, err := getFileContents(details.cpuInfoFile)
+	if err != nil {
+		return err
+	}
+
+	if strings.Contains(text, "POWER8") {
+		if !isSMTOff() {
+			return fmt.Errorf("SMT is not Off. %s", failMessage)
+		}
+	}
+
 	count, err := checkKernelModules(details.requiredKernelModules, archKernelParamHandler)
 	if err != nil {
 		return err
@@ -74,4 +92,23 @@ func archKernelParamHandler(onVMM bool, fields logrus.Fields, msg string) bool {
 
 func getCPUDetails() (vendor, model string, err error) {
 	return genericGetCPUDetails()
+}
+
+func isSMTOff() bool {
+
+	// Check if the SMT is available and off
+
+	cmd := exec.Command(ppc64CpuCmd, smtStatusOption)
+	additionalEnv := "LANG=C"
+	cmd.Env = append(cmd.Env, additionalEnv)
+	out, err := cmd.Output()
+
+	if err == nil && strings.TrimRight(string(out), "\n") == "SMT is off" {
+		return true
+	} else if err != nil {
+		kataLog.Warn("ppc64_cpu isn't installed, can't detect SMT")
+		return true
+	}
+
+	return false
 }


grahamwhaley:
Travis CI failed a static check:
```
cli/kata-check_ppc64le.go:66:7:warning: ineffectual assignment to err (ineffassign)
```

cli/kata-check_ppc64le.go
```diff
@@ -55,6 +63,13 @@ func hostIsVMContainerCapable(details vmContainerCapableDetails) error {
 		return err
 	}
 
+	out, err := exec.Command("bash", "-c", checkPower8cmd).Output()
```
jodh-intel: Please can we not do this. The code already parses `/proc/cpuinfo` - see `getCPUInfo()`.

nitkon: @jodh-intel : Addressed the issue.

cli/kata-check_ppc64le.go
```diff
@@ -75,3 +93,14 @@ func archKernelParamHandler(onVMM bool, fields logrus.Fields, msg string) bool {
 func getCPUDetails() (vendor, model string, err error) {
 	return genericGetCPUDetails()
 }
+
+func isSMTOff() bool {
+	// Check if the SMT is off
+	out, err := exec.Command(ppc64CpuCmd, smtStatusOption).Output()
```
amshinde: @nitkon Is it possible to get this through sysfs rather than execing a command here?

nitkon: @amshinde : There is no easy way of getting a conclusive evidence from sysfs. The ppc64_cpu command interprets many files in sysfs to come to this conclusion. So it's better to exec this command as well.

jodh-intel: I'm really not comfortable with this approach, but if there really is no other way, I guess we have no other alternative :disappointed:

cli/kata-check_ppc64le.go
```diff
@@ -17,6 +19,11 @@ const (
 	archCPUModelField  = genericCPUModelField
 )
 
+var (
+	ppc64CpuCmd     = "ppc64_cpu"
```
jodh-intel: Is this command guaranteed to be installed (forever) and in `$PATH`?

cli/kata-check_ppc64le.go
```diff
@@ -17,6 +19,11 @@ const (
...
+	smtStatusOption = "--smt"
```
jodh-intel: Is this option guaranteed to be available in all versions of `ppc64_cpu`?

cli/kata-check_ppc64le.go
```diff
@@ -75,3 +93,14 @@ func archKernelParamHandler(onVMM bool, fields logrus.Fields, msg string) bool {
 func getCPUDetails() (vendor, model string, err error) {
 	return genericGetCPUDetails()
 }
+
+func isSMTOff() bool {
+	// Check if the SMT is off
+	out, err := exec.Command(ppc64CpuCmd, smtStatusOption).Output()
+
+	if err == nil && strings.TrimRight(string(out), "\n") == "SMT is off" {
```
jodh-intel: Is `ppc64_cpu` i18n'd? Do we need to set `LC_ALL=C` / `LANG=C` to ensure this output will match in English?

cli/kata-check_ppc64le.go
mpe: No.

cli/kata-check_ppc64le.go
mpe: It's existed since the initial version AFAICS (~2009), and would obviously never be removed.

cli/kata-check_ppc64le.go
mpe: No it's not i18n'd.

jodh-intel: Ok - thanks for the responses @mpe. I wonder if it would be a good precaution to set the C locale anyway "just in case".

Let's get some more reviews on this PR: ping @kata-containers/runtime.

cli/kata-check_ppc64le.go
```diff
@@ -55,6 +62,17 @@ func hostIsVMContainerCapable(details vmContainerCapableDetails) error {
...
+	text, err := getFileContents(details.cpuInfoFile)
+	if err != nil {
+		return err
+	}
+
+	if strings.Contains(text, "POWER8") {
+		if !isSMTOff() {
+			return fmt.Errorf("SMT is not Off. %s", failMessage)
+		}
+	}
```
WeiZhang555: It's better to move this chunk under 
```
 	if count == 0 {
 		return nil
 	}
```

So that the judgement of `count == 0` can be near the assignment.

cli/kata-check_ppc64le.go
WeiZhang555: So I think maybe it's better to print a warning to indicate "ppc64_cpu isn't installed, can't detect SMT"?

cli/kata-check_ppc64le.go
mpe: Yeah it's probably sensible to use `LANG=C` just in case.

cli/kata-check_ppc64le.go
mpe: I just remembered `ppc64_cpu` also has logic to understand the `subcores_per_core` mode the host is in, which is an additional complication you'd have to deal with. So shelling out to it is probably the least-worst option.

jodh-intel: Understood. btw, can you run `ppc64_cpu --smt` as a non-`root` user?

nitkon: @jodh-intel : Yes you can run.

cli/kata-check_ppc64le.go
nitkon: Thanks for the inputs @WeiZhang555 @jodh-intel : I have updated my PR.

cli/kata-check_ppc64le.go
```diff
@@ -75,3 +94,23 @@ func archKernelParamHandler(onVMM bool, fields logrus.Fields, msg string) bool {
...
 	return genericGetCPUDetails()
 }
+
+func isSMTOff() bool {
+
+	// Check if the SMT is available and off
+
+	cmd := exec.Command(ppc64CpuCmd, smtStatusOption)
+	additionalEnv := "LANG=C"
+	newEnv := append(os.Environ(), additionalEnv)
```
jodh-intel: Nit: This is changing the environment in the parent but you actually only need to change the child (`cmd.Env`). This is effectively equivalent today but if we were to i18n the runtime, this would become a bug :)

nitkon: @jodh-intel : I am sending an updated PR.

nitkon: @jodh-intel : Updated my PR.

jodh-intel: Thanks @nitkon.

lgtm

cli/kata-check_ppc64le.go
grahamwhaley: I think we missed one tiny niggle - + 1 here for @WeiZhang555 comment - we should move the module check call to down underneath the cpuinfo/SMT check code I think. Right now the module check is split up by the cpuinfo/SMT code.

(sorry for the niggle - thanks for sticking with this @nitkon !)

nitkon: @grahamwhaley : Fixed!

nitkon:
@jodh-intel : Good to merge?

grahamwhaley:
Was waiting on the CIs yesterday. Looks like pullapprove has somehow got hungup. We have 2 approvals - I'll force merge...

Issue #396, opened 2018-06-13
Kata-runtime getting stuck, causing CRI-O to hang


jamiehannaford:
# Description of problem

CRI-O fails to boot sometimes when using Kata. 

# Expected result

After running `systemctl restart crio`, I expected CRI-O to restart, but it's currently stuck in the initialization procedure and doesn't seem to bind to its sock file. This causes Kubelet to hang too, since it needs `crio.sock` to launch its server.

# Actual result

Looking at the CRI-O logs, it's able to run `kata-runtime state <id>`:

```
Jun 13 12:09:01 k8s-jamie-terraform-4c6ef93f-default-pool-mv3n systemd[1]: Stopping OCI-based implementation of Kubernetes Container Runtime Interface...
Jun 13 12:09:01 k8s-jamie-terraform-4c6ef93f-default-pool-mv3n systemd[1]: Stopped OCI-based implementation of Kubernetes Container Runtime Interface.
Jun 13 12:09:01 k8s-jamie-terraform-4c6ef93f-default-pool-mv3n systemd[1]: Started OCI-based implementation of Kubernetes Container Runtime Interface.
Jun 13 12:09:01 k8s-jamie-terraform-4c6ef93f-default-pool-mv3n crio[23222]: time="2018-06-13 time="2018-06-13 12:09:01.559676555Z" level=debug msg="[graphdriver] trying provided driver "overlay""
time="2018-06-13 12:09:01.559820868Z" level=debug msg="overlay: override_kernelcheck=true"
time="2018-06-13 12:09:01.598789333Z" level=debug msg="overlay test mount with multiple lowers succeeded"
time="2018-06-13 12:09:01.645248033Z" level=debug msg="backingFs=extfs, projectQuotaSupported=false, useNativeDiff=true"
time="2018-06-13 12:09:01.647089066Z" level=warning msg="hooks path: "/usr/share/containers/oci/hooks.d" does not exist"
time="2018-06-13 12:09:01.647129086Z" level=warning msg="hooks path: "/etc/containers/oci/hooks.d" does not exist"
time="2018-06-13 12:09:01.647140964Z" level=debug msg="hooks map[]"
time="2018-06-13 12:09:01.647508237Z" level=info msg="CNI network k8s-pod-network (type=calico) is used from /etc/cni/net.d/10-calico.conf"
[...]

PR #404, opened 2018-06-17
Add cleanup task for failed creates

jamiehannaford:
Addresses a comment left by @sboeuf https://github.com/kata-containers/runtime/issues/396#issuecomment-397027292.

Not sure whether this is what you had in mind, but I've verified locally that dirs are now being cleaned up. I think a big reason `kata-runtime list` was hanging was due to the large amount of junk folders left around in `/run/vc/sbs`

### DIFF
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -45,23 +45,30 @@ func createSandboxFromConfig(sandboxConfig SandboxConfig) (*Sandbox, error) {
 		return nil, err
 	}
 
+	defer func() {
+		if err != nil {
+			s.Logger().WithError(err).WithField("sandboxid", s.id).Error("Creating sandbox failed")
+			s.Cleanup()
+		}
+	}()
+
 	// Create the sandbox network
-	if err := s.createNetwork(); err != nil {
+	if err = s.createNetwork(); err != nil {
 		return nil, err
 	}
 
 	// Start the VM
-	if err := s.startVM(); err != nil {
+	if err = s.startVM(); err != nil {
 		return nil, err
 	}
 
 	// Create Containers
-	if err := s.createContainers(); err != nil {
+	if err = s.createContainers(); err != nil {
 		return nil, err
 	}
 
 	// The sandbox is completely created now, we can store it.
-	if err := s.storeSandbox(); err != nil {
+	if err = s.storeSandbox(); err != nil {
 		return nil, err
 	}
 
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -1433,3 +1433,18 @@ func (s *Sandbox) AddVhostUserDevice(devInfo api.VhostUserDevice, devType config
 	}
 	return fmt.Errorf("unsupported device type")
 }
+
+// Cleanup will clean up any dangling mounts created during a startup
+// procedure, as well as terminate the VM. This is used when a sandbox
+// failed during create.
+func (s *Sandbox) Cleanup() {
+	// stop VM
+	s.stop()
+	// unmount and delete mount dirs
+	bindUnmountAllRootfs(kataHostSharedDir, s)
+	os.RemoveAll(filepath.Join(kataHostSharedDir, s.id))
+	// delete any other sandbox dirs (e.g. in /run/vc/sbs)
+	s.storage.deleteSandboxResources(s.id, nil)
+	// delete from shared mem
+	globalSandboxList.removeSandbox(s.id)
+}


virtcontainers/sandbox.go
```diff
@@ -1433,3 +1433,18 @@ func (s *Sandbox) AddVhostUserDevice(devInfo api.VhostUserDevice, devType config
...
+
+// Cleanup will clean up any dangling mounts created during a startup
+// procedure, as well as terminate the VM. This is used when a sandbox
+// failed during create.
+func (s *Sandbox) Cleanup() {
+	// stop VM
+	s.stop()
+	// unmount and delete mount dirs
+	bindUnmountAllRootfs(kataHostSharedDir, s)
+	os.RemoveAll(filepath.Join(kataHostSharedDir, s.id))
```
jshachm: `kataHostSharedDir` is specific to the `kata_agent.go` implementation, and should never be used by the generic  sandbox implementation. And about this, maybe something has been solved by #292

grahamwhaley:
Notes then:
 - dir cleanup is good - I think we are missing these bits. But, the hangup (also such as I see over at https://github.com/kata-containers/tests/pull/414), I suspect is not because we left dirs laying around - that might slow things down (as it walks lots of dirs maybe), but it should not hang the runtime. I suspect there is a deeper rooted issue as well, which we may well be able to treat separately.

- As this PR is asking for some input, and afaict does not quite conform to the PR and patch details found at https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md, I'm going to mark this RFC/WIP/DNM for the minute (@jamiehannaford - hope that is OK, and feel free to fix as/when ready).
- possibly due to the above, the CIs are not happy (probably expected).

virtcontainers/sandbox.go
```diff
@@ -1433,3 +1433,18 @@ func (s *Sandbox) AddVhostUserDevice(devInfo api.VhostUserDevice, devType config
 	}
 	return fmt.Errorf("unsupported device type")
 }
+
+// Cleanup will clean up any dangling mounts created during a startup
+// procedure, as well as terminate the VM. This is used when a sandbox
+// failed during create.
+func (s *Sandbox) Cleanup() {
```
bergwolf: The function does not need to be exported.

virtcontainers/sandbox.go
sboeuf: @jamiehannaford I agree with @jshachm that you cannot use `Kata` specific paths from `sandbox.go`, this is specific to the agent implementation.
Talking about #292, I have just merged it !

virtcontainers/api.go
```diff
@@ -45,23 +45,30 @@ func createSandboxFromConfig(sandboxConfig SandboxConfig) (*Sandbox, error) {
 		return nil, err
 	}
 
+	defer func() {
+		if err != nil {
+			s.Logger().WithError(err).WithField("sandboxid", s.id).Error("Creating sandbox failed")
+			s.Cleanup()
```
sboeuf: I am not sure if we can simply apply one big cleanup function because the failure could be caused by a bunch of different reasons, and using a hammer to cleanup might end up with some conflicts.
Trying to be more concrete here, let's say this function fails when creating the network, we don't want to call into `s.stop()` because this will try to communicate with the agent to stop the sandbox inside the VM, but the VM does not even exist at this point. That's why I think something with finer granularity would be more appropriate.

grahamwhaley:
I was just reading though this area of code (trying to figure out where my errant QEMU went...), and saw the teardown code. It occurred to me that we probably need some sort of 'teardown stack', and if we fail at some point we then call all the relevant teardown funcs in the reverse order they were added to the stack (so, yes, a real stack). I dunno if there is a nice idiom for doing this in golang?

sboeuf:
@grahamwhaley the way to do that in Golang is by using `defer`. Each defer is stacked up and at the end the stack is read.

grahamwhaley:
:-) I guess that works well if we have a single call chain for the whole setup. And we could use the `if err != nil` idiom inside the defer'd teardown funcs. defer didn't occur to me as I wasn't thinking this was a teardown through a single callchain, but maybe it is...

egernst:
@jamiehannaford - I appreciate the contribution here - more graceful failing is an important area for improvement imo.  I'm trying to scrub our backlog and reduce the number of open PRs in the project and  see that this PR has gone quiet and is close to getting stale.  

I see https://github.com/kata-containers/runtime/pull/351 similarly handles the error case.  That's close to merge now -- can we re-evaluate the behavior once this merges? What do you think?

jodh-intel:
@jamiehannaford - https://github.com/kata-containers/runtime/pull/351 has now landed so would be good to get some input to @egernst's query above.

jamiehannaford:
Based on the feedback and other PR, I think we can close this. Thanks!

jodh-intel:
Thanks @jamiehannaford!

Issue #344, opened 2018-05-29
Isolate hypervisor process with cgroups


sboeuf:
The hypervisor process is responsible of taking care of the VMs, and if a security breach was related to the hypervisor itself, it could give a way for a VM to escape.
In order to increase the security, we should constraint the hypervisor process inside using `cgroups` as much as possible.
For instance, the CPU cgroup can be a simple sum of the number of CPUs needed by the containers running inside the VM.

sboeuf:
Relates to https://github.com/kata-containers/runtime/issues/168

jshachm:
Good issue! 
And besides this, should we do this like what `libvirt` does: put `vcpu thread` and `emulator thread`  into different sub cgroups path.
And maybe we need a good `cgroups operation pkg` ,if we want to implementation features.
What do you think? @sboeuf  :-)

sboeuf:
@jshachm can we get this granularity (I mean identifying the threads) given that we are using the Qemu command line and not the libvirt package in our case ?

jshachm:
@sboeuf 
Yep, AFAIK, `vcpu thread` has some characteristics like ` /proc/xxx/task/comm` contains `vcpu` etc. 
We can use these to split them.

Talking about `cgroups control pkg`. Can we use [`containerd/cgroups` ](https://github.com/containerd/cgroups) since it is a native `cgroups pkg` used in `containerd`.

cc\ @WeiZhang555   @egernst  @jingxiaolu

jingxiaolu:
Discussed with @jshachm @WeiZhang555 offilined, we can work on this issue.

I do agree with @jshachm that we should introduce a pkg to maintain cgroup handlings in `runtime` :-)

sboeuf:
A global package sounds like a good idea. About the step to approach this issue, I think we can split into sub issues, with first constraining the qemu process into some cgroups, and as a second step, use a finer granularity and identify specific threads as suggested by @jshachm.

jodh-intel:
Let's also be careful to make the code generic enough to handle isolating any hypervisor, not just qemu :)

Further, we need to ensure that we can unit test this fully by launching a fake hypervisor which runs the equivalent of the following so we can check to ensure the code dtrt:

```
$ (ls -l /proc/self/ns; cat /proc/self/cgroup)
```

jingxiaolu:
Refactoring cgroup handling in runtime is almost ok, I hope a draft can be sent to you in next few days :-)

What I've done:
[...]

PR #405, opened 2018-06-18
vendor: introduce libcontainer/cgroups package

jingxiaolu:
Introduce libcontainer/cgroups package for further cgroups handling refactor

Fixes: #344

Signed-off-by: Jingxiao Lu <lujingxiao@huawei.com>

PR #416, opened 2018-06-19
[WIP] Refactor cgroup handling

jingxiaolu:
According to #344, this PR is trying to refactor cgroups handling in `runtime`.

What I've done in this PR:
1. introduce `libcontainer/cgroups` package for cgroups handling;
2. move cgroups handlings from CLI to sandbox level; refactor the handlings by the help of `libcontainer/cgroups`;

Works to be continue:
1. Add tests as @jodh-intel mentioned;
2. Add handling to `UpdateContainer()`;
3. Remove old cgroups handling codes

Fixes: #344

Signed-off-by: Jingxiao Lu lujingxiao@huawei.com

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -120,8 +120,11 @@
   name = "github.com/opencontainers/runc"
   packages = [
     "libcontainer/configs",
+    "libcontainer/cgroups",
+    "libcontainer/cgroups/fs",
     "libcontainer/seccomp",
     "libcontainer/specconv",
+    "libcontainer/system",
     "libcontainer/utils"
   ]
   revision = "0351df1c5a66838d0c392b4ac4cf9450de844e2d"
--- a/virtcontainers/cgroups.go
+++ b/virtcontainers/cgroups.go
@@ -1,0 +1,111 @@
+// +build linux
+// Copyright (c) 2018 Huawei Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package virtcontainers
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/kata-containers/runtime/virtcontainers/pkg/annotations"
+	"github.com/opencontainers/runc/libcontainer/cgroups"
+	"github.com/opencontainers/runc/libcontainer/cgroups/fs"
+	"github.com/opencontainers/runc/libcontainer/configs"
+	"github.com/opencontainers/runc/libcontainer/specconv"
+	"github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// cgroupsManager maintains cgroups implementations and config
+type cgroupsManager struct {
+	libcontainerManager cgroups.Manager
+	libcontainerConfig  *configs.Config
+}
+
+// newManager setup cgroup manager for sandbox
+func (cgm *cgroupsManager) newManager(s *Sandbox) error {
+	ociConfigStr, err := s.Annotations(annotations.ConfigJSONKey)
+	if err != nil {
+		return err
+	}
+
+	var ociSpec specs.Spec
+	if err := json.Unmarshal([]byte(ociConfigStr), &ociSpec); err != nil {
+		return err
+	}
+
+	cgm.libcontainerConfig, err = specconv.CreateLibcontainerConfig(&specconv.CreateOpts{
+		CgroupName:       s.id,
+		UseSystemdCgroup: false,
+		NoPivotRoot:      false,
+		NoNewKeyring:     false,
+		Spec:             &ociSpec,
+	})
+
+	if err != nil {
+		return fmt.Errorf("create container config for %s failed with %s", s.id, err)
+	}
+
+	state, _ := s.storage.fetchSandboxState(s.id)
+	if state.CgroupPaths == nil {
+		cgm.libcontainerManager = &fs.Manager{
+			Cgroups: cgm.libcontainerConfig.Cgroups,
+			Paths:   nil,
+		}
+	} else {
+		cgm.libcontainerManager = &fs.Manager{
+			Cgroups: cgm.libcontainerConfig.Cgroups,
+			Paths:   state.CgroupPaths,
+		}
+	}
+
+	s.cgroups = *cgm
+
+	return nil
+}
+
+// addSandbox adding shim pid to host cgroups and set the resource limitation with cgroups
+func (cgm *cgroupsManager) addSandbox(s *Sandbox) error {
+	shimPid := s.state.Pid
+	if err := cgm.libcontainerManager.Apply(shimPid); err != nil {
+		return fmt.Errorf("apply %d to host cgroups of sandbox %s failed with %s", shimPid, s.id, err)
+	}
+
+	if cgm.libcontainerConfig == nil {
+		return fmt.Errorf("setup host cgroup of sandbox %s failed with libcontainerConfig is nil", s.id)
+	}
+
+	if err := cgm.libcontainerManager.Set(cgm.libcontainerConfig); err != nil {
+		return fmt.Errorf("setup host cgroups of sandbox %s failed with %s", s.id, err)
+	}
+
+	s.state.CgroupPaths = cgm.libcontainerManager.GetPaths()
+	s.storage.storeSandboxResource(s.id, stateFileType, s.state)
+
+	s.Logger().Infof("set host cgroup for %v successful", s.id)
+	return nil
+}
+
+// addContainer adding shim pid of container to sandbox's host cgroups
+func (cgm *cgroupsManager) addContainer(c *Container) error {
+	shimPid := c.process.Pid
+	if err := cgm.libcontainerManager.Apply(shimPid); err != nil {
+		return fmt.Errorf("apply %d of container %s to host cgroups of sandbox %s failed with %s", shimPid, c.id, c.sandboxID, err)
+	}
+	return nil
+}
+
+// deleteSandbox cleanup cgroup folders
+func (cgm *cgroupsManager) deleteSandbox(s *Sandbox) {
+	if cgm.libcontainerManager == nil {
+		s.Logger().Warningf("destroying host cgroups for %s failed: libcontainerManager is nil!", s.id)
+	}
+
+	if err := cgm.libcontainerManager.Destroy(); err != nil {
+		s.Logger().Warningf("destroy host cgroups for %s failed: %v", s.id, err)
+	}
+
+	return
+}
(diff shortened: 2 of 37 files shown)


jingxiaolu:
This PR is re-submitted according to @sboeuf's comments at #405 .

@sboeuf @jodh-intel @grahamwhaley @devimc although is PR is still WIP, could you help to pay some time on it? Many thanks~

cc\ @WeiZhang555 @jshachm

jingxiaolu:
CI is failing, let me fix it first...

virtcontainers/cgroups.go
```diff
@@ -0,0 +1,111 @@
+// +build linux
```
jodh-intel: Could you remove this comment and just rename the file to be `virtcontainers/cgroups_linux.go` as that's much clearer imho.

virtcontainers/cgroups.go
```diff
@@ -0,0 +1,111 @@
...
+		NoNewKeyring:     false,
+		Spec:             &ociSpec,
+	})
+
+	if err != nil {
+		return fmt.Errorf("create container config for %s failed with %s", s.id, err)
+	}
+
+	state, _ := s.storage.fetchSandboxState(s.id)
+	if state.CgroupPaths == nil {
```
jodh-intel: I don't really understand this code. Why can't it be just...

```
state, err := s.storage.fetchSandboxState(s.id)
if err != nil {
    return err
}

cgm.libcontainerManager = &fs.Manager{
    Cgroups:    cgm.libcontainerConfig.Cgroups,
    Paths:      state.CgroupPaths,
}
```

Maybe a comment in the code would help here.

virtcontainers/cgroups.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// addSandbox adding shim pid to host cgroups and set the resource limitation with cgroups
+func (cgm *cgroupsManager) addSandbox(s *Sandbox) error {
+	shimPid := s.state.Pid
+	if err := cgm.libcontainerManager.Apply(shimPid); err != nil {
+		return fmt.Errorf("apply %d to host cgroups of sandbox %s failed with %s", shimPid, s.id, err)
+	}
+
+	if cgm.libcontainerConfig == nil {
```
jodh-intel: Shouldn't this test be the first one in the function (fail fast)?

virtcontainers/cgroups.go
```diff
@@ -0,0 +1,111 @@
...
+func (cgm *cgroupsManager) addContainer(c *Container) error {
+	shimPid := c.process.Pid
+	if err := cgm.libcontainerManager.Apply(shimPid); err != nil {
+		return fmt.Errorf("apply %d of container %s to host cgroups of sandbox %s failed with %s", shimPid, c.id, c.sandboxID, err)
+	}
+	return nil
+}
+
+// deleteSandbox cleanup cgroup folders
+func (cgm *cgroupsManager) deleteSandbox(s *Sandbox) {
```
jodh-intel: It looks like this function should return an `error` as there are error scenarios it has to deal with?

Gopkg.lock
```diff
@@ -120,8 +120,11 @@
   name = "github.com/opencontainers/runc"
   packages = [
     "libcontainer/configs",
+    "libcontainer/cgroups",
+    "libcontainer/cgroups/fs",
```
WeiZhang555: Why is "libcontainer/cgroups/systemd" not involved? Is it unnecessary?

virtcontainers/cgroups.go
WeiZhang555: And, you need a new file named "virtcontainers/cgroups_unsupported.go" contains exactly same interface with current implementation(old codes) for other platforms other than linux. Or it won't compile on ppc64le

WeiZhang555: You can imitating 
```
 vendor/github.com/opencontainers/runc/libcontainer/cgroups/cgroups_unsupported.go
@@ -0,0 +1,3 @@
 +// +build !linux
 +
 +package cgroups
```

virtcontainers/cgroups.go
```diff
@@ -0,0 +1,111 @@
...
+
+// cgroupsManager maintains cgroups implementations and config
+type cgroupsManager struct {
+	libcontainerManager cgroups.Manager
+	libcontainerConfig  *configs.Config
+}
+
+// newManager setup cgroup manager for sandbox
+func (cgm *cgroupsManager) newManager(s *Sandbox) error {
+	ociConfigStr, err := s.Annotations(annotations.ConfigJSONKey)
```
bergwolf: Where is `s.Annotations()` defined? I can't find it in the PR. Ideally we should only unmarshal the OCI json once for each container, since it is really slow... And for an empty sandbox w/o containers (e.g. in the CRI case), there is no such container OCI spec. You need to handle that case as well. So IMO the cgroup manager needs to be created upon first container creation instead.

virtcontainers/cgroups.go
```diff
@@ -0,0 +1,111 @@
...
+	s.state.CgroupPaths = cgm.libcontainerManager.GetPaths()
+	s.storage.storeSandboxResource(s.id, stateFileType, s.state)
+
+	s.Logger().Infof("set host cgroup for %v successful", s.id)
+	return nil
+}
+
+// addContainer adding shim pid of container to sandbox's host cgroups
+func (cgm *cgroupsManager) addContainer(c *Container) error {
+	shimPid := c.process.Pid
```
bergwolf: Need to check for `shimPid > 0` to exclude the builtin shim case.

virtcontainers/cgroups.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	s.cgroups = *cgm
+
+	return nil
+}
+
+// addSandbox adding shim pid to host cgroups and set the resource limitation with cgroups
+func (cgm *cgroupsManager) addSandbox(s *Sandbox) error {
+	shimPid := s.state.Pid
```
bergwolf: What is `s.state.Pid`? A sandbox does not have a corresponding shim. All shims are associated with containers instead.

virtcontainers/container.go
```diff
@@ -664,6 +664,10 @@ func createContainer(sandbox *Sandbox, contConfig ContainerConfig) (c *Container
 		sandbox.setSandboxPid(c.process.Pid)
 	}
 
+	if ann[annotations.ContainerTypeKey] == string(PodContainer) {
```
bergwolf: PodSandbox and PodContainer are both annotations in the kata CLI to get the missing sandbox abstraction from runc compatible command lines. There is no need to use such annotation in virtcontainers, where we know clearly about sandbox vs. containers.

virtcontainers/sandbox.go
```diff
@@ -463,6 +466,8 @@ type Sandbox struct {
 	wg *sync.WaitGroup
 
 	shmSize uint64
+
+	cgroups cgroupsManager
```
bergwolf: `cgroups *cgroupsManager`?

Gopkg.lock
jingxiaolu: Should we support systemd containers? If yes, I'll add "libcontainer/cgroups/systemd"~

virtcontainers/cgroups.go
jshachm: `sandbox.state.pid` is shim pid of the first container in the pod, in other word means `pause` container.

bergwolf: First, a sandbox can be empty in which case there is no container in it at all. Secondly you cannot assume the first container in a sandbox is always a `pause` container that never quits. Such assumption breaks with docker and frakti case.

virtcontainers/container.go
jingxiaolu: which means: once `createContainer()` is called, we're clearly creating a container, I don't need to check `PodContainer` or what, just add the container's pid to cgroup.
Am I clear?~

virtcontainers/sandbox.go
jingxiaolu: Yes, I adding this to enclose the implementations and data of cgroups handlings in `cgroups.go` (will be `cgroups_linux.go` and `cgroups_unsupported.go`)

virtcontainers/cgroups.go
bergwolf: Yes, because you rely on a container OCI spec to create the cgroup manager and we won't have  a container spec until the first **container** is to be created.

virtcontainers/cgroups.go
bergwolf: `shimPid == 0` happens in noop_shim case. You can just return nil IMO.

virtcontainers/container.go
bergwolf: Yes, your understanding is correct.

virtcontainers/cgroups.go
jshachm: Sry for missing for frakti case~

virtcontainers/cgroups.go
jingxiaolu: Currently the only place unmarshaling the OCI json is in [`kata-agent level`](https://github.com/kata-containers/runtime/blob/c324b55255186d22682bd74cde1beb50c1df8afc/virtcontainers/kata_agent.go#L778), but I think `newManager()` should be called at `sandbox level`.

I would like to:
1. add a pointer named `ociSpec` at `Sandbox struct`;
2. unmarshal in `newSandbox()` and assign it to `ociSpec`;
3. when `createContainer()` in `kata-agent` level, get it from `ociSpec`;

Please share your comments, thanks~

bergwolf: @jingxiaolu 

1. Yes, a pointer to the first ocispec in `Sandbox` makes sense since it avoids unmarshalling the same json twice.

2. As I stated above, one issue with `newManager()` in `newSandbox()` is that there may be no containers in sandboxConfig. Then you do not have the OCI spec you need to create the cgroup manager. The right place to do it is in `createContainer()`.

3. Yes, it makes sense.

WeiZhang555:
@jingxiaolu is currently too busy and can't get enough time on this. I'll carry his work and take over the process.

jodh-intel:
Thanks @WeiZhang555 - branch needs updating due to conflicts too btw.

bergwolf:
@WeiZhang555 Any updates on this one?

jodh-intel:
Ping @WeiZhang555 :)

sboeuf:
@WeiZhang555 I bet you're very busy, but just checking if this is something you're planning to look at?
I'm asking because we could reassign this to someone else that might have more bandwidth if someone's interested.

WeiZhang555:
@sboeuf Sorry for delay, let me try to finish it in several days. It is truly blocked for so many months!
I'll give some update in this week. But if anyone is so interested in implementing this, welcome!

WeiZhang555:
Closing this. New implementation is https://github.com/kata-containers/runtime/pull/734

PR #417, opened 2018-06-19
virtcontainers: Set ppc64le maxmem depending on qemu version

nitkon:
The "Failed to allocate HTAB of requested size,
try with smaller maxmem" error in ppc64le occurs
when maxmem allocated is very high. This got fixed
in qemu 2.10 and kernel 4.11. Hence put a maxmem
restriction of 32GB per kata-container if qemu
version less than 2.10

Fixes: #415

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -64,6 +64,9 @@ const qmpCapErrMsg = "Failed to negoatiate QMP capabilities"
 
 const defaultConsole = "console.sock"
 
+var qemuMajorVersion int
+var qemuMinorVersion int
+
 // agnostic list of kernel parameters
 var defaultKernelParameters = []Param{
 	{"panic", "1"},
@@ -509,6 +512,8 @@ func (q *qemu) waitSandbox(timeout int) error {
 	}
 
 	q.qmpMonitorCh.qmp = qmp
+	qemuMajorVersion = ver.Major
+	qemuMinorVersion = ver.Minor
 
 	q.Logger().WithFields(logrus.Fields{
 		"qmp-major-version": ver.Major,
--- a/virtcontainers/qemu_ppc64le.go
+++ b/virtcontainers/qemu_ppc64le.go
@@ -12,6 +12,7 @@ import (
 	govmmQemu "github.com/intel/govmm/qemu"
 	"github.com/kata-containers/runtime/virtcontainers/device/drivers"
 	"github.com/kata-containers/runtime/virtcontainers/utils"
+	"github.com/sirupsen/logrus"
 )
 
 type qemuPPC64le struct {
@@ -52,6 +53,11 @@ var supportedQemuMachines = []govmmQemu.Machine{
 		Type:    QemuPseries,
 		Options: defaultQemuMachineOptions,
 	},
+}
+
+// Logger returns a logrus logger appropriate for logging qemu messages
+func (q *qemuPPC64le) Logger() *logrus.Entry {
+	return virtLog.WithField("subsystem", "qemu")
 }
 
 // MaxQemuVCPUs returns the maximum number of vCPUs supported
@@ -105,12 +111,14 @@ func (q *qemuPPC64le) cpuModel() string {
 
 func (q *qemuPPC64le) memoryTopology(memoryMb, hostMemoryMb uint64) govmmQemu.Memory {
 
-	if hostMemoryMb > defaultMemMaxPPC64le {
-		hostMemoryMb = defaultMemMaxPPC64le
-	} else {
-		// align hostMemoryMb to 256MB multiples
+	if qemuMajorVersion >= 2 && qemuMinorVersion >= 10 {
+		q.Logger().Debug("Aligning maxmem to multiples of 256MB. Assumption: Kernel Version >= 4.11")
 		hostMemoryMb -= (hostMemoryMb % 256)
+	} else {
+		q.Logger().Debug("Restricting maxmem to 32GB as Qemu Version < 2.10, Assumption: Kernel Version >= 4.11")
+		hostMemoryMb = defaultMemMaxPPC64le
 	}
+
 	return genericMemoryTopology(memoryMb, hostMemoryMb)
 }
 


virtcontainers/qemu.go
```diff
@@ -64,6 +64,9 @@ const qmpCapErrMsg = "Failed to negoatiate QMP capabilities"
 
 const defaultConsole = "console.sock"
 
+var qemuMajorVersion int
+var qemuMinorVersion int
```
jodh-intel: I wonder if we want to store these in one of the types, even though they are only being used for PPC64le atm.

virtcontainers/qemu_ppc64le.go
```diff
@@ -105,12 +105,13 @@ func (q *qemuPPC64le) cpuModel() string {
 
 func (q *qemuPPC64le) memoryTopology(memoryMb, hostMemoryMb uint64) govmmQemu.Memory {
 
-	if hostMemoryMb > defaultMemMaxPPC64le {
-		hostMemoryMb = defaultMemMaxPPC64le
-	} else {
+	if qemuMajorVersion >= 2 && qemuMinorVersion >= 10 {
```
jodh-intel: I'd consider adding a couple of logging calls here, one for each scenario so it's clear what's happening from the log.

What we cannot reliably check the guest kernel version so maybe your debug entry for handling qemu 2.10+ should mention the 4.11 guest kernel assumption.

nitkon: @jodh-intel : Done.  Updated my PR.

virtcontainers/qemu.go
nitkon: @jodh-intel : Should we? I am not sure if its gonna get consumed by any other architectures atm.

nitkon:
@jodh-intel :  I see the following error in the failed CI.
```
ERROR: Invalid URL 'http://download.opensuse.org/repositories/home:clearlinux:preview:clear-containers-2.1/Fedora_25/home:clearlinux:preview:clear-containers-2.1.repo' found in the following files:
./virtcontainers/hack/virtc/README.md
ERROR: Invalid URL 'http://download.opensuse.org/repositories/home:/clearlinux:/preview:/clear-containers-2.1/xUbuntu_16.10/' found in the following files:
./virtcontainers/hack/virtc/README.md
```

Should the CI be restarted?

jodh-intel:
Jobs restarted.

nitkon:
@jodh-intel : All `required` CI are green

grahamwhaley:
On the F27 CI:
```
# time="2018-06-20 10:30:42.502996695Z" level=debug msg="event: "/var/run/crio/exits/975e6d0a7f676053db9d0bf52cbefb7b6487a0a5161366ad36270e5b8ae730b6": CREATE" 
# time="2018-06-20 10:30:42.503002488Z" level=debug msg="container or sandbox exited: 975e6d0a7f676053db9d0bf52cbefb7b6487a0a5161366ad36270e5b8ae730b6" 
# time="2018-06-20 10:30:42.503009278Z" level=debug msg="container exited and found: 975e6d0a7f676053db9d0bf52cbefb7b6487a0a5161366ad36270e5b8ae730b6" 
# rm: cannot remove '/tmp/tmp.BGiXFsa5eZ/crio-run/vfs-containers/5c06d787b94818845e6986598960770958b1c21febf5cf5f69bbf288138e41d3/userdata/shm': Device or resource busy
Build timed out (after 5 minutes). Marking the build as aborted.
Build was aborted
```

Known and wholly unrelated @jodh-intel @chavafg ? If so, then we can merge this I think (the button is not green, but is 'pressable' ;-) )

virtcontainers/qemu.go
sboeuf: I would definitely include those as part of the `qemu` structure since this is an info about the hypervisor. And I agree with @jodh-intel that even if it's only used by ppc64, this is not a reason to define some global variable there.

nitkon: @sboeuf : Okay, I shall add qemuMajorVersion and qemuMinorVersion into the qemu structure. However, I am unsure of what would be the most efficient way of exposing it in `func (q *qemuPPC64le) memoryTopology`.  One way is to get qemu version info in `func (q *qemu) init(sandbox *Sandbox) error` and pass via `newQemuArch` resulting in change of signature of `newQemuArch` everywhere.

jodh-intel: It's a bit fiddly, but we could add them to `qemuArchBase` like this:

- https://github.com/jodh-intel/runtimes/tree/set-and-get-qemu-version.

sboeuf: @nitkon oh yeah my bad, I forgot `qemu` structure was not the interface related to `qemu_ppc64le`. As @jodh-intel suggested, `qemuArchBase` should be the one.

@jodh-intel, I have taken a look at your suggestion but I thought we could simplify this by simply having those fields being called from `qemu_ppc64le` structure (no need for getters and setters). Isn't it possible since this structure inherits from `qemuArchBase` ?

jodh-intel: I'm not quite sure what you're saying tbh. I'm hoping there is a simpler way though as I'm not particularly happy with what I came up with :)

nitkon: @jodh-intel : So I was going through the implementation but noticed the following.
`Printing q.qemuMajorVersion, q.qemuMinorVersion in "func (q qemuArchBase) getQEMUVersion() (major, minor int) {" shows me the following:
level=info msg="DEBUG: versions: major: 0, minor: 0" arch=amd64 name=kata-runtime pid=6215 .`

nitkon: @jodh-intel : I am a bit confused by the reference implementation. We are setting qemuMajorVersion and qemuMinorVersion using one qemu instance and getting it with another qemu instance. (resulting it in 0,0 values). 

IIUC, we need to set/get using the same qemu instance

sboeuf:
@nitkon please backport this PR to `stable-1.1` and `stable-1.2`

sboeuf:
Nevermind, it's already part of the stable branches.

Issue #386, opened 2018-06-12
Bug: Message too long 13110


linxiulei:
# Description of problem

Create a container with an env file, which has size bigger than 10240 bytes, cmd like

```
docker run --runtime cc-runtime --env-file /tmp/env -it busybox sh
```

then

```
docker run --runtime cc-runtime busybox ls
```

it has no response anymore

# Expected result

it's supposed to run normally

# Actual result

docker cmd no response

---

# Meta details

Running `cc-collect-data.sh` version `3.0.22 (commit 4b569ac)` at `2018-06-12.17:13:02.305371930+0800`.

---

Runtime is `/usr/bin/cc-runtime`.

# `cc-env`

Output of "`/usr/bin/cc-runtime cc-env`":
```
[Meta]
  Version = "1.0.9"

[Runtime]
  Debug = false
  [Runtime.Version]
    Semver = "3.0.22"
    Commit = "4b569ac"
    OCI = "1.0.1"
  [Runtime.Config]
    Path = "/etc/clear-containers/configuration.toml"

[Hypervisor]
  MachineType = "pc"
  Version = "QEMU emulator version 2.7.1 (), Copyright (c) 2003-2016 Fabrice Bellard and the QEMU Project developers"
  Path = "/usr/local/bin/qemu-system-x86_64"
  Debug = false
  BlockDeviceDriver = "virtio-scsi"

[Image]
  Path = "/usr/share/clear-containers/cc-20640-agent-62f3f7.img"

[Kernel]
  Path = "/usr/share/clear-containers/vmlinuz-4.9.47-82.1.alios7.container"
  Parameters = ""

[Proxy]
  Type = "ccProxy"
  Version = "Version: 3.0.22+git.e7ee393-9bf9e916e8a6bc173fa71fe1fe796b37d077d62f"
  Path = "/usr/libexec/clear-containers/cc-proxy"
  Debug = false

[Shim]
  Type = "ccShim"
  Version = "shim version: 3.0.22 (commit: 10a255b)"
  Path = "/usr/libexec/clear-containers/cc-shim"
  Debug = false

[Agent]
  Type = "hyperstart"
  Version = "<<unknown>>"

[Host]
  Kernel = "4.4.95-1.al7.x86_64"
  Architecture = "amd64"
  VMContainerCapable = true
  [Host.Distro]
    Name = "Alibaba Cloud Enterprise Linux Server"
    Version = "17.01"
  [Host.CPU]
[...]

PR #425, opened 2018-06-20
fix message too long

malc0lm:
Since [hyper_ctlfd_read()](https://github.com/hyperhq/hyperstart/blob/master/src/init.c#L1290) can realloc buffer, we don't limit 10240 message size.
Fixes: #386
Signed-off-by:malc0lm malclee@outlook.com

### DIFF
--- a/virtcontainers/pkg/hyperstart/hyperstart.go
+++ b/virtcontainers/pkg/hyperstart/hyperstart.go
@@ -327,12 +327,8 @@ func ReadCtlMessage(conn net.Conn) (*DecodedMessage, error) {
 // hyperstart control serial link, use SendCtlMessage.
 func (h *Hyperstart) WriteCtlMessage(conn net.Conn, m *DecodedMessage) error {
 	length := len(m.Message) + CtlHdrSize
-	// XXX: Support sending messages by chunks to support messages over
-	// 10240 bytes. That limit is from hyperstart src/init.c,
-	// hyper_channel_ops, rbuf_size.
-	if length > 10240 {
-		return fmt.Errorf("message too long %d", length)
-	}
+	// Since hyper_ctlfd_read() can realloc buffer, we don't limit 10240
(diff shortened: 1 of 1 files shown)


gnawux:
why `hyperstart.go`? Is the `hyperstart` package called when kata agent is in use? cc @bergwolf @egernst

egernst:
@gnawux, shouldn't be AFAIU.

gnawux:
@egernst then I think #386 shouldn't reach this line, right?

bergwolf:
@gnawux No, it is not used by kata. hyperstart.go is talking to the cc-agent in hyperstart protocol.

grahamwhaley:
Just for reference, I literally happened to just see this:
https://github.com/containers/virtcontainers/issues/99

virtcontainers/pkg/hyperstart/hyperstart.go
```diff
@@ -327,12 +327,8 @@ func ReadCtlMessage(conn net.Conn) (*DecodedMessage, error) {
 // hyperstart control serial link, use SendCtlMessage.
 func (h *Hyperstart) WriteCtlMessage(conn net.Conn, m *DecodedMessage) error {
 	length := len(m.Message) + CtlHdrSize
-	// XXX: Support sending messages by chunks to support messages over
-	// 10240 bytes. That limit is from hyperstart src/init.c,
-	// hyper_channel_ops, rbuf_size.
-	if length > 10240 {
-		return fmt.Errorf("message too long %d", length)
-	}
+	// Since hyper_ctlfd_read() can realloc buffer, we don't limit 10240
```
sboeuf: Unfortunately, you're talking about hyperstart here but Clear Containers uses it's own version of hyperstart, that is called `cc-agent`, and I don't think this work...

malc0lm:
Ok，I close this pr.

Issue #426, opened 2018-06-20
Expose share sandbox pidns setting


bergwolf:
The kata-agent allows for sandbox level pidns sharing and we always set it to false at the moment. We should expose it via `SandboxConfig` and let callers decide.

PR #427, opened 2018-06-20
sandbox: expose share sandbox pidns setting

bergwolf:
So that we let callers decide if kata-agent should let all containers in
a sandbox share the same pid namespace.

Fixes: #426

Signed-off-by: Peng Tao <bergwolf@gmail.com>

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -528,7 +528,7 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
 	req := &grpc.CreateSandboxRequest{
 		Hostname:     hostname,
 		Storages:     storages,
-		SandboxPidns: false,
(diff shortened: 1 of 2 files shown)


jodh-intel:
The change looks fine but there doesn't appear to be any way to use this new setting? I thought there might need to be a change to `virtcontainers/pkg/oci/utils.go:SandboxConfig()` or something?

bergwolf:
@jodh-intel This will be first used only by frakti. And kata cli can possibly use it as well when cri-o and containerd-cri stops creating `pause` containers and just pass the `CreateSandbox` CRI requests to kata.

jodh-intel:
Hi @bergwolf - could you add a bit more detail to the commit message to explain some of that?

lgtm

bergwolf:
@jodh-intel commit message updated.

jodh-intel:
Thanks @bergwolf.

lgtm

PR #429, opened 2018-06-20
qemu: clean up qmp channel

bergwolf:
We only need one qmp channel and it is qemu internal detail thus
sandbox.go does not need to be aware of it.

This was part of #303 and but does not really relate to vm factory so I take it out.

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -7,7 +7,6 @@ package virtcontainers
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


jodh-intel:
lgtm

The CI is failing due to lint issues. I don't yet understand why since:

- we're pinning gometalinter to a particular commit (and that hasn't been changed in a while).
- this PR doesn't touch the files that the linters are complaining about.

jodh-intel:
Ugh - I've think I've just worked it out: https://github.com/kata-containers/tests/issues/439.

Issue #431, opened 2018-06-21
kata agent connection creation race


bergwolf:
If we send multiple grpcs at once before a client is being created, we end up
creating multiple connections to the agent and that breaks when using builtin
proxy since only one connection is allowed.

PR #432, opened 2018-06-21
kata_agent: fix connection race

bergwolf:
If we send multiple grpcs at once when a client is being created, we end up
creating multiple connections to the agent and it breaks when using builtin
proxy since only one connection is allowed.

Fixes: #431

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


bergwolf:
@grahamwhaley Updated. This was found when integrating kata with frakti.

virtcontainers/kata_agent.go
```diff
@@ -1089,6 +1094,14 @@ func (k *kataAgent) statsContainer(sandbox *Sandbox, c Container) (*ContainerSta
 }
 
 func (k *kataAgent) connect() error {
+	// lockless quick pass
```
WeiZhang555: nit: quick path?

WeiZhang555:
LGTM

Issue #443, opened 2018-06-25
New Storage/Device APIs to handle device-less storages 


harche:
Storages that do not have a backing device (such as `ephemeral`) need to be handled properly at the sandbox level. There are few suggestions by @bergwolf and @WeiZhang555 that need to be sorted out. 
We can use this issue as a placeholder to keep the discussion public. 

Requesting @bergwolf  and @WeiZhang555 and whoever else is interested to put their thoughts here so that a larger community can participate. 

The result of this discussion will lead to refactoring the support for `ephemeral` [PR](https://github.com/kata-containers/runtime/pull/307/)

WeiZhang555:
I'll copy some text from our previous discussion in Slack:

> For discussion, I want to share something in mind though it's not very clear either, so feel free to correct me.
> ```// Mount describes a container mount.                                               
> type Mount struct {                                                                 
>     Source      string                                                              
>     Destination string                                                              
>                                                                                     
>     // Type specifies the type of filesystem to mount.                              
>     Type string                                                                     
>                                                                                     
>     // Options list all the mount options of the filesystem.                        
>     Options []string                                                                
>                                                                                     
>     // HostPath used to store host side bind mount path                             
>     HostPath string                                                                 
>                                                                                     
>     // ReadOnly specifies if the mount should be read only or not                   
>     ReadOnly bool                                                                   
>                                                                                     
>     // BlockDevice represents block device that is attached to the                  
>     // VM in case this mount is a block device file or a directory                  
>     // backed by a block device.                                                    
>     BlockDevice *drivers.BlockDevice                                                
> }```            
> 
> For a Mount struct, besides normal information for a mountpoint, there're two additional fields: HostPath for 9p and BlockDevice for block:
> 1. 9pfs volume: There's a real existing dir on host, you can pass to many kata containers with multi bind mount, and unbind when one container leaves.
> 2. virtio-blk or virio-scsi backed file:  in use case "-v /dev/sda:/dev/sda",  we need to hotplug a real block device to hypervisor. Sometimes user would want to "bind mount" a block device to different containers, but from kata-runtime/hypervisor's perspective, it doesn't need to hotplug /dev/sda to hypervisor many times, it can hotplug once but reference it to many containers. Then how to know when should we unplug the block? --> we need to remember which container is the last reference and unplug the device when nobody is using it any more. This part is what I am working towards and not fully finished(see https://github.com/kata-containers/runtime/pull/301)
> 
[...]

PR #458, opened 2018-07-03
[WIP] [Do Not Merge] virtcontainers: Add support for ephemeral volumes

harche:
Ephemeral volumes should not be passed at 9pfs mounts.
They should be created inside the VM.

This patch disables ephemeral volumes from getting
mounted as 9pfs from the host and instead a corresponding
tmpfs is created inside the VM.

Fixes : #61

Signed-off-by: Harshal Patil <harshal.patil@in.ibm.com>


harche:
This PR implements the discussion https://github.com/kata-containers/runtime/issues/443 to avoid using devices at sandbox level.

WeiZhang555:
The design looks good to me, but need changes from kata-agent before we can merge this.

bergwolf:
@harche It seems that the PR works w/o agent change, right? The missing piece in kata agent is just the ref counting part but tmpfs handling is already there.

cli/create.go
```diff
@@ -249,6 +248,17 @@ func createSandbox(ociSpec oci.CompatOCISpec, runtimeConfig oci.RuntimeConfig,
...
 	console string, disableOutput bool) (vc.Process, error) {
 
+	// Set the mount type to 'ephemeral' if the mount source path
+	// is provisioned by k8s for ephemeral storage.
+	// For the given pod ephemeral volume is created only once
+	// backed by tmpfs inside the VM. For successive containers
+	// of the same pod the already existing volume is reused.
+	for idx, mnt := range ociSpec.Mounts {
+		if IsEphemeralStorage(mnt.Source) {
+			ociSpec.Mounts[idx].Type = "ephemeral"
```
sboeuf: I think we need to define the type `ephemeral` (as other types) inside a separate package that can be imported both from the agent and the runtime. This can be done in a follow up PR, but this would be better than simply hardcode the type at each end of the chain (runtime on one side, and the agent on the other side.)

WeiZhang555:
> @harche It seems that the PR works w/o agent change, right? The missing piece in kata agent is just the ref counting part but tmpfs handling is already there.

I believe it can work w/o agent change, as @bergwolf described, we need a reference counter in kata-agent to mark when the "ephemeral" volume can be unmounted, without this, the exit of any container could lead to umount of the tmpfs which will break the "ephemeral volume", this could happen when one container in POD failed to start and has to be restarted several times.

bergwolf:
@harche Could you please fix UT coverage and remove the wip tag? Then the PR is ready to get merged.

bergwolf:
The codecov failure `codecov/patch — 39.13% of diff hit (target 64.16%)` looks bogus since clearly the patch is already properly tested with UT.

I think we can safely merge this but I want to understand why `codecov/patch` fails. Looking at https://codecov.io/gh/kata-containers/runtime/commit/b821a5df4c1adb154613b80f981ba16684ed8b80, the change in kata_agent.go is clearly tested by UT but `diff coverage` reports `0%`.

In that case, I would suggest we either remove `codecov/patch`, or always ignore it during reviewing.

@egernst @sboeuf @jodh-intel @grahamwhaley @WeiZhang555 @gnawux WDYT?

bergwolf:
Moving `codecov/patch` discussion to kata-containers/tests/issues/508 so this PR can be merged.

Issue #461, opened 2018-07-04
ci: add specfic image-type for aarch64


Pennyzct:
# Description of problem
Default image-type(aka clearlinux) couldn't work in aarch64, so we add specifc image-type to avoid jenkins set-up failure. [#473](https://github.com/kata-containers/tests/pull/473)
@jongwu @weichen81 @jodh-intel

PR #462, opened 2018-07-04
ci: add specfic image-type for aarch64

Pennyzct:
default image-type(aka clearlinux) couldn't work in aarch64, so we add specifc image-type to avoid jenkins set-up failure.

Fixes: #461

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -76,18 +76,34 @@ assets:
       Root filesystem disk image used to boot the guest virtual
       machine.
     url: "https://github.com/kata-containers/osbuilder"
-    release: "20640"
+    architecture:
+      aarch64:
+        name: "fedora"
+        version: "latest"
+      ppc64le:
+        name: "centos"
+        version: "latest"
+      x86_64:
+        name: &default-image-name "clearlinux"
+        version: "20640"
     meta:
-      image-type: "clearlinux"
+      image-type: *default-image-name
 
   initrd:
     description: |
       Root filesystem initrd used to boot the guest virtual
       machine.
     url: "https://github.com/kata-containers/osbuilder"
-    meta:
-      base-os: "alpine"
-      os-version: "3.7"
+    architecture:
+      aarch64:
+        name: &default-initrd-name "alpine"
+        version: &default-initrd-version "3.7"
+      ppc64le:
+        name: *default-initrd-name
+        version: *default-initrd-version
+      x86_64:
+        name: *default-initrd-name
+        version: *default-initrd-version
 
   kernel:
     description: "Linux kernel optimised for virtual machines"


jcvenegas:
lgtm, I think would be nice to follow image-type-x86_64 to follow the same conversion. But that can be done in a follow up PR because will need to modify CI scripts as well.

versions.yaml
```diff
@@ -79,6 +79,7 @@ assets:
     release: "20640"
     meta:
       image-type: "clearlinux"
+      image-type-aarch64: "fedora"
```
jodh-intel: Hi @Pennyzct - thanks for raising this. I think you've hit on a limitation of the current structure of this file. As such, I think we could expand it to something like the following:

```diff
diff --git a/versions.yaml b/versions.yaml
index d616dc0..4d88730 100644
--- a/versions.yaml
+++ b/versions.yaml
@@ -76,18 +76,34 @@ assets:
       Root filesystem disk image used to boot the guest virtual
       machine.
     url: "https://github.com/kata-containers/osbuilder"
-    release: "20640"
+    architecture:
+      arm64:
+        name: "fedora"
+        version: "latest"
+      ppc64le:
+        name: "centos"
+        version: "latest"
+      x86_64:
+        name: &default-image-name "clearlinux"
+        version: "20640"
     meta:
-      image-type: "clearlinux"
+      image-type: *default-image-name
 
   initrd:
     description: |
       Root filesystem initrd used to boot the guest virtual
       machine.
     url: "https://github.com/kata-containers/osbuilder"
-    meta:
-      base-os: "alpine"
-      os-version: "3.7"
+    architecture:
+      arm64:
+        name: &default-initrd-name "alpine"
+        version: &default-initrd-version "3.7"
+      ppc64le:
+        name: *default-initrd-name
+        version: *default-initrd-version
+      x86_64:
+        name: *default-initrd-name
+        version: *default-initrd-version
 
   kernel:
     description: "Linux kernel optimised for virtual machines"
```

Note the "references" or anchors in the above so that we can still use the query below:

```sh
$ yq read versions.yaml assets.image.meta.image-type
clearlinux
```

That will ensure we don't break https://github.com/kata-containers/tests/blob/master/.ci/install_kata_image.sh#L98.

Note that in the diff above I've deleted `assets.image.release` as that doesn't appear to be used - could you confirm @jcvenegas, @chavafg? That value does "reappear" as `assets.image.architecture.x86_64.version` but since https://github.com/kata-containers/osbuilder/blob/master/rootfs-builder/clearlinux/config.sh#L9 specifies a default of `"default"`, I wonder if we can change it here too? wdyt @jcvenegas, @chavafg?

Finally, note that I've applied the same structure change to the `initrd` section. @jcvenegas - can you confirm that the two elements I deleted and replaced with a new structure (`assets.initrd.meta.base-os` and `assets.initrd.meta.os-name`) are not used?

/cc @nitkon, @grahamwhaley.

Pennyzct: This should be a template for all involved arch😊.

jcvenegas: @jodh-intel Clear Linux is a distro changes a lot every day. We probably will hit some issues if we follow their latest release all the time.  I'd prefer we dont follow latest automatically.

jodh-intel: Fair point - let's stick with a specific version for Clear Linux then.

Pennyzct:
ptal 😊@jodh-intel @jcvenegas

sboeuf:
@jodh-intel will be off for 2 weeks, let's merge this without waiting for his feedback here.

jodh-intel:
\o/ - nice! ;)

sboeuf:
@egernst clearly not a breaking change, but not a bug fix either. Should we backport it?

PR #470, opened 2018-07-05
hypervisor/qemu: add memory hotplug support

bergwolf:
As discussed in #303, I'm splitting the memory hotplug part from the vm factory PR to get more eyes on it. The govmm part of change is already merged so we just need to change qemu to call the APIs properly.

However, because this is just part of the vm factory PR, there is no actual caller for the internal memory device hotplug API yet. That said, I did test it roughly with following hack and saw qmp hotplug commands succeed:
```
index 87f8eda..6bba236 100644
@@ -947,6 +947,14 @@ func (s *Sandbox) startVM() error {

        s.Logger().Info("VM started")

+       // HACK: test memory hotplug
+       defer func() {
+               _, err := s.hypervisor.hotplugAddDevice(&memoryDevice{0, 128}, memoryDev)
+               if err != nil {
+                       s.Logger().WithError(err).Warn("bergwolf: memory hotplug failed")
+               }
+
+       }()
```

### DIFF
--- a/vendor/github.com/docker/go-units/duration.go
+++ b/vendor/github.com/docker/go-units/duration.go
@@ -1,0 +1,35 @@
+// Package units provides helper function to parse and print size and time units
+// in human-readable format.
+package units
+
+import (
+	"fmt"
+	"time"
+)
+
+// HumanDuration returns a human-readable approximation of a duration
+// (eg. "About a minute", "4 hours ago", etc.).
+func HumanDuration(d time.Duration) string {
+	if seconds := int(d.Seconds()); seconds < 1 {
+		return "Less than a second"
+	} else if seconds == 1 {
+		return "1 second"
+	} else if seconds < 60 {
+		return fmt.Sprintf("%d seconds", seconds)
+	} else if minutes := int(d.Minutes()); minutes == 1 {
+		return "About a minute"
+	} else if minutes < 46 {
+		return fmt.Sprintf("%d minutes", minutes)
+	} else if hours := int(d.Hours() + 0.5); hours == 1 {
+		return "About an hour"
+	} else if hours < 48 {
+		return fmt.Sprintf("%d hours", hours)
+	} else if hours < 24*7*2 {
+		return fmt.Sprintf("%d days", hours/24)
+	} else if hours < 24*30*2 {
+		return fmt.Sprintf("%d weeks", hours/24/7)
+	} else if hours < 24*365*2 {
+		return fmt.Sprintf("%d months", hours/24/30)
+	}
+	return fmt.Sprintf("%d years", int(d.Hours())/24/365)
+}
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -7,6 +7,7 @@ package virtcontainers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -39,12 +40,15 @@ type CPUDevice struct {
 type QemuState struct {
 	Bridges []Bridge
 	// HotpluggedCPUs is the list of CPUs that were hot-added
-	HotpluggedVCPUs []CPUDevice
-	UUID            string
+	HotpluggedVCPUs  []CPUDevice
+	HotpluggedMemory int
+	UUID             string
 }
 
 // qemu is an Hypervisor interface implementation for the Linux qemu hypervisor.
 type qemu struct {
+	vmConfig Resources
+
 	config HypervisorConfig
 
 	qmpMonitorCh qmpChannel
@@ -169,6 +173,7 @@ func (q *qemu) init(sandbox *Sandbox) error {
 		return err
 	}
 
+	q.vmConfig = sandbox.config.VMConfig
 	q.config = sandbox.config.HypervisorConfig
 	q.sandbox = sandbox
 	q.arch = newQemuArch(q.config)
@@ -204,20 +209,27 @@ func (q *qemu) cpuTopology() govmmQemu.SMP {
 	return q.arch.cpuTopology(q.config.DefaultVCPUs, q.config.DefaultMaxVCPUs)
 }
 
-func (q *qemu) memoryTopology(sandboxConfig SandboxConfig) (govmmQemu.Memory, error) {
+func (q *qemu) hostMemMB() (uint64, error) {
 	hostMemKb, err := getHostMemorySizeKb(procMemInfo)
 	if err != nil {
-		return govmmQemu.Memory{}, fmt.Errorf("Unable to read memory info: %s", err)
+		return 0, fmt.Errorf("Unable to read memory info: %s", err)
 	}
 	if hostMemKb == 0 {
-		return govmmQemu.Memory{}, fmt.Errorf("Error host memory size 0")
+		return 0, fmt.Errorf("Error host memory size 0")
 	}
 
-	hostMemMb := uint64(float64(hostMemKb / 1024))
+	return hostMemKb / 1024, nil
+}
+
+func (q *qemu) memoryTopology() (govmmQemu.Memory, error) {
+	hostMemMb, err := q.hostMemMB()
+	if err != nil {
+		return govmmQemu.Memory{}, err
+	}
 
 	memMb := uint64(q.config.DefaultMemSz)
-	if sandboxConfig.VMConfig.Memory > 0 {
-		memMb = uint64(sandboxConfig.VMConfig.Memory)
+	if q.vmConfig.Memory > 0 {
+		memMb = uint64(q.vmConfig.Memory)
 	}
 
 	return q.arch.memoryTopology(memMb, hostMemMb), nil
@@ -271,7 +283,7 @@ func (q *qemu) createSandbox(sandboxConfig SandboxConfig) error {
 
 	smp := q.cpuTopology()
 
-	memory, err := q.memoryTopology(sandboxConfig)
+	memory, err := q.memoryTopology()
 	if err != nil {
 		return err
 	}
@@ -705,6 +717,9 @@ func (q *qemu) hotplugDevice(devInfo interface{}, devType deviceType, op operati
 		// TODO: find a way to remove dependency of deviceDrivers lib @weizhang555
 		device := devInfo.(deviceDrivers.VFIODevice)
 		return nil, q.hotplugVFIODevice(device, op)
+	case memoryDev:
+		memdev := devInfo.(*memoryDevice)
+		return nil, q.hotplugMemory(memdev, op)
 	default:
 		return nil, fmt.Errorf("cannot hotplug device: unsupported device type '%v'", devType)
 	}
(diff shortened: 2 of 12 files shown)


virtcontainers/qemu.go
```diff
@@ -204,20 +209,27 @@ func (q *qemu) cpuTopology() govmmQemu.SMP {
...
-		return govmmQemu.Memory{}, fmt.Errorf("Unable to read memory info: %s", err)
+		return 0, fmt.Errorf("Unable to read memory info: %s", err)
 	}
 	if hostMemKb == 0 {
-		return govmmQemu.Memory{}, fmt.Errorf("Error host memory size 0")
+		return 0, fmt.Errorf("Error host memory size 0")
 	}
 
-	hostMemMb := uint64(float64(hostMemKb / 1024))
+	return uint64(float64(hostMemKb / 1024)), nil
```
jodh-intel: This doesn't look right to me - should this be `float64(hostMemKb)`? Or can we just drop the `float64()`?

virtcontainers/qemu.go
```diff
@@ -837,6 +852,61 @@ func (q *qemu) hotplugRemoveCPUs(amount uint32) (uint32, error) {
 	return amount, q.sandbox.storage.storeHypervisorState(q.sandbox.id, q.state)
 }
 
+func (q *qemu) hotplugMemory(slot uint32, memMB int, op operation) error {
+	if memMB <= 0 {
+		return errors.New("cannot hotplug zero or negative MB memory")
```
jodh-intel: Could you make the `memMB` a `uint64` to avoid having to add this check? Same question for the function below.

virtcontainers/qemu.go
bergwolf: Yes, indeed. I was following the old code but you are right that `float64` is unnecessary here. I'll fix it. Thanks!

virtcontainers/qemu.go
bergwolf: Sure. But I'm keeping the one in `q.hotplugAddMemory` since the govmm requires `int` and we need a sanity check better early than later.

bergwolf:
@jodh-intel PR updated. PTAL.

sboeuf:
@jcvenegas PTAL!

virtcontainers/qemu.go
```diff
@@ -204,20 +210,27 @@ func (q *qemu) cpuTopology() govmmQemu.SMP {
 	return q.arch.cpuTopology(q.config.DefaultVCPUs, q.config.DefaultMaxVCPUs)
 }
 
-func (q *qemu) memoryTopology(sandboxConfig SandboxConfig) (govmmQemu.Memory, error) {
+func (q *qemu) hostMemMB() (uint64, error) {
 	hostMemKb, err := getHostMemorySizeKb(procMemInfo)
 	if err != nil {
-		return govmmQemu.Memory{}, fmt.Errorf("Unable to read memory info: %s", err)
+		return 0, fmt.Errorf("Unable to read memory info: %s", err)
```
jcvenegas: Unable to read *host* memory info: %s  may help for debug information.

gnawux:
lgtm

virtcontainers/qemu.go
```diff
@@ -837,6 +853,64 @@ func (q *qemu) hotplugRemoveCPUs(amount uint32) (uint32, error) {
...
+	// Don't exceed the maximum amount of memory
+	if currentMemory+int(memMB) > int(maxMem) {
+		return fmt.Errorf("Unable to hotplug %d MB memory, currently this SB has %d memory and the maximum amount is %d",
+			memMB, currentMemory, q.config.DefaultMemSz)
+	}
+
+	return q.hotplugAddMemory(slot, int(memMB))
+}
+
+func (q *qemu) hotplugAddMemory(slot uint32, amount int) error {
```
jcvenegas: the qmp function takes an ID I wonder if for clarity would be better name it `id` or to if want to keep it refer that indeed we are adding an slot.

virtcontainers/qemu.go
```diff
@@ -837,6 +853,64 @@ func (q *qemu) hotplugRemoveCPUs(amount uint32) (uint32, error) {
...
+		return fmt.Errorf("cannot hotplug %d MB memory: size too large", memMB)
+	}
+
+	maxMem, err := q.hostMemMB()
+	if err != nil {
+		return err
+	}
+
+	// calculate current memory
+	currentMemory := int(q.config.DefaultMemSz)
```
jcvenegas: This configuration is the one that comes from `kata-containers.toml` if that is the case this may change if the user changes the value. In that case current memory wont be the same memory that we used to boot.

If that is the case we have a few options here:

- When we start the VM save the memory that was used to start it.
- We can query to qmp to get the current memory in the guest.

jcvenegas:
@bergwolf I think this is the initial implementation to make VM templating work right?

Do we expect to add more changes to PR or you only want to add this by now?

virtcontainers/qemu.go
```diff
@@ -837,6 +853,64 @@ func (q *qemu) hotplugRemoveCPUs(amount uint32) (uint32, error) {
...
 
+func (q *qemu) hotplugMemory(slot uint32, memMB uint64, op operation) error {
+	// We do not support memory hot unplug.
+	if op == removeDevice {
+		return errors.New("cannot hot unplug memory device")
+	}
+
+	// rule out extremely large parameter
+	if memMB > math.MaxInt32 {
+		return fmt.Errorf("cannot hotplug %d MB memory: size too large", memMB)
```
jcvenegas: Nit: The config file for kata ask for memory in MiB we may want to follo the same metric.

virtcontainers/qemu.go
```diff
@@ -39,12 +41,15 @@ type CPUDevice struct {
 type QemuState struct {
 	Bridges []Bridge
 	// HotpluggedCPUs is the list of CPUs that were hot-added
-	HotpluggedVCPUs []CPUDevice
-	UUID            string
+	HotpluggedVCPUs  []CPUDevice
+	HotpluggedMemory int
```
sboeuf: Why aren't you using a `memoryDevice` type instead here? Also, this would force this structure to be public.

virtcontainers/qemu.go
```diff
@@ -837,6 +853,64 @@ func (q *qemu) hotplugRemoveCPUs(amount uint32) (uint32, error) {
 	return amount, q.sandbox.storage.storeHypervisorState(q.sandbox.id, q.state)
 }
 
+func (q *qemu) hotplugMemory(slot uint32, memMB uint64, op operation) error {
```
sboeuf: You could use `memoryDevice` instead of `slot uint32, memMB uint64` parameters.

virtcontainers/qemu.go
```diff
@@ -169,6 +174,7 @@ func (q *qemu) init(sandbox *Sandbox) error {
 		return err
 	}
 
+	q.vmConfig = sandbox.config.VMConfig
```
sboeuf: The memory coming through the `Resources` structure in `sandbox.config.VMConfig` is part of some legacy code that should be removed IMO. We need a unified way to provide the resources, and that's using `HypervisorConfig` that we already have through `q.config`.
It is a bit odd to have CPU and Memory resources provided through 2 different structures.

virtcontainers/qemu.go
sboeuf: I think saving the amount of memory used when the VM was started is the best option here. Or we could simply save the current amount of memory plugged to the VM (but this is because we cannot unplug that I am thinking about this).

bergwolf: @jcvenegas 
> This configuration is the one that comes from kata-containers.toml if that is the case this may change if the user changes the value. In that case current memory wont be the same memory that we used to boot.

Sorry, I don't quit follow it. Do you mean the sandbox config we used to boot can change later when someone connects to the sandbox again? It doesn't sound like a solid solution. But I think we only use `kata-containers.toml` when creating the sandbox and later access to the sandbox all fetches the sandbox config from persistent metadata with the sandbox ID. Am I missing something?

virtcontainers/qemu.go
bergwolf: The idea is to convert the slot number to an ID, either we do it in this function, or we need to do it somewhere else.

vendor/github.com/docker/go-units/duration.go
```diff
@@ -0,0 +1,35 @@
+// Package units provides helper function to parse and print size and time units
+// in human-readable format.
+package units
+
+import (
+	"fmt"
+	"time"
+)
+
```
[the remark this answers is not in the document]
bergwolf: Indirect dependency? These were added by `dep` anyway.

bergwolf:
@jcvenegas 
> Do we expect to add more changes to PR or you only want to add this by now?

This is all I need for vm factory. Feel free to add more PR based on it to support memory hotplug in other use cases. I assume that you want to work on that, right?

virtcontainers/qemu.go
jcvenegas: @bergwolf yes, well I am considering a future case were someone want to update the memory in a container. And we want to increase the memory. In that case exist the case that the config file was modified.

Another case is when we create a container inside a Pod. In that case the pod may be created with one config file, but then may be updated with a different default memory, and then create a container inside a pod. 

But agree for the current use case does not have any issue with that.

virtcontainers/qemu.go
jcvenegas: I am fine with any case. I ask because maybe  a reader can think that is it is slot number where the memory will be inserted in the VM. This maybe not true all the time, qemu  inserts it into the first empty slot.

bergwolf:
@jcvenegas @sboeuf PR updated. PTAL.

virtcontainers/qemu.go
bergwolf: I did not change this because `memoryDevice.slot` is internal knowledge. Even if we have `sandbox.HotplugMemory`, we should not expose it. @sboeuf let me know if you think otherwise.

sboeuf: Sorry, what I meant is that we could have something like `HotpluggedMemory []MemoryDevice` the same way we have `HotpluggedVCPUs []CPUDevice`.

virtcontainers/qemu.go
sboeuf: Why don't you do that directly as part of this PR (separate commit of course). Do you think that's a lot of work ?

virtcontainers/qemu.go
bergwolf: No, I don't think we need that. We only need to track how much memory has been hotplugged. Unlike CPU devices, we do not support memory hot unplug, nor need to unplug memory the same chunk as they were plugged in even if we support memory hot unplug in future.

sboeuf: Fair enough !

virtcontainers/qemu.go
bergwolf: As discussed in the arch meeting, it will break ondisk metadata compatibility so I prefer to defer it to a standalone PR at least.

sboeuf: As discussed during Arch committee meeting, this rework will be done in a separate PR to prevent from API breakage.

virtcontainers/qemu.go
250665088: @bergwolf     Sorry I don't quite understand why you put "slot" field in struct MemoryDevice rather than struct QemuState , as an input from user will not involve "slot" field. When we hotplug memory device, we use a qmp command like "object_add ...  mem1 ...", in which "1" is memory slot number. We should store "slot" in QemuState which is hypervisor.json so that we could get memory slot number when we need to hotplug memory.

Issue #472, opened 2018-07-06
When do pause a sandbox, it's monitor always get timeout errors


lifupan:
# Description of problem

When monitor a sandbox using sandbox.Monitor(), it always receive
the timeout error after pause the sandbox. Thus the monitor cannot
tell its a crash caused timeout or just a paused timeout.

Thus it's better to stop its monitor after paused the sandbox, then
the caller can re monitor it once it resume this sandbox.

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #473, opened 2018-07-06
api: To stop its monitor after a sandbox paused

lifupan:
After the sandbox is paused, it's needed to stop its monitor,
Otherwise, its monitors will receive timeout errors if it is
paused for a long time, thus its monitor will not tell it's a
crash caused timeout or just a paused timeout.

Fixes: #472

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -1203,6 +1203,14 @@ func (s *Sandbox) Pause() error {
 		return err
 	}
 
+	//After the sandbox is paused, it's needed to stop its monitor,
+	//Otherwise, its monitors will receive timeout errors if it is
+	//paused for a long time, thus its monitor will not tell it's a
+	//crash caused timeout or just a paused timeout.
+	if s.monitor != nil {
+		s.monitor.stop()
+	}
+
 	return s.pauseSetStates()
 }
 


jodh-intel:
This change looks fine. However, do we have tests to cover this scenario?

/cc @chavafg, @sboeuf.

bergwolf:
Still timeout on `docker pull alpine`. Kicked that one again.
```
Running command '/usr/bin/docker [docker pull alpine]'
[docker pull alpine]
Timeout: 600 seconds
Exit Code: 1
Stdout: Using default tag: latest
latest: Pulling from library/alpine
```

devimc:
lgtm

virtcontainers/sandbox.go
```diff
@@ -1203,6 +1203,14 @@ func (s *Sandbox) Pause() error {
 		return err
 	}
 
+	//After the sandbox is paused, it's needed to stop its monitor,
+	//Otherwise, its monitors will receive timeout errors if it is
+	//paused for a long time, thus its monitor will not tell it's a
+	//crash caused timeout or just a paused timeout.
+	if s.monitor != nil {
+		s.monitor.stop()
```
sboeuf: This looks like half the fix we need. If we stop the monitor when pausing the sandbox (literally pausing the VM), we need to re-init a watcher when resuming the VM, don't you think ?
I'd prefer to merge a PR solving the whole picture here.

lifupan: Hi, Sboeuf

I think the sandbox pause caller should responsible for re-init a monitor/watcher after it resume the sandbox.

sboeuf: I cannot think of a case where the caller would not want to re-enable the monitor after it resumes the sandbox, for this reason, it's better if we get this as part of the Kata API.
@bergwolf WDYT ?

sboeuf:
@lifupan please backport to `stable-1.1` and `stable-1.2`

PR #482, opened 2018-07-12
versions: Update CRI-O supported version

### LINKED ISSUE #481: update cri-o version
some fixes to the cri-o tests were introduced. we need to update our cri-o version to have a more stable CI.

chavafg:
Some test fixes were introduced into the 1.9 and 1.10 branches
of cri-o. These fixes will help us minimize random failures.

Fixes #481.

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -132,9 +132,9 @@ externals:
     description: |
       OCI-based Kubernetes Container Runtime Interface implementation
     url: "https://github.com/kubernetes-incubator/cri-o"
-    version: "e0dd8a3d4c9e5ebc8b25299950bd5b222e3783d3"
+    version: "6273bea4c9ed788aeb3d051ebf2d030060c05b6c"
     meta:
-      openshift: "v1.9.11"
+      openshift: "7fcfaf1ec696d14615069707fbe8b36f19cbb8d7"
 
   cri-containerd:
     description: |


versions.yaml
```diff
@@ -132,9 +132,9 @@ externals:
     description: |
       OCI-based Kubernetes Container Runtime Interface implementation
     url: "https://github.com/kubernetes-incubator/cri-o"
-    version: "e0dd8a3d4c9e5ebc8b25299950bd5b222e3783d3"
+    version: "6273bea4c9ed788aeb3d051ebf2d030060c05b6c"
     meta:
-      openshift: "v1.9.11"
+      openshift: "7fcfaf1ec696d14615069707fbe8b36f19cbb8d7"
```
sboeuf: Why have you updated the openshift version, it's not mentioned in the commit message.

chavafg: it is not the openshift version, but the cri-o version that we use for openshift.

sboeuf: Ah yeah I missed that... Thanks !

sboeuf:
@amshinde ptal

sboeuf:
@egernst what do you think about this one? Not really a bug fix, but not a feature either.
It would be a good stable candidate though.
Let @chavafg know if this one needs to be backported.

PR #495, opened 2018-07-18
kata_agent: print request details

bergwolf:
It helps tracking each request that is sent and we can match with the
one printed by kata-agent on the guest side to find out any stack
requests in the middle.


sboeuf:
@egernst since this PR is an enhancement, do you think it should be backported?
Since it won't break any compatibility, I think it'd be a good stable candidate so that stable branches does not diverge too much.

PR #501, opened 2018-07-23
virtcontainers: keep qmp connection whenever possible

bergwolf:
For each time a sandbox structure is created, we ensure s.Release()
is called. Then we can keep the qmp connection as long as Sandbox
pointer is alive.

All VC interfaces are still stateless as s.Release() is called before
each API returns.

OTOH, for VCSandbox APIs, FetchSandbox() must be paired with s.Release,
the same as before.


virtcontainers/qemu.go
```diff
@@ -628,7 +596,11 @@ func (q *qemu) togglePauseSandbox(pause bool) error {
 	return nil
 }
 
-func (q *qemu) qmpSetup() (*govmmQemu.QMP, error) {
+func (q *qemu) qmpSetup() error {
+	if q.qmpMonitorCh.qmp != nil {
+		return nil
```
sboeuf: Just a thought here, do we need this to be idempotent? Is there any reason we would call qmpSetup() with `qmp` being not nil?

bergwolf: `qmpSetup()` is now always called before any qmp talks. But `qmpShutdown()` is only called explicitly when qemu.disconnect() is called. So yes, there are lots of cases when qmpSetup() is called with qmp being not nil.

sboeuf:
@bergwolf @egernst I would not put this one into the stable branches, but I'm not sure.
WDYT?

bergwolf:
@sboeuf No, I don't think it's a stable candidate.

Issue #502, opened 2018-07-23
cleanup sandbox.containers slice


bergwolf:
Container IDs are supposed to be unique within a sandbox. Use a map to describe containers.

PR #503, opened 2018-07-23
sandbox: change container slice to a map

bergwolf:
ContainerID is supposed to be unique within a sandbox. It is better to use
a map to describe containers of a sandbox.

Fixes: #502

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -453,7 +453,7 @@ type Sandbox struct {
 
 	volumes []Volume
 
-	containers []*Container
+	containers map[string]*Container
 
 	runPath    string
 	configPath string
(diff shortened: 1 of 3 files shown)


virtcontainers/sandbox.go
```diff
@@ -870,15 +873,14 @@ func (s *Sandbox) removeContainer(containerID string) error {
 		return errNeedContainerID
 	}
 
-	for idx, c := range s.containers {
-		if containerID == c.id {
-			s.containers = append(s.containers[:idx], s.containers[idx+1:]...)
-			return nil
-		}
+	if s.containers[containerID] == nil {
```
sboeuf: the way to write this in Golang is:
```go
if _, exist := s.containers[containerID]; !exist {
        return fmt.Errorf("Could not remove the container %q from the sandbox %q containers list", containerID, s.id)
}
```

virtcontainers/sandbox.go
```diff
@@ -978,7 +980,10 @@ func (s *Sandbox) startVM() error {
 }
 
 func (s *Sandbox) addContainer(c *Container) error {
-	s.containers = append(s.containers, c)
+	if s.containers[c.id] != nil {
```
sboeuf: Same thing here to check a container entry.

bergwolf: Good point. Fixed both. Thanks!

Issue #504, opened 2018-07-23
kata-runtime create automatically creates vm factory if it does not exists yet


bergwolf:
As suggested by @WeiZhang555, we can automatically create vm factory during container creation.

PR #505, opened 2018-07-23
cli: create vm factory if failed to load existing one

bergwolf:
When vm factory is enabled, create vm factory automatically
during creation.

Fixes: #504


bergwolf:
Looks like travis-ci is having issues with linux-ppc64le platform. It's stuck in creating state and then timeout after a very long period. Let's merge ignoring it since all the other CIs have passed.

PR #507, opened 2018-07-23
network: Error out when host-networking is requested

### LINKED ISSUE #499: Error out when running container with net=host
We cannot support host networking with Kata as listed in the limitations doc. But on running with --net=host, we actually end up modifying the host network namespace. We should error out early instead, by trying to identify that host networking was requested.

amshinde:
Instead of continuing with the network setup, we should detect
if host network namespace was requested and error out early.

Fixes #499

### DIFF
--- a/virtcontainers/network.go
+++ b/virtcontainers/network.go
@@ -6,6 +6,7 @@
 package virtcontainers
 
 import (
+	"bufio"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
@@ -579,6 +580,107 @@ func newNetwork(networkType NetworkModel) network {
 	}
 }
 
+const procMountInfoFile = "/proc/self/mountinfo"
+
+// getNetNsFromBindMount returns the network namespace for the bind-mounted path
+func getNetNsFromBindMount(nsPath string, procMountFile string) (string, error) {
+	netNsMountType := "nsfs"
+
+	// Resolve all symlinks in the path as the mountinfo file contains
+	// resolved paths.
+	nsPath, err := filepath.EvalSymlinks(nsPath)
+	if err != nil {
+		return "", err
+	}
+
+	f, err := os.Open(procMountFile)
+	if err != nil {
+		return "", err
+	}
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		text := scanner.Text()
+
+		// Scan the mountinfo file to search for the network namespace path
+		// This file contains mounts in the eg format:
+		// "711 26 0:3 net:[4026532009] /run/docker/netns/default rw shared:535 - nsfs nsfs rw"
+		//
+		// Reference: https://www.kernel.org/doc/Documentation/filesystems/proc.txt
+
+		// We are interested in the first 9 fields of this file,
+		// to check for the correct mount type.
+		fields := strings.Split(text, " ")
+		if len(fields) < 9 {
+			continue
+		}
+
+		// We check here if the mount type is a network namespace mount type, namely "nsfs"
+		mountTypeFieldIdx := 8
+		if fields[mountTypeFieldIdx] != netNsMountType {
+			continue
+		}
+
+		// This is the mount point/destination for the mount
+		mntDestIdx := 4
+		if fields[mntDestIdx] != nsPath {
+			continue
+		}
+
+		// This is the root/source of the mount
+		return fields[3], nil
+	}
+
+	return "", nil
+}
+
+// hostNetworkingRequested checks if the network namespace requested is the
+// same as the current process.
+func hostNetworkingRequested(configNetNs string) (bool, error) {
+	var evalNS, nsPath, currentNsPath string
+	var err error
+
+	// Net namespace provided as "/proc/pid/ns/net" or "/proc/<pid>/task/<tid>/ns/net"
+	if strings.HasPrefix(configNetNs, "/proc") && strings.HasSuffix(configNetNs, "/ns/net") {
+		if _, err := os.Stat(configNetNs); err != nil {
+			return false, err
+		}
+
+		// Here we are trying to resolve the path but it fails because
+		// namespaces links don't really exist. For this reason, the
+		// call to EvalSymlinks will fail when it will try to stat the
+		// resolved path found. As we only care about the path, we can
+		// retrieve it from the PathError structure.
+		if _, err = filepath.EvalSymlinks(configNetNs); err != nil {
+			nsPath = err.(*os.PathError).Path
+		} else {
+			return false, fmt.Errorf("Net namespace path %s is not a symlink", configNetNs)
+		}
+
+		_, evalNS = filepath.Split(nsPath)
+
+	} else {
+		// Bind-mounted path provided
+		evalNS, _ = getNetNsFromBindMount(configNetNs, procMountInfoFile)
+	}
+
+	currentNS := fmt.Sprintf("/proc/%d/task/%d/ns/net", os.Getpid(), unix.Gettid())
+	if _, err = filepath.EvalSymlinks(currentNS); err != nil {
+		currentNsPath = err.(*os.PathError).Path
+	} else {
+		return false, fmt.Errorf("Unexpected: Current network namespace path is not a symlink")
+	}
+
+	_, evalCurrentNS := filepath.Split(currentNsPath)
+
+	if evalNS == evalCurrentNS {
+		return true, nil
+	}
+
+	return false, nil
+}
+
 func initNetworkCommon(config NetworkConfig) (string, bool, error) {
 	if !config.InterworkingModel.IsValid() || config.InterworkingModel == NetXConnectDefaultModel {
 		config.InterworkingModel = DefaultNetInterworkingModel
@@ -591,6 +693,15 @@ func initNetworkCommon(config NetworkConfig) (string, bool, error) {
 		}
 
 		return path, true, nil
+	}
+
+	isHostNs, err := hostNetworkingRequested(config.NetNSPath)
+	if err != nil {
+		return "", false, err
+	}
+
+	if isHostNs {
+		return "", false, fmt.Errorf("Host networking requested, not supported by runtime")
 	}
 
 	return config.NetNSPath, false, nil
--- a/virtcontainers/network_test.go
+++ b/virtcontainers/network_test.go
@@ -7,12 +7,16 @@ package virtcontainers
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net"
 	"os"
+	"path/filepath"
 	"reflect"
+	"syscall"
 	"testing"
 
 	"github.com/containernetworking/plugins/pkg/ns"
+	"github.com/stretchr/testify/assert"
 	"github.com/vishvananda/netlink"
 	"github.com/vishvananda/netns"
 )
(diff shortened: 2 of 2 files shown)


egernst:
good idea!

virtcontainers/cnm.go
```diff
@@ -19,6 +24,10 @@ func cnmLogger() *logrus.Entry {
 
 // init initializes the network, setting a new network namespace for the CNM network.
 func (n *cnm) init(config NetworkConfig) (string, bool, error) {
+	if config.NetNSPath == hostNetworkPath {
```
sboeuf: Why do you need to introduce this hardcoded check ? You have a generic solution in `initNetworkCommon()`.

virtcontainers/network.go
```diff
@@ -593,6 +593,33 @@ func initNetworkCommon(config NetworkConfig) (string, bool, error) {
...
 	}
 
+	// Check if the network namespace requested is the same as the runtime
+	var nsPath, currentNsPath string
+
+	// Here we are trying to resolve the path but it fails because
+	// namespaces links don't really exist. For this reason, the
+	// call to EvalSymlinks will fail when it will try to stat the
+	// resolved path found. As we only care about the path, we can
+	// retrieve it from the PathError structure.
```
sboeuf: This looks good to me. I had exactly the same case some while ago, and I acknowledge this is the way to solve it.

jodh-intel:
Can you add either a unit test or a test in the tests repo to assert that we do not modify the host networking please?

virtcontainers/cnm.go
amshinde: I have got rid of the the docker specific check.

amshinde:
@jodh-intel I had created an issue in the tests repo for this:
https://github.com/kata-containers/tests/issues/528

WeiZhang555:
This is so good to me, but need some test cases .

WeiZhang555:
Oops, sorry, wrong button

virtcontainers/network.go
```diff
@@ -579,6 +580,51 @@ func newNetwork(networkType NetworkModel) network {
 	}
 }
 
+func getNetNsFromBindMount(nsPath string) (string, error) {
```
jodh-intel: You could add a unit test for this function if you made `mountInfoFile` a global `var` (that the tests could modify) or an extra param to this function.

virtcontainers/network.go
```diff
@@ -593,6 +639,45 @@ func initNetworkCommon(config NetworkConfig) (string, bool, error) {
...
+	var evalNS, nsPath, currentNsPath string
+	var err error
+
+	// Here we are trying to resolve the path but it fails because
+	// namespaces links don't really exist. For this reason, the
+	// call to EvalSymlinks will fail when it will try to stat the
+	// resolved path found. As we only care about the path, we can
+	// retrieve it from the PathError structure.
+
+	if strings.HasPrefix(config.NetNSPath, "/proc") && strings.HasSuffix(config.NetNSPath, "/ns/net") {
```
jodh-intel: This block is effectively self-contained so I think you should put it into a new function so you can say:

```
if hostNetworkingRequested(config.NetNSPath) {
    return "", false, fmt.Errorf("Host networking requested, not supported by kata-runtime")
}
```

That will of course also allow you to fully unit test the new functionality.

virtcontainers/network.go
```diff
@@ -579,6 +580,51 @@ func newNetwork(networkType NetworkModel) network {
...
+		if len(fields) < 5 {
+			continue
+		}
+
+		index := strings.Index(text, " - ")
+		if index < 0 {
+			continue
+		}
+
+		postSeparatorFields := strings.Fields(text[index+3:])
```
devimc: Q. why `index+3` ?

virtcontainers/network.go
```diff
@@ -579,6 +580,51 @@ func newNetwork(networkType NetworkModel) network {
...
+		if postSeparatorFields[0] != netNsMountType {
+			continue
+		}
+
+		if fields[4] != nsPath {
+			continue
+		}
+
+		return fields[3], nil
```
devimc: `0, 4, 3` could you please add comments? why these numbers? examples ...

virtcontainers/network.go
```diff
@@ -593,6 +639,45 @@ func initNetworkCommon(config NetworkConfig) (string, bool, error) {
...
+	currentNS := fmt.Sprintf("/proc/%d/task/%d/ns/net", os.Getpid(), unix.Gettid())
+	_, err = filepath.EvalSymlinks(currentNS)
+	if err != nil {
+		currentNsPath = err.(*os.PathError).Path
+	}
+
+	_, evalCurrentNS := filepath.Split(currentNsPath)
+
+	if evalNS == evalCurrentNS {
+		return "", false, fmt.Errorf("Host networking requested, not supported by kata-runtime")
```
devimc: `kata-runtime`? this file is part of `virtcontainers`

virtcontainers/network.go
jodh-intel: Good catch @devimc!

virtcontainers/network.go
```diff
@@ -593,6 +694,10 @@ func initNetworkCommon(config NetworkConfig) (string, bool, error) {
 		return path, true, nil
 	}
 
+	if hostNetworkingRequested(config.NetNSPath) {
```
bergwolf: Move it before `createNetNS()`? No need to check the same ns created by us. Otherwise `createNetNS()` is the broken one.

virtcontainers/network.go
```diff
@@ -579,6 +580,106 @@ func newNetwork(networkType NetworkModel) network {
...
+
+	// Net namespace provided as "/proc/pid/ns/net"
+	if strings.HasPrefix(configNetNs, "/proc") && strings.HasSuffix(configNetNs, "/ns/net") {
+		// Here we are trying to resolve the path but it fails because
+		// namespaces links don't really exist. For this reason, the
+		// call to EvalSymlinks will fail when it will try to stat the
+		// resolved path found. As we only care about the path, we can
+		// retrieve it from the PathError structure.
+
+		_, err := filepath.EvalSymlinks(configNetNs)
```
devimc: `err` was declared at line 647 and you can do something like this

```go
if _, err = filepath.EvalSymlinks(configNetNs); err != nil
```

virtcontainers/network.go
```diff
@@ -579,6 +580,106 @@ func newNetwork(networkType NetworkModel) network {
...
+			continue
+		}
+
+		// Field following the separator contains the mount type
+		// We check here if the mount type is a network namespace mount type, namely "nsfs"
+
+		postSeparatorFields := strings.Fields(text[index+3:])
+		if postSeparatorFields[0] != netNsMountType {
+			continue
+		}
```
devimc: I'd change this to

```go
    postSeparatorFields := strings.Fields(text[index:])
    if len(postSeparatorFields) > 1 && postSeparatorFields[1] != netNsMountType {
        continue
    }
```

virtcontainers/network.go
amshinde: @bergwolf I didnt quite get you. `createNetNS` creates the netns is no netns is provided and returns immediately.

jodh-intel: Hi @amshinde - I think he's saying do something like this:

```go
if config.NetNSPath != "" {
        if hostNetworkingRequested(config.NetNSPath) {
                return "", false, fmt.Errorf("Host networking requested, not supported by runtime")
        }
} else if config.NetNSPath == "" {
        path, err := createNetNS()
        if err != nil {
                return "", false, err
        }

        return path, true, nil
}
```

The reason being that the runtime itself wouldn't be creating a host ns (via `createNetNS()`) so why call `hostNetworkingRequested()` unconditionally?

amshinde: @jodh-intel The `hostNetworkingRequested` path will run only if no net namespace is provided.
I had something like this before
```
   if config.NetNSPath == "" {
            path, err := createNetNS()
            if err != nil {
                   return "", false, err
            }
           return path, true, nil
   } else {
            if hostNetworkingRequested(config.NetNSPath) {
                  return "", false, fmt.Errorf("Host networking requested, not supported by runtime")
            }
   }
```

But golint specified that I take out the else since all code paths in the if block  return.

jodh-intel: Ah - sorry. I'd missed that 2nd `return` when `NetNsPath == ""`.

virtcontainers/network.go
jodh-intel: That does look better to me.

jodh-intel:
lgtm.

grahamwhaley:
CI's are not happy - this is from the ubuntu 16.04:
```
=== RUN   TestGetNetNsFromBindMount
Tmpdir : /tmp/700119547
--- FAIL: TestGetNetNsFromBindMount (0.00s)
	assertions.go:239: 
                          
	Error Trace:	network_test.go:529
		
	Error:      	Not equal: 
		
	            	expected: "net:[4026532009]"
		
	            	actual: ""
		
	Messages:   	Test 0, expected net:[4026532009], got 
```

amshinde:
@grahamwhaley Yes, saw that. I ran those tests locally and they pass. I dont get it why they fail in our CI!

grahamwhaley:
I can only guess they have a quite different setup in the network space - dunno. Maybe you can temporarily add a commit to this PR that echo out a bunch of useful info (a shell script sounds useful). I know @jodh-intel has an Issue open for us to add running fuller system data collection tools which would then be attached or dumped when a CI fails - that would have been handy right now!

amshinde:
I have added some debug logs in the tests for now, will wait to see if it helps me find out what the issue is.

bergwolf:
LGTM

jodh-intel:
Hi @amshinde - what distro are you running on locally? I've just tested on Ubuntu 16.04 and 18.04 and both fail this test as the CI does:

```
$ (cd virtcontainers; go test -v -run TestGetNetNsFromBindMount)
=== RUN   TestGetNetNsFromBindMount
Tmpdir : /tmp/153090120
Run : 0, File contents : 711 26 0:3 net:[4026532009] /run/docker/netns/default rw shared:535 - nsfs nsfs rw
Run : 1, File contents : 711 26 0:3 net:[4026532009] /run/netns/ns123 rw shared:535 - tmpfs tmpfs rw
Run : 2, File contents : a a a a a a a - b c d
Run : 3, File contents : 
--- FAIL: TestGetNetNsFromBindMount (0.00s)
        Error Trace:    network_test.go:533
        Error:          Not equal: 
                        expected: "net:[4026532009]"
                        actual: ""
        Messages:       Test 0, expected net:[4026532009], got 
FAIL
exit status 1
FAIL    github.com/kata-containers/runtime/virtcontainers       0.007s
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.5 LTS
Release:        16.04
Codename:       xenial
$ cat /proc/version
Linux version 4.4.0-131-generic (buildd@lgw01-amd64-015) (gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.10) ) #157-Ubuntu SMP Thu Jul 12 15:51:36 UTC 2018
```

BTW - centos "passed", but it didn't really:

```
INFO: unit tests skipped for github.com/kata-containers/runtime in centos
INFO: issue https://github.com/kata-containers/runtime/issues/228
```

virtcontainers/network.go
```diff
@@ -579,6 +580,111 @@ func newNetwork(networkType NetworkModel) network {
...
+
+	// Net namespace provided as "/proc/pid/ns/net"
+	if strings.HasPrefix(configNetNs, "/proc") && strings.HasSuffix(configNetNs, "/ns/net") {
+		// Here we are trying to resolve the path but it fails because
+		// namespaces links don't really exist. For this reason, the
+		// call to EvalSymlinks will fail when it will try to stat the
+		// resolved path found. As we only care about the path, we can
+		// retrieve it from the PathError structure.
+
+		if _, err = filepath.EvalSymlinks(configNetNs); err != nil {
```
sboeuf: What if the `EvalSymlinks()` returns `err = nil` ? Should we simply return false ? Or return an error ?

virtcontainers/network.go
```diff
@@ -579,6 +580,111 @@ func newNetwork(networkType NetworkModel) network {
...
+		}
+		_, evalNS = filepath.Split(nsPath)
+
+	} else {
+		// Bind-mounted path provided
+		evalNS, _ = getNetNsFromBindMount(configNetNs, procMountInfoFile)
+	}
+
+	currentNS := fmt.Sprintf("/proc/%d/task/%d/ns/net", os.Getpid(), unix.Gettid())
+	if _, err = filepath.EvalSymlinks(currentNS); err != nil {
```
sboeuf: What if the `EvalSymlinks()` returns `err = nil` ? Should we simply return false ? Or return an error ?

virtcontainers/network.go
```diff
@@ -579,6 +580,111 @@ func newNetwork(networkType NetworkModel) network {
...
+
+		// We are interested in the first 9 fields of this file,
+		// to check for the correct mount type.
+		fields := strings.Split(text, " ")
+		if len(fields) < 9 {
+			fmt.Printf("Found fields less than 9\n")
+			continue
+		}
+
+		index := strings.Index(text, " - ")
```
sboeuf: Is it rock solid to test `" - "` ? Why don't we keep this simple as it seems to me the mount type will always be index 8 ?

virtcontainers/network.go
```diff
@@ -579,6 +580,111 @@ func newNetwork(networkType NetworkModel) network {
...
+
+		// This is the mount point/destination for the mount
+		if fields[4] != nsPath {
+			fmt.Printf("Did not match the network ns path provided : %s\n", fields[4])
+			continue
+		}
+
+		// This is the root/source of the mount
+		fmt.Printf("Returning result : %s\n", fields[3])
+		return fields[3], nil
```
sboeuf: Based on the comment showing the format expected `"711 26 0:3 net:[4026532009] /run/docker/netns/default rw shared:535 - nsfs nsfs rw"`, I feel this function could be simpler:

```go
                ...
		fields := strings.Split(text, " ")
		if len(fields) < 9 {
			fmt.Printf("Found fields less than 9\n")
			continue
		}

                mountTypeFieldIdx = 8
                if fields[mountTypeFieldIdx] != "nsfs" {
                        continue
                }

                nsIdx := 3
                nsField := strings.Split(fields[nsIdx], ":")
                if nsField[0] != "net" {
                        continue
               }

               return fields[3]
}
```

virtcontainers/network.go
WeiZhang555: I think @amshinde already did this in else clause which handles `err == nil` case.

WeiZhang555:
LGTM
Nice unit tests! let's wait Jenkins' report

grahamwhaley:
This thread is slightly confusing :-) - there are the questions around 'did the CIs really pass', and 'do we need more tests here' I think.
I'm going to mark as DNM for the moment and request some feedback from @jodh-intel and @amshinde on their thoughts. Hopefully I'm just being over cautious and we can merge this rsn...

amshinde:
@grahamwhaley This is good for merging now. I had added unit tests and now figured out why the tests were failing in CI and not locally for me.
@sboeuf @grahamwhaley @jodh-intel PTAL

virtcontainers/network.go
```diff
@@ -579,6 +580,108 @@ func newNetwork(networkType NetworkModel) network {
...
+func getNetNsFromBindMount(nsPath string, procMountFile string) (string, error) {
+	netNsMountType := "nsfs"
+
+	// Resolve all symlinks in the path as the mountinfo file contains
+	// resolved paths.
+	resolvedPath, err := filepath.EvalSymlinks(nsPath)
+	if err != nil {
+		return "", err
+	}
+	nsPath = resolvedPath
```
sboeuf: I don't see the reason why we reassign this value here. Simply use `resolvedPath` from the loop below.

virtcontainers/network.go
```diff
@@ -579,6 +580,108 @@ func newNetwork(networkType NetworkModel) network {
...
+	} else {
+		// Bind-mounted path provided
+		evalNS, _ = getNetNsFromBindMount(configNetNs, procMountInfoFile)
+	}
+
+	currentNS := fmt.Sprintf("/proc/%d/task/%d/ns/net", os.Getpid(), unix.Gettid())
+	if _, err = filepath.EvalSymlinks(currentNS); err != nil {
+		currentNsPath = err.(*os.PathError).Path
+	} else {
+		return false, fmt.Errorf("Unexpected:Current network namespace path is not a symlink")
```
sboeuf: I don't think we need "Unexpected:" prefix here, since the log is an error message.

amshinde: I just added that here as this should never happen. /proc/{pid}/task/{tid}/ns/net  are links that are created in the kernel. So we should really not land in the else case.

sboeuf: Fair enough. Could you please add a space between `Unexpected:` and `Current ...`

virtcontainers/network.go
amshinde: done

virtcontainers/network.go
amshinde: done

virtcontainers/network.go
```diff
@@ -579,6 +580,107 @@ func newNetwork(networkType NetworkModel) network {
...
 
+const procMountInfoFile = "/proc/self/mountinfo"
+
+// getNetNsFromBindMount returns the network namespace for the bind-mounted path
+func getNetNsFromBindMount(nsPath string, procMountFile string) (string, error) {
+	netNsMountType := "nsfs"
+
+	// Resolve all symlinks in the path as the mountinfo file contains
+	// resolved paths.
+	nsPath, err := filepath.EvalSymlinks(nsPath)
```
sboeuf: Does this build ?

amshinde: Yes It does. Why?

sboeuf: I thought that using `:=` on the existing `nsPath` variable would not work, but I forgot that `err` was not declared yet!

sboeuf:
LGTM, thanks @amshinde

Issue #508, opened 2018-07-24
When createNetwork failed, it should rollback the NetNs it created


lifupan:
# When createNetwork failed, it should rollback the NetNs it created

When createNetwork failed, kata will leave the redundant NetNs in the system

# Expected result

When createNetwork failed, it should rollback the NetNs it created

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #509, opened 2018-07-24
virtconainers: rollback the NetNs when createNetwork failed

lifupan:
When createNetwork failed, cleanup the NetNs if it created.

Fixes: #508

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -906,8 +906,20 @@ func (s *Sandbox) Delete() error {
 }
 
 func (s *Sandbox) createNetwork() error {
+	var netNsPath string
+	var netNsCreated bool
+	var networkNS NetworkNamespace
+	var err error
+
+	//rollback the NetNs when createNetwork failed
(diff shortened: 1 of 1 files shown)


virtcontainers/sandbox.go
```diff
@@ -906,8 +906,19 @@ func (s *Sandbox) Delete() error {
 }
 
 func (s *Sandbox) createNetwork() error {
+	var netNsPath string
+	var netNsCreated bool
+	var err error
+
+	//rollback the NetNs when createNetwork failed
+	defer func() {
+		if err != nil && netNsPath != "" && netNsCreated {
```
bergwolf: To make the check work in a defer closure, you need to make sure `err` is never redefined in this function.

Also the error from the last line `return s.storage.storeSandboxNetwork(s.id, networkNS)` should also be checked.

lifupan: Yeah, I had cleanup the err redefined lines and checked the last return code.

Thanks

Fupan

PR #517, opened 2018-07-26
agent: Increase timeout for check request.

jcvenegas:
In slow systems when run  multiples containers in parallel.  The start-up time
takes longer and when we do the initial gprc server check, we got a time-out
because is not ready.

This PR adds a better error message when this happens and increase the time-out
to 10 seconds.

Notice that this fix the CI but  we still need to find the route cause why
is taking longer and try to reduce this delay.

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -1229,8 +1229,12 @@ func (k *kataAgent) disconnect() error {
 	return nil
 }
 
+// check grpc server is serving
 func (k *kataAgent) check() error {
 	_, err := k.sendReq(&grpc.CheckRequest{})
+	if err != nil {
+		err = fmt.Errorf("Failed to check if grpc server is working: %s", err)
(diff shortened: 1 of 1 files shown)


virtcontainers/kata_agent.go
```diff
@@ -34,6 +34,7 @@ import (
 )
 
 var (
+	checkRequestTimeout   = 10 * time.Second
```
sboeuf: You can bump this to 30s IMO, this won't hurt, and will cover for extreme cases.

jodh-intel:
lgtm.

I wonder if it would be worth logging how long it actually *did* take for the check to complete though?

sboeuf:
Just merged it as this will unblock the CI.

sboeuf:
@jcvenegas please backport to stable branches.

jcvenegas:
@sboeuf sure

Issue #520, opened 2018-07-26
vm factory should use the same logger as virtcontainers


bergwolf:
We need a log setter API for vm factory to use the same logger as other part of runtime.

PR #521, opened 2018-07-26
factory: add SetLogger API

bergwolf:
So that we actually use the same logger as other packages when being
invoked by CLI.

Fixes: #520

### DIFF
--- a/virtcontainers/factory/factory.go
+++ b/virtcontainers/factory/factory.go
@@ -67,6 +67,15 @@ func NewFactory(config Config, fetchOnly bool) (vc.Factory, error) {
 	return &factory{b}, nil
 }
 
+// SetLogger sets the logger for the factory.
+func SetLogger(logger logrus.FieldLogger) {
+	fields := logrus.Fields{
+		"source": "virtcontainers",
+	}
+
+	factoryLogger = logger.WithFields(fields)
+}
+
 func (f *factory) log() *logrus.Entry {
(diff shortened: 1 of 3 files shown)


virtcontainers/factory/factory.go
```diff
@@ -67,6 +68,16 @@ func NewFactory(config Config, fetchOnly bool) (vc.Factory, error) {
 	return &factory{b}, nil
 }
 
+// SetLogger sets the logger for the factory.
+func SetLogger(logger logrus.FieldLogger) {
+	fields := logrus.Fields{
+		"source": "virtcontainers",
+		"arch":   runtime.GOARCH,
```
jodh-intel: I see this is basically doing the same as `virtcontainers/api.go:SetLogger()`. But I'm wondering why we don't just add the `arch` field here:

- https://github.com/kata-containers/runtime/blob/master/cli/main.go#L146

But also, you might want to add a `"subsystem": "factory"` field as we've done that in various parts of vc (like `network.go`).

bergwolf: Good point about moving arch to main.go.

As for `subsystem`, it is set in `factory.log()` function.

bergwolf: Updated. PTAL @jodh-intel .

jodh-intel:
lgtm

devimc:
lgtm

bergwolf:
fedora CI failed on #525. restarting.

PR #524, opened 2018-07-27
do not append sandbox is to kernel parameters

bergwolf:
CreateSandbox() is the point we know which sandbox id will use
this guest. If we rely on the kernel command line for it, we have
to put sandbox id in qemu args and it breaks vm factory.

Also fix a typo in create.go that fails to assign the factory pointer.

The vendor commit will be re-vendored once kata-containers/agent/pull/309 is merged.


jodh-intel:
lgtm

sboeuf:
`openstack/third-party-check` is still not required for now, so we can move forward with this!

PR #528, opened 2018-07-27
Ensure containerID and sandboxID fields logged as early as possible

### LINKED ISSUE #519: vc: virtLog() does not log container/pod IDs
# Description of problem

During some debug, I found entries such as:
```
(Jul 26 08:37:36 ubnt1604Master kata-runtime[7605]: time="2018-07-26T08:37:36.022838989Z" level=info msg="Virtual endpoint unmarshalled [{{c2d201b7-0c9b-4594-84f9-f6bf3721c951 br0 {eth0 02:00:ca:fe:00:00 [172.17.0.13/16 eth0]} {tap0 02:42:ac:11:00:0d []} 2 [0xc42000e210 0xc42000e220 0xc42000e228 0xc42000e230 0xc42000e238 0xc42000e240 0xc42000e248 0xc42000e250] [0xc42000e258 0xc42000e260 0xc42000e268 0xc42000e270 0xc42000e278 0xc42000e280 0xc42000e288 0xc42000e290]} {{{26 1500 0 eth0 02:42:ac:11:00:0d up|broadcast|multicast 69699 27 0 <nil>  0xc420198240 0 <nil> ether <nil> up 0 0 0} veth} [172.17.0.13/16 eth0] [{Ifindex: 26 Dst: <nil> Src: <nil> Gw: 172.17.0.1 Flags: [] Table: 254} {Ifindex: 26 Dst: 172.17.0.0/16 Src: 172.17.0.13 Gw: <nil> Flags: [] Table: 254}] {[]  [] []}} false virtual}]" arch=amd64 command=kill name=kata-runtime pid=7605 source=virtcontainers
```

in the system journal. The come from around:
https://github.com/kata-containers/runtime/blob/master/virtcontainers/network.go#L502

Those lines have no container or pod IDs in them. Can I ask if we think it is possible to add those (which I can see might involve some additions to the func calls etc.). Also, these are not using the `networkLogger` that is defined higher up in that same file - do we think maybe they can/should be?

# Expected result

Ideally it would be easy to correlate these log entries to the relevant pod/container/VM they were acting upon.

# Actual result

I can tell the association at present, as I am performing a pretty linear test. If there were parallelism involved here then I think it would be much more difficult.

Yeah, I'll understand if it's hard to get that info into the right place to put out in the logs.

jodh-intel:
Improve containerID and sandboxID logging by:

- Ensuring that `setExternalLoggers()` saves any previously-set log fields
  (this required a `SetLogger()` API change).

- Calling `setExternalLoggers()` as early as possible, then again once new fields become available.

  This is necessary to ensure that all virtcontainers functions atleast have access to the containerID for logging purposes.

PR also performs various log cleanup.

Fixes #519.
    
### DIFF
--- a/cli/delete.go
+++ b/cli/delete.go
@@ -53,11 +53,16 @@ EXAMPLE:
 }
 
 func delete(containerID string, force bool) error {
+	kataLog = kataLog.WithField("container", containerID)
+	setExternalLoggers(kataLog)
+
 	// Checks the MUST and MUST NOT from OCI runtime specification
 	status, sandboxID, err := getExistingContainerInfo(containerID)
 	if err != nil {
 		return err
 	}
+
+	containerID = status.ID
 
 	kataLog = kataLog.WithFields(logrus.Fields{
 		"container": containerID,
@@ -65,8 +70,6 @@ func delete(containerID string, force bool) error {
 	})
 
 	setExternalLoggers(kataLog)
-
-	containerID = status.ID
 
 	containerType, err := oci.GetContainerType(status.Annotations)
 	if err != nil {
--- a/cli/events.go
+++ b/cli/events.go
@@ -141,6 +141,9 @@ information is displayed once every 5 seconds.`,
 			return fmt.Errorf("container id cannot be empty")
 		}
 
+		kataLog = kataLog.WithField("container", containerID)
+		setExternalLoggers(kataLog)
+
 		duration := context.Duration("interval")
 		if duration <= 0 {
 			return fmt.Errorf("duration interval must be greater than 0")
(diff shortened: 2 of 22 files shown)


grahamwhaley:
I too wondered if the [1] might get trimmed - maybe when it merges. If it were a kernel thing, then that is exactly what you do for text you don't want to end up in the commit iirc (that or very similar ;-) )

jodh-intel:
@egernst - fixed.

jodh-intel:
Rebased!

jodh-intel:
CI's are green. Ignoring the openstack failing CI + tiny codecov change (0.01% :)...

PR #536, opened 2018-07-31
qemu: clear qmp state before wait for qemu process

bergwolf:
So that if there is any remaining state, we do not let it interfere
with the new one. This should fix the occasional vm factory hang.

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -517,6 +517,8 @@ func (q *qemu) waitSandbox(timeout int) error {
 	var ver *govmmQemu.QMPVersion
 	var err error
 
+	// clear any possible old state before trying to connect again.
+	q.qmpShutdown()
 	timeStart := time.Now()
 	for {
 		disconnectCh := make(chan struct{})
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -25,9 +25,10 @@ import (
 )
 
 type qmpChannel struct {
-	ctx  context.Context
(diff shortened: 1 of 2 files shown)


jodh-intel:
lgtm

/cc @markdryan.

bergwolf:
@markdryan At the point of `hypervisor.waitsandbox()`, I don't think anyone should connect/disconnect to the qemu process concurrently -- we wait for the qemu process to be up before actually talk to it in any means.

markdryan:
@bergwolf Can we be sure no one else has called Shutdown on this instance?  If so it should be okay.  You could also read from the disconnectedChannel which will block until the connection is truly closed.  This is what all the test cases do, e.g., https://github.com/intel/govmm/blob/master/qemu/qmp_test.go#L497

bergwolf:
@markdryan `q.qmpShutdown()` checks for nil qmp handler before calling shutdown. So if shutdown is called before, we are still safe. The only case we need to exclude is concurrent connect/disconnect.

markdryan:
@bergwolf, I'd missed that.  I thought you were calling Shutdown directly.  Reading from the disconnected channel might still be a good idea though.

bergwolf:
@markdryan Yes, you are right. It still makes sense to wait on the disconnected channel. I've added a new commit to do it. PTAL.

jodh-intel:
lgtm

jodh-intel:
Ignoring 0.01% drop in code cov...

PR #539, opened 2018-07-31
Reseed guest random number generator after booting from vm factory

bergwolf:
Because vm templating would share the initial kernel memory among guests, we need to reseed CRNG when new guests are created from template.

This is put on top of #524 because they would conflict on agent vendor. Also I labeled it wip since the agent change github.com/kata-containers/agent/pull/312 is not merge yet.


jodh-intel:
#524 and https://github.com/kata-containers/agent/pull/312 are now merged so this just needs a rebase to resolve the conflict I think.

bergwolf:
@jodh-intel Yes, indeed. I have rebased the PR. ptal.

jodh-intel:
Thanks @bergwolf.

lgtm

jodh-intel:
Ping @kata-containers/runtime .

PR #544, opened 2018-08-02
virtcontainers: fix the issue of cleanup the vm's path

### LINKED ISSUE #543: the VM specific path doesn't cleanup when the sandbox is stopped
# Description of problem

The sandbox's VM path hasn't been cleanup when it stops, such as:

$sudo ls /run/vc/vm/
alpine134  alpine135

# Expected result

The sandbox's vm path should be cleanup when the
sandbox is stopped.

# Actual result

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

lifupan:
To use the filepath.Join() instead of the simple
string append method to form the file path, otherwise
it will lose the "/" between the two parts.

Fixes #543.

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -579,7 +579,7 @@ func (q *qemu) stopSandbox() error {
 		return err
 	}
 
-	err = os.RemoveAll(RunVMStoragePath + q.id)
+	err = os.RemoveAll(filepath.Join(RunVMStoragePath, q.id))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


jodh-intel:
Nice!

lgtm

grahamwhaley:
I've kicked centos-7.4 CI to rebuild, as it looked like a network error.

jshachm:
LGTM Good catch~

jodh-intel:
Ignoring openstack CI fail. All else is green, so merging...

Issue #545, opened 2018-08-02
cli: support kata-runtime factory status


bergwolf:
As suggested by @grahamwhaley, introduce `kata-runtime factory status` subcommand to check and print the vm factory status.

PR #546, opened 2018-08-02
cli: support factory status command

bergwolf:
It checks vm factory status and prints the result.

Fixes: #545

### DIFF
--- a/cli/factory.go
+++ b/cli/factory.go
@@ -18,6 +18,7 @@ import (
 var factorySubCmds = []cli.Command{
 	initFactoryCommand,
 	destroyFactoryCommand,
+	statusFactoryCommand,
 }
 
 var factoryCLICommand = cli.Command{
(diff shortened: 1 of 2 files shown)


cli/factory.go
```diff
@@ -96,3 +97,37 @@ var destroyFactoryCommand = cli.Command{
...
+					HypervisorType:   runtimeConfig.HypervisorType,
+					HypervisorConfig: runtimeConfig.HypervisorConfig,
+					AgentType:        runtimeConfig.AgentType,
+					AgentConfig:      runtimeConfig.AgentConfig,
+				},
+			}
+			kataLog.WithField("factory", factoryConfig).Info("load vm factory")
+			f, err := vf.NewFactory(factoryConfig, true)
+			if err != nil {
+				fmt.Println("vm factory is off")
```
jodh-intel: It might be worth changing this (and the other functions in this file) to be something like:

```
fmt.Fprintln(defaultOutputFile, "vm factory is off")
```

jodh-intel:
lgtm

cli/factory.go
bergwolf: Done.

jodh-intel: Thanks!

Issue #547, opened 2018-08-03
vm directory is not removed after docker rm


bergwolf:
```
[macbeth@agent]$sudo ls /var/lib/vc/sbs/ /run/vc/sbs/ /run/vc/vm/
/run/vc/sbs/:

/run/vc/vm/:

/var/lib/vc/sbs/:
[macbeth@agent]$docker run --rm -it --runtime kata busybox echo foobar
foobar
[macbeth@agent]$sudo ls /var/lib/vc/sbs/ /run/vc/sbs/ /run/vc/vm/
[sudo] password for bergwolf:
/run/vc/sbs/:

/run/vc/vm/:
647cf0342937f3b852c5707045bed62be79fabe7853b594fcb0255d36216be30

/var/lib/vc/sbs/:
```

grahamwhaley:
Is this different from #544 that landed yesterday?

bergwolf:
Nop, it still exists on master tip.

PR #548, opened 2018-08-03
qemu: create vm directory before launching qemu

bergwolf:
Right now we create it in `createsandbox` and it would
create the vm dir unnecessarily for fetchsandbox() and
it ends up leaving an empty vm dir behind even after
DeleteSandbox.

Fixes: #547


jodh-intel:
lgtm

PR #549, opened 2018-08-03
virtcontainers: hotplug communication channel

### LINKED ISSUE #532: vsock support planning
As discussed with @sboeuf, we should add vsock support [in several steps](https://github.com/kata-containers/agent/pull/299#issuecomment-407462580):

1.  @devimc should finish his work on coldplugging the vsock (from the runtime) since this is the easiest thing to do. This means the agent would simply have to check for serial of vsock. I think this is an important first step to validate that vsock works properly as expected with the whole chain/stack.
2. As a second step, we make both serial and vsock sockets hotpluggable from the runtime codebase. This will need a corresponding PR on the agent side to detect if serial or vsock get hotplugged (could be based on uevent or simply checking some known paths).
In the meantime, we can add the new gRPC function (something like changeCommChannel()), with the appropriate agent implementation which will take care of stopping the gRPC server, close the current connection and connect to the new socket/port in order to re-spin a new gRPC server. The agent should have a timeout to check for this new socket/port for a certain amount of time since the hotplug might take some time.
3. As a third step, we would use leverage the changeCommChannel() from the runtime in case of VM templating.

And #497 (#383) serves as the implementation for the 1st step above.

devimc:
Do not cold plug communication channel (unix socket or vsock)
instead hot plug it after launching QEMU.
VM templating requires this feature.
    
fixes #532

### DIFF
--- a/virtcontainers/agent.go
+++ b/virtcontainers/agent.go
@@ -230,6 +230,9 @@ type agent interface {
 	// reseedRNG will reseed the guest random number generator
 	reseedRNG(data []byte) error
 
+	// waitForAgentReady waits until the agent is ready
+	waitForAgentReady(sandbox *Sandbox) error
+
 	// updateInterface will tell the agent to update a nic for an existed Sandbox.
 	updateInterface(inf *grpc.Interface) (*grpc.Interface, error)
 
--- a/virtcontainers/factory/factory.go
+++ b/virtcontainers/factory/factory.go
@@ -30,7 +30,8 @@ type Config struct {
 }
 
 type factory struct {
-	base base.FactoryBase
+	base             base.FactoryBase
+	hypervisorConfig vc.HypervisorConfig
 }
 
 func trace(parent context.Context, name string) (opentracing.Span, context.Context) {
@@ -73,7 +74,7 @@ func NewFactory(ctx context.Context, config Config, fetchOnly bool) (vc.Factory,
 		b = cache.New(ctx, config.Cache, b)
 	}
 
-	return &factory{b}, nil
+	return &factory{base: b}, nil
 }
 
 // SetLogger sets the logger for the factory.
@@ -144,7 +145,7 @@ func (f *factory) GetVM(ctx context.Context, config vc.VMConfig) (*vc.VM, error)
 	span, _ := trace(ctx, "GetVM")
 	defer span.Finish()
 
-	hypervisorConfig := config.HypervisorConfig
+	f.hypervisorConfig = config.HypervisorConfig
 	err := f.validateNewVMConfig(config)
 	if err != nil {
 		f.log().WithError(err).Error("invalid hypervisor config")
@@ -172,46 +173,48 @@ func (f *factory) GetVM(ctx context.Context, config vc.VMConfig) (*vc.VM, error)
 		}
 	}()
 
-	err = vm.Resume()
-	if err != nil {
-		return nil, err
-	}
-
-	// reseed RNG so that shared memory VMs do not generate same random numbers.
-	err = vm.ReseedRNG()
-	if err != nil {
-		return nil, err
-	}
-
-	online := false
-	baseConfig := f.base.Config().HypervisorConfig
-	if baseConfig.NumVCPUs < hypervisorConfig.NumVCPUs {
-		err = vm.AddCPUs(hypervisorConfig.NumVCPUs - baseConfig.NumVCPUs)
-		if err != nil {
-			return nil, err
-		}
-		online = true
-	}
-
-	if baseConfig.MemorySize < hypervisorConfig.MemorySize {
-		err = vm.AddMemory(hypervisorConfig.MemorySize - baseConfig.MemorySize)
-		if err != nil {
-			return nil, err
-		}
-		online = true
-	}
-
-	if online {
-		err = vm.OnlineCPUMemory()
-		if err != nil {
-			return nil, err
-		}
-	}
-
-	return vm, nil
+	return vm, vm.Resume()
 }
 
 // CloseFactory closes the factory.
 func (f *factory) CloseFactory(ctx context.Context) {
 	f.base.CloseFactory(ctx)
+}
+
+// InitializeVM initializes the VM, reseed random, adds CPUs and memory.
+func (f *factory) InitializeVM(ctx context.Context, vm *vc.VM, sandbox *vc.Sandbox) error {
+	var err error
+
+	span, _ := trace(ctx, "InitializeVM")
+	defer span.Finish()
+
+	// reseed RNG so that shared memory VMs do not generate same random numbers.
+	if err = vm.ReseedRNG(); err != nil {
+		return err
+	}
+
+	online := false
+	baseConfig := f.base.Config().HypervisorConfig
+	if baseConfig.NumVCPUs < f.hypervisorConfig.NumVCPUs {
+		err = vm.AddCPUs(f.hypervisorConfig.NumVCPUs - baseConfig.NumVCPUs)
+		if err != nil {
+			return err
+		}
+		online = true
+	}
+
+	if baseConfig.MemorySize < f.hypervisorConfig.MemorySize {
+		err = vm.AddMemory(f.hypervisorConfig.MemorySize - baseConfig.MemorySize)
+		if err != nil {
+			return err
+		}
+		online = true
+	}
+
+	if online {
+		if err = vm.OnlineCPUMemory(); err != nil {
+			return err
+		}
+	}
+	return nil
 }
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -904,6 +904,9 @@ func (q *qemu) hotplugDevice(devInfo interface{}, devType deviceType, op operati
 	case netDev:
 		device := devInfo.(*VirtualEndpoint)
 		return nil, q.hotplugNetDevice(device, op)
+	case serialPortDev:
+		socket := devInfo.(Socket)
+		return nil, q.hotplugSocket(socket, op)
 	default:
 		return nil, fmt.Errorf("cannot hotplug device: unsupported device type '%v'", devType)
 	}
@@ -1244,4 +1247,27 @@ func genericMemoryTopology(memoryMb, hostMemoryMb uint64) govmmQemu.Memory {
 	}
 
 	return memory
+}
+
+func (q *qemu) hotplugSocket(socket Socket, op operation) error {
+	if err := q.qmpSetup(); err != nil {
+		return err
+	}
+
+	switch op {
+	case addDevice:
+		if err := q.qmpMonitorCh.qmp.ExecuteCharDevUnixSocketAdd(q.qmpMonitorCh.ctx, socket.ID, socket.HostPath, false, true); err != nil {
+			return err
+		}
+
+		if err := q.qmpMonitorCh.qmp.ExecuteVirtSerialPortAdd(q.qmpMonitorCh.ctx, socket.DeviceID, socket.Name, socket.ID); err != nil {
+			return err
+		}
+	case removeDevice:
+		return errors.New("Remove socket device is not supported")
+	default:
+		return fmt.Errorf("Unknown socket operation: %d", op)
+	}
+
+	return nil
 }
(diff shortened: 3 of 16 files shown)


devimc:
ready for review

devimc:
ping @bergwolf @sboeuf

devimc:
depends-on https://github.com/intel/govmm/pull/31

virtcontainers/qemu_arch_base.go
```diff
@@ -24,6 +24,9 @@ type qemuArch interface {
 	// disableNestingChecks nesting checks will be ignored
 	disableNestingChecks()
 
+	// nestingChecks return true if nesting checks are enabled, otherwise false
+	nestingChecks() bool
```
sboeuf: I would call this function `isNested()` or `isRunningNested()`

virtcontainers/qemu.go
```diff
@@ -1095,3 +1098,25 @@ func genericMemoryTopology(memoryMb, hostMemoryMb uint64) govmmQemu.Memory {
...
+		if err != nil {
+			return err
+		}
+
+		err = q.qmpMonitorCh.qmp.ExecuteVirtSerialPortAdd(q.qmpMonitorCh.ctx, socket.DeviceID, socket.Name, socket.ID)
+		if err != nil {
+			return err
+		}
+	} else {
+		return errors.New("Remove socket device is not supported")
```
sboeuf: You're making the assumption that `operation` has only two types here. If we get a third possible `operation`, we might want to throw a more generic error here with the stringified name of the operation that's not supported.
Also, we could decide that we won't get more than 2 operations, and in this case, we should use a boolean.

virtcontainers/qemu.go
```diff
@@ -1095,3 +1098,25 @@ func genericMemoryTopology(memoryMb, hostMemoryMb uint64) govmmQemu.Memory {
...
 	return memory
 }
+
+func (q *qemu) hotplugSocket(socket Socket, op operation) error {
+	if err := q.qmpSetup(); err != nil {
+		return err
+	}
+
+	if op == addDevice {
+		err := q.qmpMonitorCh.qmp.ExecuteCharDevUnixSocketAdd(q.qmpMonitorCh.ctx, socket.ID, socket.HostPath, false, true)
```
sboeuf: Please use the format:
```go
if err := ...; err != nil {
```
whenever possible.

virtcontainers/qemu.go
```diff
@@ -1095,3 +1098,25 @@ func genericMemoryTopology(memoryMb, hostMemoryMb uint64) govmmQemu.Memory {
...
+		if err != nil {
+			return err
+		}
+
+		err = q.qmpMonitorCh.qmp.ExecuteVirtSerialPortAdd(q.qmpMonitorCh.ctx, socket.DeviceID, socket.Name, socket.ID)
```
sboeuf: Same here.

virtcontainers/qemu.go
```diff
@@ -1120,3 +1123,29 @@ func (q *qemu) hotplugSocket(socket Socket, op operation) error {
...
 }
+
+func (q *qemu) hotplugVSockPCI(vsock kataVSOCK, op operation) error {
+	if err := q.qmpSetup(); err != nil {
+		return err
+	}
+
+	if op == addDevice {
+		vhostfdname := fmt.Sprintf("vhostfd-%d", vsock.contextID)
+		err := q.qmpMonitorCh.qmp.ExecuteGetFD(q.qmpMonitorCh.ctx, vhostfdname, vsock.vhostFd)
```
sboeuf: Same comments than serial hotplug: `operation` and `if err :=....`

sboeuf:
https://github.com/intel/govmm/pull/31 got merged, you can rebase ;)

virtcontainers/qemu_arch_base.go
devimc: `nestingChecks` returns true if nesting checks are enabled or not, `isNested()` or `isRunningNested()` sounds more like a fact that we are running in a nested environment or not.

devimc:
@sboeuf changes applied, thanks

virtcontainers/qemu_arch_base.go
sboeuf: Ok I can see that, but this is very confusing that you're using such a function to check determine `DisableModern`. The thing is, we want to disable modern when we run inside a VM, that is when the flag `nestedRun` is true.
Maybe something is wrong with the whole naming of `nestedRun` initially, but I don't expect that we decide `DisableModern` based on the fact that we check or not the nested environment. WDYT ?

devimc: agree we should rename those functions

devimc:
depends-on https://github.com/kata-containers/agent/pull/299

virtcontainers/agent.go
```diff
@@ -221,4 +221,7 @@ type agent interface {
 
 	// reseedRNG will reseed the guest random number generator
 	reseedRNG(data []byte) error
+
+	// hotplugCommunicationChannel hot plugs the communication channel in the hypervisor
+	hotplugCommunicationChannel(h hypervisor) error
```
sboeuf: I was thinking we could have a more generic function here, something like `waitForAgentReady()` or `completeAgentInitialization()`.
This would allow several different actions to be gathered and handled by this same function, and from a higher perspective (`sandbox.go`), I think the only thing we want to make sure is that both VM and agent are ready before we can proceed further.
```go
waitForAgentReady(s *Sandbox) error
```

virtcontainers/sandbox.go
```diff
@@ -1013,6 +1013,8 @@ func (s *Sandbox) startVM() error {
 
 	s.Logger().Info("VM started")
 
+	s.agent.hotplugCommunicationChannel(s.hypervisor)
+
```
sboeuf: Please add one log here:
```go
s.Logger().Info("Agent initialization completed")
```

virtcontainers/sandbox.go
```diff
@@ -1013,6 +1013,10 @@ func (s *Sandbox) startVM() error {
 
 	s.Logger().Info("VM started")
 
+	s.agent.hotplugCommunicationChannel(s.hypervisor)
```
bergwolf: I'm wondering how it works with the vm factory code above. By the time we hotplug the communication channel here, it is too late for vm factory because vm factory requires the communication channel for several things: reseed the random number generator, hotplug necessary memory/cpu and online them.

Maybe we should at least check if the requested communication channel is already hotplugged in `hotplugCommunicationChannel` so that it can be called both at the vm factory and here.

virtcontainers/agent.go
amshinde: Agree, lets make this api more generic. `waitForAgentReady()` sounds good to me. 
Agents that dont need to do additional steps can simply return true.

devimc:
@bergwolf @sboeuf changes applied, thanks

virtcontainers/sandbox.go
```diff
@@ -1104,6 +1104,10 @@ func (s *Sandbox) startVM() error {
 
 	s.Logger().Info("VM started")
 
+	s.agent.waitForAgentReady(s)
```
bergwolf: Copying my previous comment (which I don't think is addressed in the new commits):

I'm wondering how it works with the vm factory code above. By the time we hotplug the communication channel here, it is too late for vm factory because vm factory requires the communication channel for several things: reseed the random number generator, hotplug necessary memory/cpu and online them.

Maybe we should at least check if the requested communication channel is already hotplugged in `hotplugCommunicationChannel` so that it can be called both at the vm factory and here.

devimc:
@sboeuf  @bergwolf changes applied, thanks

devimc:
@bergwolf  @sboeuf @egernst  @jodh-intel @kata-containers/runtime please take a look

devimc:
recheck

devimc:
merge me

caoruidong:
> cli/create.go:94::warning: cyclomatic complexity 16 of function create() is high (> 15) (gocyclo)

Could you please fix this?

devimc:
@caoruidong done, thanks

virtcontainers/vm.go
```diff
@@ -71,13 +71,8 @@ func NewVM(ctx context.Context, config VMConfig) (*VM, error) {
...
 
 	agent := newAgent(config.AgentType)
-	agentConfig := newAgentConfig(config.AgentType, config.AgentConfig)
-	// do not keep connection for temp agent
-	if c, ok := agentConfig.(KataAgentConfig); ok {
-		c.LongLiveConn = false
-	}
 	vmSharePath := buildVMSharePath(id)
-	err = agent.configure(hypervisor, id, vmSharePath, true, agentConfig)
+	err = agent.configure(hypervisor, id, vmSharePath, true, nil)
```
jcvenegas: why the configuration is nil here?

jcvenegas: nvm, just check agent code, It can not  be removed because agent interface.

jcvenegas:
lgtm

jcvenegas:
@bergwolf I'ld like you could provide last feedback related with vm templating.

virtcontainers/factory/factory.go
```diff
@@ -34,7 +34,8 @@ func (f *Config) validate() error {
 }
 
 type factory struct {
-	base base.FactoryBase
+	base             base.FactoryBase
+	hypervisorConfig vc.HypervisorConfig
```
bergwolf: @devimc Please do not assume that there is only one user of a vm factory at one time. A VM's config is a property of the vm, not the factory. It is the caller's responsibility to call `InitializeVM` with a `hypervisorconfig` matching with the VM to be initialized. Or, you can add a `VMConfig` filed to struct VM in vm.go to record the config of a created vm.

virtcontainers/vm.go
```diff
@@ -244,5 +233,8 @@ func (v *VM) assignSandbox(s *Sandbox) error {
 
 	s.hypervisor = v.hypervisor
 
+	// Use sanbox's agent since it points to the right communication channel.
+	v.agent = s.agent
```
bergwolf: After `assignSandbox`, we should **never** use the VM pointer again. Why do you need the vm agent pointing to the right communication channel?

virtcontainers/sandbox.go
```diff
@@ -1105,45 +1105,70 @@ func (s *Sandbox) ListRoutes() ([]*grpc.Route, error) {
...
 				AgentType:        s.config.AgentType,
 				AgentConfig:      s.config.AgentConfig,
-			})
-			if err != nil {
+			}); err != nil {
 				return err
 			}
-			err = vm.assignSandbox(s)
-			if err != nil {
+			if err = vm.assignSandbox(s); err != nil {
```
bergwolf: `assignSandbox` should be called **after** `InitializeVM`. The purpose is to assign a fully functional guest to a sandbox.

virtcontainers/vm.go
```diff
@@ -99,12 +94,6 @@ func NewVM(ctx context.Context, config VMConfig) (*VM, error) {
 		if err != nil {
 			return nil, err
 		}
-
-		virtLog.WithField("vm id", id).Info("check agent status")
-		err = agent.check()
```
bergwolf: Sorry, but you cannot simply remove this. The template factory needs the check to make sure agent is fully up and running before taking a snapshot of the vm. It ensures that we share the initial states of initramfs, guest kernel and kata-agent among all the guests created by the template factory.

Let's take a step back and look at the whole picture. -- You need to hotplug all comm channels, but vm factory requires a working agent. So could you add comm channel hotplug to `NewVM()` instead? Then it still returns a fully working guest and you do not need to split factory's `GetVM` API.

devimc:
@bergwolf thanks for reviewing this, but I think @egernst and @mcastelino have some concerns with this change.

devimc:
@egernst and @mcastelino can you comment in this pr?

sboeuf:
@devimc looks fine to me.
Please address remaining comments from @bergwolf, and rebase the PR.

virtcontainers/vm.go
devimc: the communication channel is used to initialize the VM (add cpus, memory, reseed RNG, etc), we are going to use the VM pointer again when the VM is initialized

PR #550, opened 2018-08-06
devices: add interface "sandbox.AddDevice"

### LINKED ISSUE #50: virtcontainers: Implement Storage hotplug API

WeiZhang555:
Fixes #50

Add new interface sandbox.AddDevice, then for Frakti use case, a device
can be attached to sandbox before container is created.

### DIFF
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -11,6 +11,7 @@ import (
 	"syscall"
 
 	deviceApi "github.com/kata-containers/runtime/virtcontainers/device/api"
+	deviceConfig "github.com/kata-containers/runtime/virtcontainers/device/config"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/sirupsen/logrus"
 )
@@ -759,4 +760,25 @@ func PauseContainer(sandboxID, containerID string) error {
 // ResumeContainer is the virtcontainers container resume entry point.
 func ResumeContainer(sandboxID, containerID string) error {
 	return togglePauseContainer(sandboxID, containerID, false)
+}
+
+// AddDevice will add a device to sandbox
+func AddDevice(sandboxID string, info deviceConfig.DeviceInfo) (deviceApi.Device, error) {
+	if sandboxID == "" {
+		return nil, errNeedSandboxID
+	}
+
+	lockFile, err := rwLockSandbox(sandboxID)
+	if err != nil {
+		return nil, err
+	}
+	defer unlockSandbox(lockFile)
+
+	s, err := fetchSandbox(sandboxID)
+	if err != nil {
+		return nil, err
+	}
+	defer s.releaseStatelessSandbox()
+
+	return s.AddDevice(info)
 }
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -1133,12 +1133,21 @@ func (c *Container) attachDevices() error {
 
 func (c *Container) detachDevices() error {
 	for _, dev := range c.devices {
-		if err := c.sandbox.devManager.DetachDevice(dev.ID, c.sandbox); err != nil {
-			if err == manager.ErrDeviceNotAttached {
-				// skip if device isn't attached
-				continue
-			}
+		err := c.sandbox.devManager.DetachDevice(dev.ID, c.sandbox)
+		if err != nil && err != manager.ErrDeviceNotAttached {
 			return err
+		}
+
+		if err = c.sandbox.devManager.RemoveDevice(dev.ID); err != nil {
+			c.Logger().WithFields(logrus.Fields{
+				"container": c.id,
+				"device-id": dev.ID,
+			}).WithError(err).Error("remove device failed")
+
+			// ignore the device not exist error
+			if err != manager.ErrDeviceNotExist {
+				return err
+			}
 		}
 	}
 
--- a/virtcontainers/device/manager/manager.go
+++ b/virtcontainers/device/manager/manager.go
@@ -93,7 +93,7 @@ func (dm *deviceManager) createDevice(devInfo config.DeviceInfo) (api.Device, er
 	}
 }
 
-// NewDevice creates bundles of devices based on array of DeviceInfo
+// NewDevice creates a device based on specified DeviceInfo
 func (dm *deviceManager) NewDevice(devInfo config.DeviceInfo) (api.Device, error) {
 	dm.Lock()
 	defer dm.Unlock()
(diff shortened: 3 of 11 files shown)


WeiZhang555:
@bergwolf This is final interface for hotpluging storage for Frakti, the store in my mind should be:

1. `VC.CreateSandbox` to create a sandbox object.
_2. `VCSandbox.AddDevice()` to hotplug block device before container creation._
3. Create `ContainerConfig`, modify its `Mounts []Mount`,  fill `BlockDeviceID string` field for block based storage with deviceID got from step 2.
4. `VC.CreateContainer()` to create a container with `ContainerConfig` got from step 3.

The function name `AddDevice()` in this PR is different from in  design spec(`AddStorage` from https://github.com/kata-containers/documentation/blob/master/design/kata-api-design.md#sandbox-hotplug-api ) , we can find a better name if you think it's not good.

Tell me what do you think of this @bergwolf

virtcontainers/container.go
```diff
@@ -1142,6 +1142,11 @@ func (c *Container) detachDevices() error {
 		}
 	}
 
+	for _, dev := range c.devices {
+		// ignore the device not exist error
+		c.sandbox.devManager.RemoveDevice(dev.ID)
```
devimc: ignore it, but at least log the error please

amshinde: Yes, please log the error.

amshinde:
Have just one comment, 
lgtm otherwise

virtcontainers/api.go
```diff
@@ -11,6 +11,7 @@ import (
 	"syscall"
 
 	deviceApi "github.com/kata-containers/runtime/virtcontainers/device/api"
+	"github.com/kata-containers/runtime/virtcontainers/device/config"
```
sboeuf: You could follow the same thing you did with the above import:
```go
import (
	deviceConfig "github.com/kata-containers/runtime/virtcontainers/device/config"
)
```

virtcontainers/container.go
sboeuf: @WeiZhang555 I think we should identify one type of error that we can ignore on purpose (and log something), so that we don't ignore other errors.

Also, is there any reason why we cannot call this `RemoveDevice()` function right after we detached the device with `DetachDevice()` ? This would prevent from running twice through the same list of devices.

virtcontainers/sandbox.go
```diff
@@ -1546,3 +1546,26 @@ func (s *Sandbox) AppendDevice(device api.Device) error {
...
+	if s.devManager == nil {
+		return nil, fmt.Errorf("device manager isn't initialized")
+	}
+
+	b, err := s.devManager.NewDevice(info)
+	if err != nil {
+		return nil, err
+	}
+
+	err = s.devManager.AttachDevice(b.DeviceID(), s)
```
sboeuf:
```go
if err := s.devManager.AttachDevice(b.DeviceID(), s); err != nil {
```

virtcontainers/sandbox.go
```diff
@@ -1546,3 +1546,26 @@ func (s *Sandbox) AppendDevice(device api.Device) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	if err = s.storeSandboxDevices(); err != nil {
```
sboeuf:
```go
if err := s.storeSandboxDevices(); err != nil {
```

virtcontainers/container.go
WeiZhang555: Will log this.

@sboeuf We can, let me modify it a bit.

virtcontainers/api.go
WeiZhang555: OK

virtcontainers/sandbox.go
WeiZhang555: Oh, you're right.

I once used a global `err` for doing the rolling back, then I realized we may not need a roll back for this. 
I will change it, thanks!

virtcontainers/container.go
WeiZhang555: And `ErrDeviceNotExist` is the only error RemoveDevice can return, but I will add a judgement for future extension.

WeiZhang555:
Comments are addressed, please take another look at this , thank you all! @sboeuf @amshinde @devimc

jodh-intel:
lgtm

grahamwhaley:
will wait for a re-ack from either @sboeuf or @amshinde on the changes.
I'm going to presume we can ignore the -0.25% codecov drop (I'd call that 'in the noise')...

jshachm:
LGTM

virtcontainers/container.go
```diff
@@ -1142,6 +1142,21 @@ func (c *Container) detachDevices() error {
 		}
 	}
 
+	for _, dev := range c.devices {
+		if err := c.sandbox.devManager.RemoveDevice(dev.ID); err != nil {
```
sboeuf: Thanks for addressing the first comment here, but you didn't answer/addressed the second one:
> is there any reason why we cannot call this `RemoveDevice()` function right after we detached the device with `DetachDevice()` ? This would prevent from running twice through the same list of devices.

virtcontainers/device/manager/manager.go
```diff
@@ -104,6 +104,17 @@ func (dm *deviceManager) NewDevice(devInfo config.DeviceInfo) (api.Device, error
 	return dev, err
 }
 
+// RemoveDevice creates bundles of devices based on array of DeviceInfo
```
amshinde: @WeiZhang555 This comment needs to be changed.

amshinde:
@WeiZhang555 I have one minor comment. Please address @sboeuf 's comment as well.

virtcontainers/device/manager/manager.go
WeiZhang555: Done

amshinde:
@sboeuf ptal

virtcontainers/container.go
WeiZhang555: Nice suggestion, I think this could work.
Will modify

WeiZhang555:
@sboeuf Your comments are addressed, PTAL

sboeuf:
Thanks @WeiZhang555 !

WeiZhang555:
Weird, why this piece of modification could break the CI?  https://github.com/kata-containers/runtime/pull/550#discussion_r209831742 , that makes no sense...

sboeuf:
@WeiZhang555 I think the CI was not stable for some reasons... Could you rebase and repush? This should pass without issues.

Issue #563, opened 2018-08-09
cli: Make message of using initrd OR rootfs clearer


nitkon:
# Description of problem

When starting a kata container, if initrd and rootfs both are provided in the configuration.toml then the error message presented is "cannot specify an image and an initrd in configuration file" which might be a bit confusing.

# Expected result

A suggestion is to have a clearer message like: "specify an image or an initrd in the configuration file, not both"

jodh-intel:
Hi @nitkon - again, please put details in the issue and remove the (untouched) template text :)

nitkon:
Hi @jodh-intel , done!

jodh-intel:
Thanks!

PR #564, opened 2018-08-09
cli: Make message of using initrd OR rootfs clearer

nitkon:
Fixes:  #563

### DIFF
--- a/cli/config.go
+++ b/cli/config.go
@@ -322,7 +322,7 @@ func newQemuHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 
 	if image != "" && initrd != "" {
 		return vc.HypervisorConfig{},
-			errors.New("cannot specify an image and an initrd in configuration file")
+			errors.New("having both an image and an initrd defined in the configuration file is not supported")
 	}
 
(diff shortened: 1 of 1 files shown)


nitkon:
I feel that this commit makes the error message more explicit. I may be wrong. Feel free to NACK it if its not worth it. :-)

grahamwhaley:
I can see the original message is a little confusing. Your new message is fine, but does not read too much like an `Error` to me though.

Combining the two a bit, how about:
`having both an image and an initrd defined in the configuration file is not supported`

/cc @jodh-intel for wordy input...

sboeuf:
I am fine with this either way, but I would prefer @jodh-intel's opinion on how to turn this sentence the best way (as he's a native English speaker).

jodh-intel:
+1 to @grahamwhaley's wording.

@nitkon - please add detail to the commit message though - a "bare fixes" is a bit too terse for my liking ;)

nitkon:
@jodh-intel : Noted your comments. Updated my PR.

jodh-intel:
Thanks @nitkon.

lgtm

PR #565, opened 2018-08-09
tracing: Add initial opentracing support

### LINKED ISSUE #557: Add initial opentracing support
Add the basic framework to enable opentracing support using [jaeger](https://godoc.org/github.com/uber/jaeger-client-go).

Related: https://github.com/kata-containers/kata-containers/issues/27.

jodh-intel:
Add initial support for opentracing by using the `jaeger` package.
Since opentracing uses the `context` package, add a `context.Context`
as the first parameter to all the functions that we might want to
trace. Trace "spans" (trace points) are then added by extracting the
trace details from the specified context parameter.

Notes:

- Although the tracer is created in `main()`, the "root span"
  (aka the first trace point) is not added until `beforeSubcommands()`.

  This is by design and is a compromise: by delaying the creation of the
  root span, the spans become much more readable since using the web-based
  JaegerUI, you will see traces like this:

  ```
    kata-runtime: kata-runtime create
    ------------  -------------------
        ^                ^
        |                |
    Trace name        First span name
                     (which clearly shows the CLI command that was run)
  ```

  Creating the span earlier means it is necessary to expand 'n' spans in
  the UI before you get to see the name of the CLI command that was run.
  In adding support, this became very tedious, hence my design decision to
  defer the creation of the root span until after signal handling has been
  setup and after CLI options have been parsed, but still very early in
  the code path.

- At this stage, the tracing stops at the `virtcontainers` call boundary.

- Tracing is "always on" as there doesn't appear to be a way to toggle
it. However, its resolves to a "nop" unless the tracer can talk to a
jaeger agent.

Fixes #557.


jodh-intel:
To test this PR with Jaeger:

- Configure docker to use the code in this PR.
- Run the docker command shown on https://www.jaegertracing.io/docs/getting-started/ to start all the Jaeger bits.
- Run something like:
  ```
  $ docker run -ti --runtime kata-runtime busybox true
  ```
   
- Browse to http://localhost:16686.
- Click the "Service" pull down and select `kata-runtime`.
- Click "Find Traces".
- Click on a trace on the right hand pane!

jodh-intel:
Note: I'll add further tracing granularity (for virtcontainers) on #566.

jodh-intel:
```
cli/tracing.go:45:25:warning: cfg.New is deprecated: use NewTracer() function  (SA1019) (staticcheck)
```

I was aware of this, but the new API doesn't appear to work... Let's take a closer look...

jodh-intel:
Right, we're now using the new Jaeger API. Jaeger helpfully provides two API's called `NewTracer()` - one works, one doesn't (atleast for me). Guess which I picked first? :)

Gopkg.lock
```diff
@@ -2,191 +2,248 @@
 
 
 [[projects]]
+  digest = "1:167b6f65a6656de568092189ae791253939f076df60231fdd64588ac703892a1"
```
sboeuf: I am thinking the dep file format change should be in its own PR, to avoid it being hidden through this PR, and to make sure ppl realize it's been updated.

cli/create.go
```diff
@@ -17,6 +18,7 @@ import (
 	vc "github.com/kata-containers/runtime/virtcontainers"
 	vf "github.com/kata-containers/runtime/virtcontainers/factory"
 	"github.com/kata-containers/runtime/virtcontainers/pkg/oci"
+	opentracing "github.com/opentracing/opentracing-go"
```
sboeuf: I don't think you need to specify a custom name for the import here. `opentracing-go` will be automatically translated into `opentracing` to be used from the code.

cli/create.go
```diff
@@ -85,12 +92,16 @@ var createCLICommand = cli.Command{
 // Use a variable to allow tests to modify its value
 var getKernelParamsFunc = getKernelParams
 
-func create(containerID, bundlePath, console, pidFilePath string, detach bool,
+func create(ctx context.Context, containerID, bundlePath, console, pidFilePath string, detach bool,
 	runtimeConfig oci.RuntimeConfig) error {
 	var err error
 
+	span, ctx := opentracing.StartSpanFromContext(ctx, "create")
```
sboeuf: I'm curious how much overhead (time wise) this tracing will add ? Given the fact that we have a jeager agent running (as you said this would resolve into no-op otherwise).

cli/create.go
jodh-intel: I didn't specify the name - it was added automatically by the go tooling.

Gopkg.lock
jodh-intel: Agreed. I'll raise a PR for that later today...

jodh-intel: Right - #574 raised. I'll add dnm to this PR so when #574 is merged I can remove my first commit from this PR...

jodh-intel:
Branch updated - I've now found the magic knob to toggle tracing so now you need to set `enable_tracing=true` in `configuration.toml` if you want to trace the runtime.

BTW, I plan to raise a PR on the dev guide explaining how to enable and view traces. I'll then raise a runtime PR to add a Tracing section to the `README`, referencing the new dev guide URL for tracing info.

cli/create.go
jodh-intel: Good question. I've written a test program to create spans using a "real" tracer and a "NOP" tracer (as this branch now uses by default). Creating `100,000` spans on a NUC:

| | nanonseconds / span |
|----|---|
| **Real tracer** | `2860` |
| **NOP tracer** | `550` |

So, there is some overhead even with tracing "disabled" but when you consider that a `docker run -ti busybox true` with this branch currently only creates `17` spans, I'd say the overhead is unlikely to be an issue, particularly when comparing that with the value tracing brings.

sboeuf:
#574 just got merged, @jodh-intel you can rebase ;)

cli/create.go
sboeuf: Thanks for providing the data, this looks negligible indeed.

jodh-intel:
On it. Just found a bug-ette so will re-push and comment once that's resolved...

jodh-intel:
@sboeuf, @grahamwhaley, @egernst - branch updated ;)

jodh-intel:
Nope :( I'd rather no re-force-push to avoid restarting all the long-running CI's again.

grahamwhaley:
Confirmed - it really doesn't like you does it! It also does not seem to have a 'refresh' button on their gui :-(.
I'd leave it over the weekend and see if it feels better on Monday :-)

jodh-intel:
I was hoping this might be merged by Monday so looking for reviewers... ;)

sboeuf:
@jodh-intel I'll do another review, but let's ask @amshinde @egernst @bergwolf to review this too ;)

grahamwhaley:
pullapprove has woken from its slumber it seems @jodh-intel ...

cli/tracing.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (t traceLogger) Error(msg string) {
+	kataLog.Error(msg)
+}
+
+func (t traceLogger) Infof(msg string, args ...interface{}) {
+	kataLog.Infof(msg, args...)
+}
+
+func createTracer(name string) (opentracing.Tracer, error) {
```
bergwolf: Check if `tracing` is enabled first? Or am I misunderstanding it that `tracing` is checked somewhere else?

jodh-intel:
There is one issue that we need to resolve though - how to specify details of a remote Jaeger Collector (which reads all the traces + spans). Since this branch doesn't do anything special, the jaeger package will try to send the tracer+spans to the local system where the runtime is running.

I think this means we'll need to expand the config options for tracing to something like:

```
[runtime]

[runtime.tracing]
enabled = true
# Set to value of JAEGER_ENDPOINT if you trace collector is not local
# (See https://github.com/jaegertracing/jaeger-client-go#environment-variables)
endpoint = ""
```

However, as you can see from https://github.com/jaegertracing/jaeger-client-go#environment-variables, there are a lot of variables. I don't think we can rely on just variables as getting those set in the *runtime*'s docker environment is not easy (unless we make the runtime read [`process.env`](https://github.com/opencontainers/runtime-spec/blob/master/config.md#process) from `config.json` and set them itself or something).

I think we can probably handle this on a separate issue+PR, although note that if we do this, the `enable_tracing` option might change real soon now (TM) :smile:

cli/tracing.go
bergwolf: @jodh-intel Yeah, I think we need to decide who controls the tracing functionality -- kata config or Jaeger config? Right now, each of them has a switch and it looks like only the Jaeger one is in effect.

And do we really need to create these hierarchical spanners? If not, we can put in `app.Before` and `app.After` something like:
```
if (tracing) {
    setup tracer
    start tracer span
}
```
and
```
if (tracing) {
     stop tracer span
     cleanup tracer
}
```
So we have one tracer and one span for each command line invocation. We can pass around `span` so that we can call `SetTag` on it. And it looks much cleaner than having every one to call
```
span := opentracing.StartSpanFromContext
defer span.Finish()
```

Another question about opentracing (sorry to be dumb), do we need to plant trace points on our own? Or is it integrated with logrus somehow? Or can Jaeger insert new trace points dynamically?

jodh-intel: I agree. But the recommendation is to always pass a `context.Context`, and not to pass the spans directly as that "pollutes" the codebase too much. See:

- https://github.com/yurishkuro/opentracing-tutorial/tree/master/go/lesson02#propagate-the-in-process-context
 (note that that was written by Yuri, the tech lead for Jaeger tracing @ Uber).

jodh-intel:
To be clear about https://github.com/kata-containers/runtime/pull/565#issuecomment-412446736, I think we want to minimise the tracing options in `configuration.toml`. However, atleast when running under docker, allowing the Jaeger client library to access environment variables isn't actually that easy.

cli/tracing.go
bergwolf: @jodh-intel Thanks for the presentation. I now understand better how Jaeger works and I agree we need to create span for each function we want to trace. And Jaeger looks more like a function tracing system to me. I have two more questions though:

1. Can we use it to trace any point in the code? If so, we can use it to replace most of the logs.
2. Do users have to install Jaeger (server?) to be able to disable tracing?

jodh-intel: Hi @bergwolf.

> Can we use it to trace any point in the code?

Yes, assuming the code has access to a `context.Context`.

> Do users have to install Jaeger (server?) to be able to disable tracing?

To disable tracing, they just never uncomment `enable_tracing` in the config file. So if `enable_tracing == false`, the runtime will use the "NOP tracer". But also, if `enable_tracing == true` but no Jaeger agent is running, that will not actually generate any errors either - it will just make the runtime a little slower as the jaeger-client-go code tries (unsucessfully but quietly :) to talk to the Jaeger agent over UDP.

jodh-intel: ... And yes, I'm starting to think that once we have the tracing support working well, we can remove some of the logging. It would be interesting to know how [`zap`](https://github.com/uber-go/zap) performs compared to jaeger since if jaeger is faster (and I'm guessing that it is), we don't need tracing any more in theory. Some users might be concerned about the UDP comms, but that's what syslog uses I believe ;)

jodh-intel:
Hi @bergwolf - agreed and agreed :)

grahamwhaley:
This is **all green**. I don't have time to review or test for any noticeable impact right now, but... I'll merge on the acks of others (is that 'standing on the ACKS of giants' I wonder ;-) ).
I liked @bergwolf suggestion that maybe we can carry the context around in one of the other existing data structures :-)

PR #572, opened 2018-08-10
Implement containerd shim v2 API for Kata Containers

### LINKED ISSUE #485: Implement containerd shim v2 API for Kata Containers
In the arch committee meeting last week, Michael Crosby introduced the containerd shim API v2. After the meeting, read related documents, patches, and discussed with @bergwolf @lifupan and related folks. And I think we could begin to implement the kata shim for containerd v2 API now.

## Current kata & containerd working pattern

Right now, containerd and runtime work in the following pattern:

containerd-kata-shimv1

Whenever containerd tries to create a container, 

- It creates a containerd-shim
- The containerd-shim call runtime cli -- the cli is specified by the `runtime=kata` flag, and the cli should be compatible with runc
- The containerd-shim served at the address specified by containerd, and feedback events to containerd’s listening address

Though the containerd-shim designed to be an replaceable component

- There still has to be one containerd shim per container (or process?)

## Benefits from shim v2 API

With the proposed shim v2 API, the most significant change is

- The shim will write the serving address back to containerd through stdout, instead of specified by containerd

As a result

- We may replace the containerd-shim with a customized shim, and don’t need to implement a runc compatible CLI for containerd-shim.
- We may use a single shim for a couple of containers, such as one shim per pod

Then the architecture may change to the follows:

containerd-shimv2-per-pod

- One shim for containerd per pod
- No proxies and kata-shim processes any more

And in v2 API, stats function is moved to shim, which makes the shim more self-contained.

## Where we start

- This routing pattern (built-in proxy + built-in shim) has already been implemented in hyperhq/hyperd#727 
- The OCI Spec processing part exists in kata runtime
- And the other stuff could reference the current kata-shim and v2 shim PR (containerd/containerd#2434).

lifupan:
This is the init code of containerd shim v2 for kata containers, and I had done the general 
test with the containerd's client command ctr and cri's crictl, both of them worked well. 

Fixes #485


jodh-intel:
Hi @lifupan - thanks for raising! This is clearly a very large PR. I'm not sure I'll be able to review it today but will review on Monday if not.

sboeuf:
@lifupan Thanks for submitting the implementation of Kata for containerd-shim-v2!
One thing that comes to mind, we need to introduce a bunch of integration tests for this.
You mentioned 
>I had done the general test with the containerd's client command ctr and cri's crictl, both of them worked well.

Could you please work with @chavafg and @GabyCT in order to introduce more testing to our CI, because I want to make sure we get no regression regarding your implementation once it gets merged.

lifupan:
@sboeuf   Sure, no problem.

containerd-shim/kata/utils.go
```diff
@@ -0,0 +1,168 @@
...
+	status, err := sandbox.StatusContainer(cid)
+	if err != nil {
+		logrus.WithError(err).Warnf("failed to get container %s status", cid)
+		return err
+	}
+
+	if oci.StateToOCIState(status.State) != oci.StateStopped {
+		err := vci.KillContainer(ctx, sid, cid, syscall.SIGKILL, true)
+		if err != nil {
+			logrus.WithError(err).Warnf("failed to kill container %s", cid)
```
jodh-intel: Please could you use structured logging everywhere, so prefer something like:

```
logrus.WithError(err).WithField("container", cid).Warn("failed to kill container")
```

containerd-shim/kata/wait.go
```diff
@@ -0,0 +1,64 @@
...
+		//This wait could be triggered before exec start which
+		//will get the exec's id, thus this assignment must after
+		//the exec exit, to make sure it get the exec's id.
+		processID = execs.id
+		pid = execs.pid
+	}
+
+	ret, err := s.sandbox.WaitProcess(c.id, processID)
+	if err != nil {
+		logrus.Errorf("Wait for process cid=%s,processID=%s  failed with err: %v", c.id, processID, err)
```
jodh-intel: Another good example of why we prefer structured logging:

```
logrus.WithError(err).WithFields(logrus.Fields{
    "container": c.id,
    "pid": processID,
}).Error("Wait for process failed")

```

containerd-shim/kata/create.go
```diff
@@ -0,0 +1,257 @@
...
+
+	"context"
+	vc "github.com/kata-containers/runtime/virtcontainers"
+	vf "github.com/kata-containers/runtime/virtcontainers/factory"
+	"github.com/kata-containers/runtime/virtcontainers/pkg/oci"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/sirupsen/logrus"
+)
+
+func create(ctx context.Context, s *service, containerID, bundlePath, netns string, detach bool,
```
jodh-intel: Does this context contain an opentracing span do you know (https://godoc.org/github.com/opentracing/opentracing-go)? I suspect that the answer is either "no" or "we don't know"). It would be good to be able to enable tracing for the containerd-shim as we now have for the runtime (#565, #617, #641). But since the containerd-shim is not creating the context I suspect we'll have to create our own spans in a later PR.

containerd-shim/kata/create.go
```diff
@@ -0,0 +1,257 @@
...
+		factoryConfig := vf.Config{
+			Template: true,
+			VMConfig: vc.VMConfig{
+				HypervisorType:   runtimeConfig.HypervisorType,
+				HypervisorConfig: runtimeConfig.HypervisorConfig,
+				AgentType:        runtimeConfig.AgentType,
+				AgentConfig:      runtimeConfig.AgentConfig,
+			},
+		}
+		logrus.WithField("factory", factoryConfig).Info("load vm factory")
```
jodh-intel: This seems to be using a default logger meaning all log output will go to stderr. I don't think we want to do that? Can you  create a `logger()` function that returns a `*logrus.Entry` and leverage our logging routines to get nanosecond timestamps, journald logging, etc?

lifupan:
@jodh-intel  Hmm, I will try to split those duplicated codes into a module thus both of the kata cli and shimv2 can use them. 

Thanks.

Fupan

containerd-shim/kata/create.go
```diff
@@ -0,0 +1,257 @@
...
+		if err != nil {
+			vci.SetFactory(ctx, f)
+		}
+	}
+}
+
+var systemdKernelParam = []vc.Param{
+	{
+		Key:   "init",
+		Value: "/usr/lib/systemd/systemd",
```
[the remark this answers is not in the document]
lifupan: Hi @jodh-intel , since we had thought the kata shimv2 would be independent from kata runtime and placed into containerd's shimv2 directory, thus I think there is no need to split those little common codes into a a separated library from cli.

containerd-shim/kata/service.go
```diff
@@ -0,0 +1,786 @@
...
+	"github.com/containerd/typeurl"
+	ptypes "github.com/gogo/protobuf/types"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+	"path/filepath"
+)
+
+const bufferSize = 32
```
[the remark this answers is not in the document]
lifupan: I'll add a comment line to it.

containerd-shim/kata/service.go
```diff
@@ -0,0 +1,786 @@
...
+
+var (
+	empty   = &ptypes.Empty{}
+	bufPool = sync.Pool{
+		New: func() interface{} {
+			buffer := make([]byte, 32<<10)
```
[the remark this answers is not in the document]
lifupan: Actually they  are not the same thing. I'll remove this var to stream.go where it should belong to.

containerd-shim/kata/service.go
```diff
@@ -0,0 +1,786 @@
...
+func (s *service) Pids(ctx context.Context, r *taskAPI.PidsRequest) (*taskAPI.PidsResponse, error) {
+	var id string
+	var pid uint32
+	var processes []*task.ProcessInfo
+	for pid, id = range s.processes {
+		pInfo := task.ProcessInfo{
+			Pid: pid,
+		}
+
+		if id != "" {
```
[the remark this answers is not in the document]
lifupan: I'll retune this.

Fupan

containerd-shim/kata/utils.go
lifupan: Ok, NP

containerd-shim/kata/create.go
lifupan: Hi @jodh-intel 
Since containerd-shim hasn't support opentracing,  If we put the kata shimv2 into containerd repo at last, I don't think  they can accept adding opentracing span there.

containerd-shim/kata/create.go
lifupan: Hi @jodh-intel 
No, the logger has been setup and redirected to a file in containerd-shim.

Makefile
```diff
@@ -135,6 +135,10 @@ DEFHOTPLUGVFIOONROOTBUS := false
 SED = sed
 
 CLI_DIR = cli
+SHIMV2 = containerd-shim-kata-v1
```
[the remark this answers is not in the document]
lifupan: Hi @jodh-intel 
I had communicated with crosby.michael@gmail.com about the shim name style and he explained the reason as below: 

Because we want to version the shims separately from the runtime/manager in containerd.  The containerd implementation is at v2 but the runc shim is the first release so it's v1.  If we do a large change, we can update the runc shim to a v2 binary while we make sure that v1 shims continue to work and run the same.  You would be able to do this with kata as well without impacting existing containers on a v1 kata shim.

bergwolf:
Walked through this PR with @lifupan f2f and it looks really good. Thanks @lifupan for such extraordinary work! And of course it
LGTM!

WeiZhang555:
51 commits is a large number, but great work! Need lots of time to read this.

containerd-shim/kata/service.go
```diff
@@ -0,0 +1,787 @@
...
+
+	"github.com/containerd/containerd/api/types/task"
+	"github.com/containerd/containerd/runtime/v2/runc/options"
+	"github.com/containerd/typeurl"
+	ptypes "github.com/gogo/protobuf/types"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+	"path/filepath"
```
WeiZhang555: It's better to tune the import to format:

```
import(
    "context"
    "path/filepath"

    "github.com/kata-containers/xxxx"
    "github.com/kata-containers/xxxx"

     "github.com/projectA/xxx"
     "github.com/projectB/xxx"

     "golang.org/xxx"
)
```

Same to all below

containerd-shim/kata/container.go
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) 2018 HyperHQ Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
```
WeiZhang555: Hmmm, this package is named "containerd-shim/kata/", since this is kata-runtime, I'm wonder if this `kata` is necessary.

Quesion is, in future, will there be any `containerd-shim/gvisor` or `containerd-shim/xxx` ?

WeiZhang555:
Just have a quick glance(Too many codes to read, can't finish it in a while).

One question in my mind: I see package `containerd-shim/kata` has lots of similar logics(and differences) with `cli`, how many codes can be shared between these two different packages?

If they can share some common codes, it will be easier to maintain for new bugfixes and new feature. I can see there will be some refactor if we want to share codes between two packages, and it already takes you lots of time on this, so this is not a strong opinion for this PR, maybe we can do it later, just want to know the possibility. 

@lifupan @bergwolf

jodh-intel:
I agree - we simply cannot duplicate code or we'll end up with bitrot.

containerd-shim/kata/container.go
lifupan: Ok, I'll remove directory "kata"

Makefile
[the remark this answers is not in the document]
gnawux: @lifupan I think you may simply remove the `-v1` here. For the containerd side, the `-v1` means using the v1 implementation to support v2 API, which is not the case for us. On the contrary, we don't have a v1 shim, we have only this one.

gnawux:
@WeiZhang555 @jodh-intel 
I think to reduce duplication could be addressed in a separated issue. This PR introduced a lot of new logic, and it is better to limit the code in its local dir. And guarantee the logic with enough of test cases. Later a new issue/PR could be issued to move the similar logic together if needed.

Makefile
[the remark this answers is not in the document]
resouer: +1, we don't need to stick to "v1" sth, it's confusing

WeiZhang555:
@gnawux This sounds good to me. Seperating common codes can cost lots of effort and will make this PR harder to review. Please just don't forget :-)

lifupan:
Hi @resouer @jodh-intel , after discussing with @gnawux , I'd prefer to use containerd-shim-kata-v2 as the shim name,  and here's "v2" means the  runtime api version, thus common users will not confused.

gnawux:
LGTM

jodh-intel:
@gnawux, @WeiZhang555 - yes, I appreciate that restructuring the code is going to take some effort, but we do currently work hard to avoid duplicated code where it is reasonable to do so.

The concern is that if this lands -- as @WeiZhang555 is hinting at -- the two copies of code will start to diverge and that increases the overall cost of maintaining this repository.

Rather than "fix up" the code after this PR lands, unless there is a burning need to get this landed fast, I'd rather we do one of the following:

- Fix it on this PR.
- Raise a fresh PR to refactor the code and get that landed *before* this PR lands (so that this PR can be rebased to take advantage of that).

On a related topic, I really think the commits on this PR need to be squashed - 54 is too many imho.

/cc @egernst, @sboeuf, @grahamwhaley.

gnawux:
@jodh-intel agree on the squash request

gnawux:
In the meanwhile, as most of the CI has passed, I think we could remove the WIP flag though there still some changes are requested.

lifupan:
Hi @jodh-intel @sboeuf 
Thanks for your suggestions, I'll extract the common codes into a clilib package which can be used by
both of the cli and shimv2 first and then send an init PR which only deal with the shimv2's frame work.
At last I'll send PRs one by one which address the specific service.

gnawux:
It's easy to say something like

> another one focusing on the declaration of the new skeleton with a bunch of empty functions, and the testing associated

However, for a new big task, a common thing is coming back to change the definition when developing the content, which makes the review and discussion of the former PR become a waste of time.

Right now, the job is here which works fine in general and @lifupan has already add bunch of tests for it, we should not waste time on re-doing the development progress again.

**I agree** we should *squash the commits* to help to review. However, **I don't think** we should *split it into several PRs* like @sboeuf said. It's simply too formalism and doesn't help we landed the implementation soon. If you simply think the code is too much to read, split to PRs doesn't help to reduce code.

What do you think about this, @egernst @bergwolf @WeiZhang555 ?

bergwolf:
Yes, I agree with @gnawux. It is smaller commits in a PR that can help reviewing. It doesn't help reviewing by splitting a large PR into smaller PRs. If someone with an area of expertise only wants to review part of a PR, they can certainly do so and only ACK on the particular part. But it is not reasonable to ask for splitting a PR to fit each reviewers' area of expertise. For this PR, it is already working and the code doesn't involve much change to the existing code. So @sboeuf can you please review it closely? It will cost sometime but I think it's faster than redoing everything again.

WeiZhang555:
To copy some text from Architecture Meeting:

I think it's fine to merge this before CLI refactor, seperating the containerd-shim-v2 logic in standalone directory might be easier for reviewer because the logic is all about shim-v2 adaption and won't involve too many components. 

And we all agree on : 
* squashing the commits as it's so many.
* do CLI refactor. Please don't forget this, and it'be better to do this before shim-v2 and CLI diverge

sboeuf:
@WeiZhang555 I'm confused, I thought it was agreed that the refactoring should happen first.

jodh-intel:
Yes, I thought we'd agreed to do the refactor first too. That will simplify the remaining code added by this PR since it can then just use the refactored code (rather than copying code as it does today).

gnawux:
Based on the discussion in the arch committee meeting, we created issue #714 to track the refactor of the cli, which will remove the duplicated code between current cli and this PR.

As @WeiZhang555 said

> do CLI refactor. Please don't forget this, and it'be better to do this before shim-v2 and CLI diverge

This issue is working on this.

WeiZhang555:
@sboeuf @jodh-intel 

Sorry for confusion, I'm not expressing myself very clear, that is my own thought, not commonly agreed idea on Arch Meeting. I'm saying that I'm OK with either refactoring CLI first or later.

According to Arch Meeting, it should be:
* Squash commits
* Refactor first before merging.

crosbymichael:
when can I test this? :)

jodh-intel:
Hi @crosbymichael - since this PR duplicates code, @lifupan is working on refactoring that code on #714 to allow this PR to be simplified.

I've added an entry to the Architecture Committee agenda for next week so we can get a status update on that task:

- https://etherpad.openstack.org/p/katacontainers-2018-architecture-committee-mtgs

lifupan:
Hi @crosbymichael,  just as @gnawux said, the  kata's shimv2's functions have been completed and I'm refactoring the kata's cli codes to share those codes between cli and shimv2.

amshinde:
@lifupan Any updates on this? I think there were plans to refactor this PR a bit?

lifupan:
Hi, Ace-tang

By now the functions have been completed and there is some code refactor needed. I will start the cli code refactor in a few days to make the codes much neat. You can have a try and please let me know if you meet any issues.

lifupan:
Hi, Ace-tang

By now the functions have been completed and there is some code refactor needed. I will start the cli code refactor in a few days to make the codes much neat. You can have a try and please let me know if you meet any issues.

lifupan:
Hi, Ace-tang

By now the functions have been completed and there is some code refactor needed. I will start the cli code refactor in a few days to make the codes much neat. You can have a try and please let me know if you meet any issues.

lifupan:
Hi, Ace-tang

By now the functions have been completed and there is some code refactor needed. I will start the cli code refactor in a few days to make the codes much neat. You can have a try and please let me know if you meet any issues.

lifupan:
Hi, Ace-tang

By now the functions have been completed and there is some code refactor needed. I will start the cli code refactor in a few days to make the codes much neat. You can have a try and please let me know if you meet any issues.

lifupan:
Hi, Ace-tang

By now the functions have been completed and there is some code refactor needed. I will start the cli code refactor in a few days to make the codes much neat. You can have a try and please let me know if you meet any issues.

Ace-Tang:
Thanks, @lifupan , for current pr, I test good work with containerd.

lifupan:
Hi, Ace-tang

By now the functions have been completed and there is some code refactor needed. I will start the cli code refactor in a few days to make the codes much neat. You can have a try and please let me know if you meet any issues.

Ace-Tang:
Thanks, @lifupan , for current pr, I test good work with containerd.

raravena80:
@lifupan ping, any updates? (from your weekly Kata herder)

lifupan:
Hi, @raravena80 
Sorry for this late reply, since I mainly focused on the kata benchmark testing in the past few days, and hadn't noticed this PR. I think I'll go on this PR from the cli codes refactoring from next week.

egernst:
We now have refactoring of some kata-utils merged -- what's next for this PR!?  Excited to get testing with this.

lifupan:
Hi @egernst , I'm rebasing this PR plus some additional refactors.

containerd-shim/kata/service.go
```diff
@@ -0,0 +1,786 @@
...
+func (s *service) pid() uint32 {
+	for true {
+		_, ok := s.processes[pidCount]
+		if !ok {
+			break
+		} else {
+			pidCount++
+			//if it overflows, recount from 5
+			if pidCount < 5 {
+				pidCount = 5
```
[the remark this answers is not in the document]
jodh-intel: Nit: this issue is still outstanding.

jodh-intel:
Thanks @lifupan - very nice refactor! I appreciate how long this must have taken to get right. But it's looks really good now so was worth the effort imho :)

lgtm

egernst:
Great work @lifupan - I appreciate this isn't trivial, as @jodh-intel pointed out.

In the current state, you added a commit at end which refactored the code to take into account the package katautils that merged.  I'd like to see this applied for the cli code first, in a seperate PR maybe even, and then the refactoring squashed into the commits.  I realize this will be painful and end result will be the same, but it'll help for review and maintenance, imo.

egernst:
@lifupan - in case it helps, here's what I was suggesting re: CLI changes: https://github.com/kata-containers/runtime/pull/897/commits/939e3a44b381ab25eefbfacf965ce18829ee72bb

I'd rather see this part in place first, and then apply your containerd-shim on top of this.

Makefile
```diff
@@ -253,7 +257,9 @@ define SHOW_ARCH
   $(shell printf "\\t%s%s\\\n" "$(1)" $(if $(filter $(ARCH),$(1))," (default)",""))
 endef
 
-all: runtime netmon
+all: runtime shim netmon
+
+shim: $(SHIMV2_OUTPUT)
```
sboeuf: I would prefer `containerd-shim-v2`

Makefile
```diff
@@ -413,11 +422,14 @@ check-go-static:
 coverage:
 	$(QUIET_TEST).ci/go-test.sh html-coverage
 
-install: default runtime install-scripts install-completions install-config install-bin install-bin-libexec
+install: default runtime install-scripts install-completions install-config install-bin install-shim install-bin-libexec
 
 install-bin: $(BINLIST)
 	$(foreach f,$(BINLIST),$(call INSTALL_EXEC,$f,$(BINDIR)))
 
+install-shim: $(SHIMV2)
```
sboeuf: I would prefer `install-containerd-shim-v2`

cli/containerd-shim-kata-v2/main.go
```diff
@@ -1,13 +1,16 @@
-
+// Copyright (c) 2018 HyperHQ Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
```
sboeuf: Please rework the commit:

    containerd-shim-kata: Support shim's start subcommand
    
    When creating a container in an existing sandbox,
    this shim should return its sock address and exit
    directly.

There is some rework of a bunch of structures that are not related to the support of the `start` subcommand.

crosbymichael:
@sboeuf the pids are opaque to containerd.  They are only used via the client to implement things like `docker ps`, etc.  For the main pid, I think the sandbox or whatever else you are already doing is fine.  We don't depend on the pids in containerd so it should be safe to omit or use the sandbox.

sboeuf:
@crosbymichael 
>the pids are opaque to containerd. They are only used via the client to implement things like docker ps, etc.

So you mean that we might break some use cases where, again, some assumptions about PIDs are made?

>For the main pid, I think the sandbox or whatever else you are already doing is fine. We don't depend on the pids in containerd so it should be safe to omit or use the sandbox.

That's a good news, thanks :)

crosbymichael:
I mean, there is always the risk of breaking certain usecases for clients but they should handle omitted pids just fine.  They can also introspect the runtime for a container and marshal the results as needed to handle differences.

sboeuf:
@crosbymichael makes sense, thanks!

Random-Liu:
@sboeuf For Kubernetes containerd integration, we don't use that pid to do anything on the host.

However, we do assume that each container/exec process has a different "pid number". And the "pid number" is used to figure out whether a `TaskExit` event is for the "init process" of the container/exec. For kata, I think you can just use the internal pid inside the sandbox.

sboeuf:
@Random-Liu 
>However, we do assume that each container/exec process has a different "pid number". And the "pid number" is used to figure out whether a TaskExit event is for the "init process" of the container/exec. For kata, I think you can just use the internal pid inside the sandbox.

Ok so if I follow correctly, you mean that you expect different PID numbers for all those processes, but you don't care if those PIDs are real (exist on the host). Do you think providing the same PID for every process of the same sandbox won't work?

Random-Liu:
> Ok so if I follow correctly, you mean that you expect different PID numbers for all those processes, but you don't care if those PIDs are real (exist on the host). Do you think providing the same PID for every process of the same sandbox won't work?

@sboeuf Actually, I just realized that we don't need the pid for that. I'm going to send a PR and cherry-pick it into containerd 1.2 as well.

So for Kubernetes integration, the only use case of `pid` is to show it in `crictl inspect` for debugging. You can put whatever there which works best for you. :)

sboeuf:
@Random-Liu 
>You can put whatever there which works best for you. :)

Nice! I have acked your PR on cri/containerd ;)

lifupan:
Hi @sboeuf, rebase had been done!

sboeuf:
Thanks @lifupan, I'll take a look on Monday!

egernst:
@lifupan - I think some of the rebase request was to squash https://github.com/kata-containers/runtime/pull/572/commits/1dcd50c17978ed13b8d91debc44d283a43051515 into the individual commits which introduce those codes.

containerd-shim/kata/service.go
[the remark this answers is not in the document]
egernst: @lifupan can you address this? ^

lifupan:
Hi @egernst, the rebase had been done.

gnawux:
and should ask @WeiZhang555 for reviewing as he had commented in this thread.

containerd-shim-v2/utils.go
```diff
@@ -29,6 +33,59 @@ func cReap(s *service, status int, id, execid string, exitat time.Time) {
 	}
 }
 
+func cleanupContainer(ctx context.Context, sid, cid, bundlePath string) error {
```
sboeuf: I don't see why cleanup is doing so many things here. I would expect `Delete()` to take care of most of the cleanup, since we have all the right information there. No need to dig through the OCI spec that we stored.

Also, when the `Delete()` will happen for a container which is in fact a sandbox, it should take care of removing everything.

containerd-shim-v2/service.go
```diff
@@ -537,7 +537,14 @@ func (s *service) Checkpoint(ctx context.Context, r *taskAPI.CheckpointTaskReque
 
 // Connect returns shim information such as the shim's pid
 func (s *service) Connect(ctx context.Context, r *taskAPI.ConnectRequest) (*taskAPI.ConnectResponse, error) {
-	return nil, errdefs.ErrNotImplemented
+	s.Lock()
+	defer s.Unlock()
+
+	return &taskAPI.ConnectResponse{
```
sboeuf: This shouldn't be used in our case because even the response that we're returning here is kind of wrong.

containerd-shim-v2/service.go
```diff
@@ -548,7 +548,27 @@ func (s *service) Connect(ctx context.Context, r *taskAPI.ConnectRequest) (*task
...
+	s.Lock()
+	defer s.Unlock()
+
+	if len(s.containers) != 0 {
+		return empty, nil
+	}
+
+	defer os.Exit(0)
+
+	err := s.sandbox.Stop()
```
sboeuf: What if the sandbox has been already stopped and deleted? We should check for that, and also check that `s.sandbox` is valid before we use it.

containerd-shim-v2/metrics.go
```diff
@@ -0,0 +1,66 @@
...
+				Failcnt: v.Failcnt,
+			})
+	}
+
+	var perCPU []uint64
+	for _, v := range cgStats.CPUStats.CPUUsage.PercpuUsage {
+		perCPU = append(perCPU, v)
+	}
+
+	metrics := &cgroups.Metrics{
```
sboeuf: Please factorize this in a converting function. Something like `statsToMetrics()` that would take care of converting the stats returned by Kata into the metrics structure expected by the containerd API.

containerd-shim-v2/service.go
```diff
@@ -313,7 +321,26 @@ func (s *service) Delete(ctx context.Context, r *taskAPI.DeleteRequest) (*taskAP
...
+	c, err := s.getContainer(r.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	if execs := c.execs[r.ExecID]; execs != nil {
+		return nil, errdefs.ToGRPCf(errdefs.ErrAlreadyExists, "id %s", r.ExecID)
+	}
+
+	execs, err := newExec(c, r.Stdin, r.Stdout, r.Stderr, r.Terminal, r.Spec)
```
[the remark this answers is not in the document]
sboeuf: Nevermind... I realized that was indeed the expected behavior.

containerd-shim-v2/service.go
```diff
@@ -316,7 +316,39 @@ func (s *service) Start(ctx context.Context, r *taskAPI.StartRequest) (*taskAPI.
...
+		}
+
+		return &taskAPI.DeleteResponse{
+			ExitStatus: c.exit,
+			ExitedAt:   c.time,
+			Pid:        PID,
+		}, nil
+	}
+	//deal with the exec case
+	execs, err := c.getExec(r.ExecID)
```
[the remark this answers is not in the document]
sboeuf: Nevermind... I realized that was indeed the expected behavior.

cli/main.go
```diff
@@ -307,7 +304,7 @@ func beforeSubcommands(c *cli.Context) error {
 
 	katautils.SetConfigOptions(name, defaultRuntimeConfiguration, defaultSysConfRuntimeConfiguration)
 
-	configFile, runtimeConfig, tracing, err = katautils.LoadConfiguration(c.GlobalString(configFilePathOption), ignoreLogging, false)
```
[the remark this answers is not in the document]
lifupan: Since the mainly tracing functions also had been removed into katautils pkg, thus the tracing trigger had been dealt in katautils pkg, and there was no need to return it here.

containerd-shim-v2/service.go
```diff
@@ -235,7 +239,45 @@ func (s *service) Cleanup(ctx context.Context) (*taskAPI.DeleteResponse, error)
 
 // Create a new sandbox or container with the underlying OCI runtime
 func (s *service) Create(ctx context.Context, r *taskAPI.CreateTaskRequest) (_ *taskAPI.CreateTaskResponse, err error) {
-	return nil, errdefs.ErrNotImplemented
+	s.Lock()
+	defer s.Unlock()
+
+	ns, err := namespaces.NamespaceRequired(ctx)
```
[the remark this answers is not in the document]
lifupan: Here's namespace is not the "namespace" in OCI.  This "namespace" is the kubernetes term, for much more info, please see https://kubernetes.io/blog/2016/08/kubernetes-namespaces-use-cases-insights/.

containerd-shim-v2/service.go
```diff
@@ -235,7 +239,45 @@ func (s *service) Cleanup(ctx context.Context) (*taskAPI.DeleteResponse, error)
...
+	rootfs := filepath.Join(r.Bundle, "rootfs")
+	defer func() {
+		if err != nil {
+			if err2 := mount.UnmountAll(rootfs, 0); err2 != nil {
+				logrus.WithError(err2).Warn("failed to cleanup rootfs mount")
+			}
+		}
+	}()
+	for _, rm := range r.Rootfs {
+		m := &mount.Mount{
```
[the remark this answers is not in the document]
lifupan: The root filesystem for the containers is provided by on the Create rpc. Shims are responsible for managing the lifecycle of the filesystem mount during the lifecycle of a container.  For much more info, please see https://github.com/containerd/containerd/blob/master/runtime/v2/README.md#root-filesystems.

containerd-shim-v2/service.go
```diff
@@ -235,7 +239,45 @@ func (s *service) Cleanup(ctx context.Context) (*taskAPI.DeleteResponse, error)
...
+
+	rootfs := filepath.Join(r.Bundle, "rootfs")
+	defer func() {
+		if err != nil {
+			if err2 := mount.UnmountAll(rootfs, 0); err2 != nil {
+				logrus.WithError(err2).Warn("failed to cleanup rootfs mount")
+			}
+		}
+	}()
+	for _, rm := range r.Rootfs {
```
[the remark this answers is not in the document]
lifupan: I had no idea why containerd defined the rootfs as "repeated" as below:

message CreateTaskRequest {
	string id = 1;
	string bundle = 2;
	repeated containerd.types.Mount rootfs = 3;
	...
}

containerd-shim-v2/service.go
```diff
@@ -235,7 +236,53 @@ func getTopic(ctx context.Context, e interface{}) string {
...
+	}
+
+	switch containerType {
+	case vc.PodSandbox:
+		err = cleanupContainer(ctx, s.id, s.id, path)
+		if err != nil {
+			return nil, err
+		}
+	case vc.PodContainer:
+		sandboxID, err := ociSpec.SandboxID()
```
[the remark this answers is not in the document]
lifupan: Here the Cleanup() api isn't called from the GRPC, instead it is called by shimv2 binary subcommand "delete", This command allows containerd to delete any container resources created, mounted, and/or run by a shim when containerd can no longer communicate over rpc. This happens if a shim is SIGKILL'd with a running container. These resources will need to be cleaned up when containerd looses the connection to a shim. This is also used when containerd boots and reconnects to shims. If a bundle is still on disk but containerd cannot connect to a shim, the delete command is invoked. At this moment, the shimv2 didn't have load the sandbox structs in memory thus here it should reload it from filesystem.

containerd-shim-v2/utils.go
lifupan: The same answer as above, the cleanupContainer() is only called by Cleanup() api, in which it should reload the sandbox info from filesystem.

containerd-shim-v2/service.go
lifupan: Yes, here the connect will ask for the shim's pid and the containers's pid, for the container's pid, I think it's better to return the shim's pid as the task pid instead of return any other values.

containerd-shim-v2/service.go
lifupan: No, this case will not happen unless there is a bug. Since the Shutdown() api is the only one entrance of stop and delete the sandbox, once it stop and delete it, this shimv2 process will also
exit and there is no chance for a shimv2 process running with sandbox has been stopped.

sboeuf:
@lifupan thanks for reworking. I'll be OOO until next Monday. I hope you get more reviews from other ppl this week. I'm planning to do a final (I hope it will be final :)) review on Monday when I get back.
Thanks!

gnawux:
Thanks for your careful review @sboeuf, which helps much for the project. Have a nice vacation.

And I have pinged @WeiZhang555, who will review the patch soon. Then we only wait for your approval before merging it.

containerd-shim-v2/service.go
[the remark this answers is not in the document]
WeiZhang555: Here we need some help from @crosbymichael :smiley: 

Any explanation on this? @crosbymichael

pkg/katautils/hook.go
```diff
@@ -1,9 +1,10 @@
-// Copyright (c) 2018 Intel Corporation
+// Copyright (c) 2017 Intel Corporation
```
WeiZhang555: s/2017/2018/ ? :wink:

containerd-shim-v2/create.go
```diff
@@ -0,0 +1,177 @@
...
+	//it has been ignored by kata-agent in sandbox.
+
+	for _, ns := range []specs.LinuxNamespaceType{
+		specs.UserNamespace,
+		specs.UTSNamespace,
+		specs.IPCNamespace,
+		specs.PIDNamespace,
+		specs.CgroupNamespace,
+	} {
+		removeNameSpace(&ociSpec, ns)
```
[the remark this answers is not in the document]
lifupan: Yes, it must be removed. For kubernetes,  all of the containers in a pod will share the same namespaces except mnt namespace,  thus the container's ocispec contains the pod's namespace info such as "/proc/<pid>/ns/pid", but the "pid" is wrong, shimv2 returned a fake pid to containerd. Thus we must remove them otherwise the kata agent will complained of "cannot find the pid" error.

containerd-shim-v2/create.go
```diff
@@ -0,0 +1,177 @@
...
+
+	container, err := newContainer(s, r, containerType, &ociSpec)
+	if err != nil {
+		return nil, err
+	}
+
+	return container, nil
+}
+
+func createSandbox(ctx context.Context, ociSpec oci.CompatOCISpec, runtimeConfig oci.RuntimeConfig,
```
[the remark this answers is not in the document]
lifupan: Indeed there are some codes shared between those two functions, but there are some difference too. For example, shimv2 will set the sandbox config as stateful and cli will not, shimv2 isn't care about the container ID mapping with the sandbox ID and shimv2 will also need the console dealing.  Thus I think merged those two functions into one will add much more "if else" logic which will make this function  much more complex. That's why I think it's better to keep them separated and as simple as possible.

containerd-shim-v2/delete.go
```diff
@@ -0,0 +1,49 @@
...
+	"context"
+	"path"
+
+	"github.com/containerd/containerd/mount"
+	"github.com/kata-containers/runtime/pkg/katautils"
+	vc "github.com/kata-containers/runtime/virtcontainers"
+	"github.com/sirupsen/logrus"
+)
+
+func deleteContainer(ctx context.Context, s *service, c *container) error {
```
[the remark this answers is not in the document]
lifupan: In fact those two functions only share the logic of stopContainer() and deleteContainer(),  I don't  think there is a necessary to wrap those two calling into another new function.

containerd-shim-v2/create.go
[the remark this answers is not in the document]
WeiZhang555: I see. Then I think it's not the containerd-shim-v2 specific problem, kata-runtime cli must have same problem. I think it's better to do some converting(to right pid) or removing in somewhere as `virtcontainers/kata-agent.go`, as a common handler

containerd-shim-v2/delete.go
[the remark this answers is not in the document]
WeiZhang555: @lifupan How about `createContainer` then?

These should be classified to one kind of issue: de-duplicate , which you have done partly but not covered all yet.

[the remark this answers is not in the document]
lifupan: For createContainer and createSandbox I had tried to merged them into a common function.

containerd-shim-v2/create.go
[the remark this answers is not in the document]
lifupan: I had double checked that kata-agent.go had dealt  the  namespaces.  Thus here it shouldn't deal with those namespaces, but there is a bug in dealing with Pid namespace in virtcontainers/kata-agent.go and there is also a bug in kata agent for dealing with "sharedPidNs", which will complained error as below: 
  Could not run process: container_linux.go:348: starting container process caused "process_linux.go:301: running exec setns process for init caused \"exit status 40\"": unknownill 

Thus here I will just remove the pid namespace and once I fixed the above two issues then I will delete these "remove" lines codes.

containerd-shim-v2/service.go
```diff
@@ -0,0 +1,782 @@
...
+		"-namespace", ns,
+		"-address", containerdAddress,
+		"-publish-binary", containerdBinary,
+		"-id", id,
+		"-debug",
+	}
+	cmd := sysexec.Command(self, args...)
+	cmd.Dir = cwd
+
+	// Set the go max process to 2 in case the shim forks two much process
```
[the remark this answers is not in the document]
lifupan: done

containerd-shim-v2/delete.go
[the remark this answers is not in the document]
WeiZhang555: Will you do more? Or this is already final version?

Take `createSandbox` as an example, I find they are almost same, except the return value type and missing 3 LOC.

So if we merge them to `katautils.CreateContainer()`, will this bring big problem?

containerd-shim-v2/create.go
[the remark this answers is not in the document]
WeiZhang555: It's better to file an issue tracking this, in case we leave it behind

containerd-shim-v2/create.go
```diff
@@ -0,0 +1,177 @@
...
+		specs.CgroupNamespace,
+	} {
+		removeNameSpace(&ociSpec, ns)
+	}
+
+	//set the network namespace path
+	//this set will be applied to sandbox's
+	//network config and hasn't nothing to
+	//do with containers in the sandbox since
+	//networkNamesapce hasn't been ignored by
```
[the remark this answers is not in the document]
lifupan: Done

containerd-shim-v2/create.go
```diff
@@ -0,0 +1,177 @@
...
+		specs.IPCNamespace,
+		specs.PIDNamespace,
+		specs.CgroupNamespace,
+	} {
+		removeNameSpace(&ociSpec, ns)
+	}
+
+	//set the network namespace path
+	//this set will be applied to sandbox's
+	//network config and hasn't nothing to
```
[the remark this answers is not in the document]
lifupan: Done

containerd-shim-v2/create.go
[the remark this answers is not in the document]
lifupan: had merged

containerd-shim-v2/create.go
```diff
@@ -0,0 +1,177 @@
...
+
+	containers := sandbox.GetAllContainers()
+	if len(containers) != 1 {
+		return nil, fmt.Errorf("BUG: Container list from sandbox is wrong, expecting only one container, found %d containers", len(containers))
+	}
+
+	return sandbox, nil
+}
+
+func createContainer(ctx context.Context, sandbox vc.VCSandbox, ociSpec oci.CompatOCISpec, containerID, bundlePath string,
```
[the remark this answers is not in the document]
lifupan: Had merged

containerd-shim-v2/metrics.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"github.com/containerd/cgroups"
+	"github.com/containerd/typeurl"
+
+	google_protobuf "github.com/gogo/protobuf/types"
+	vc "github.com/kata-containers/runtime/virtcontainers"
+)
+
+func marshalMetrics(s *service, containerID string) (*google_protobuf.Any, error) {
+	stats, err := s.sandbox.StatsContainer(containerID)
```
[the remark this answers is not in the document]
Ace-Tang: I also like it

pkg/katautils/hook.go
lifupan: Done

containerd-shim-v2/create_test.go
```diff
@@ -0,0 +1,362 @@
...
+	taskAPI "github.com/containerd/containerd/runtime/v2/task"
+
+	vc "github.com/kata-containers/runtime/virtcontainers"
+	"github.com/kata-containers/runtime/virtcontainers/pkg/vcmock"
+
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCreateCreateSandboxSuccess(t *testing.T) {
```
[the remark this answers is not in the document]
lifupan: Done

containerd-shim-v2/create.go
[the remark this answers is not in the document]
lifupan: yes, had opened an issue as https://github.com/kata-containers/runtime/issues/930

lifupan:
Hi, @WeiZhang555 the rebase and refactor has been done.

WeiZhang555:
LGTM overall.

I'm leaving final review to @sboeuf

containerd-shim-v2/stream.go
```diff
@@ -5,11 +5,119 @@
...
+	"sync"
+	"syscall"
+
+	"github.com/containerd/fifo"
+)
+
+var (
+	bufPool = sync.Pool{
+		New: func() interface{} {
+			buffer := make([]byte, 32<<10)
```
sboeuf: Please makes this a constant, and explain why it is needed.

containerd-shim-v2/service.go
```diff
@@ -298,12 +299,19 @@ func (s *service) Start(ctx context.Context, r *taskAPI.StartRequest) (*taskAPI.
 		}
 
 		return &taskAPI.StartResponse{
-			Pid: PID,
+			Pid: s.pid,
```
sboeuf: Nit: This is supposed to be part of the "start" implementation commit.

containerd-shim-v2/service.go
```diff
@@ -298,12 +299,19 @@ func (s *service) Start(ctx context.Context, r *taskAPI.StartRequest) (*taskAPI.
...
 		}, nil
 	}
 
 	//start an exec
-	return nil, errdefs.ErrNotImplemented
+	_, err = startExec(ctx, s, r.ID, r.ExecID)
```
sboeuf: Please use if/else here, this way you can factorize by calling:
```go
        return &taskAPI.StartResponse{
		Pid: s.pid,
	}, nil
```
only once at the end.

containerd-shim-v2/service.go
```diff
@@ -403,7 +403,30 @@ func (s *service) Update(ctx context.Context, r *taskAPI.UpdateTaskRequest) (*pt
 
 // Wait for a process to exit
 func (s *service) Wait(ctx context.Context, r *taskAPI.WaitRequest) (*taskAPI.WaitResponse, error) {
-	return nil, errdefs.ErrNotImplemented
+	var ret uint32
+
+	s.Lock()
+	c, err := s.getContainer(r.ID)
+	s.Unlock()
```
sboeuf: Don't we want to use the same global locking strategy for this one?
```go
s.Lock()
defer s.Unlock()

c, err := s.getContainer(r.ID)

...
```

containerd-shim-v2/service.go
```diff
@@ -350,7 +350,46 @@ func (s *service) ResizePty(ctx context.Context, r *taskAPI.ResizePtyRequest) (*
...
+	c, err := s.getContainer(r.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	if r.ExecID == "" {
+		return &taskAPI.StateResponse{
+			ID:         c.id,
+			Bundle:     c.bundle,
+			Pid:        PID,
```
sboeuf: Please replace `PID` with `s.pid`.

containerd-shim-v2/service.go
```diff
@@ -350,7 +350,46 @@ func (s *service) ResizePty(ctx context.Context, r *taskAPI.ResizePtyRequest) (*
...
+	//deal with exec case
+	execs, err := c.getExec(r.ExecID)
+	if err != nil {
+		return nil, err
+	}
+
+	return &taskAPI.StateResponse{
+		ID:         execs.id,
+		Bundle:     c.bundle,
+		Pid:        PID,
```
sboeuf: Same here.

containerd-shim-v2/service.go
```diff
@@ -403,7 +403,30 @@ func (s *service) Update(ctx context.Context, r *taskAPI.UpdateTaskRequest) (*pt
...
+	c, err := s.getContainer(r.ID)
+	s.Unlock()
+
+	if err != nil {
+		return nil, err
+	}
+
+	//wait for container
+	if r.ExecID == "" {
+		ret = <-c.exitCh
```
sboeuf: I think here it would be very nice to add `s.sandbox.StopContainer()`. We want to make sure the container is completely stopped inside the VM. The container process might terminate because it properly returned, and we want to ensure there's nothing left inside the VM.

containerd-shim-v2/service.go
```diff
@@ -350,7 +350,46 @@ func (s *service) ResizePty(ctx context.Context, r *taskAPI.ResizePtyRequest) (*
 
 // State returns runtime state information for a process
 func (s *service) State(ctx context.Context, r *taskAPI.StateRequest) (*taskAPI.StateResponse, error) {
```
sboeuf: Just noticing the way this function is implemented brings duplication additionally to the states of the Kata API. Instead, it maintains its own states.
@bergwolf @lifupan WDYT?

containerd-shim-v2/service.go
```diff
@@ -550,7 +550,35 @@ func (s *service) Resume(ctx context.Context, r *taskAPI.ResumeRequest) (*ptypes
...
+		processID = execs.id
+	}
+
+	err = s.sandbox.SignalProcess(c.id, processID, signum, r.All)
+	if err != nil {
+		return nil, err
+	}
+
+	if signum == syscall.SIGTERM {
+		err = s.sandbox.SignalProcess(c.id, processID, syscall.SIGKILL, r.All)
```
sboeuf: Why are we sending a `SIGKILL` here? The signal is a `SIGTERM` and has already been sent.
If there's a good explanation for this, please add a comment in the code explaining this.

containerd-shim-v2/metrics.go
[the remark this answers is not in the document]
sboeuf: That's not the only reason why `containerd-shim-v2` is very cool, but one of the main ones :)

containerd-shim-v2/service.go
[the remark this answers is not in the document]
lifupan: Hi @sboeuf, I'm sorry for the wrong explanation above. I had just dived into the containerd and found that here the "namespace" is the network namespace created by the containerd's CNI plugin, and that's why I passed it to kata's network ns path.

containerd-shim-v2/service.go
```diff
@@ -235,7 +239,45 @@ func (s *service) Cleanup(ctx context.Context) (*taskAPI.DeleteResponse, error)
...
+			Type:    rm.Type,
+			Source:  rm.Source,
+			Options: rm.Options,
+		}
+		if err := m.Mount(rootfs); err != nil {
+			return nil, errors.Wrapf(err, "failed to mount rootfs component %v", m)
+		}
+	}
+
+	container, err := create(ctx, s, r, ns, s.config)
```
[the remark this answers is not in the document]
lifupan: Sorry for the wrong explanation above. I had just dived into the containerd and found that here the "namespace" is the network namespace created by the containerd's CNI plugin, and that's why I passed it to kata's network ns path. Sorry for the confusion.

containerd-shim-v2/service.go
lifupan: No, here we cannot use a global lock. Since there would be multiple wait service requested in a pod, such as multi-containers  or execs run in this pod. The global lock would make the wait serialized, this isn't what we want.

containerd-shim-v2/service.go
lifupan: I thinks this states is needed. First the containerd requested states isn't entirely match to kata's states. there must be a transfer between them and this states does it; Second, there is some states value kata doesn't keep thus shimv2 needs to store it in this states.

containerd-shim-v2/service.go
lifupan: Since kubernetes will use SIGTERM signal to stop a container by default, but some container process such as "shell" will not response to this signal, thus container stop will always get timeout response, in order to terminate the container successfully, here send another SIGKILL to make sure the process will be terminated successfully.  I'll add it to comments.

containerd-shim-v2/create.go
```diff
@@ -0,0 +1,99 @@
...
+	containerType, err := ociSpec.ContainerType()
+	if err != nil {
+		return nil, err
+	}
+
+	// Todo:
+	// Since there is a bug in kata for sharedPidNs, here to
+	// remove the pidns to disable the sharePidNs temporarily,
+	// once kata fixed this issue, we can remove this line.
+	removeNameSpace(&ociSpec, specs.PIDNamespace)
```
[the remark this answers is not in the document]
lifupan: Done

containerd-shim-v2/create.go
```diff
@@ -0,0 +1,99 @@
...
+	//do with containers in the sandbox since
+	//networkNamespace has been ignored by
+	//kata-agent in sandbox.
+	for _, n := range ociSpec.Linux.Namespaces {
+		if n.Type != specs.NetworkNamespace {
+			continue
+		}
+
+		if n.Path == "" {
+			n.Path = netns
```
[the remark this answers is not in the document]
lifupan: The reason I had explained above for why replacing the network namespace with netns.

Since the "netns" is created by containerd's plugin, thus if containerd doesn't enabled any network plugin, then here the "netns" will be empty and yes it will be done later by CreateSandbox().

containerd-shim-v2/service.go
lifupan: Done

containerd-shim-v2/service.go
lifupan: Done

containerd-shim-v2/service.go
lifupan: Done

containerd-shim-v2/service.go
lifupan: Done

containerd-shim-v2/stream.go
lifupan: Done

containerd-shim-v2/service.go
sboeuf: Oh yeah good point, this would end up in a deadlock.

containerd-shim-v2/service.go
sboeuf: Yes I realized that, and I saw that you tried to translate the Kata states when possible, but I understand containerd expects more states than what we have with Kata. I'm not asking to change anything, it's just sad that we have to redefine another layer of states again...

containerd-shim-v2/service.go
sboeuf: Well I understand but I would expect the CRI layer (CRI-containerd or CRI-O) to actually send a SIGKILL themselves if they realize that SIGTERM didn't work. Is it something that we want to handle at this level?
/cc @bergwolf ^

containerd-shim-v2/service.go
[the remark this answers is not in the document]
sboeuf: Oh okay then that's nice because CNI is handled in a very straightforward way by doing so, I like it!
However, could you change the naming of the variable from `ns` to `netns` to make sure it's very obvious that we retrieved the network namespace from containerd.

containerd-shim-v2/service.go
sboeuf: Ah yes that's cool! Compared to the madness that we have to handle with our OCI CLI implementation, this is a real improvement here.

lifupan: Yes, if it's needed in the future  I'll add it.

lifupan: I had renamed it to netns and add  a line comment.

containerd-shim-v2/service.go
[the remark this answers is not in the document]
lifupan: I had renamed it to netns and a line of comment.

containerd-shim-v2/service.go
bergwolf: I think we should follow runc's practice and only send `SIGTERM`. If the container process chooses to ignore `SIGTERM`, the upper layer should send other signals after timeout. @lifupan do you see any test cases that runc passes but shimv2 fails due to `SIGTERM` timeout?

sboeuf: @bergwolf exactly, the upper layer knows about the workload and should be the one that knows if SIGTERM is appropriate or not.

lifupan: @bergwolf @sboeuf  The reason I added this KILL signal because for kata cli it also will send another KILL signal to make sure stop the container once it received SIGTERM and SIGKILL signal.

sboeuf: @lifupan let's not duplicate some fix that we had to use with kata CLI since they might not be needed in this case. Do you have a concrete test case that will actually fail if you don't send a SIGKILL after a SIGTERM ?

gnawux: Yes, IIRC the interactive shell ignores `SIGTERM`, I think it would respond to the `SIGTERM` if you run without `-it`.

sboeuf: Fair enough, it's not ideal, but we'll fix that later once we understand better the issue. I don't want to delay this PR because of this detail.

gnawux:
Finally merged, 
Thanks all the reviewers.

Time to have a beer @lifupan

containerd-shim-v2/service.go
lifupan: Thanks @sboeuf @jodh-intel @WeiZhang555 @bergwolf  @egernst. Thanks for your patient and careful help and review.

PR #580, opened 2018-08-14
[WIP] k8s memory resource hotplug

miaoyq:
At present, kata-runtime already supports memory hotplug in hypervisor side，however, the limitation of pod resources in k8s cannot be satisfied.

This pr try to satisfy the usage of k8s. 

Signed-off-by: Yanqiang Miao <miao.yanqiang@zte.com.cn>

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -1181,6 +1181,19 @@ func (c *Container) addResources() error {
 		return c.sandbox.agent.onlineCPUMem(vcpusAdded)
 	}
 
+	// Container is being created, try to add the memory specified
+	mem := c.config.Resources.Mem
+	if mem != 0 {
+		virtLog.Debugf("hot adding %d B memory", mem)
+		sizeMB := int(mem / 1024 / 1024)
+		// sizeMB needs to be divisible by 2
+		sizeMB = sizeMB + sizeMB%2
+		_, err := c.sandbox.hypervisor.hotplugAddDevice(&memoryDevice{1, sizeMB}, memoryDev)
+		if err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 
@@ -1202,6 +1215,18 @@ func (c *Container) removeResources() error {
 	if vCPUs != 0 {
 		virtLog.Debugf("hot removing %d vCPUs", vCPUs)
 		if _, err := c.sandbox.hypervisor.hotplugRemoveDevice(vCPUs, cpuDev); err != nil {
+			return err
+		}
+	}
+
+	mem := c.config.Resources.Mem
+	if mem != 0 {
+		virtLog.Debugf("hot removing %d B memory", mem)
+		sizeMB := int(mem / 1024 / 1024)
+		// sizeMB needs to be divisible by 2
+		sizeMB = sizeMB + sizeMB%2
+		_, err := c.sandbox.hypervisor.hotplugRemoveDevice(&memoryDevice{1, sizeMB}, memoryDev)
+		if err != nil {
 			return err
 		}
 	}
--- a/virtcontainers/pkg/oci/utils.go
+++ b/virtcontainers/pkg/oci/utils.go
@@ -605,6 +605,11 @@ func ContainerConfig(ocispec CompatOCISpec, bundlePath, cid, console string, det
 			resources.VCPUs = uint32(utils.ConstraintsToVCPUs(*ocispec.Linux.Resources.CPU.Quota, *ocispec.Linux.Resources.CPU.Period))
 		}
 	}
+	if ocispec.Linux.Resources.Memory != nil {
+		if ocispec.Linux.Resources.Memory.Limit != nil {
+			resources.Mem = uint32(*ocispec.Linux.Resources.Memory.Limit)
+		}
+	}
 
 	containerConfig := vc.ContainerConfig{
 		ID:             cid,


amshinde:
@miaoyq Can you add bit more details about what problem you are solving here and how you are solving it.

WeiZhang555:
You need to add `Fixes #470` and some detailed description of the commit in your commit message, or the CI won't pass.

virtcontainers/pkg/oci/utils.go
```diff
@@ -605,6 +605,11 @@ func ContainerConfig(ocispec CompatOCISpec, bundlePath, cid, console string, det
 			resources.VCPUs = uint32(utils.ConstraintsToVCPUs(*ocispec.Linux.Resources.CPU.Quota, *ocispec.Linux.Resources.CPU.Period))
 		}
 	}
+	if ocispec.Linux.Resources.Memory != nil {
+		if ocispec.Linux.Resources.Memory.Limit != nil {
```
WeiZhang555: You can make a short sentense:

```
if ocispec.Linux.Resources.Memory != nil && ocispec.Linux.Resources.Memory.Limit != nil {
```

virtcontainers/container.go
```diff
@@ -1206,6 +1219,18 @@ func (c *Container) removeResources() error {
 		}
 	}
 
+	mem := c.config.Resources.Mem
+	if mem != 0 {
+		virtLog.Debugf("hot removing %d B memory", mem)
+		sizeMB := int(mem / 1024 / 1024)
+		// sizeMB needs to be divisible by 2
+		sizeMB = sizeMB + sizeMB%2
+		_, err := c.sandbox.hypervisor.hotplugRemoveDevice(&memoryDevice{1, sizeMB}, memoryDev)
```
bergwolf: To support hot-unplug memory devices, we need to support guest ballooning first (which I don't think we have added yet).

jodh-intel:
Hi @miaoyq - any update on this?

Related: #624.

linzichang:
@miaoyq @jodh-intel @WeiZhang555 We should think about this carefully. We face 3 problems:
1. Each memory online size on Linux must be memblock aligned. That also mean the hotplug memory size need to be multiple of memblock size. We should read `/sys/devices/system/memoryblock_size_bytes` on GuestOS to get this size.
2. “Memory Hot-add” has two phase. First phase know as "Physical Hotplug", Qemu virtual hotplug or physical hotplug a memory device, cause an ACPI notify IRQ to linux kernel. Kernel handle IRQ and setup memory section, allocate page struct, etc. Second phase know as "Logical Hotplug",  using the sysfs interface to online memory and kernel will let the Buddy memory system to manage those new memory. In the first phase, kernel need to allocate page struct. The size is (hos-add memory size/4KB) * sizeof(struct page). Nearly 40:1 or 50:1(I haven't calculate it very accurately). This maybe not a problem but it affect the choosing of default vm boot memory size and per-hot-add size. For example, a vm with 128MB, the maximum hot-ad size once maybe 4G.
3. How to choose the default boot size, now is 2G. If pod don't need 2G, we may use hot-remove. But this face the `memblock` problem. And in addition, we need to care about ballooning and page migration which is not always succeed.

clarecch:
Indeed it's a better solution to hot add memory aligned to guestos memory block size. As now we have the fact:
1. Hot add memory should be aligned to memory section size, but the size of a memory section is architecture dependent. For example, power uses 16MB, ia64 uses 1GB, x86_64 uses 128MB,  ppc64le uses 256MB.
2.  Each memory block is described under /sys/devices/system/memory as `/sys/devices/system/memory/memoryXXX (XXX is the memory block id.)`, which is the unit of memory online/offline. In sparse memory model, memory block size is a multiple of memory section size, [reference](http://www.infradead.org/~mchehab/kernel_docs/unsorted/memory-hotplug.html#unit-of-memory-online-offline-operation). We can online a memory block only if we hotplug a memory-block-sized dimm, see `pages_correctly_reserved` in [reference](https://elixir.bootlin.com/linux/v4.14.49/source/drivers/base/memory.c#L230).

We can get the size of guestos memory block by reading `/sys/devices/system/memoryblock_size_bytes` in vm.

devimc:
My two cents,

there is still some points that haven't been discussed:

* How many mem slots should have a VM?
* Should number of mem slots  be configurable?
* Same as cpu sockets each mem slot consumes memory, how many MBs?

Since memory hot remove is almost impossible (or at least not reliable), we should track how much memory was assigned/required to/by a container, for example, if a container has 2GB but it's updated or removed (PODs scenario) its memory still can be used by other new or existing containers

miaoyq:
>Hot add memory should be aligned to memory section size, but the size of a memory section is architecture dependent. For example, power uses 16MB, ia64 uses 1GB, x86_64 uses 128MB, ppc64le uses 256MB.

Maybe we can make the memory section size configurable, for different architecture kata will use the different `qemu` binary. For exemple:  `x86_64`  to `qemu-system-x86_64`(or others related), `ppc64le` to `qemu-system-ppc64le`.

linzichang:
@miaoyq 
1. Physical hot-add need memory section aligned. Logical hot-add(online) need memory block aligned. Memory block size is a multiple of memory section size. So we need “memory block” aligned.
2. memory section size and memory block size is already configurable. It's in kernel configure/code. Qemu don't know what is memory block/section. x86 kernel code, when total memory is less than 64G, memory block is equal to memory section. Otherwise memory block is 2G.
```
static unsigned long probe_memory_block_size(void)
{
	/* start from 2g */
	unsigned long bz = 1UL<<31;

	if (totalram_pages >= (64ULL << (30 - PAGE_SHIFT))) {
		pr_info("Using 2GB memory block size for large-memory system\n");
		return 2UL * 1024 * 1024 * 1024;
	}

	/* less than 64g installed */
	if ((max_pfn << PAGE_SHIFT) < (16UL << 32))
		return MIN_MEMORY_BLOCK_SIZE;

	/* get the tail size */
	while (bz > MIN_MEMORY_BLOCK_SIZE) {
		if (!((max_pfn << PAGE_SHIFT) & (bz - 1)))
			break;
		bz >>= 1;
	}

	printk(KERN_DEBUG "memory block size : %ldMB\n", bz >> 20);

	return bz;
}
```
Two way to clear “memblock size”：
1. Read /sys/devices/system/memoryblock_size_bytes on GuestOS to get the size though `agent`.
2. Modify kernel configure and code to make the memory block size to what we want.

clarecch:
@miaoyq 
```For example, power uses 16MB, ia64 uses 1GB, x86_64 uses 128MB, ppc64le uses 256MB.```
This is just always not absolute. Memory section size may be different according to different kernel config. So it's difficult to make it configurable in host.

**Comment by miaoyq** [comment]
@linzichang @clarecch 
Thanks for your explanation. 
> 1 .Read /sys/devices/system/memoryblock_size_bytes on GuestOS to get the size though agent.
>2. Modify kernel configure and code to make the memory block size to what we want.

If so, the first option is a little better I think, we can add a interface in `agent`,  via which `runtime` can ask the `block_size` for `agent`.

**Comment by clarecch** [comment]
@miaoyq @devimc @linzichang Maybe we can get memory block size from agent when kata-runtime create or kata-runtime start, then store it in file. We don't need to  ask the memory block_size for agent everytime hotplug.

**Comment by miaoyq** [comment]
@clarecch Agree.

**Comment by clarecch** [comment]
```
there is still some points that haven't been discussed:

How many mem slots should have a VM?
Should number of mem slots be configurable?
Same as cpu sockets each mem slot consumes memory, how many MBs?
```
@grahamwhaley Do you have any idea about memory slot, which I'm not quite familiar with? I have read https://github.com/clearcontainers/runtime/issues/380. Is there any result of memory slot discussion?

**Comment by grahamwhaley** [comment]
> How many mem slots should have a VM?
> Should number of mem slots be configurable?
> Same as cpu sockets each mem slot consumes memory, how many MBs?
> @grahamwhaley Do you have any idea about memory slot, which I'm not quite familiar with? I have read clearcontainers/runtime#380. Is there any result of memory slot discussion?

Hi @clarecch - heh, good digging, finding that 1yr old thread! I think any measures we did before on number-of-slot overheads (~1.5yr ago now) will be out of date - we should/would have to re-run them if we need that data.

But, I think this will always be difficult - I very much doubt we can find a default that will suite all situations and users. I think we need to make this configurable (in the runtime toml config file), and set some sensible defaults. The only sensible default I can think of might be:
- split MAXMEM into a fixed number of slots (say 16 or 32 - with some math/rounding to make sure we end up on nice boundaries etc.). This has the flexibility that it will work on all systems probably (big and small memory systems).

In the config file I think we should probably offer two config options to give the user flexibility:
- the ability to set the number of slots (so MAXMEM/n slots)
- the ability to set the slot size (so nslots = MAXME/slotsize)

**Comment by clarecch** [comment]
@grahamwhaley very enlightening!

**Comment by grahamwhaley** [comment]
Maybe I'm thinking about this wrong as well.... I don't think slots have a fixed size do they? (that is, you can plug different sized dimms into each slot I think in QEMU?).
In that case, maybe the number of slots does not really relate to the MAXSIZE of memory, but does relate to how many times we expect to hotplug add memory to a pod.
Do we only expect to hotplug one dimm for each extra container added to a pod? In that case, the number of slots relates directly to the maximum number of containers we ever expect to add to a pod. That may change our view here. 8 or 16 slots still feels sensible to me - but, does anybody have evidence or numbers around how many containers make up a pod?

**Comment by clarecch** [comment]
I think in some case we will hotplug multiple dimms for every container in a pod, which contains multiple containers.

**Comment by linzichang** [comment]
I change `defaultMemSlots` from 2 to 255. The PSS of qemu process will increase 11MB.

File: virtcontainers/container.go [hunk 3]
```diff
@@ -1181,6 +1181,19 @@ func (c *Container) addResources() error {
 		return c.sandbox.agent.onlineCPUMem(vcpusAdded)
 	}
 
+	// Container is being created, try to add the memory specified
+	mem := c.config.Resources.Mem
+	if mem != 0 {
+		virtLog.Debugf("hot adding %d B memory", mem)
+		sizeMB := int(mem / 1024 / 1024)
+		// sizeMB needs to be divisible by 2
+		sizeMB = sizeMB + sizeMB%2
```
jodh-intel: It might be better to move the log call to here and log `sizeMB` (and maybe `mem` too):

```
virtLog.WithField("memory-mb", sizeMB).Info("hot adding memory")
```

Same comment for the log call in `removeResources()`.

virtcontainers/pkg/oci/utils.go
```diff
@@ -605,6 +605,11 @@ func ContainerConfig(ocispec CompatOCISpec, bundlePath, cid, console string, det
...
+			resources.Mem = uint32(*ocispec.Linux.Resources.Memory.Limit)
```
jodh-intel: This is casting an int64 into an int32:

- https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md

Maybe `ContainerResources.Mem` should be changed to `int64` too?

grahamwhaley:
Thanks @linzichang, thanks for grabbing that data. 11Mb over how much - what is to total PSS (so we can see what % increase there is).
Also, was that with KSM enabled or not on the host?, and how many containers did you run?

Our PSS footprint can be between maybe 48Mb (20 containers, KSM) to 135Mb (20 containers, no KSM) for instance on my local system - in which case 11Mb could represent anywhere from a 23% to 8% footprint increase. Both of those I would call significant enough to need a discussion :-)

You could maybe use the reportgen from the tests repo to generate us a report.
Over at: https://github.com/kata-containers/tests/tree/master/metrics/report
If you run the `grabdata.sh` with and without your changes installed and then `makereport.sh`, the resulting pdf report may help us assess.

`grabdata` can take some time, and if you are only looking at the density/footprint then the version on the pending PR https://github.com/kata-containers/tests/pull/650 allows you to tell `grabdata` to only run a subset of tests (like, just density). I think it would also be good in this case to see if there is any significant change in boot time as well though.

devimc:
@lifupan 

> @devimc I change defaultMemSlots from 2 to 255. The PSS of qemu process will increase 11MB.

thanks, have you measure how much memory consume the guest kernel? I wouldn't like to see issue https://github.com/kata-containers/runtime/issues/295 again but now with memory slots

linzichang:
@devimc You have @ the wrong guy twice, :)

devimc:
@linzichang :smile:  I'm sorry

linzichang:
@sboeuf @devimc If @miaoyq has no time to work on this, I and @clarecch can help to rework this.

miaoyq:
@sboeuf @linzichang @clarecch  I'm working on this. 
If I have any doubts, I will discuss with you. Thanks. :-)

linzichang:
@miaoyq I need this feature in short time. We have already talk about the solution in https://github.com/kata-containers/runtime/pull/624#issuecomment-419656853 . I will implement some both needed method in update memory to help speedup this PR. Let'us get this done quicker together. Thank you to rework this very much :)

linzichang:
@miaoyq I think you should first take care the memory footprint we mentions in https://github.com/kata-containers/runtime/pull/580#issuecomment-418219019 https://github.com/kata-containers/runtime/pull/580#issuecomment-418305751 https://github.com/kata-containers/runtime/pull/580#issuecomment-419934060. Because I haven't tested it clear yet.

miaoyq:
@linzichang I think you're a little bit more familiar with this, so in order not to affect your usage, you can rework this freely. :-)

raravena80:
Hi, @linzichang @miaoyq any updates on this? thx!

miaoyq:
@raravena80 
I think @linzichang and @clarecch have finished this locally.
@linzichang Could you update this feature in a new PR?

linzichang:
@miaoyq @raravena80 I will open a new PR in the last few days.

linzichang:
Rework PR #786

miaoyq:
>Rework PR #786

@linzichang I will close this PR, thanks!

PR #600, opened 2018-08-17
[DNM] network: Move to network hotplug

### LINKED ISSUE #598: network: Move to fully hotplugged network
We need to move to fully hotplugged network if we want to support the use case of booting a VM from a template, and also to allow more flexibility.

### LINKED ISSUE #599: cli: Move OCI hooks handling from virtcontainers to the CLI
OCI hooks are obviously OCI specific and very related to the way the network is setup. Once the hotplug gets introduced in Kata, we should move the hook processing to the CLI level. This allows for an OCI agnostic Kata API.

sboeuf:
This patchset introduces all the changes needed to move to a fully hotplugged network instead of the legacy coldplug that was in place.
Also, going along with those changes, this patchset moves the handling of the OCI hooks out of virtcontainers, allowing for an OCI agnostic Kata API.

Fixes #598
Fixes #599


sboeuf:
This PR is blocked by the fact that we cannot move to a fully hotplugged model until we get both VFIO network and vhost-user hotplug support.

egernst:
will want to make sure we test this with all CNI/CNM plugins (incl vhost-user based and SRIOV/VFIO based).

egernst:
@sboeuf - are there changes in the gRPC protocol which this is reliant on since 1.2?

sboeuf:
@egernst no, nothing related to GRPC changes here.

raravena80:
Hi, @sboeuf any updates on this? why is it marked do not merge? Thx!

sboeuf:
@raravena80 this PR can be dropped as I don't think we'll move to a fully hotplugged model by default. We already have Kata able to hotplug new network.

See https://github.com/kata-containers/runtime/issues/598 for reference about the discussion we already had on the topic.

PR #602, opened 2018-08-17
network: fix vhost-user net creation

egernst:
When creating a device structure to be added to the hypervisor, make
sure that the device includes the vhost-user type.  In particular,
for network devices, specific it is VhostUserNet

Fixes: #601

### DIFF
--- a/virtcontainers/network.go
+++ b/virtcontainers/network.go
@@ -286,6 +286,7 @@ func (endpoint *VhostUserEndpoint) Attach(h hypervisor) error {
 		ID:         id,
 		SocketPath: endpoint.SocketPath,
 		MacAddress: endpoint.HardAddr,
+		Type:       config.VhostUserNet,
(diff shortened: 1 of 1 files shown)


amshinde:
lgtm

jodh-intel:
Good catch. This sort of issue is nasty and the only way to avoid these sorts of bug I think is to (somehow) enforce we only ever create structs using a `NewThing() Thing` type idiom (so that the *only* way to create a `Thing` is to make that call).

jodh-intel:
lgtm

bergwolf:
Good catch! And I think we should also fail `h.addDevice` properly (instead of always returning success there). Surely, this
LGTM!

sboeuf:
Just restarted the density build, we can merge right after

sboeuf:
@egernst please backport to stable branches.

Issue #614, opened 2018-08-21
Kata-runtime start failed on Arm host with GICv4 support


Weichen81:
# Description of problem

QEMU supports GICv2, GICv3 and host parameters for guest gic-version. The "host" parameter will let QEMU detect GIC version by itself. This parameter will work properly when host GIC version is GICv2 or GICv3. But the detection will failed when host GIC is GICv4 or higher:
"Unable to determine GIC version supported by host"

# Expected result

kata-runtime start successfully

# Actual result

"Unable to determine GIC version supported by host"

---

@Pennyzct

Weichen81:
Different GIC supports different QEMU parameters for virtual GIC and max VCPUs. So this issue also affects
https://github.com/kata-containers/runtime/issues/584

PR #585, opened 2018-08-15
qemu: refactor maximum vcpus supported in aarch64

Pennyzct:
on aarch64, we support different gic interrupt controllers.
The maximum number of vCPUs depends on the GIC version, or on how many redistributors we can fit into the memory map.

Fixes: #584

Signed-off-by: Penny Zheng <penny.zheng@arm.com>
Signed-off-by: Wei Chen <Wei.Chen@arm.com>

PR #616, opened 2018-08-21
qemu/arm64: Detect host GIC version to configure guest GIC

Weichen81:
QEMU supports GICv2, GICv3 and host parameters for guest gic-version. The "host" parameter will let QEMU detect GIC version by itself. This parameter will work properly when host GIC version is GICv2 or GICv3. But the detection will failed when host GIC is GICv4 or higher:
"Unable to determine GIC version supported by host"

In this case, we have to detect the host GIC version manually and force QEMU to use GICv3 when host GIC is GICv4 or higher.

Fixes: #614

Signed-off-by: Wei Chen <wei.chen@arm.com>

### DIFF
--- a/virtcontainers/qemu_arm64.go
+++ b/virtcontainers/qemu_arm64.go
@@ -6,9 +6,12 @@
 package virtcontainers
 
 import (
+	"io/ioutil"
 	"runtime"
+	"strings"
 
 	govmmQemu "github.com/intel/govmm/qemu"
+	"github.com/sirupsen/logrus"
 )
 
 type qemuArm64 struct {
@@ -20,7 +23,7 @@ const defaultQemuPath = "/usr/bin/qemu-system-aarch64"
 
 const defaultQemuMachineType = QemuVirt
 
-const defaultQemuMachineOptions = "gic-version=host,usb=off,accel=kvm"
+var defaultQemuMachineOptions = "usb=off,accel=kvm,gic-version=" + getGuestGICVersion()
 
 // Not used
 const defaultPCBridgeBus = ""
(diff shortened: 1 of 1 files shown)


virtcontainers/qemu_arm64.go
```diff
@@ -20,7 +23,7 @@ const defaultQemuPath = "/usr/bin/qemu-system-aarch64"
 
 const defaultQemuMachineType = QemuVirt
 
-const defaultQemuMachineOptions = "gic-version=host,usb=off,accel=kvm"
+var defaultQemuMachineOptions = "usb=off,accel=kvm" + "," + guestGICversion
```
caoruidong: Can it just be "usb=off,accel=kvm,"

virtcontainers/qemu_arm64.go
```diff
@@ -45,6 +48,65 @@ var supportedQemuMachines = []govmmQemu.Machine{
...
 }
 
+// Logger returns a logrus logger appropriate for logging qemu-aarch64 messages
+func qemuArmLogger() *logrus.Entry {
+	return virtLog.WithField("subsystem", "qemu-aarch64")
+}
+
+// On ARM platform, we have different GIC interrupt controllers. Different
+// GIC supports different QEMU parameters for virtual GIC and max VCPUs
+var hostGICVersion = getHostGICVersion()
```
devimc: I think this should be a local variable in `getGuestGICVersion`

virtcontainers/qemu_arm64.go
Weichen81: @caoruidong Yes, that would be better, I will fix it.

virtcontainers/qemu_arm64.go
Weichen81: @devimc Emm, yoiu are right. I had thought that @Pennyzct would use this variable to adjust the maxvcpu number for QEMU. It seems we are in the same package. I will fix it.

Weichen81: @devimc Sorry, I had got a mis-understanding about your comment just now, hostGICVersion would be used in @Pennyzct 's PR https://github.com/kata-containers/runtime/pull/585 later to detect maxvcpu. So I would prefer to keep it as var.

virtcontainers/qemu_arm64.go
```diff
@@ -45,6 +48,65 @@ var supportedQemuMachines = []govmmQemu.Machine{
...
+
+	return 0
+}
+
+// QEMU supports GICv2, GICv3 and host parameters for gic-version. The host
+// parameter will let QEMU detect GIC version by itself. This parameter
+// will work properly when host GIC version is GICv2 or GICv3. But the
+// detection will failed when host GIC is gicv4 or higher. In this case,
+// we have to detect the host GIC version manually and force QEMU to use
+// GICv3 when host GIC is GICv4 or higher.
```
bergwolf: Looks like this is something that should be fixed in QEMU.

virtcontainers/qemu_arm64.go
```diff
@@ -45,6 +48,65 @@ var supportedQemuMachines = []govmmQemu.Machine{
...
+// We will access this file on host to detect host GIC version
+var gicProfile = "/proc/interrupts"
+
+// Detect the host GIC version.
+// Success: return the number of GIC version
+// Failed: return 0
+func getHostGICVersion() (version uint32) {
+	bytes, err := ioutil.ReadFile(gicProfile)
+	if err != nil {
+		qemuArmLogger().WithField("GIC profile", gicProfile).WithError(err).Error("Failed to parse GIC profile")
```
jodh-intel: Nit: It's an extra line of code, but I'd still be tempted to add `return 0` here to fail fast.

virtcontainers/qemu_arm64.go
```diff
@@ -45,6 +48,65 @@ var supportedQemuMachines = []govmmQemu.Machine{
...
+
+// QEMU supports GICv2, GICv3 and host parameters for gic-version. The host
+// parameter will let QEMU detect GIC version by itself. This parameter
+// will work properly when host GIC version is GICv2 or GICv3. But the
+// detection will failed when host GIC is gicv4 or higher. In this case,
+// we have to detect the host GIC version manually and force QEMU to use
+// GICv3 when host GIC is GICv4 or higher.
+var guestGICversion = getGuestGICVersion()
+
+func getGuestGICVersion() (version string) {
```
jodh-intel: Nit: For parity with `getHostGICVersion()`, it might be better if `getGuestGICVersion()` simply returned `2`, `3` or `host`. That would make `defaultQemuMachineOptions` a little clearer as it could become:

```
var defaultQemuMachineOptions = "usb=off,accel=kvm," + "gic-version=" + guestGICversion
```

or maybe just:

```
var defaultQemuMachineOptions = "usb=off,accel=kvm," + "gic-version=" + getGuestGICVersion()
```

virtcontainers/qemu_arm64.go
```diff
@@ -45,6 +48,65 @@ var supportedQemuMachines = []govmmQemu.Machine{
...
+	bytes, err := ioutil.ReadFile(gicProfile)
+	if err != nil {
+		qemuArmLogger().WithField("GIC profile", gicProfile).WithError(err).Error("Failed to parse GIC profile")
+	}
+
+	if strings.Contains(string(bytes), "GICv2") {
+		return 2
+	}
+
+	if strings.Contains(string(bytes), "GICv3") {
```
jodh-intel: Nit: Duplicated string conversion. Maybe:

```
s := string(bytes)

if strings.Contains(s, "GICv2") {
    return 2
}

if strings.Contains(s, "GICv3") {
    return 3
}
    :
```

jodh-intel:
Thanks for raising @Weichen81!

Weichen81:
@jodh-intel Thank you! I have updated the PR to address your comments : )

jodh-intel:
Thanks @Weichen81! Let's just wait for the CI to re-run...

Weichen81:
@jodh-intel @bergwolf @devimc
Can we merge this PR ? @Pennyzct 's PR is pending ; )

jodh-intel:
@Weichen81 - I've just restarted the two failing CI's. Let's see how they go so we can get this landed asap...

PR #617, opened 2018-08-21
Add virtcontainers trace support

### LINKED ISSUE #566: Add tracing to virtcontainers
As a follow-on from #557, add tracing to the virtcontainers packages.

jodh-intel:
This PR expands the opentracing support in the runtime, providing increased granularity by adding trace spans to the `virtcontainers` packages.

Like the initial PR (#565), the strategy here is to add `context.Context` parameters to all `virtcontainers` API calls. The specified context replaces the default context created in certain parts of the code and allows trace spans to be created.

I've added a `trace()` function in `cli/tracing.go` but most subsystems and subsystem implementations also create their own `trace()` method.

Tags have been added to spans, mostly via the `trace()` functions:

- `component`

  This is equivalent to the `source` log field (see https://github.com/kata-containers/tests/tree/master/cmd/log-parser#logfile-requirements) and is named this way to conform to the standard
  (https://github.com/opentracing/specification/blob/master/semantic_conventions.md#span-tags-table).

- `type`

  This is used to provide more information for interface implementations.

Examples:

- For `component=network`, `type` could be `cnm`.
- For `component=hypervisor`, `type` could be `qemu`.

Fixes #566.
    
### DIFF
--- a/cli/create_test.go
+++ b/cli/create_test.go
@@ -293,7 +293,7 @@ func TestCreateInvalidArgs(t *testing.T) {
 		},
 	}
 
-	testingImpl.CreateSandboxFunc = func(sandboxConfig vc.SandboxConfig) (vc.VCSandbox, error) {
+	testingImpl.CreateSandboxFunc = func(ctx context.Context, sandboxConfig vc.SandboxConfig) (vc.VCSandbox, error) {
 		return sandbox, nil
 	}
 
@@ -488,7 +488,7 @@ func TestCreateProcessCgroupsPathSuccessful(t *testing.T) {
 	defer os.RemoveAll(path)
 	ctrsMapTreePath = path
 
-	testingImpl.CreateSandboxFunc = func(sandboxConfig vc.SandboxConfig) (vc.VCSandbox, error) {
+	testingImpl.CreateSandboxFunc = func(ctx context.Context, sandboxConfig vc.SandboxConfig) (vc.VCSandbox, error) {
 		return sandbox, nil
 	}
 
@@ -583,7 +583,7 @@ func TestCreateCreateCgroupsFilesFail(t *testing.T) {
 	defer os.RemoveAll(path)
 	ctrsMapTreePath = path
 
-	testingImpl.CreateSandboxFunc = func(sandboxConfig vc.SandboxConfig) (vc.VCSandbox, error) {
+	testingImpl.CreateSandboxFunc = func(ctx context.Context, sandboxConfig vc.SandboxConfig) (vc.VCSandbox, error) {
 		return sandbox, nil
 	}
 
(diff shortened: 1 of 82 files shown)


jodh-intel:
Given the amount of code this touches, it would be *great* to get this reviewed sooner rather than later to avoid "rebase hell" if code changes underneath this PR... :)

cli/create_test.go
```diff
@@ -293,7 +293,7 @@ func TestCreateInvalidArgs(t *testing.T) {
 		},
 	}
 
-	testingImpl.CreateSandboxFunc = func(sandboxConfig vc.SandboxConfig) (vc.VCSandbox, error) {
+	testingImpl.CreateSandboxFunc = func(ctx context.Context, sandboxConfig vc.SandboxConfig) (vc.VCSandbox, error) {
```
devimc: I wonder if wouldn't be better to define a global context variable instead of modify all the functions

devimc: might be something like `traceContext` ..

jodh-intel: I did think about that (and in fact did use that technique in `virtcontainers/hack/virtc/main.go`), but I'd rather we use [the recommended approach](https://github.com/yurishkuro/opentracing-tutorial/tree/master/go/lesson02#propagate-the-in-process-context) and minimise globals where possible.

devimc: make sense, thanks for the info

devimc:
recheck

jodh-intel:
Ping @amshinde, @sboeuf, @bergwolf.

jodh-intel:
Ping @egernst.

egernst:
awesome.  @jodh-intel to easy stable handling, do you think we can seperate out the agent change from the VC changes?  It makes it a bit easier to track a whole PR for applicability.  WDYT?

bergwolf:
@egernst Is this going to be a stable candidate?

jodh-intel:
@egernst - I agree. I've removed that commit and will raise a follow-on PR when this PR gets merged...

@egernst, @bergwolf  - Agreed - I don't think the tracing code should be in stable yet as it's brand new (and hence "unproven" at this point).

jodh-intel:
Ping @kata-containers/runtime - please can you review??

jodh-intel:
Errm, checking the F27 CI log:

```
Repository  was already defined in a 'Depends-on:' tag.
Only one repository per tag is allowed.
```

WTF? I didn't add any depends-on...? That message looks like "repository" is actually `""` (note the double space before "was").

/me puts another coin in the slot and pulls the handle...

jodh-intel:
@amshinde, @bergwolf, @sboeuf, @egernst, @WeiZhang555 - ptal....

devimc:
recheck

devimc:
restarting jenkins-ci-fedora-27-vsocks

PR #623, opened 2018-08-22
network: Move OCI hooks and network namespace creation out of virtcontainers

### LINKED ISSUE #599: cli: Move OCI hooks handling from virtcontainers to the CLI

sboeuf:
Before we decide to move to a fully hotpluggable network model, we want to go step by step by making sure we can still support the network being coldplugged in case of `kata-runtime`, but also letting the possibility to consumer of the API (such as `frakti`) to hotplug the network interfaces, without having virtcontainers doing some prework there.

This PR introduces a few rework and fixes, but the most important part is about the creation of the network namespace and the execution of the OCI hooks being handled directly from the CLI. This allows the Kata API to be OCI agnostic.

Fixes #599


sboeuf:
@egernst @WeiZhang555 @bergwolf @amshinde @caoruidong PTAL

FYI, this PR does not supersede #600, but shares part of the commits. The approach here is less aggressive as we're not dropping the network coldplug, but we let the possibility for the network not to be created if we don't want to (frakti case with the factory).

We need to discuss if we want eventually to move to a fully hopluggable model, but this is beyond this PR scope.

jodh-intel:
This is going to conflict with #617, so please could folk take a look at that one :)

cli/create.go
```diff
@@ -267,6 +267,20 @@ func createSandbox(ctx context.Context, ociSpec oci.CompatOCISpec, runtimeConfig
...
 
+	// Important to create the network namespace before the sandbox is
+	// created, because it is not responsible for the creation of the
+	// netns if it does not exist.
+	if err := setupNetworkNamespace(&sandboxConfig.NetworkConfig); err != nil {
+		return vc.Process{}, err
+	}
+
+	// Execute prestart OCI hooks
+	if err := enterNetNS(sandboxConfig.NetworkConfig.NetNSPath, func() error {
```
jodh-intel: This is a bit of a mind-bender :) I think it would be clearer if you assigned to `err` and then checked `err` on a separate line.

virtcontainers/noop_resource_storage.go
```diff
@@ -0,0 +1,112 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package virtcontainers
+
+import (
+	"github.com/kata-containers/runtime/virtcontainers/device/api"
+)
+
+type noopResourceStorage struct{}
```
jodh-intel: Nice. Since you've added this though, how about splitting `filesystem.go` into:

- `resource_storage.go` (`type resourceStorage interface` only).
- `resource_storage_filesystem.go`

And renaming `filesystem_test.go` to `resource_storage_filesystem_test.go` ?

cli/hook.go
```diff
@@ -0,0 +1,139 @@
...
+			}).Error("hook error")
+
+			return err
+		}
+	}
+
+	return nil
+}
+
+func postStartHooks(spec oci.CompatOCISpec, cid, bundlePath string) error {
```
jodh-intel: This function and `postStopHooks()` are almost identical to `preStartHooks()`, so how about having them all call a generic function. Something like:

```
func runHooks(cid, bundlePath string, hooks []Hook, name string) error { }

func preStartHooks(spec oci.CompatOCISpec, cid, bundlePath string) error {
    if spec.Hooks == nil {
        return nil
    }

    return runHooks(cid, bundlePath, spec.Hooks.Prestart, "pre-start")
}
```

cli/start.go
```diff
@@ -78,7 +79,23 @@ func start(ctx context.Context, containerID string) (vc.VCSandbox, error) {
...
+		}
+
+		ociSpec, err := oci.GetOCIConfig(status)
+		if err != nil {
+			return nil, err
+		}
+
+		if err := enterNetNS(s.GetNetNs(), func() error {
+			return postStartHooks(ociSpec, sandboxID, status.Annotations[vcAnnot.BundlePathKey])
+		}); err != nil {
```
jodh-intel: Same comment as above - this is compact, but rather difficult to understand ;)

virtcontainers/default_network.go
```diff
@@ -0,0 +1,83 @@
...
+	"fmt"
+
+	"github.com/containernetworking/plugins/pkg/ns"
+	"github.com/sirupsen/logrus"
+)
+
+type defNetwork struct {
+}
+
+func defNetworkLogger() *logrus.Entry {
```
jodh-intel: Why not?:

```
func (n *defNetwork) logger() *logrus.Entry
```

virtcontainers/network.go
```diff
@@ -1339,6 +1177,93 @@ func createVirtualNetworkEndpoint(idx int, ifName string, interworkingModel NetI
 	return endpoint, nil
 }
 
+func generateInterfacesAndRoutes(networkNS NetworkNamespace) ([]*grpc.Interface, []*grpc.Route, error) {
```
jodh-intel: I'm pretty surprised the linters don't complain about the length/complexity of this function ;)

egernst:
This is a big PR - nice work.  I feel like some of the commits would be better off in a seperate PR?  In particular, 
virtcontainers: qemu: Don't shutdown QMP from hotplug

virtcontainers/network.go
```diff
@@ -598,8 +599,8 @@ const (
 	// NoopNetworkModel is the No-Op network.
 	NoopNetworkModel NetworkModel = "noop"
 
-	// CNMNetworkModel is the CNM network.
-	CNMNetworkModel NetworkModel = "CNM"
+	// DefaultNetworkModel is the default network.
+	DefaultNetworkModel NetworkModel = "default"
```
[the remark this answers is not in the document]
sboeuf: Yeah tell me which name I could use, but I agree "default" is definitely not great...
A bit more background here, initially I wanted to remove the interface since it was not really needed as we now have only one real implementation, but I realized that our unit testing would be way more complicated/impossible? if we decide to remove the noop implementation.

caoruidong:
Most of changes are code movement, generally good

cli/create.go
sboeuf: Ok fair enough ;)

virtcontainers/noop_resource_storage.go
sboeuf: Yes, this makes sense, let me take care of this.

virtcontainers/default_network.go
sboeuf: sure, that works !

cli/hook.go
sboeuf: Okay I can work on this

cli/create.go
```diff
@@ -267,6 +267,20 @@ func createSandbox(ctx context.Context, ociSpec oci.CompatOCISpec, runtimeConfig
...
+		return preStartHooks(ociSpec, containerID, bundlePath)
```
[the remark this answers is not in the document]
sboeuf: This makes sense, even if I don't expect any hook for containers following the sandbox container.

cli/delete.go
```diff
@@ -107,6 +108,10 @@ func delete(ctx context.Context, containerID string, force bool) error {
 
 	switch containerType {
 	case vc.PodSandbox:
+		if err := postStopHooks(ociSpec, sandboxID, status.Annotations[vcAnnot.BundlePathKey]); err != nil {
```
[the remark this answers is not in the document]
sboeuf: Oh yes good point, I have to move this slightly further in the code.

cli/hook.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	stateJSON, err := json.Marshal(state)
+	if err != nil {
+		return err
+	}
+
+	var stdout, stderr bytes.Buffer
+	cmd := &exec.Cmd{
+		Path:   hook.Path,
```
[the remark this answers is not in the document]
sboeuf: I am trying not to introduce too much new code, simply doing a clean move from virtcontainers to the CLI. I think your proposal should go through a follow up PR.

virtcontainers/network.go
```diff
@@ -1277,6 +1278,93 @@ func createVirtualNetworkEndpoint(idx int, ifName string, interworkingModel NetI
 	return endpoint, nil
 }
 
+func generateInterfacesAndRoutes(networkNS NetworkNamespace) ([]*grpc.Interface, []*grpc.Route, error) {
```
amshinde: This code is very Kata agent  specific since it deals with grpc and not applicable for other agents. So I think it should be it in kata agent.

cli/hook.go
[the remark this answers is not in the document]
amshinde: @sboeuf Please open an issue to track this.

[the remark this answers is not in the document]
sboeuf: https://github.com/kata-containers/runtime/issues/631

virtcontainers/network.go
sboeuf: I understand, but if you look at the Kata API, we're using both `grpc.Interface` and `grpc.Route` structures as the generic structures to add interfaces or routes.
I think moving this function out of `kata_agent.go` is the first step, and the second logical step would be to move those very generic structure definitions out of `kata-containers/agent/protocols/grpc`. It'd might be better to have them in a separate package so that they don't get associated with the kata-agent.

sboeuf:
Rebased! I'll address the comments now.

virtcontainers/noop_resource_storage.go
sboeuf: addressed in other PR (already merged)

cli/create.go
sboeuf: done

cli/start.go
sboeuf: done

virtcontainers/default_network.go
sboeuf: done

cli/hook.go
sboeuf: done

sboeuf:
@jodh-intel all your comments addressed. The rebase on top of the opentracing was trickier than I thought it'd be :)

sboeuf:
@bergwolf I have addressed your comments, PTAL

cli/create.go
```diff
@@ -266,6 +266,21 @@ func createSandbox(ctx context.Context, ociSpec oci.CompatOCISpec, runtimeConfig
...
+	// Important to create the network namespace before the sandbox is
+	// created, because it is not responsible for the creation of the
+	// netns if it does not exist.
+	if err := setupNetworkNamespace(&sandboxConfig.NetworkConfig); err != nil {
+		return vc.Process{}, err
+	}
+
+	// Run pre-start OCI hooks.
+	err = enterNetNS(sandboxConfig.NetworkConfig.NetNSPath, func() error {
+		return preStartHooks(ctx, ociSpec, containerID, bundlePath)
```
bergwolf: While it can work, I wonder why we are not calling it in `start.go`? After all, they are `preStart` hooks not `preCreate` hooks.

amshinde: @bergwolf This is what the OCI spec calls them. We have simply chosen to call it in create as we need the network hook to be called. @sboeuf and I had a discussion about moving to a hotplug model, but lets address that in a separate PR.

I also believe we need to evaluate if there are certain hooks that need to be run inside the VM.

virtcontainers/network.go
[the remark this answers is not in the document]
amshinde: @egernst I think naming the network to Default is a bit awkward, lets rework this in a separate PR if required.

egernst:
change is fine with me, I'll want to get @bergwolf 's ACK.  I'd prefer a less awkward naming scheme (defaultNetwork when we only have one network is a bit odd), but don't want to delay this too much longer when I'm not providing a better alternative...

cli/create.go
jodh-intel: I think it's worth pointing out that both the current code, and this PR are somewhat abusing the standard:

- https://github.com/opencontainers/runtime-spec/blob/master/config.md#prestart

amshinde:
@bergwolf Can you take another look at this?  Are you ok with merging this as is and address the hooks issue in a separate PR. Else we can wait for  @sboeuf when he is back next week.

cli/create.go
bergwolf: @amshinde @sboeuf I think we will need to move it to start.go otherwise vm factory will still be lacking network configurations. But I agree with you that it can be handled in a following up PR.

bergwolf:
Well, it looks like `pullapprove` really dislikes my approval and just keeps ignoring it. I'm going to ignore it as well, again.

Issue #158, opened 2018-04-02
Support Memory constrains


egernst:
_From @devimc on January 22, 2018 15:1_

- [x] hot add/remove memory
    issue https://github.com/kata-containers/runtime/issues/625
    pull request https://github.com/kata-containers/runtime/pull/624
- [x] apply memory cgroups in the guest
    issue https://github.com/kata-containers/runtime/issues/653
    pull request https://github.com/kata-containers/runtime/pull/654
- [x] track how much memory is being used for all containers and decide if a new chunk of memory must be added or not
https://github.com/kata-containers/runtime/pull/624
- [x] check if the maximum number of memory slots impacts container memory footprint
https://github.com/kata-containers/runtime/pull/580#issuecomment-418219019

Notes:
* on x86 only chucks/multiples of 128mb can be added
* memory hot remove is not possible

_Copied from original issue: containers/virtcontainers#578_

jcvenegas:
WIP: https://github.com/jcvenegas/runtime-1/tree/memory-hotplug

linzichang:
@jcvenegas @bergwolf Hi, I have seen cpu-hotplug support for "docker update" is done.  About "memory-hotplug", I find there is already some implementations like "hotplugMemory", but they haven't combined with kata-runtime update yet. What is the plan for this?

PR #624, opened 2018-08-23
virtcontainers: hotplug memory with kata-runtime update command.

### LINKED ISSUE #625: virtcontainers: hotplug memory with kata-runtime update command
We don't support using kata-runtime update command to hotplug memory to a vm, so it's needed to 
connect kata-runtime update command with hypervisor memory hotplug feature.

clarecch:
Add support for using update command to hotplug memory to vm.
Connect kata-runtime update interface with hypervisor memory hotplug
feature.

Fixes #625

Signed-off-by: Clare Chen <clare.chenhui@huawei.com>


virtcontainers/qemu.go
```diff
@@ -1010,6 +1011,12 @@ func (q *qemu) hotplugMemory(memDev *memoryDevice, op operation) error {
 			memDev.sizeMB, currentMemory, q.config.DefaultMemSz)
 	}
 
+	//memory slot 0 is occupied
```
jodh-intel: I don't understand this comment - can you provide more detail?

virtcontainers/container.go
```diff
@@ -1267,6 +1267,56 @@ func (c *Container) updateResources(oldResources, newResources ContainerResource
...
 		newResources.VCPUs = oldVCPUs - vcpusRemoved
 	}
+	return nil
+}
+
+func (c *Container) updateMemoryResources(oldResources, newResources ContainerResources) error {
+	oldMem := oldResources.Mem
+	newMem := newResources.Mem
+
+	if newMem == 0 || oldMem == newMem {
```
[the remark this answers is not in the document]
clarecch: It wouldn't be en error.  The newMem comes from resources.Memory.Limit. When we use update command to update vcpu without update memory, for example "docker update --cpus 5 [container-id]", the value newMem will be zero, which just means the user don't want to update memory.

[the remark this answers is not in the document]
jodh-intel: Ah - right. I find combining the handling of two different resources rather confusing tbh ;)

virtcontainers/qemu.go
grahamwhaley: Is mem0 maybe either the nvdimm or an initial fixed block that can never be plug/unplugged - @devimc , I suspect you may remember?

devimc: yep, mem0 is occupied by other components like nvdimm or vmfactory share memory

virtcontainers/container.go
```diff
@@ -972,6 +972,7 @@ func (c *Container) update(resources specs.LinuxResources) error {
 
 	newResources := ContainerResources{
 		VCPUs: uint32(utils.ConstraintsToVCPUs(*resources.CPU.Quota, *resources.CPU.Period)),
+		Mem:   uint32(*resources.Memory.Limit >> 20),
```
devimc: please add a comment explaining this

virtcontainers/qemu.go
grahamwhaley: Great - @clarecch , can you add that to the comment please then :-) thx!

virtcontainers/container.go
```diff
@@ -1267,6 +1267,56 @@ func (c *Container) updateResources(oldResources, newResources ContainerResource
...
+			"old-mem": fmt.Sprintf("%dMB", oldMem),
+			"new-mem": fmt.Sprintf("%dMB", newMem),
+		}).Debug("the actual number of Mem will not be modified")
+		return nil
+	}
+
+	if oldMem < newMem {
+		// hot add memory
+		addMemMB := newMem - oldMem
+		// hot add memory must be aligned to 128MB
```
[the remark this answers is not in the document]
devimc: this is a limitation in the kernel https://elixir.bootlin.com/linux/v4.18.4/source/include/linux/mmzone.h#L1082 (PAGE_SECTION_MASK), we can deal with this limitation but we need support for memory align in QEMU, I think those patches haven't been merged yet (not sure).

virtcontainers/qemu.go
devimc: the problem I see with this implementation is that you can update the memory just once since currently we only have 2 slots, @clarecch have you consider to add an option in the configuration file to make the number of slots configurable? , what is the impact (boot time, memory footprint) of each extra slot? ,  In the case of PODs a memory constraint have to be applied, please take a look to this PR https://github.com/kata-containers/runtime/issues/158

bergwolf: For initrd w/o vm factory case, slot 0 is available for hotplug. You can proceed with current code for now. But as @WeiZhang555 mentioned, we should track the slot id somewhere so that we actually know which slot are used and which are not. Then we can drop the hardcoded starting point of slot number.

virtcontainers/container.go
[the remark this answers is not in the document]
clarecch: Yes, I don't find any limitation of memory section align in QEMU, so I do judgment here. I will define a variable of 128 and create a function to encapsulate the memory checking, thanks.

virtcontainers/qemu.go
clarecch: @bergwolf I am using QemuState.Slots, which is stored in hypervisor.json to store the slot id if it has changed now. I think if we have use any mempry slot we should update QemuState.Slots, and then store it to hypervisor.json. I will take a look at components  which will occupied the mem0

clarecch: @devimc Yes, I also think it's needed to add an option in the configuration file to make the number of slots configurable. But maybe create a new issue to do this as well as check if the maximum number of memory slots impacts container memory footprint?

virtcontainers/container.go
clarecch: Please tell me what do I need to explain. Why we need Mem field or why we have 20 right shift?  ^^

virtcontainers/container.go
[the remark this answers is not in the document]
clarecch: @devimc @jodh-intel The reason why hot add memory must be aligned to 128MB is that guestos will check whether hotplug memory range is aligned with memory section when we hot add memory to the guest, with reference to https://elixir.bootlin.com/linux/v4.18.4/source/mm/memory_hotplug.c#L1075.
On x86_64 the size of section is 128MB, but on other platform section may not be 128MB. If we want to keep compatible with other platform, we need to read the section size of guest rather than use 128MB.

virtcontainers/container.go
jodh-intel: It is a little unclear as `resources.Memory.Limit` is an `int64` byte value. It would help if `ContainerResources.Mem` was called `ContainerResources.MemMb` imho.

clarecch: Yes, you are rignt. I will change the name.

virtcontainers/qemu.go
clarecch: @devimc @bergwolf How about use qmp command to track slot number at present, and then initialize the QemuState.Slots field before hotplug memory?
Just like this.
(QEMU) query-memory-devices
{
    "return": [
        {
            "data": {
                "slot": 0,
                "node": 0,
                "addr": 4294967296,
                "memdev": "/objects/mem0",
                "id": "nv0",
                "hotpluggable": true,
                "hotplugged": false,
                "size": 536870912
            },
            "type": "dimm"
        }
    ]
}

virtcontainers/container.go
devimc: @clarecch 
> why we have 20 right shift?

virtcontainers/qemu.go
devimc: @clarecch sounds good, in that way we can see what slots are available

bergwolf: @clarecch Yes, indeed. We can even remove the slot assignment from kata. We only need to make sure every time we use different memory device id to hotplug it. QEMU always plugs a new dimm in the first available slot. We either query qemu for available slots, or rely on qmp hotplug failure to detect unavailable dimm slots.

virtcontainers/container.go
```diff
@@ -1267,6 +1267,56 @@ func (c *Container) updateResources(oldResources, newResources ContainerResource
...
+		if err != nil {
+			return err
+		}
+		newResources.Mem = newMem
+		if err := c.sandbox.agent.onlineCPUMem(0); err != nil {
+			return err
+		}
+	} else {
+		// hot remove memory
+		return fmt.Errorf("can't hot remove memory, we don't support this feature yet")
```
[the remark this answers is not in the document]
ydjainopensource: We do have memory hotplug on Z :smile:  How about this for the error message : ` Memory hotplug unsupported` feels much more concise.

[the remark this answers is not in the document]
clarecch: Do you mean `Memory hot remove unsupported` ? I think it would be a good choice.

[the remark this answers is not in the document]
ydjainopensource: No I think hotplug would be the better choice here. I say this because, almost all of IBM Z docs use plug/unplug and even ovirt calls it [hotplug](https://ovirt.org/develop/release-management/features/virt/hot-plug-memory/). Looks like Vmware is the only one calling it hotadd. But I am okay with hot add if the community agrees.

virtcontainers/container.go
```diff
@@ -1267,6 +1267,56 @@ func (c *Container) updateResources(oldResources, newResources ContainerResource
...
+		}).Debug("the actual number of Mem will not be modified")
+		return nil
+	}
+
+	if oldMem < newMem {
+		// hot add memory
+		addMemMB := newMem - oldMem
+		// hot add memory must be aligned to 128MB
+		if addMemMB%128 != 0 {
+			return fmt.Errorf("can not hot add memory which isn't aligned to 128MB")
```
ydjainopensource: Can you change the error message to something like `Memory must be aligned to 128 MB for hotplug` ? or `hot add` ?

clarecch: ok， Updated

virtcontainers/container.go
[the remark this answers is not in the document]
nitkon: @clarecch @jodh-intel : You are right. For ppc64le, the hot plugged memory must be aligned to 256MB.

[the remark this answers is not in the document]
clarecch: @nitkon @devimc @jodh-intel We have some discuss about hot add memory align in https://github.com/kata-containers/runtime/pull/580

virtcontainers/container.go
jodh-intel: /cc @miaoyq regarding the int64 / int32 `Mem` size issue.

virtcontainers/container.go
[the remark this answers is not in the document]
jodh-intel: @clarecch - rather than https://github.com/kata-containers/agent/issues/354, is there are reason that we can't just update the agent to align to the correct boundary when it's given a memory adjustment by the runtime? That way we don't need a new gRPC call and all the logic is encapsulated in the agent.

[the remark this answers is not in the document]
clarecch: @jodh-intel The qmp command that hotplug memory needs a memory size aligned to correct boundary, so runtime should know the memory block, which gets from the agent.

bergwolf:
@clarecch Could you proceed this w/o the alignment query from guest? It can be fixed in a follow up PR which would depend on querying memory alignment from the guest. That will speed up the development of this feature.

clarecch:
@bergwolf 
> Could you proceed this w/o the alignment query from guest

Do you mean the memory section alignment? I have open a issue for that, which is going to implement querying memory block size from the agent https://github.com/kata-containers/agent/issues/354. Is it exactly what you mean?

clarecch:
The jenkins-ci-centos-7-4 failed with:
```
Summarizing 3 Failures:

[Fail] CPUs and CPU set updating cpus and cpuset of a running container [It] should have the right number of vCPUs 
/tmp/jenkins/workspace/kata-containers-runtime-centos-7-4-PR/go/src/github.com/kata-containers/tests/integration/docker/cpu_test.go:432

[Fail] memory constraints run container and update its memory constraints [It] should have applied the constraints 
/tmp/jenkins/workspace/kata-containers-runtime-centos-7-4-PR/go/src/github.com/kata-containers/tests/integration/docker/mem_test.go:157

[Fail] docker volume remove bind-mount source before container exits [It] should exit cleanly without leaking process 
/tmp/jenkins/workspace/kata-containers-runtime-centos-7-4-PR/go/src/github.com/kata-containers/tests/integration/docker/volume_test.go:175
```
I know the memory constraints failed is because of memory hot remove. But I can't figure out why the first and the third testcase failed. Pls tell me!

devimc:
@clarecch 

```
Stderr: Error response from daemon: Cannot update container 2b4090f7c82e239c210f8d8efd5b7f0615386459f934276a53e726e4e66f3114: /usr/local/bin/kata-runtime did not terminate sucessfully: rpc error: code = Unknown desc = failed to write 0-1 to cpuset.cpus: write /sys/fs/cgroup/cpuset/docker/2b4090f7c82e239c210f8d8efd5b7f0615386459f934276a53e726e4e66f3114/cpuset.cpus: invalid argument
```

linzichang:
@devimc 
```
Stderr: Error response from daemon: Cannot update container 2b4090f7c82e239c210f8d8efd5b7f0615386459f934276a53e726e4e66f3114: /usr/local/bin/kata-runtime did not terminate sucessfully: rpc error: code = Unknown desc = failed to write 0-1 to cpuset.cpus: write /sys/fs/cgroup/cpuset/docker/2b4090f7c82e239c210f8d8efd5b7f0615386459f934276a53e726e4e66f3114/cpuset.cpus: invalid argument
```
This failed because now we have both udev and onlinecpumem RPC working. 
1. udev has some probability to make some of CPUs online. Making onlineCPUResources return error. See https://github.com/kata-containers/agent/blob/master/grpc.go#L145
2. Then `onlineCPUMem` will return. See https://github.com/kata-containers/agent/blob/master/grpc.go#L206
3. And can't update the parent cgroup cpuset. See https://github.com/kata-containers/agent/blob/master/grpc.go#L235

bergwolf:
@linzichang Since we handle device online ourselves, I don't think we should let udev service run in the guest. That means we should disable udev service if systemd is the init process when building the rootfs image.

Another approach is to simply remove the onlineCPUMem call, and add cpu/memory requirements in the `CreateContainer` grpc. Then the agent can manage cpu/memory assignment and wait for requested amount of resources (if any is specified) to be ready with a timeout.

@sboeuf @devimc @WeiZhang555 wdyt?

clarecch:
Does `the hot add memory should be aligned to memory section size` and `hot remove memory should return fail`  break the OCI compatibility? It seems that the CI failed due to CRI-O compatibility. If so, maybe we should find some way to keep OCI compatible.

clarecch:
I think in order to keep the OCI compatible we could:
1. For `hot remove memory should return fail`, we don't return fail, just do nothing and update the container memory cgroup.
2. For `hot add memory should be aligned to memory section size`, it's much more complicated. We can premise the `vm_memory =  vm_reserved_memory + all_container_memory + vm_residual_memory`.

 - `vm_memory` is the total memory of vm.
 - `vm_reserved_memory` is the minimal memory vm should use to keep running. Let's make it to be 1G size first.
 - `all_container_memory` is the sum of all container memory limit.
 - `vm_residual_memory` is the residual memory of vm except for `vm_reserved_memory` and `all_container_memory`.

Let's explain it with a case:
When we start katacontainer with `docker run -ti ubuntu`, we will have the `vm_memory`=2G, `vm_reserved_memory`=1G, `all_container_memory`=0（container has no memory limit）， so `vm_residual_memory`=2G-1G=1G. We can do many things with `vm_residual_memory`.

- When we update a memory with adding 300M or create a container with 300M memory, we don't have to hotplug memory with qmp(wich will keep us from making memory section align judgement), just update the memory limit of container, so `vm_residual_memory` will decrease 300M, `all_container_memory` will add 300M, others remain no change. 

- When we delete a container with a memory limit of 300M, `all_container_memory` will decrease 300M, `vm_residual_memory` will add 300M, which can slove vm memory recycling problem of https://github.com/kata-containers/runtime/pull/580#issuecomment-417006097 @devimc 

- When `vm_residual_memory` draines, for example remains 100M. If we want to add a new container with 300M or update a exist container with 300M, the `vm_residual_memory` is not enough. We can hotplug memory with 2*128MB(two memory section aligned memory), so `vm_memory` will increase 256MB, `vm_residual_memory` will increase  256MB, which will have 356MB in total to create a container or update container memory.

In this way we can solve hotplug memory align problem and vm memory recycling problem etc. WDYT

This idea is originally from @linzichang 

If CI is not failed for OCI compatibility, we can implement this feature in https://github.com/kata-containers/runtime/pull/580. Let's move on this PR first.

clarecch:
@devimc I think your idea is reasonable for memory slot, but not properly for the chucks of memory to hotplug. With the help of `vm_residual_memory` I had mentioned, users can update any size of memory as they want. Kata-runtime will do the job to make the hotplug memory to be aligned with memory section size. So the chucks of memory should not be configurable.

clarecch:
I made some change to make kata-runtime update interface OCI compatible. It seems the CI has no problem now. Let's push on this PR. PTAL!

clarecch:
@bergwolf @jodh-intel PTAL

caoruidong:
Please fix the CI error. Add a function to gRPCProxy.

virtcontainers/qemu.go
```diff
@@ -1047,16 +1052,18 @@ func (q *qemu) hotplugMemory(memDev *memoryDevice, op operation) error {
 			memDev.sizeMB, currentMemory, q.config.MemorySize)
 	}
 
+	memoryDevices, err := q.qmpMonitorCh.qmp.ExecQueryMemoryDevices(q.qmpMonitorCh.ctx)
+	if err != nil {
+		return fmt.Errorf("failed to query memory devices: %v", err)
+	}
+
+	memDev.slot = memoryDevices[len(memoryDevices)-1].Data.Slot + 1
```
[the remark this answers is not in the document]
bergwolf: I see, `query-memory-devices` only returns slots that are in use. Then it seems your code would crash here by trying to access negative slice index if there is no hotplugged memory.

```
[macbeth@qmp]$sudo ./qmp-shell -p /run/vc/vm/1deba5c96d9f2490cf7eafb03ba8219a442248d0f716b5954957103e7394cb75/qmp.sock
Welcome to the QMP low-level shell!
Connected to QEMU 2.11.1

(QEMU) query-memory-devices
{
    "return": []
}
```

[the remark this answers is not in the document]
clarecch: Oh, I will fix it.

linzichang:
@bergwolf In https://github.com/kata-containers/runtime/blob/a5f05bf3e131722b3c2df1dea88e61700a801136/virtcontainers/kata_agent.go#L1259 , why test `err != golangGrpc.ErrClientConnClosing`? I think k.client.Close() only return `nil` or `golangGrpc.ErrClientConnClosing`. Will this test make the code always success?

bergwolf:
@linzichang 
`golangGrpc.ErrClientConnClosing` means the client is in closing state and we do not care about that error so it is ignored.

clarecch:
It ready to read memory block size of guest from agent and check hot add memory to be aligned with memory block size of guest.

virtcontainers/qemu.go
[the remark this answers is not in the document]
clarecch: updated， PTAL

clarecch:
I think the CI has no problem now, PTAL

clarecch:
The CI is green now.

virtcontainers/container.go
```diff
@@ -1296,8 +1296,90 @@ func (c *Container) updateResources(oldResources, newResources ContainerResource
...
+		return fmt.Errorf("Cound not get the state of sandbox")
+	}
+
+	memorySectionSizeMB := state.GuestMemoryBlockSizeMB
+	if memorySectionSizeMB == 0 {
+		return nil
+	}
+
+	// TODO: hot add memory aligned to memory section should be more properly. See https://github.com/kata-containers/runtime/pull/624#issuecomment-419656853
+	*mem = uint32(math.Ceil(float64(*mem)/float64(memorySectionSizeMB))) * memorySectionSizeMB
```
caoruidong: I wonder if it is proper to change mem in this function because it says "check" . You can add newmem as a return value or rename this function.

clarecch: Ok, I think add newmem as a return value will be more proper.

clarecch:
The CI is green and conflict is solved. PTAL

virtcontainers/container.go
clarecch: updated

virtcontainers/container.go
```diff
@@ -1296,8 +1296,89 @@ func (c *Container) updateResources(oldResources, newResources ContainerResource
 		// recalculate the actual number of vCPUs if a different number of vCPUs was removed
 		newResources.VCPUs = oldVCPUs - vcpusRemoved
 	}
+	return nil
+}
+
+func (c *Container) memHotplugValid(mem uint32) (uint32, error) {
+	state, err := c.sandbox.storage.fetchSandboxState(c.sandbox.id)
```
bergwolf: sandbox state is already saved in the sandbox structure. Please use that one if available so we can possibly avoid disk IO here.

clarecch: Oh, yes you are right. I will fix it.

clarecch:
The CI is green and conflict is solved. PTAL @jodh-intel @devimc @bergwolf

jodh-intel:
lgtm

clarecch:
@bergwolf of course

caoruidong:
Codecov is stuck. Merge this because this PR updates some tests. If coverage drops, @clarecch please add more tests, thank you!

PR #639, opened 2018-08-24
config: Detect if VM memory smaller than image

### LINKED ISSUE #636: Add sanity check for image size and VM memory size
Add a basic check to catch the scenario where the default memory for the VM has been set too low (smaller than the image size).

This will allow the runtime to display a helpful message rather than just falling over - see #626.

jodh-intel:
Add a heuristic to ensure the amount of memory allocated to the
hypervisor is bigger than the size of the image.

This catches simple configuration issues where `default_memory=` is set
to a smaller value than the size of either the `image=` or `initrd=`
files.

If the configured image type is `initrd`, fail but only warn in the
logs for `image` as although it seems a highly unlikely scenario, it is
permitted.

Update tests to ensure that created resources have `>0` bytes.

Fixes #636.

### DIFF
--- a/cli/config.go
+++ b/cli/config.go
@@ -577,7 +577,77 @@ func loadConfiguration(configPath string, ignoreLogging bool) (resolvedConfigPat
 		config.ProxyConfig = vc.ProxyConfig{}
 	}
 
+	if err := checkHypervisorConfig(config.HypervisorConfig); err != nil {
+		return "", config, err
+	}
+
 	return resolved, config, nil
+}
+
+// checkHypervisorConfig performs basic "sanity checks" on the hypervisor
+// config.
+func checkHypervisorConfig(config vc.HypervisorConfig) error {
+	type image struct {
+		path   string
+		initrd bool
+	}
+
+	images := []image{
+		{
+			path:   config.ImagePath,
+			initrd: false,
+		},
+		{
+			path:   config.InitrdPath,
+			initrd: true,
+		},
+	}
+
+	memSizeMB := int64(config.DefaultMemSz)
+
+	if memSizeMB == 0 {
+		return errors.New("VM memory cannot be zero")
+	}
+
+	mb := int64(1024 * 1024)
+
+	for _, image := range images {
+		if image.path == "" {
+			continue
+		}
+
+		imageSizeBytes, err := fileSize(image.path)
+		if err != nil {
+			return err
+		}
+
+		if imageSizeBytes == 0 {
+			return fmt.Errorf("image %q is empty", image.path)
+		}
+
+		if imageSizeBytes > mb {
+			imageSizeMB := imageSizeBytes / mb
+
+			msg := fmt.Sprintf("VM memory (%dMB) smaller than image %q size (%dMB)",
+				memSizeMB, image.path, imageSizeMB)
+
+			if imageSizeMB >= memSizeMB {
+				if image.initrd {
+					// Initrd's need to be fully read into memory
+					return errors.New(msg)
+				}
+
+				// Images do not need to be fully read
+				// into memory, but it would be highly
+				// unusual to have an image larger
+				// than the amount of memory assigned
+				// to the VM.
+				kataLog.Warn(msg)
+			}
+		}
+	}
+
+	return nil
 }
 
 // getDefaultConfigFilePaths returns a list of paths that will be
--- a/cli/config_test.go
+++ b/cli/config_test.go
@@ -6,6 +6,7 @@
 package main
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -119,8 +120,8 @@ func createAllRuntimeConfigFiles(dir, hypervisor string) (config testRuntimeConf
 	files := []string{hypervisorPath, kernelPath, imagePath, shimPath, proxyPath}
 
 	for _, file := range files {
-		// create the resource
-		err = createEmptyFile(file)
+		// create the resource (which must be >0 bytes)
+		err := writeFile(file, "foo", testFileMode)
 		if err != nil {
 			return config, err
 		}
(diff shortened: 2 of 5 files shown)


jodh-intel:
An example of running with an incorrect config:

```
$ docker run --runtime kata-runtime -ti busybox true
docker: Error response from daemon: OCI runtime create failed: VM memory (8MB) smaller than image "/usr/share/kata-containers/kata-containers-initrd_clearlinux_1.2.0_agent_fcfa054a757.initrd" size (13MB): unknown.
```

devimc:
@jodh-intel  CI is not happy

```
--- FAIL: TestConfigLoadConfiguration (0.02s)
	assertions.go:239: 
                          
	Error Trace:	config_test.go:230
		
	            	
			config_test.go:258
		
	Error:      	Received unexpected error:
		
	            	image "/tmp/kata-runtime-125062757/load-config-998631040/test/image" is empty
	assertions.go:239: 
                          
	Error Trace:	config_test.go:236
```

jodh-intel:
@devimc - branch updated.

jodh-intel:
Ugh, that looks bad:

```
Cloning into 'tests'...
fatal: Couldn't find remote ref pull/639/head
```

jodh-intel:
That was an F27 CI failure btw. Prodded failing CI's...

devimc:
```
--- FAIL: TestMinimalRuntimeConfig (0.00s)
	config_test.go:512: no such file or directory
```

cli/config.go
```diff
@@ -577,9 +577,51 @@ func loadConfiguration(configPath string, ignoreLogging bool) (resolvedConfigPat
...
+		}
+
+		if imageSizeBytes == 0 {
+			return fmt.Errorf("image %q is empty", image)
+		}
+
+		if imageSizeBytes > mb {
+			imageSizeMB := imageSizeBytes / mb
+
+			if imageSizeMB >= memSizeMB {
```
WeiZhang555: I didn't test more on this, can this work if `(imageSizeMB+1) == memSizeMB`? Should we leave some MB for applications?

for example: 

`if imageSizeMB >= memSizeMB + 64MB`

jodh-intel: I did think about that but thought I'd leave this PR as a very basic heuristic. I'm not sure how we can reliably determine "how much more" memory we need so kept it simple for this first PR.

jodh-intel:
I think I've fixed the tests now - see the new commit. I did run tests locally, but if you look at that commit you'll understand why/how they passed! :smile:

cli/config.go
jodh-intel: Any thoughts on this @grahamwhaley, @bergwolf, @sboeuf?

bergwolf: I don't think we can guess how much memory a container app would use. In stead, we should try to catch OOM events from guest.

cli/config.go
```diff
@@ -577,9 +577,51 @@ func loadConfiguration(configPath string, ignoreLogging bool) (resolvedConfigPat
...
+	if err := checkHypervisorConfig(config.HypervisorConfig); err != nil {
+		return "", config, err
+	}
+
 	return resolved, config, nil
 }
 
+// checkHypervisorConfig performs basic "sanity checks" on the hypervisor
+// config.
+func checkHypervisorConfig(config vc.HypervisorConfig) error {
```
bergwolf: While initramfs needs to be read all into the guest memory, the disk image does not need to be. So the sanity check can only apply to the `initrd` case.

And it can only be a best effort check, because we do not know how much memory the uncompressed initramfs and guest kernel will actually use.

cli/config.go
jodh-intel: `imageSize` is the size in bytes of the configured image type (`kata-containers.img` or `kata-containers-initrd.img`).

And yes, as already mentioned, this is a very simple check - we need something better going forward but this will catch some simple misconfiguration cases.

grahamwhaley: So, yes, I think as a simple premise to catch the basic simple cases, this will work:
> If your VM RAM is < your kata.img size, you will always fail to load the VM

cli/config.go
jodh-intel: Right. I can change the PR to only check the image, but I wonder if we want to just comment the initrd behaviour and still perform the check. Wouldn't it still be pretty unusual / crazy to configure the VM to have less memory than the size of the initrd (particularly given initrd's are tiny? :)

bergwolf: I'm fine with either way you choose :)

cli/config.go
jshachm: As what I test and find..  ramdisk size might take a little trick here. Ramdisk size is half of vm memory and as we tested before. If   initrd not gizpped before take 100M, U at least need 200M vm memory. So if we just judge the imgSize, we will get an error of `Ramdisk doesn't exsit ` or other boot error.

@jodh-intel @grahamwhaley @bergwolf 
Tell me sth deeper if you guys have any ideas on this ~

bergwolf: @jshachm Except for the space consumed by unpressed initramfs, you also need memory for the guest kernel, and all the stack/heap needed to run them. So I don't think you can simply guess a minimal guest memory size and expect all the binaries to run w/o issues.

@grahamwhaley 
> If your VM RAM is < your kata.img size, you will always fail to load the VM

This only applies to initramfs. For disk image, it is not fully loaded into the memory. Besides we plug it as a nvdimm device so if dax is on, it has very small memory footprint (for the file system metadata) in the guest.

jodh-intel:
Hi @bergwolf - branch updated. It now fails for initrd but only warns for a configured image.

jodh-intel:
Branch updated with crude test rework to lower cyclomatic complexity.

bergwolf:
@jodh-intel That looks like a OOM situation. Is the failing test using small guest memory?

jodh-intel:
Kicked density CI (network timeout).

jodh-intel:

jodh-intel:
Ping @bergwolf, @egernst, @devimc, @grahamwhaley.

jodh-intel:
Ping @amshinde.

jodh-intel:
Hi @grahamwhaley - since you were part of this conversation, could you tal please? :)

cli/config.go
```diff
@@ -577,9 +577,79 @@ func loadConfiguration(configPath string, ignoreLogging bool) (resolvedConfigPat
...
+				if image.initrd {
+					// Initrd's need to be fully read into memory
+					return errors.New(msg)
+				}
+
+				// Images do not need to be fully read
+				// into memory, but it would be highly
+				// unusual to have an image larger
+				// than the amount of memory assigned
+				// to the VM.
```
bergwolf: I don't think the comment is accurate. It's actually quite common to have a rootfs image larger than the guest memory.

jodh-intel: The default memory is 2G and most images (not initrd's) are much smaller than that. However, image sizes can vary a lot I guess and users can modify `default_memory` [1], so how about I just remove the comment and the log warning?

---

[1] - @grahamwhaley - OOI, what's the minimum value we can use?

jcvenegas:
lgtm

PR #655, opened 2018-08-28
Add ref counter for devices

### LINKED ISSUE #635: device will be hotplugged twice when multiple volume with same source specified
A device shouldn't be plugged to VM twice, this can lead to data corruption with multiple writers inside VM.

Instead we should add a reference count in device manager, one device should be plugged once, and reference to many writers with "binding mount" inside Guest, this can avoid data corruption.

I'm doing this work now, adding this issue to track.

WeiZhang555:
devices: add reference count for devices.

Fixes #635

Remove `Hotplugged bool` field from device and add two new fields
instead:
* `RefCount`: how many references to this device. One device can be
referenced(`NewDevice()`) many times by same/different container(s),
two devices are regarded identical if they have same hostPath
* `AttachCount`: how many times this device has been attached. A device
can only be hotplugged once to the qemu, every new Attach command will
add the AttachCount, and real `Detach` will be done only when
`AttachCount == 0`


virtcontainers/device/drivers/generic.go
```diff
@@ -27,38 +30,49 @@ func NewGenericDevice(devInfo *config.DeviceInfo) *GenericDevice {
...
-	return config.DeviceGeneric
+// HostPath return host path of device
+func (device *GenericDevice) GetHostPath() string {
+	return device.DeviceInfo.HostPath
 }
 
-// GetDeviceInfo returns device information used for creating
-func (device *GenericDevice) GetDeviceInfo() interface{} {
-	return device.DeviceInfo
+// Reference adds one reference to device
```
jodh-intel: I'd remove these comments as you've already handled them in the `Device` interface comments.

virtcontainers/device/api/interface.go
```diff
@@ -49,13 +49,20 @@ type Device interface {
...
 	// Caller could cast the return value to device specific struct
 	// e.g. Block device returns *config.BlockDrive and
 	// vfio device returns []*config.VFIODev
 	GetDeviceInfo() interface{}
 	// IsAttached checks if the device is attached
-	IsAttached() bool
+	GetAttachCount() int
+
+	// Reference adds one reference to device
+	Reference() int
```
jodh-intel: Please can you document here what the return values are for these two new functions?

jodh-intel:
Oh - also Travis is complaining as there are some lint issues to resolve.

virtcontainers/device/api/interface.go
WeiZhang555: Ah, good point!

virtcontainers/device/drivers/generic.go
WeiZhang555: Let's me check whether lint is OK without this comments

WeiZhang555:
There're some issues not fixed, mark this as Work In Progress before fixing it

virtcontainers/device/drivers/generic.go
WeiZhang555: Just tried. Golint won't allow this, so have to keep a duplicate comment here.

virtcontainers/device/drivers/block.go
```diff
@@ -101,26 +110,19 @@ func (device *BlockDevice) Detach(devReceiver api.DeviceReceiver) error {
...
-	return device.ID
-}
-
 // GetDeviceInfo returns device information used for creating
 func (device *BlockDevice) GetDeviceInfo() interface{} {
 	return device.BlockDrive
 }
+
+// It should implement GetAttachCount() and DeviceID() as api.Device implementation
+// here it shares function from *GenericDevice so we don't need duplicate codes
```
bergwolf: What does this comment address to?

virtcontainers/device/drivers/generic.go
```diff
@@ -27,38 +32,98 @@ func NewGenericDevice(devInfo *config.DeviceInfo) *GenericDevice {
...
+	}
+	return device.RefCount
+}
+
+// bumpAttachCount is used to add/minus attach count for a device
+// * attach bool: true means attach, false means detach
+// return values:
+// * skip bool: no need to do real attach/detach, skip following actions.
+// * err error: error while do attach count bump
+func (device *GenericDevice) bumpAttachCount(attach bool) (skip bool, err error) {
```
bergwolf: Could you please add a lock here (to protect AttachCount calculation) so that we allow parallel hotplug of these devices?

virtcontainers/device/drivers/generic.go
```diff
@@ -27,38 +32,98 @@ func NewGenericDevice(devInfo *config.DeviceInfo) *GenericDevice {
...
+// GetMajorMinor returns device major and minor numbers
+func (device *GenericDevice) GetMajorMinor() (int64, int64) {
+	return device.DeviceInfo.Major, device.DeviceInfo.Minor
 }
 
-// GetDeviceInfo returns device information used for creating
-func (device *GenericDevice) GetDeviceInfo() interface{} {
-	return device.DeviceInfo
+// Reference adds one reference to device
+func (device *GenericDevice) Reference() uint {
```
bergwolf: @WeiZhang555 I'm wondering why you need both `RefCount` and `AttachCount`? Can they be combined into one reference count on the device?

virtcontainers/device/drivers/block.go
WeiZhang555: Just tell the developers that as a `api.Device` implementation, it should define two function `GetAttachCount` and `DeviceID` but in this file there're no these function. Because it shares function s from embedded `*GenericDevice`

virtcontainers/device/drivers/generic.go
WeiZhang555: There's a `deviceManager` in "virtcontainers/device/manager/manager.go", it is designed to do Attach/Detach via `deviceManager.AttachDevice/DetachDevice` which has a global lock.

Do you think we need a per-device lock here? If everyone use the device manager as global entry, it should be fine to do parallel operations.

jodh-intel: I wonder if we could use https://golang.org/pkg/sync/atomic instead of locks?

virtcontainers/device/drivers/generic.go
WeiZhang555: Hmm, not sure, maybe I can remove the RefCount

Because we have 4 steps in container creation process, `NewDevice`, `AttachDevice`, `DetachDevice`, `RemoveDevice`,   we do the real `RemoveDevice` when `refCount==0`, and we do real `DetachDevice` when `AttachCount==0`.

Maybe I can remove the `refCount`, and only allow device to be deleted after `AttachCount==0`.
Let me think about it.

WeiZhang555: Removing `RefCount` can break one case:

many containers reference one device, but no one has done the `AttachDevice()`, then one containers calls `RemoveDevice`, the shared device will be removed from list but other containers could still want to use it(plug it later). This may happen under parallel operations.

So I think it's better to keep both `RefCount` and `AttachCount`.

WeiZhang555:
Add another commit to manage rootfs block via device manager, this could make devices more clear in persist data.

virtcontainers/device/drivers/generic.go
WeiZhang555: @jodh-intel This is a switch case with 3 conditions, I don't think we can protect it with a atomic lock.

For example, if I modify this to:

```
if attach {
    if swapped:=atomic.CompareAndSwapUint64(&device.AttachCount, 0, 0); swapped {
        return false, nil
   }else if if swapped:=atomic.CompareAndSwapUint64(&device.AttachCount, intMax, intMax); swapped {
        return true, fmt.Errorf("device was attached too many times")
    } else {
        atomic.AddUint64(&device.AttachCount, 1)
        return true, nil
    }
} else { // detach use case
...
}
```

But this logic can't protect `device.AttachCount` because between each `if...else...`, it could changed. So I don't have a good way with `Atomic`

jodh-intel: Yep - thanks for checking though @WeiZhang555!.

jodh-intel:
lgtm

WeiZhang555:
ping @kata-containers/runtime

virtcontainers/device/drivers/generic.go
bergwolf: Fair enough. Thanks @WeiZhang555

bergwolf:
LGTM

Issue #663, opened 2018-08-29
Fix static-checks on travis


chavafg:
Currently seeing this:
```
fatal: ambiguous argument 'origin/': unknown revision or path not in the working tree.
```

PR #664, opened 2018-08-29
travis: Add variable needed to run static checks

chavafg:
Now that we support multiple branches, we changed how
static-checks.sh compares branches. We now need a
variable called $target_branch to make the correct
comparison when testing a PR.

Fixes: #663.


jodh-intel:
lgtm

PR #684, opened 2018-08-31
factory: start proxy before assign vm to a sandbox

bergwolf:
Right now proxy is always started **after** we assign the guest vm to a sandbox. But then we will have to wait again for the proxy to start up. We can move the creation of proxy to vm factory so that new vms are created with configured proxy type, and then wait for both proxy and vm/agent in parallel. This will save about 300ms startup time for vm factory.

### DIFF
--- a/virtcontainers/cc_proxy.go
+++ b/virtcontainers/cc_proxy.go
@@ -6,6 +6,7 @@
 package virtcontainers
 
 import (
+	"fmt"
 	"os/exec"
 )
 
@@ -14,8 +15,12 @@ type ccProxy struct {
 
 // start is the proxy start implementation for ccProxy.
 func (p *ccProxy) start(sandbox *Sandbox, params proxyParams) (int, string, error) {
-	config, err := newProxyConfig(sandbox.config)
-	if err != nil {
+	if sandbox.config == nil {
+		return -1, "", fmt.Errorf("Sandbox config cannot be nil")
+	}
+
+	config := sandbox.config.ProxyConfig
+	if err := validateProxyConfig(config); err != nil {
 		return -1, "", err
 	}
 
--- a/virtcontainers/cc_proxy.go
+++ b/virtcontainers/cc_proxy.go
@@ -5,33 +5,27 @@
 
 package virtcontainers
 
-import (
-	"fmt"
-	"os/exec"
-)
+import "os/exec"
 
 type ccProxy struct {
 }
 
 // start is the proxy start implementation for ccProxy.
-func (p *ccProxy) start(sandbox *Sandbox, params proxyParams) (int, string, error) {
-	if sandbox.config == nil {
-		return -1, "", fmt.Errorf("Sandbox config cannot be nil")
-	}
-
-	config := sandbox.config.ProxyConfig
-	if err := validateProxyConfig(config); err != nil {
+func (p *ccProxy) start(params proxyParams) (int, string, error) {
+	if err := validateProxyParams(params); err != nil {
 		return -1, "", err
 	}
 
+	params.logger.Info("Starting cc proxy")
+
 	// construct the socket path the proxy instance will use
-	proxyURL, err := defaultProxyURL(sandbox, SocketTypeUNIX)
+	proxyURL, err := defaultProxyURL(params.id, SocketTypeUNIX)
 	if err != nil {
 		return -1, "", err
 	}
 
-	args := []string{config.Path, "-uri", proxyURL}
-	if config.Debug {
+	args := []string{params.path, "-uri", proxyURL}
+	if params.debug {
 		args = append(args, "-log", "debug")
 	}
 
@@ -43,7 +37,7 @@ func (p *ccProxy) start(sandbox *Sandbox, params proxyParams) (int, string, erro
 	return cmd.Process.Pid, proxyURL, nil
 }
 
-func (p *ccProxy) stop(sandbox *Sandbox, pid int) error {
+func (p *ccProxy) stop(pid int) error {
 	return nil
 }
 
--- a/virtcontainers/factory/factory.go
+++ b/virtcontainers/factory/factory.go
@@ -29,10 +29,6 @@ type Config struct {
 	VMConfig vc.VMConfig
 }
 
-func (f *Config) validate() error {
-	return f.VMConfig.Valid()
-}
-
 type factory struct {
 	base base.FactoryBase
 }
@@ -50,7 +46,7 @@ func NewFactory(ctx context.Context, config Config, fetchOnly bool) (vc.Factory,
 	span, _ := trace(ctx, "NewFactory")
 	defer span.Finish()
 
-	err := config.validate()
+	err := config.VMConfig.Valid()
 	if err != nil {
 		return nil, err
 	}
@@ -93,13 +89,15 @@ func (f *factory) log() *logrus.Entry {
 	return factoryLogger.WithField("subsystem", "factory")
 }
 
-func resetHypervisorConfig(config *vc.HypervisorConfig) {
-	config.NumVCPUs = 0
-	config.MemorySize = 0
-	config.BootToBeTemplate = false
-	config.BootFromTemplate = false
-	config.MemoryPath = ""
-	config.DevicesStatePath = ""
+func resetHypervisorConfig(config *vc.VMConfig) {
+	config.HypervisorConfig.NumVCPUs = 0
+	config.HypervisorConfig.MemorySize = 0
+	config.HypervisorConfig.BootToBeTemplate = false
+	config.HypervisorConfig.BootFromTemplate = false
+	config.HypervisorConfig.MemoryPath = ""
+	config.HypervisorConfig.DevicesStatePath = ""
+	config.ProxyType = vc.NoopProxyType
+	config.ProxyConfig = vc.ProxyConfig{}
 }
 
 // It's important that baseConfig and newConfig are passed by value!
@@ -113,8 +111,8 @@ func checkVMConfig(config1, config2 vc.VMConfig) error {
 	}
 
 	// check hypervisor config details
-	resetHypervisorConfig(&config1.HypervisorConfig)
-	resetHypervisorConfig(&config2.HypervisorConfig)
+	resetHypervisorConfig(&config1)
+	resetHypervisorConfig(&config2)
 
 	if !reflect.DeepEqual(config1, config2) {
 		return fmt.Errorf("hypervisor config does not match, base: %+v. new: %+v", config1, config2)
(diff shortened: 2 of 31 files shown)


virtcontainers/hyperstart_agent.go
```diff
@@ -461,7 +478,18 @@ func (h *hyper) stopSandbox(sandbox *Sandbox) error {
...
-	return h.proxy.stop(sandbox, h.state.ProxyPid)
+	if err := h.proxy.stop(h.state.ProxyPid); err != nil {
+		return err
+	}
+
+	h.state.ProxyPid = -1
+	h.state.URL = ""
+	if err := sandbox.storage.storeAgentState(sandbox.id, h.state); err != nil {
+		// ignore error
+		h.Logger().WithError(err).WithField("sandbox id", sandbox.id).Errorf("failed to clean up agent state")
```
jodh-intel: Can you make that field name just `sandbox` to avoid the space in it? Same comment for `kata_agent.go`.

virtcontainers/cc_proxy.go
```diff
@@ -5,28 +5,27 @@
 
 package virtcontainers
 
-import (
-	"os/exec"
-)
+import "os/exec"
```
jodh-intel: I don't think we're testing this any more in the CI? In fact, can we even drop the `cc_*` stuff now I wonder?

virtcontainers/kata_agent.go
```diff
@@ -512,6 +528,44 @@ func (k *kataAgent) startProxy(sandbox *Sandbox) error {
...
+	}
+
+	logger := k.Logger().WithFields(logrus.Fields{
+		"proxy-pid": pid,
+		"proxy-url": url,
+	})
+	if sandbox != nil {
+		logger = logger.WithField("sandbox", sandbox.id)
+	}
+	logger.Infof("set proxy")
```
jodh-intel: I'm not sure what this is for? If we need it, it can be `logger.Info()` too. Might be better to drop it (and potentially add some trace spans in these files instead :)

virtcontainers/kata_agent.go
```diff
@@ -602,7 +656,19 @@ func (k *kataAgent) stopSandbox(sandbox *Sandbox) error {
...
+	if err := k.proxy.stop(k.state.ProxyPid); err != nil {
+		return err
+	}
+
+	// clean up agent state
+	k.state.ProxyPid = -1
+	k.state.URL = ""
+	if err := sandbox.storage.storeAgentState(sandbox.id, k.state); err != nil {
+		// ignore error
+		k.Logger().WithError(err).WithField("sandbox id", sandbox.id).Errorf("failed to clean up agent state")
```
jodh-intel: `s/"sandbox id"/"sandbox"/`

virtcontainers/sandbox.go
```diff
@@ -1600,10 +1602,10 @@ func (s *Sandbox) HotplugAddDevice(device api.Device, devType config.DeviceType)
 			if _, err := s.hypervisor.hotplugAddDevice(dev, vfioDev); err != nil {
 				s.Logger().
 					WithFields(logrus.Fields{
-						"sandboxid":       s.id,
-						"vfio device ID":  dev.ID,
-						"vfio device BDF": dev.BDF,
-					}).WithError(err).Error("failed to hotplug VFIO device")
+					"sandboxid":       s.id,
+					"vfio device ID":  dev.ID,
```
jodh-intel: Again, I think it's best to avoid field names with spaces "just in case", so maybe `vfio-device-id` ;)

virtcontainers/sandbox.go
```diff
@@ -1600,10 +1602,10 @@ func (s *Sandbox) HotplugAddDevice(device api.Device, devType config.DeviceType)
 			if _, err := s.hypervisor.hotplugAddDevice(dev, vfioDev); err != nil {
 				s.Logger().
 					WithFields(logrus.Fields{
-						"sandboxid":       s.id,
-						"vfio device ID":  dev.ID,
-						"vfio device BDF": dev.BDF,
-					}).WithError(err).Error("failed to hotplug VFIO device")
+					"sandboxid":       s.id,
```
jodh-intel: This should be just `sandbox` for consistency with the rest of the codebase I think.

virtcontainers/vm.go
```diff
@@ -148,9 +215,24 @@ func (v *VM) Start() error {
 	return v.hypervisor.startSandbox()
 }
 
+// Disconnect agent and proxy connections to a VM
+func (v *VM) Disconnect() error {
+	v.logger().Info("kill vm")
+
+	if err := v.agent.disconnect(); err != nil {
+		v.logger().WithError(err).Errorf("failed to disconnect agent")
```
jodh-intel: `s/Errorf/Error/` as the last param is a simple string (same comment for below).

virtcontainers/cc_proxy.go
bergwolf: Surely I agree. Created #694 to track it.

virtcontainers/factory/factory.go
```diff
@@ -129,13 +127,25 @@ func (f *factory) checkConfig(config vc.VMConfig) error {
 	return checkVMConfig(config, baseConfig)
 }
 
+func (f *factory) validateNewVMConfig(config vc.VMConfig) error {
+	if len(config.AgentType.String()) == 0 {
+		fmt.Errorf("Invalid agent type")
```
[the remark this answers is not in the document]
bergwolf: Good point!

virtcontainers/factory/template/template.go
```diff
@@ -100,20 +100,27 @@ func (t *template) createTemplateVM(ctx context.Context) error {
...
-	err = vm.Pause()
-	if err != nil {
+	if err = vm.Disconnect(); err != nil {
 		return err
 	}
 
-	err = vm.Save()
-	if err != nil {
+	// Sleep a bit to let the agent grpc server clean up
+	time.Sleep(2 * time.Second)
```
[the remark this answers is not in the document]
bergwolf: This is only used when creating the very first templating vm. And it actually saves more time for every new vm created from the template later on.

virtcontainers/hyperstart_agent.go
bergwolf: Good point!

virtcontainers/kata_agent.go
bergwolf: ok, I'll drop it.

bergwolf:
FYI, the PR is verified in kata-containers/tests#578 and it all passed.

bergwolf:
@jodh-intel when we close connection to the agent, it needs sometime to cleanup and restart listening on the communication( serial or vsock) port. That time can be saved if we sleep a bit to wait for the agent to come around and start listening again. The sleep is only done when creating new vm templates and saves time for every new vm that are created from template, so it worth the invest.

bergwolf:
//recheck

virtcontainers/hyperstart_agent.go
```diff
@@ -390,7 +390,11 @@ func (h *hyper) startProxy(sandbox *Sandbox) error {
 	}
 
 	// Start the proxy here
-	pid, uri, err := h.proxy.start(sandbox, proxyParams{})
+	pid, uri, err := h.proxy.start(proxyParams{
+		id:     sandbox.id,
+		path:   sandbox.config.ProxyConfig.Path,
```
sboeuf: You forgot to add `debug`.
```go
debug: sandbox.config.ProxyConfig.Debug,
```

virtcontainers/factory/template/template.go
```diff
@@ -100,35 +104,44 @@ func (t *template) createTemplateVM(ctx context.Context) error {
...
-	err = vm.Pause()
-	if err != nil {
+	if err = vm.Disconnect(); err != nil {
 		return err
 	}
 
-	err = vm.Save()
-	if err != nil {
+	// Sleep a bit to let the agent grpc server clean up
+	time.Sleep(templateWaitForAgent)
```
sboeuf: Why a sleep here? Any way we could provide some more robust way of actively waiting?

virtcontainers/factory/template/template.go
```diff
@@ -100,35 +104,44 @@ func (t *template) createTemplateVM(ctx context.Context) error {
...
+		return err
+	}
+
+	if err = vm.Save(); err != nil {
 		return err
 	}
 
 	// qemu QMP does not wait for migration to finish...
-	time.Sleep(1 * time.Second)
+	time.Sleep(templateWaitForMigration)
```
sboeuf: Why a sleep here? Any way we could provide some more robust way of actively waiting?

sboeuf:
@bergwolf would be nice if you could reduce the coverage drop ;)

virtcontainers/factory/template/template.go
amshinde: Agree. Also the timeout values seem a bit arbitrary. Is there a possibility that these timeouts may not be sufficient on on slower machines?

virtcontainers/hyperstart_agent.go
bergwolf: nice catch!

virtcontainers/factory/template/template.go
bergwolf: As I replied to @jodh-intel here (https://github.com/kata-containers/runtime/pull/684#issuecomment-419480476), it is an optimization worth investing.

OTOH, I don't see any possible active waiting method since we cannot connect to the agent to make sure it is up.

bergwolf:
UT added and CI green. @sboeuf @amshinde PTAL.

bergwolf:
@sboeuf @amshinde Are you happy with the PR now? Please merge it if so.

virtcontainers/factory/template/template.go
sboeuf: Ok now I understand what you're doing here. I understand this is out of the regular path since it's about VM template creation that might happen hours before the creation of an actual container.
And this second makes sure the agent is already in a proper waiting state for when the template is going to be used.
That's hard to think how of a better way, I understand...
Could you open an issue to see if we can think about something else after this got merged.

virtcontainers/factory/template/template.go
sboeuf: Could we check the status of the template files being created by `vm.Save()` as a way to validate the migration is done?

sboeuf:
@bergwolf also, Travis seems completely stuck, could you re-push this PR to unblock it?

virtcontainers/factory/template/template.go
bergwolf: No, we do not know the actual size of the file either. We might be able to monitor the file close event but it's very hacky and as you said in the other comment, it is on the non-regular path and it does not worth so much effort to optimize.

virtcontainers/factory/template/template.go
bergwolf: Sure, I'll open an issue for this but I don't think there is an easy way to do it and TBF I don't think it worth much effort to optimize the sleep time since it only happens once.

bergwolf:
@sboeuf Thanks for reviewing. Repushed and opened #728.

sboeuf:
@amshinde PTAL

sboeuf:
@jodh-intel @egernst @WeiZhang555 PTAL, we need another approval to this PR if you think it's ok to be merged. And please merge it as Travis is stuck for no reason, otherwise the CI is all green.

WeiZhang555:
Not familiar with this part, so I think I should trust @sboeuf 's judgement :smile: 

LGTM

virtcontainers/factory/template/template.go
bergwolf: @laijs just mentioned to me that we can use qmp `query-migrate` do check the migration status. Then `templateWaitForMigration` can be dropped. I'll send a new PR to remove it after this one is merged. Does it sound good to you @WeiZhang555 ?

w.r.t. `templateWaitForAgent`, it just affects new normal VM creation speed if your machine is too slow for the agent to come around.

WeiZhang555: `query-migrate` sounds quite good.
Then can you add a "// TODO: " as a reminder? In case everyone of us forgot this, at least in some days someone can find this `TODO` :laughing:

bergwolf: @WeiZhang555 here is our #728 ;-)

WeiZhang555: @bergwolf Oh, we already got one! Nice!

Issue #688, opened 2018-09-03
virtcontainers: Seccomp profile passed has no effect, silently fails. 


nitkon:
# Description of problem

When we pass a seccomp profile to a Kata Container, it silently fails and has no effect on the Kata
container. 

# Expected result

```
# cat sleep.json 
{
        "defaultAction": "SCMP_ACT_ALLOW",
        "syscalls": [
                {
                        "name": "nanosleep",
                        "action": "SCMP_ACT_ERRNO"
                                
                }               
                        ]
}

# docker run -it --security-opt seccomp=sleep.json --runtime kata-runtime busybox sh
/ # sleep 6   ( This should have no effect if seccomp profile got applied successfully)
/ # 
```

# Actual result
```
# docker run -it --security-opt seccomp=sleep.json --runtime kata-runtime busybox sh
/ # sleep 6   ( Sleeps for 6 seconds)
/ # 
```

PR #381, opened 2018-06-10
Fix golint ppc64le issues and eventually Travis-CI

nitkon:

PR #689, opened 2018-09-03
virtcontainers: Pass seccomp profile inside VM

nitkon:
Seccomp profile was dropped as agent, inside the VM,
was not capable of handling it. Allow it to be passed
as agent binary is now build with -tags "seccomp" and
the image on the guest would have libseccomp installed.

Fixes: #688


jodh-intel:
lgtm.

But blocked on:

- https://github.com/kata-containers/agent/pull/353
- https://github.com/kata-containers/osbuilder/pull/154

sboeuf:
Not sure the agent support `seccomp`. We need to fix the agent, and I think we need libseccomp to be part of the VM image if we want to be able to use seccomp inside the VM.

nitkon:
@sboeuf : PTAL https://github.com/kata-containers/agent/pull/353#issuecomment-418425975

sboeuf:
@jodh-intel @nitkon I was thinking, shouldn't this PR introduce a new config option inside `configuration.toml` under the `[agent]` section so that we can say if the agent supports or not `libseccomp`.
And of course, by default the agent will not support it, so we won't go through this code path.

jodh-intel:
@sboeuf - ah you mean we should have that option so the runtime is clear that seccomp support *must* be provided by the agent in the image, or else it needs to fail?

Yes, I agree. This is the same sort of issue as:

- https://github.com/kata-containers/agent/pull/365#discussion_r217973026 (and the runtime component which adds the new option https://github.com/kata-containers/runtime/pull/720).

nitkon:
@sboeuf @jodh-intel : 
Case 1:
I just want to understand this. As of today if the agent does not provide the seccomp support we fail with the apt message. Isn't that a sufficient message for the user?
```
docker run -it --security-opt seccomp=sleep.json --runtime kata-runtime busybox sh
docker: Error response from daemon: OCI runtime create failed: rpc error: code = Internal desc = Could not run process: container_linux.go:348: starting container process caused "**seccomp: config provided but seccomp not supported"**: unknown.

```

Case 2:

If we are thinking of implementing via configuration.toml, then ....


1. The `configuration.toml` file would have something like this
```
[agent.kata]
#Set this to true if  seccomp support is provided by the agent
#Seccomp = true
```

2.  If seccomp is not set or set to false in config.toml, but seccomp is passed in the command line then disable the seccomp 
https://github.com/kata-containers/runtime/blob/a5e82c1d4df8ff417878583600723d34e87cf771/virtcontainers/kata_agent.go#L724
https://github.com/kata-containers/runtime/blob/a5e82c1d4df8ff417878583600723d34e87cf771/virtcontainers/kata_agent_test.go#L454

and error out saying "agent does not support seccomp.  "
If agent was built with seccomp support and not set in sonfig.toml, still it would fail, which would not have been the case in CASE 1. 

3. If seccomp is set to true in config.toml, then enable the seccomp

https://github.com/kata-containers/runtime/blob/a5e82c1d4df8ff417878583600723d34e87cf771/virtcontainers/kata_agent.go#L724
https://github.com/kata-containers/runtime/blob/a5e82c1d4df8ff417878583600723d34e87cf771/virtcontainers/kata_agent_test.go#L454

nitkon:
@jodh-intel @sboeuf : 
1. We can pass the seccomp profile either explicitly using `--security-opt ` option OR docker by default passes a seccomp profile (https://github.com/moby/moby/blob/master/profiles/seccomp/default.json)

2. In either case if seccomp is not supported by agent, it would fail with the following error:

```
docker: Error response from daemon: OCI runtime create failed: rpc error: code = Internal desc = Could not run process: container_linux.go:348: starting container process caused "seccomp: config provided but seccomp not supported": unknown.
```

3. Thus, shouldn't we have agent built with seccomp support by default rather than make it optional?

sboeuf:
@nitkon the problem with enabling this by default is that even when you're not passing any seccomp from the CLI, the spec won't be nil, and this will trigger libcontainer to try to use seccomp.

Please try to look into the way seccomp is actually carried from the CLI to the agent, and identify where it's providing a `LinuxSeccomp` structure, even when nothing is provided. This way, we could override the "empty" structure with a nil pointer.

My advice would be to dump the spec in various places so that you can figure this out.

nitkon:
> @nitkon the problem with enabling this by default is that even when you're not passing any seccomp from the CLI, the spec won't be nil, and this will trigger libcontainer to try to use seccomp.
> 
If the user does not provide any seccomp profile explicitly, then do we not want the default seccomp provided by docker to be applied to our container? If we want the default seccomp profile to be applied to each container, we should have both agent and rootfs seccomp capable always and not optional. It might affect size and boot time,  but will provide additional security. 
> Please try to look into the way seccomp is actually carried from the CLI to the agent, and identify where it's providing a `LinuxSeccomp` structure, even when nothing is provided. This way, we could override the "empty" structure with a nil pointer.
> 
You mean to say when no seccomp profile is provided by the user, make sure the default seccomp is dropped inside the agent, resulting in libcontainer not trying to use the default seccomp? 
My question is, why to drop default seccomp, isn't it good for the security of Kata Containers? Are we trying to save boot time and rootfs size? 
> My advice would be to dump the spec in various places so that you can figure this out.

Sure. Thanks for the input.

sboeuf:
As you mentioned, we don't want to affect our boot time and memory footprint by enabling seccomp. At least for now. But I understand that some people might want this extra feature and that's what you're trying to enable here.

Also, because the spec that we get might have a default seccomp, then we have no other choice than letting the user decide if he wants to use seccomp or not. And in our case, this translates as a configuration option.

jodh-intel:
I'd appreciate it if the team could review the following (**very** carefully), but I think this summarises the main scenarios we need to handle in code (and tests!!)...

| seccomp profile provided? (<sup>[1](#f1)</sup>) | runtime seccomp support | runtime seccomp enabled | agent/image seccomp support | seccomp enforced? | run workload? | notes |
|-|-|-|-|-|-|-|
| no | no | - | no | no | yes | |
| no | no | - | yes | no | yes | |
| no | yes | no | no | no | yes | |
| no | yes | yes | no | no | ? | Config option enabled so runtime should warn or error (<sup>[2](#f2)</sup>) |
| no | yes | no | yes | no | yes | |
| no | yes | yes | yes | no | ? | Config option enabled so runtime should warn or error (<sup>[2](#f2)</sup>) |
| yes | no | - | no | no | yes | (<sup>[3](#f3)</sup>) |
| yes | no | - | yes | no | yes | (<sup>[3](#f3)</sup>) |
| yes | yes | no | no | no | ? | Profile provided, but config disabled so runtime should warn or error (<sup>[2](#f2)</sup>) |
| yes | yes | no | yes | no | no? | seccomp disabled by config, so runtime should warn or error (<sup>[2](#f2)</sup>) |
| yes | yes | yes | no | no | no? | Cannot enforce seccomp, so error (<sup>[2](#f2)</sup>) |
| yes | yes | yes | yes | yes | yes | Finally, we got there! :smile: |

Footnotes:

- <b id="f1">1</b>: Implicit or explicit ("`docker run --security-opt seccomp=...`").
- <b id="f2">2</b>: We could make the runtime seccomp option value either `enabled` or `enforced` to modify the behaviour (log or error respectively)?
- <b id="f3">3</b>: Runtime sets (or should set) seccomp profile to `nil` before sending to agent.

grahamwhaley:
@jodh-intel - would it be easy for you to add maybe an `(S)` after the run workload column `yes`, where it was run with seccomp enabled? Just so we can distinguish when seccomp is enforced or not?

jodh-intel:
@grahamwhaley - you mean drop the existing "seccomp enforced?" column and bundle into the next column?

jodh-intel:
Table updated with more details on the column meanings.

grahamwhaley:
Ah, that is what that column means! ;-) - then, np @jodh-intel as it is. thx!

nitkon:
Hi @jodh-intel , thanks for the table :+1: 

 1. So, why cannot we take a simple approach that:--- if a user mentions in configuration.toml file that agent/rootfs supports seccomp, then simply allow the config to be passed by the runtime to the agent else drop it. 

2. When we say "Whether the runtime version can "handle" seccomp fully " , what does that mean exactly? Isn't the role of runtime to just pass/drop the seccomp profile?

3.  What's the difference between enabled(yes/true in config file?) and enforced?

4. Could you please clarify the difference between "runtime seccomp support " and "runtime seccomp enabled" ?

jodh-intel:
Hi @nitkon 

> So, why cannot we take a simple approach that:--- if a user mentions in configuration.toml file that agent/rootfs supports seccomp, then simply allow the config to be passed by the runtime to the agent
> else drop it.

Having thought about this a little more, I'm not sure I like the idea of the config file determining whether seccomp support should be enabled. Instead, I think that when Kata fully supports seccomp, the config file should be used to tell the runtime how to behave depending on what sort of seccomp environment it's running in. And we should probe to determine whether the agent/image is seccomp-capable.

If we rely on the config file to determine whether seccomp should be used, there are three possible scenarios when a seccomp profile is provided but the config file disables seccomp:

- the config setting is correct and the image really doesn't support seccomp.
- the image in fact does support seccomp but the admin explicitly wanted to disable it for some reason [*].
- the config file is mis-configured (the admin forgot to update it, forgetting or not realising that the image does support seccomp).

That last one is the nasty one of course. Also, it begs the question how will the admin know if the image supports seccomp? Even if the image itself has the seccomp libraries installed (which `kata-collect-data.sh` will show in the contents of the `osbuilder.yaml` file), that doesn't necessarily mean the agent was built for seccomp.

> When we say "Whether the runtime version can "handle" seccomp fully " , what does that mean exactly? Isn't the role of runtime to just pass/drop the seccomp profile?

Yes, but it doesn't currently have that capability - it just drops it. So by "handling" I mean that the runtime can heuristically decide what to do.

> What's the difference between enabled(yes/true in config file?) and enforced?

See above.

> Could you please clarify the difference between "runtime seccomp support " and "runtime seccomp enabled" ?

Again, see above ;)

I think a better design might be to introduce a new gRPC agent call that the runtime can make *before* it sends over the OCI spec (including the seccomp profile). That gRPC call would tell the runtime whether the agent can deal with seccomp or not. That would be a *pure* agent call - again, we don't want libcontainer getting involved at this stage. Something like:

```
message AgentFeaturesResponse {
    bool supports_seccomp = 1;
};

rpc GetAgentFeatures(AgentFeaturesRequest) returns (AgentFeaturesResponse);
```

(Aside: In fact, I wonder if we want to replace `Check()` which is basically "ping for gRPC" with a call like this - we need to make a ping-type call before requesting the agent create containers, so why not get that call to return more information than just "I'm alive"? :)

Once that `GetAgentFeatures()` API is available:

- If a seccomp profile is specified and the agent supports seccomp, pass the profile verbatim (no config settings checked [*]).

- If a seccomp profile is specified and the agent does not support seccomp, the runtime can check the config file option:
  - if `seccomp = permissive`, log a message and pass a `nil` profile to the agent (current behaviour).
  - if `seccomp = enforcing`, log and error and fail.

The problem is that the "current behaviour" != "best behaviour" in the scenario where the agent does support seccomp since the default behaviour should probably be to enforce, so I think we'll need to break the current behaviour (by making seccomp enforced if available) to ensure the users get the safest possible default config.

---
[*] - However, we may need to come up with a config option to handle this scenario if the admin has requested that seccomp be disabled even though the agent supports it.


wdyt @sboeuf, @bergwolf, @egernst, @gnawux?

bergwolf:
@jodh-intel Yes, I agree. `GetAgentFeatures` sounds good to me. It can be useful for many use cases.

jodh-intel:
@nitkon - I've raised https://github.com/kata-containers/agent/issues/381 for adding a new gRPC call. I'm happy to look at that if you want?

raravena80:
@nitkon @jodh-intel ping looks like #381 is closed. Was that what we needed for this?

jodh-intel:
Yes, https://github.com/kata-containers/agent/issues/381 is now closed. I'm probably going to raise a PR to move that call in the runtime, but the call will still be made.

As such, @nitkon can re-vendor the agent into the runtime to get the latest version of `GetGuestDetails()` which includes whether seccomp is available in the image or not.

raravena80:
@nitkon ping (From your weekly Kata herder)

grahamwhaley:
Hi @nitkon @egernst et. al.
I said in a kata arch call a couple of weeks ago I'd have a look at the metrics impact of this PR to help push along the decision if we make this on or off by default etc. @egernst nudged me yesterday to pick that up again (some other things had come up that pushed me off it).
I did a run today trying to compare the 4-way mix of with/without seccomp and .img/initrd.
I have some results for img/img-seccomp/initrd, but am having trouble getting initrd-seccomp (alpine) up and running.
Also, my results don't look like I'd expect - so next week I'll re-run them to try and verify and investigate.
Thought I'd keep you informed at least that we are still looking at this :-)

egernst:
@grahamwhaley we should definitely move forward with merging  #689 and the osbuilder PR156.

I like the idea of making this configurable in .toml (it isn't today), and having this enabled by default, allowing the end user to make the tradeoff on security v perf.  @sameo @jon @WeiZhang555 @gnawux  thoughts?

egernst:
I'd like to get this in for 1.4 if possible!

sboeuf:
@grahamwhaley @egernst 
I'd say:
- let's merge https://github.com/kata-containers/osbuilder/pull/156
- then, we can modify the packaging scripts to enable seccomp with `SECCOMP=yes`

In the meantime, @nitkon could rework this current PR to make the use of libseccomp configurable from `configuration.toml`.

WeiZhang555:
> I like the idea of making this configurable in .toml (it isn't today), and having this enabled by default, allowing the end user to make the tradeoff on security v perf.

This makes lots of sense :+1:

sameo:
> I like the idea of making this configurable in .toml (it isn't today), and having this enabled by default, allowing the end user to make the tradeoff on security v perf. @sameo @jon @WeiZhang555 @gnawux thoughts?

This should definitely be a configuration knob, but I'd argue it should be turned **off** by default. I feel having seccomp enabled inside should not be the general use case.

egernst:
@nitkon ping -- are you able to rework this per suggestions above?  I'd *love* to get this in before 1.4 (end of week).

raravena80:
@nitkon nudge from your weekly Kata herder

nitkon:
Can someone start the Jenkins CI, please?

grahamwhaley:
I think there was another push....
/test

nitkon:
Ping for review.

virtcontainers/kata_agent.go
```diff
@@ -720,16 +720,16 @@ func (k *kataAgent) replaceOCIMountsForStorages(spec *specs.Spec, volumeStorages
...
 	// no reason to send them to the agent. It would make no sense to try
 	// to apply them on the guest.
 	grpcSpec.Hooks = nil
 
-	// Disable Seccomp since they cannot be handled properly by the agent
-	// until we provide a guest image with libseccomp support. More details
-	// here: https://github.com/kata-containers/agent/issues/104
-	grpcSpec.Linux.Seccomp = nil
+	// Set grpcSpec.Linux.Seccomp as per the runtime configuration file
+	if DisableGuestSeccomp == true {
```
bergwolf: Since we have exported guest seccomp capability in kata-containers/agent#384, please make use of it when making decisions about seccomp on the runtime side.

nitkon:
Travis-CI is passing. Can someone start the jenkins CI test?
ping for review @jodh-intel @bergwolf @egernst

gnawux:
/test

On Dec 5, 2018 20:43, "nitkon" <notifications@github.com> wrote:

Travis-CI is passing. Can someone start the jenkins CI test?
ping for review @jodh-intel <https://github.com/jodh-intel> @bergwolf
<https://github.com/bergwolf> @egernst <https://github.com/egernst>

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/kata-containers/runtime/pull/689#issuecomment-444472265>,
or mute the thread

nitkon:
The test failures are not related to the PR. Let's have a retest.
/retest

virtcontainers/kata_agent.go
```diff
@@ -986,9 +987,18 @@ func (k *kataAgent) createContainer(sandbox *Sandbox, c *Container) (p *Process,
 		return nil, err
 	}
 
+	request := &grpc.GuestDetailsRequest{
```
bergwolf: We already query guest details in `getAndStoreGuestDetails()`. Please just save the secomp results got from there and then we can use it everywhere. By saving it, please add it to `Sandbox` and also save it on disk so that we can access it again after the very first create sandbox call.

raravena80:
@nitkon ready to merge?

ydjainopensource:
ping @nitkon

nitkon:
@raravena80 : Sorry for the late reply. I was caught up in some personal work. Yes the PR is good to merge. @bergwolf I shall optimize it in a follow-up PR later ...

nitkon:
Updated my PR. Not started the tests as all tests are failing for a couple of days...

nitkon:
Re-pushed the PR as there was some `maligned structs` warning while doing static check.

Issue #690, opened 2018-09-03
make sure process has exited before to RemoveContainer


lifupan:
# Description of problem

RemoveContainer is called right after SignalProcess(SIGKILL), the container process might be still running and container Destroy() will fail, thus it's better to wait on this process to exited before to
issue  RemoveContainer.

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #691, opened 2018-09-03
virtcontainers: wait until process exited before RemoveContainer

lifupan:
RemoveContainer is called right after SignalProcess(SIGKILL), the container
process might be still running and container Destroy() will fail, thus it's better
to wait on this process exited before to issue RemoveContainer.

Fixes: #690

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -890,6 +890,12 @@ func (c *Container) stop() error {
 	// stopContainer() to succeed in such particular case.
 	c.kill(syscall.SIGKILL, true)
 
+	// Since the agent has supported the MultiWaitProcess, it's better to
+	// wait the process here to make sure the process has exited before to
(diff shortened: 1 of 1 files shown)


jodh-intel:
The PR the comment is referring to is https://github.com/kata-containers/agent/pull/344 which has just landed.

It's a shame we can't add a test here for this change but it's possible as it can't be unit-tested.

lgtm

sboeuf:
@lifupan @egernst This one is a bug fix, but unfortunately cannot be backported since we need the agent to support multiple `WaitProcess()` calls, and this is not part of the agent 1.1 or 1.2, right ?

bergwolf:
@sboeuf Yes, you are right. And the fix is only required for containerd-shimv2 so stabe 1.1 and 1.2 do not need it either.

PR #701, opened 2018-09-05
virtcontainers: apply devices constraints

### LINKED ISSUE #656: Support devices constraints
we need to support devices constraints

devimc:
Apply devices constraints to the container in the virtual machine

Depends-on: github.com/kata-containers/agent#352

fixes #656

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -666,10 +666,9 @@ func constraintGRPCSpec(grpcSpec *grpc.Spec) {
 	// here: https://github.com/kata-containers/agent/issues/104
 	grpcSpec.Linux.Seccomp = nil
 
-	// By now only CPU constraints are supported
+	// By now only CPU, Memory and Devices constraints are supported
 	// Issue: https://github.com/kata-containers/runtime/issues/158
 	// Issue: https://github.com/kata-containers/runtime/issues/204
-	grpcSpec.Linux.Resources.Devices = nil
 	grpcSpec.Linux.Resources.Pids = nil
 	grpcSpec.Linux.Resources.BlockIO = nil
 	grpcSpec.Linux.Resources.HugepageLimits = nil
--- a/virtcontainers/kata_agent_test.go
+++ b/virtcontainers/kata_agent_test.go
@@ -450,7 +450,7 @@ func TestConstraintGRPCSpec(t *testing.T) {
 	// check nil fields
 	assert.Nil(g.Hooks)
 	assert.Nil(g.Linux.Seccomp)
-	assert.Nil(g.Linux.Resources.Devices)
+	assert.NotNil(g.Linux.Resources.Devices)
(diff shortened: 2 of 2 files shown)


amshinde:
@devimc Lets verify this change works with device passthrough.

virtcontainers/kata_agent.go
```diff
@@ -669,7 +669,6 @@ func constraintGRPCSpec(grpcSpec *grpc.Spec) {
 	// By now only CPU constraints are supported
```
bergwolf: The comment is outdated even before this PR.

devimc: good catch! thanks

devimc:
@amshinde do you know if currently we have a device passthrough test? 

https://github.com/kata-containers/tests/blob/master/integration/docker/run_test.go#L85-L138

xindazhao:
Hi, is there any difference in the use of docker command by using device constrain? I use following command to enable INTEL GPU device support, with this commit added in, the GPU device node "/dev/dri/card0" and "/dev/dri/renderD128" can be seen in container, but returns error if open() the device node. Without this commit, everything works well. 
What changes do I have to make in the command line to adapt to device constrains?
# docker run -it --runtime=kata-runtime --rm  --device=/dev/vfio/0  -v /dev:/dev  ubuntu /bin/bash

devimc:
@xindazhao this patch is to honour the devices cgroup, for example if you want to grant access to `/dev/dri/card0` then you have to use `--device /dev/dri/card0` in the command line. The problem with vfio devices is that once they are `unbind`'d from the host other devices like `--device /dev/dri/card0` are no more visible, hence `docker run --runtime kata-runtime --device /dev/dri/card0 -it --rm ubuntu bash` fails with following error  `docker: Error response from daemon: linux runtime spec devices: error gathering device information while adding custom device "/dev/dri/card0": no such file or directory.`
@xindazhao I have some ideas to fix the issue your are facing, is there a way to know which devices were `bind`'d or are part of a vfio device?, for example if you passthrough `/dev/vfio/0`, how I can identify that `/dev/dri/card0` and `/dev/dri/renderD128` were created because of `/dev/vfio/0` was attached?

xindazhao:
There is no direct relationship between the VFIO node and GPU device node. In the GVT-g (VFIO Mdev) scenario, multiple virtual GPUs share a physical GPU. We append different VFIO nodes (/dev/vfio/0, /dev/vfio/1 /dev/vfio/2 ....) to the docker command line, but in the guest VM, all the GPU device nodes that you see might be the same "/dev/dri/card0" and "/dev/dri/renderD128"

sboeuf:
@devimc what's the story on this PR? What is the blocker to make CI passing so that we can move to a mergeable state?

devimc:
if device constrains are applied, we'll face issues with vfio devices see https://github.com/kata-containers/runtime/pull/701#issuecomment-419443916

Issue #709, opened 2018-09-11
master build failure


bergwolf:
Sadly CI failed to catch the broken line due to the fact that it is introduced by a different
PR that passed w/o the naming PR.

```
# github.com/kata-containers/runtime/cli
./config.go:604:27: config.DefaultMemSz undefined (type virtcontainers.HypervisorConfig has no field or method DefaultMemSz)
Makefile:331: recipe for target '/golang/src/github.com/kata-containers/runtime/kata-runtime' failed
make: *** [/golang/src/github.com/kata-containers/runtime/kata-runtime] Error 2
```

PR #710, opened 2018-09-11
cli: fix build

bergwolf:
Sadly CI failed to catch the broken line due to the fact that it is introduced by a different
PR that passed w/o the naming PR.
```
./config.go:604:27: config.DefaultMemSz undefined (type virtcontainers.HypervisorConfig has no field or method DefaultMemSz)
Makefile:331: recipe for target '/golang/src/github.com/kata-containers/runtime/kata-runtime' failed
make: *** [/golang/src/github.com/kata-containers/runtime/kata-runtime] Error 2
```
Fixes: #709

### DIFF
--- a/cli/config.go
+++ b/cli/config.go
@@ -601,7 +601,7 @@ func checkHypervisorConfig(config vc.HypervisorConfig) error {
 		},
 	}
 
-	memSizeMB := int64(config.DefaultMemSz)
+	memSizeMB := int64(config.MemorySize)
 
 	if memSizeMB == 0 {
 		return errors.New("VM memory cannot be zero")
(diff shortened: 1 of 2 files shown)


WeiZhang555:
Codes LGTM

Why CI didn't catch the broken line? Weird

bergwolf:
@WeiZhang555 It was merged in as a different PR which passed itself w/o the hypervisor config field rename. CI validation only merges the current PR to master. So even if two PRs both pass independently but when they are both merged to master, build may still break.

WeiZhang555:
I see. This is hard to resolve, unless we force each PR to rebase and run another round of CI test before merging, even so there's still potential break possibility.

jodh-intel:
Merging as we kinda need a buildable source tree... ;)

sboeuf:
@bergwolf @egernst correct me if I'm wrong, but even if this is a bug fix, it cannot be backported as the previous PR introducing the bug is not part of the stable branch (and should not be part of it), right?

bergwolf:
@sboeuf Yes, you are right.

sboeuf:
Thanks for confirmation.

Issue #711, opened 2018-09-11
disable cgroup namespaces


bergwolf:
Right now we do not support it and it should be cleared otherwise container creation will fail.

PR #712, opened 2018-09-11
kata_agent: disable cgroup namespace

bergwolf:
We do not support it yet. If we pass it to the agent, container creation
will fail.

Depends-on: github.com/kata-containers/runtime#710
Fixes: #711

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -684,6 +684,7 @@ func constraintGRPCSpec(grpcSpec *grpc.Spec) {
 	var tmpNamespaces []grpc.LinuxNamespace
 	for _, ns := range grpcSpec.Linux.Namespaces {
 		switch ns.Type {
+		case specs.CgroupNamespace:
 		case specs.NetworkNamespace:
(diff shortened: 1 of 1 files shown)


jodh-intel:
lgtm

amshinde:
@bergwolf I dont understand, you mentioned that container creation will fail, but we have been passing this all along right?

bergwolf:
@amshinde Do you mean we have passed tests with a container spec containing cgroup namespace? Which case are you referring to? AFAICT, we'll always fail such a case.

amshinde:
@bergwolf Never mind. I misunderstood, I thought the croup namespace is added by default.
lgtm.

sboeuf:
@bergwolf please backport to stable branches.

bergwolf:
@sboeuf sure, I'm not sure which branch to send against though. `stable-1.x` or the ones with `-candidate` suffix?

bergwolf:
See #722 and #723

sboeuf:
Thanks @bergwolf !

PR #719, opened 2018-09-12
Set the queue size for Multiqueue virtio-net as the number of vCPUs on the guest.

amshinde:
Instead of using a default queue size of 8 for Multiqueue virtio-net,
use the number of vCPUs on the guest as the queue size.
This is the recommended approach for setting up the queue size.
This also shown better performance results.

### DIFF
--- a/virtcontainers/network.go
+++ b/virtcontainers/network.go
@@ -95,7 +95,6 @@ const (
 	defaultRouteLabel = "default"
 	defaultFilePerms  = 0600
 	defaultQlen       = 1500
-	defaultQueues     = 8
 )
 
 // DNSInfo describes the DNS setup related to a network interface.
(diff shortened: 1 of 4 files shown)


virtcontainers/network.go
```diff
@@ -226,7 +225,7 @@ func networkLogger() *logrus.Entry {
 func (endpoint *VirtualEndpoint) Attach(h hypervisor) error {
 	networkLogger().WithField("endpoint-type", "virtual").Info("Attaching endpoint")
 
-	if err := xconnectVMNetwork(&(endpoint.NetPair), true); err != nil {
+	if err := xconnectVMNetwork(&(endpoint.NetPair), true, h.hypervisorConfig().NumVCPUs); err != nil {
```
bergwolf: Do we care about hotplugged CPUs? Properly not since they might be unplugged later. Just want to know your opinions about it.

egernst: Hey @bergwolf  - under what use-cases would we be actively hotplugging/unplugging?

bergwolf: @egernst Currently no. I don't know any use case that would require removing a container while having other containers still running in the same sandbox. It is just that our APIs and CLI interfaces allow such behavior. nvm. It's fine with me to just use the default CPU numbers for the queue size.

bergwolf:
LGTM!

virtcontainers/network.go
egernst: Well, I think we *do* care about the resulting number of vCPUS.  This shouldn't default to the default CPUs, but to what we request.

@amshinde @sboeuf

PR #720, opened 2018-09-12
cli: add guest hook path option

eguzman3:
Blocked on (https://github.com/kata-containers/agent/pull/365), will need to revendor.


jodh-intel:
Added `do-not-merge` label which will disallow this branch from landing - see `.pullapprove.yml`.

Once the dependent branch is merged, we can remove the label ;)

cli/config.go
```diff
@@ -275,6 +276,13 @@ func (h hypervisor) useVSock() bool {
 	return h.UseVSock
 }
 
+func (h hypervisor) guestHookPath() string {
```
jodh-intel: Please could you add a minimal test in `config_test.go` for this. Looking around your change, it seems we have been a little remiss recently as some of the newer calls are missing unit tests.

You could re-use code from `TestHypervisorDefaultsImage()` - we just need to test the two scenarios - either the default value is returned or the `h.GuestHookPath`.

egernst:
Hey @eguzman3  - it seems like this PR has more than just adding the guest hooks - can you split into multiple well described commits, and perhaps as seperate PRs?  Please let me know if I'm mistaken...

egernst:
And, please disregard my comment -- I see that I was in the weeds looking at the vendor code.  Thanks!

sboeuf:
@egernst @eguzman3 once kata-containers/agent#365 lands, it's still cleaner if you split the vendoring code in its own commit ;)

jodh-intel:
Still blocked on https://github.com/kata-containers/agent/pull/365.

raravena80:
@eguzman3 ping (From your weekly Kata herder). It is still blocked?

caoruidong:
I think we can drop this one cause there is a new version #834

PR #734, opened 2018-09-15
Add cgroup support

### LINKED ISSUE #344: Isolate hypervisor process with cgroups

WeiZhang555:
cgroups: add host cgroup support

Fixes #344

Add host cgroup support for kata.

This commits only adds cpu.cfs_period and cpu.cfs_quota support.

It will create 3-level hierarchy, take "cpu" cgroup as an example:

```
/sys/fs/cgroup
|---cpu
   |---vc
      |---<sandbox-id>
         |--vcpu
      |---<sandbox-id>
```

* `vc` cgroup is common parent for all kata-container sandbox, it won't be removed
after sandbox removed. This cgroup has no limitation.
* `<sandbox-id>` cgroup is the layer for each sandbox, it contains all other qemu
threads except for vcpu threads. In future, we can consider putting all shim
processes and proxy process here. This cgroup has no limitation yet.
* `vcpu` cgroup contains vcpu threads from qemu. Currently cpu quota and period
constraint applies to this cgroup.

Signed-off-by: Jingxiao Lu <lujingxiao@huawei.com>


WeiZhang555:
Replace https://github.com/kata-containers/runtime/pull/416

WeiZhang555:
Note 1:

Currently I imported "github.com/WeiZhang555/cgroups", this repo is actually a fork of master branch of `github.com/containerd/cgroups`. The reason I made a fork instead of original code is:
1. latest code of `github.com/containerd/cgroups` uses a higher version of `runtime-spec`, which contains new RDMA cgroup support: https://github.com/containerd/cgroups/blob/master/rdma.go, importing this repo requires we bump runtime-spec for both kata-runtime and kata-agent.
2. latest code has bug according to test, raised a PR to fix: https://github.com/containerd/cgroups/pull/54
3. another way is we use a older version of `github.com/containerd/cgroups` without RDMA codes, according to my test, the latest code without RDMA support also misses an important function: `AddTask()`, without this, I can't do fine-granularity  resource limitation.

Combining the issues above, I will suggest:
1. we fork latest code from `github.com/containerd/cgroups` under `kata-containers`, but removes RDMA codes, including the bugfix I mentioned.

-----------------------
Note 2:

`govmm` needs a vendor, will do it once this PR looks good

WeiZhang555:
Because there's cgroup support in kata-agent, so we need to do some trick to see if this really works.

## test steps with docker:

```
# docker run --rm -it --cpu-quota 60000 --cpu-period 100000 --runtime kata progrium/stress --cpu 4  --timeout 600s
```

Top on host, you can see the qemu process takes about 60% cpu usage.

then enter `/sys/fs/cgroup/cpu/vc/<container-id>/vcpu`, modify content of `cpu.cfs_quota_us` from 60000 to 40000, you are expected to see `qemu` cpu cost will go down from 60% to 40%.  This can let you know it really works.

## test steps with k8s + cri-containerd

Use this pod spec:

```
apiVersion: v1
kind: Pod
metadata:
  name: nginx-untrusted
  annotations:
spec:
  containers:
  - name: nginx
    image: nginx
    resources:
      requests:
        memory: "64Mi"
        cpu: "250m"
      limits:
        memory: "128Mi"
        cpu: "500m"
  - name: busybox
    image: busybox
    command: ['top']
    resources:
      requests:
        memory: "64Mi"
        cpu: "100m"
      limits:
        memory: "128Mi"
        cpu: "200m"
```

There will be 2 containers inside the POD, `nginx` with a cpu quota/period of "50000/100000", and `busybox` with a cpu quota/period of "20000/100000", so value of `cpu.cfs_quota_us` and `cpu.cfs_period_us` in `/sys/fs/cgroup/cpu/vc/<sandbox-id>/vcpu/` would be "70000" and "10000",
that's expected.

## issue found

When tested with cri-containerd with above pod spec, the resource limit in cgroup is set to 0.7 core as expected, but the VM got 3 vcpu, that's not right.

Guess that's because we hotplugged two vcpus for two containers, plus default one vcpu. We should do more accurate calculation for VCPU numbers.

bergwolf:
A huge PR but it is really something we need. I'll take a closer look later today. Thanks @WeiZhang555 !

virtcontainers/api.go
```diff
@@ -122,6 +122,11 @@ func createSandboxFromConfig(ctx context.Context, sandboxConfig SandboxConfig, f
 		return nil, err
 	}
 
+	// Setup host cgroups
+	if err := s.setupCgroups(); err != nil {
```
bergwolf: This is already called in `createContainers()`. Are you trying to make sure there is host cpu cg even for an empty sandbox?

virtcontainers/cgroups.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+
+		var spec specs.Spec
+		if err := json.Unmarshal([]byte(config), &spec); err != nil {
+			return nil, err
+		}
+
+		// TODO: how to handle empty/unlimited resource?
+		// maybe we should add a default CPU/Memory delta when no
+		// resource limit is given. -- @WeiZhang555
```
bergwolf: Default cpu quota might conflict globally on the host if we exceed the total available CPU time. While users might set it to exceed, we should not make it happen unintentionally. docker/runc does not set default quota for containers either.

virtcontainers/cgroups.go
```diff
@@ -0,0 +1,192 @@
...
+				resource.CPU.Period = rc.CPU.Period
+				resource.CPU.Quota = rc.CPU.Quota
+			} else {
+				// this is an example to show how it works:
+				// container A and `resource` has quota: 5000 and period 10000
+				// here comes container B with quota 40 and period 100,
+				// so use previous period 10000 as a baseline, container B
+				// has proportional resource of quota 4000 and period 10000, calculated as
+				// delta := 40 / 100 * 10000 = 4000
+				// and final `*resource.CPU.Quota` = 5000 + 4000 = 9000
```
bergwolf: Nice! I like the proportional calculation and the comments!

bergwolf:
@WeiZhang555 Nice patch, thanks! w.r.t. the containerd bug, is it possible to push containerd/cgroups#54 forward? Also I think we should update our spec version even though we do not support rdma right now.  Then we can import from containerd directly ;)

virtcontainers/api.go
WeiZhang555: No, this `s.setupCgroups()` in `createSandboxFromConfig` is for the initial sandbox container, in this sandbox it has exactly one container. Another call of `s.setupCgroups()` is in `s.CreateContainer()`. The setup process only happens once for each container(including sandbox container), there's no duplicate call.

virtcontainers/sandbox.go
```diff
@@ -31,7 +31,9 @@ import (
 
 // vmStartTimeout represents the time in seconds a sandbox can wait before
 // to consider the VM starting operation failed.
-const vmStartTimeout = 10
+const (
+	vmStartTimeout = 10
```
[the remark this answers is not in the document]
WeiZhang555: Got it!

virtcontainers/cgroups.go
WeiZhang555: The worry is, if there's two containers inside a POD, one with quota 6000, the other one with quota -1 (unlimited), total quota will be 6000.

That means with my calculation, one container with 0.6 core + one container with unlimited cores = 0.6 core.
Not sure if this can satisfy everyone, I can only suggest user, if you set a limit for one container, you should set limit for every container too...

devimc: please file an issue

crosbymichael:
We bumped the spec in containerd, you shouldn't have any backward incompat issues

WeiZhang555:
Hi @crosbymichael , thank you for your response! Then I think bumping kata-runtime's runtime spec version is the right way, kata-runtime/agent should use same/similiar version of runtime spec with containerd :smile:

WeiZhang555:
Depends-on: https://github.com/kata-containers/agent/pull/374
Bump runtime-spec for kata-agent

sboeuf:
@WeiZhang555 kata-containers/agent#374 has been merged, I think you can revendor and rebase your commit.
Also, please remove `do-not-merge` label.

virtcontainers/api.go
bergwolf: @WeiZhang555 There is a code path `createSandboxFromConfig` -> `s.createContainers` -> `createContainer` -> `setupCgroups` for the initial container in the sandbox as well.

WeiZhang555:
Code is ready, need vendor `govmm` https://github.com/intel/govmm/pull/52 later once it's merged

virtcontainers/api.go
WeiZhang555: `createContainer` -> `setupCgroups` doesn't exist.

Maybe you mixed `createContainer()` from `virtcontainers/container.go` with the one from `cli/create.go` ?

bergwolf: oops, you are right! I mixed the two createcontainer() functions in sandbox.go and container.go, though they have a initial capital difference. Sry for the noise...

virtcontainers/cgroups.go
```diff
@@ -0,0 +1,192 @@
...
+	commonParent cgroups.Cgroup
+	sandboxSub   cgroups.Cgroup
+	vcpuSub      cgroups.Cgroup
+}
+
+func (s *Sandbox) newCgroups() error {
+	// New will still succeed when cgroup exists
+	// create common parent for all kata-containers
+	// e.g. /sys/fs/cgroup/cpu/vc
+	parent, err := cgroups.New(cgroups.V1,
```
devimc: cgroup should be created in the path specified by `cgroupsPath` in the `config.json`, Q. why you are not using it?

jshachm: @WeiZhang555 @devimc  maybe we should consider  `docker cli` `Cgroup Parent` option.  ^^

WeiZhang555: Considering one valid use case though we don't recommend it at all:
one POD with 3 containers: c1, c2, c3

c1 has cgroups path: `/mycontainer/c1`
c2 has cgroups path: `/hello/c2`
c3 has cgroups path: `/world/c3`

So how to find a shared parent for 3 containers to put the qemu in?

virtcontainers/cgroups.go
```diff
@@ -0,0 +1,192 @@
...
+	tids, err := s.hypervisor.getThreadIDs()
+	if err != nil {
+		return err
+	}
+
+	if tids == nil {
+		// TODO: we only support qemu in this scenario, for unsupported hypervisor,
+		// just ignore and print a warning
+		s.Logger().Warning("failed to get thread ids from hypervisor")
+		return nil
```
jodh-intel: I wonder if this should be an error? That way, we'll know as soon as we try to add support for a new hypervisor what the problem is :)

virtcontainers/cgroups.go
```diff
@@ -0,0 +1,192 @@
...
+		// maybe we should add a default CPU/Memory delta when no
+		// resource limit is given. -- @WeiZhang555
+		if spec.Linux == nil || spec.Linux.Resources == nil {
+			continue
+		}
+		// calculate cpu quota and period
+		rc := spec.Linux.Resources
+		if rc.CPU != nil && rc.CPU.Quota != nil && rc.CPU.Period != nil &&
+			*rc.CPU.Quota > 0 && *rc.CPU.Period > 0 {
+			if resource.CPU.Period == nil {
```
jodh-intel: I'd consider putting this logic into a function. That would:

- encapsulate the calculation more clearly.
- allow you to write a unit test specifically for the calculation.
- reduce code complexity.

WeiZhang555:
I will be in vacation in next 3 weeks, will apply changes after I'm back!

virtcontainers/cgroups.go
WeiZhang555: @jodh-intel The function is created, since `TestMergeSpecResource` covers all the cases, I didn't add new unit tests.

WeiZhang555:
@jodh-intel Your comments are addressed. PTAL.

virtcontainers/cgroups.go
```diff
@@ -0,0 +1,192 @@
...
+	"fmt"
+
+	"github.com/containerd/cgroups"
+	"github.com/kata-containers/runtime/virtcontainers/pkg/annotations"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+const (
+	vcpuGroupName       = "vcpu"
+	defaultCgroupParent = "/vc"
```
amshinde: @WeiZhang555 Maybe add this under `kata` to not cause any conflicts?

virtcontainers/cgroups.go
amshinde: @devimc I know we were seeing some file descriptor leaks when ignoring `cgroupsPath` in the `config.json` back in the day. Do we have tests in the CI for detecting that? I dont recall the root-cause, but do you think this implementation should still take care of that?

virtcontainers/cgroups.go
```diff
@@ -0,0 +1,192 @@
...
+func (s *Sandbox) destroyCgroups() error {
+	// first move all processes in subgroup to parent in case live process blocks
+	// deletion of cgroup
+	if s.cgroup == nil {
+		// no need to destroy
+		s.Logger().Warningf("cgroup not initialized, no need to destroy")
+		return nil
+	}
+
+	if err := s.cgroup.sandboxSub.MoveTo(s.cgroup.commonParent); err != nil {
```
amshinde: @WeiZhang555 Can you add some comments explaining what you are doing here?

virtcontainers/cgroups.go
WeiZhang555: There're no CI tests for this yet. In normal cases, `destroyCgroups` will ensure all cgroup files being deleted if this function can be invoked.

virtcontainers/cgroups.go
WeiZhang555: This sounds good.

virtcontainers/cgroups.go
WeiZhang555: ok

WeiZhang555:
/test

@amshinde comments addressed. PTAL.

jshachm:
LGTM.  Greate job and have a nice vacation~~ ^ ^

amshinde:
@devimc Please take a look.

virtcontainers/cgroups.go
jodh-intel: Reading https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#control-groups, we do have to honour `cgroupsPath`.

But I'm a little confused here...

Docker doesn't have the concept of pod/sandbox so we create one VM instance per container.

- If the user runs "`docker run ...`" Docker will create a `config.json` specifying a cgroup of `/docker/${cid}/` we can create `/sys/fs/cgroup/docker/${cid}/` for the hypervisor and the shim.

- If the user then runs "`docker exec -ti ${cid} sh`", the second shim process we create should also be added to the same cgroup (which doesn't appear to be happening today!?!)

In other words, why does it matter than Docker doesn't have the concept of pods? K8s does, but then it wouldn't be generating a `cgroupsPath = "/docker/${cid}/`" ...?

devimc: I don't know what are going to be the consequences of whether `cgroupsPath` is ignored, what I'm sure is that systemd-cgroups won't be honoured and cgroup can't be updated with `systemctl`

WeiZhang555: @jodh-intel I do want to honor `cgroupsPath` but I didn't find a reasonable logic for handling this.

> Docker doesn't have the concept of pod/sandbox so we create one VM instance per container.

That's not 100% true. We do want to run K8S+docker, so we should still regard docker as one VM instance per POD.

>In other words, why does it matter than Docker doesn't have the concept of pods? K8s does, but then it wouldn't be generating a cgroupsPath = "/docker/${cid}/" ...?

I can't find a common logic for handling this, consider a "best" situation:

```
c1 has cgroups path: /POD1/c1
c2 has cgroups path: /POD1/c2
c3 has cgroups path: /POD1/c3
```

According to this configuration, we can construct a cgroup hierarchy as:

```
/sys/fs/cgroup/cpu
|--POD1
     |--vcpu
     |--c1
     |--c2
     |--c3
```

we can put VM process in `vcpu` cgroup to limit the qemu processes, and put shim process of container c1/c2/c3 into cgroup c1/c2/c3, this makes lots of sense and makes a perfect logic.

But take the worst situation as this:
```
c1 has cgroups path: /mycontainer/c1
c2 has cgroups path: /hello/c2
c3 has cgroups path: /world/c3
```

Then how to hornor the `cgroupPath` and how to find a good common parent? What I can think of is we only honor cgroupPath of `c1` and ignores c2 and c3.

```
/sys/fs/cgroup/cpu
|--mycontainer
     |--vcpu
     |--c1
     |--c2
     |--c3
```

I'm not confident enough that this makes sense or not, and as we create the container of each pod one by one, we can't control how user will define the cgroupsPath for each container.

raravena80:
@WeiZhang555 ping from your weekly Kata herder.

WeiZhang555:
Rebased. 

ping @devimc , what do you think of this https://github.com/kata-containers/runtime/pull/734#discussion_r224051455 ?

WeiZhang555:
It seems we got 2 LGTMs now. 
@sboeuf @jodh-intel @egernst do you want to take another look at this?

Two legacy issue needs to be resolved in following PR:
* honor `cgroupsPath` in config.json
* honor `--systemd-cgroup` flags

virtcontainers/cgroups.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil || tids == nil {
+		return fmt.Errorf("failed to get thread ids from hypervisor: %v", err)
+	}
+
+	// use Add() to add vcpu thread to s.cgroup, it will write thread id to
+	// `cgroup.procs` which will move all threads in qemu process to this cgroup
+	// immediately as default behaviour.
+	if len(tids.vcpus) > 0 {
+		if err := s.cgroup.sandboxSub.Add(cgroups.Process{
+			Pid: tids.vcpus[0],
```
caoruidong: So are vcpus[] thread ids or cpu ids? I'm a little confused.

WeiZhang555: They are thread ids

WeiZhang555:
CI passed. Merging this as we have enough LGTMs and it has been pending for long time.

crosbymichael:
Yeah! Congrats

liangxianlong:
@WeiZhang555 Hello，I have a question. I want to realize  “vcpupin” in kata-runtime , can i Can I reuse your code? 
/sys/fs/cgroup
|---cpuset
   |---vc
      |---<sandbox-id>
         |--vcpu
      |---<sandbox-id>

liangxianlong:
@WeiZhang555 I test your code, i run this command  "docker run -ti --cpuset-cpus 1 busybox /bin/sh",and the i see the directory on my host: "/sys/fs/cgroup/cpuset/kata/e76fe6c34d9e75333b06091e0a68095a470ba4333c3de4440e99010029f1674a/vcpu". But if we have two vcpus, i think the directory should like this "/sys/fs/cgroup/cpuset/kata/e76fe6c34d9e75333b06091e0a68095a470ba4333c3de4440e99010029f1674a/vcpu0" and "/sys/fs/cgroup/cpuset/kata/e76fe6c34d9e75333b06091e0a68095a470ba4333c3de4440e99010029f1674a/vcpu0"

WeiZhang555:
@liangxianlong So you are trying to support cpuset, currently only cfs_quota and cfs_period are supported.

cpuset support could be more complicated, it depends on our cgroup setting policy.

1. we have host cgroup and guest cgroup support, supporting cpuset need coordinate between guest and host cgroup
2. suppose a container with 1.5 core and cpuset "0-1", we don't need to set cpuset seperately for vcpu 0 and 1, we can put vcpus in same cgroup and write "0-1" in cgroup config, you don't have to put them in seperate dir

liangxianlong:
@WeiZhang555 Now，i don't care container. In my test, after "docker run -ti --cpuset-cpus 1 busybox /bin/sh ", there will be two results, (1)the container's process in vm is bound to vcpu1, (2) build a diretory on host "/sys/fs/cgroup/cpuset/kata/${sandbox-id}/vcpu". The vcpu is just a qemu thread,so if we want to realize this: Bind the vcpu to the physical cpu;Does the code need some modifications?

WeiZhang555:
@liangxianlong This is achievable, you can enhance the code to add cpuset support, it should be easy.

liangxianlong:
> @liangxianlong This is achievable, you can enhance the code to add cpuset support, it should be easy.

thanks. Another question, if i run "docker run -ti busybox /bin/sh ", two directory will be built on my host, 
(1) /sys/fs/cgroup/cpuset/kata/${sandbox-id}/vcpu; (2)  /sys/fs/cgroup/cpu/kata/${sandbox-id}/vcpu.  I think it should only build "/sys/fs/cgroup/cpu/kata/${sandbox-id}/vcpu". Why does the code create two directories?

WeiZhang555:
@liangxianlong That's because `github,com/containerd/cgroups` doesn't give us an interface to only build `cpu/kata/vcpu` without building `cpuset/kata/vcpu`, it's a miss of cgroups lib api.

By the way, it's better to open another issue for discussing and tracking this, discussing under a closed PR may be ignored by other people, this is not right place :smile:

liangxianlong:
@WeiZhang555 think s! I'm new to kata and interested in it，so please bear with me. xixi.

PR #744, opened 2018-09-18
versions: Update golang to 1.10.4

### LINKED ISSUE #148: namespaces: How to enter them in a foolproof way

jodh-intel:
Move to golang version 1.10.4 -- the oldest stable golang release at the
time of writing -- since golang 1.10+ is needed to make namespace
handling safe.

See:

- https://github.com/golang/go/issues/20676
- https://github.com/golang/go/commit/2595fe7fb6f272f9204ca3ef0b0c55e66fb8d90f

Also bumped `languages.golang.meta.newest-version` to golang version
1.11, which is the newest stable release at the time of writing.

Fixes #148.

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -138,9 +138,12 @@ externals:
 
   cri-containerd:
     description: |
-      Containerd Plugin for Kubernetes Container Runtime Interface
+      Containerd Plugin for Kubernetes Container Runtime Interface.
+
+      Note that the current version is required for golang 1.10.2
+      (see https://github.com/containerd/cri/pull/941).
     url: "https://github.com/containerd/cri"
-    version: "v1.0.5"
+    version: "8b0d53c09c41d9fbc3b3896548ecf011518e3c42"
     meta:
       containerd-version: "1.1.3"
 
--- a/versions.yaml
+++ b/versions.yaml
@@ -186,9 +186,10 @@ languages:
   golang:
     description: "Google's 'go' language"
     notes: "'version' is the default minimum version used by this project."
-    version: "1.9.7"
+    issue: "https://github.com/golang/go/issues/20676"
+    version: "1.10.4"
     meta:
-      newest-version: "1.10"
+      newest-version: "1.11"
 
 specs:
   description: "Details of important specifications"
--- a/virtcontainers/monitor.go
+++ b/virtcontainers/monitor.go
@@ -18,9 +18,9 @@ type monitor struct {
 	sandbox       *Sandbox
 	checkInterval time.Duration
 	watchers      []chan error
+	wg            sync.WaitGroup
 	running       bool
 	stopCh        chan bool
-	wg            sync.WaitGroup
 }
 
 func newMonitor(s *Sandbox) *monitor {
--- a/virtcontainers/pkg/mock/cc_proxy_mock.go
+++ b/virtcontainers/pkg/mock/cc_proxy_mock.go
@@ -25,7 +25,6 @@ type CCProxyMock struct {
 	sync.Mutex
 
 	t              *testing.T
-	wg             sync.WaitGroup
 	connectionPath string
 
 	// proxy socket
@@ -43,6 +42,8 @@ type CCProxyMock struct {
 	Signal           chan ShimSignal
 	ShimDisconnected chan bool
 	StdinReceived    chan bool
+
+	wg sync.WaitGroup
 
 	stopped bool
 }


egernst:
Like the idea - let's see if the CI can be agreeable.

caoruidong:
I think this can only be fixed by reordering field in this structs.

gnawux:
lgtm

I think the maligned crying is not related to this patch, right?

jodh-intel:
/me tweaks branch to hopefully stop the linters bleating...

jodh-intel:
CI failed due to a factory test failure:

```
--- FAIL: TestNewFactory (0.01s)
    testing.go:771: race detected during execution of test
```

/cc @bergwolf.

jodh-intel:
Sigh - go 1.10 is spotting issues in other projects code and failing our tests ;((

```
# github.com/containerd/cri/integration
integration/image_load_test.go:39: T.Skip call has possible formatting directive %v
```

grahamwhaley:
CI turbines bombed @jodh-intel Is this still something you will pursue?

bergwolf:
@jodh-intel Any updates?

grahamwhaley:
@bergwolf See #806 - @amshinde has posted some patches upstream to containerd/cri - we need to get them pulled in so we can bump the golang version and not fail the build.

grahamwhaley:
@bergwolf - yes, that is the merged PR. Over on #806 I have the patches to bump that versions.yaml and golang version change. I'll chase that up - I need to fix at least the `yq` dep for travis.

raravena80:
@jodh-intel ping from your weekly Kata herder.

jodh-intel:
We now have a cyclic dependency since this PR is failing due to https://github.com/kata-containers/tests/pull/843 (and *vice versa*).

grahamwhaley:
OK, we know this will fail on the xurl stuff, which we need to land this to fix:

```
INFO: Installing xurls utility
package mvdan.cc/xurls/v2: cannot find package "mvdan.cc/xurls/v2" in any of:
	/usr/local/go/src/mvdan.cc/xurls/v2 (from $GOROOT)
	/tmp/jenkins/workspace/kata-containers-runtime-ubuntu-16-04-PR/go/src/mvdan.cc/xurls/v2 (from $GOPATH)
```

@jodh-intel and I propose we force merge. Of course it carries some risk, but right now the CI is broken on the xurl item, and we can merge nothing else anyway...

Issue #756, opened 2018-09-19
The builtin kata proxy will be missed to start 


lifupan:
# Description of problem

It shouldn't set kataAgent.state.URL in its configure() method
for builtin kata proxy, otherwise the following check of whether
is it nil in startProxy() will return directly and failed to
start builtin proxy which will log the qemu's console.

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #757, opened 2018-09-19
virtcontainers: fix the issue of missing starting builtin proxy

lifupan:
It shouldn't set kataAgent.state.URL in its configure() method
for builtin kata proxy, otherwise the following check of whether
is it nil in startProxy() will return directly and failed to
start builtin proxy which will log the qemu's console.

Fixes: #756

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -247,7 +247,6 @@ func (k *kataAgent) configure(h hypervisor, id, sharePath string, builtin bool, 
 
 	if builtin {
 		k.proxyBuiltIn = true
-		k.state.URL, _ = k.agentURL()
 	}
 
 	// Adding the shared volume.
--- a/virtcontainers/kata_agent_test.go
+++ b/virtcontainers/kata_agent_test.go
@@ -649,6 +649,7 @@ func TestAgentConfigure(t *testing.T) {
 
 	err = k.configure(h, id, dir, true, c)
 	assert.Nil(err)
+	assert.Empty(k.state.URL)
 
 	err = k.configure(h, id, dir, false, c)
 	assert.Nil(err)


virtcontainers/vm.go
```diff
@@ -128,7 +128,7 @@ func NewVM(ctx context.Context, config VMConfig) (*VM, error) {
 	// 2. setup agent
 	agent := newAgent(config.AgentType)
 	vmSharePath := buildVMSharePath(id)
-	err = agent.configure(hypervisor, id, vmSharePath, isProxyBuiltIn(config.ProxyType), config.AgentConfig)
```
bergwolf: This breaks vm factory because vm template needs to use a builtin proxy and the builtin flag and proxy url is not set. The factory code cannot call agent.init() because that call is bound with a sandbox. That's the reason I reinitialize agent's proxy fields in the `configure` method.

lifupan: It seems the vm factory only needs to set the builtin flag in the configure() method and
the proxy url will be set in the following setProxy(), so what I should do is just remove the 
proxy url initialization to fix this issue?

bergwolf: Yes, I think so. Please make sure the PR passes kata-containers/tests#578 though.

virtcontainers/kata_agent.go
```diff
@@ -246,7 +246,6 @@ func (k *kataAgent) configure(h hypervisor, id, sharePath string, builtin bool,
 
 	if builtin {
 		k.proxyBuiltIn = true
-		k.state.URL, _ = k.agentURL()
```
jodh-intel: It's a tiny change, but it would be great if you could add a tiny unit test to `TestAgentConfigure()` :)

lifupan: Hi @jodh-intel Done

jodh-intel:
Thanks @lifupan.

lgtm

PR #761, opened 2018-09-19
virtcontainers: support vhost and physical endpoints in AddInterface

### LINKED ISSUE #758: network: AddInterface only supports VirtualEndpoints
If we look at the way `AddInterface()` has been implemented, it will always create a `virtualEndpoint`, no matter the type of device backing the interface that we're trying to add. Unfortunately, this means neither `vhost-user` nor `vfio` can be leveraged in this case.

This needs to be fixed either by adding a new field `devType` to the `grpc.Interface` structure passed as an argument of `AddInterface()`, or by adding a new argument `devType` directly to the function itself.

@bergwolf @caoruidong @WeiZhang555 @amshinde WDYT?

caoruidong:
Support to hotplug vhost and physical endpoints by AddInterface API

Fixes #758

### DIFF
--- a/virtcontainers/network.go
+++ b/virtcontainers/network.go
@@ -152,6 +152,7 @@ type Endpoint interface {
 	Name() string
 	HardwareAddr() string
 	Type() EndpointType
+	PciAddr() string
 
 	SetProperties(NetworkInfo)
 	Attach(hypervisor) error
@@ -178,6 +179,7 @@ type PhysicalEndpoint struct {
 	BDF                string
 	Driver             string
 	VendorDeviceID     string
+	PCIAddr            string
 }
 
 // VhostUserEndpoint represents a vhost-user socket based network interface
@@ -189,6 +191,7 @@ type VhostUserEndpoint struct {
 	IfaceName          string
 	EndpointProperties NetworkInfo
 	EndpointType       EndpointType
+	PCIAddr            string
 }
 
 // Properties returns properties for the veth interface in the network pair.
@@ -210,6 +213,11 @@ func (endpoint *VirtualEndpoint) HardwareAddr() string {
 // Type identifies the endpoint as a virtual endpoint.
 func (endpoint *VirtualEndpoint) Type() EndpointType {
 	return endpoint.EndpointType
+}
+
+// PciAddr returns the PCI address of the endpoint.
+func (endpoint *VirtualEndpoint) PciAddr() string {
+	return endpoint.PCIAddr
 }
 
 // SetProperties sets the properties for the endpoint.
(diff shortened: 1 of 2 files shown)


virtcontainers/network.go
```diff
@@ -212,6 +215,11 @@ func (endpoint *VirtualEndpoint) Type() EndpointType {
 	return endpoint.EndpointType
 }
 
+// PciAddr returns the PCI address of the endpoint.
+func (endpoint *VirtualEndpoint) PciAddr() string {
```
devimc: `s/PciAddr/PCIAddr` ?

jcvenegas:
CI failing on `vm template tests`  https://github.com/kata-containers/tests/issues/760 not related with this PR.

virtcontainers/sandbox.go
```diff
@@ -17,8 +17,8 @@ import (
 	"syscall"
 
 	"github.com/containernetworking/plugins/pkg/ns"
-	specs "github.com/opencontainers/runtime-spec/specs-go"
```
jcvenegas: I think this change is not needed, is this an error from a formatting tool?

virtcontainers/sandbox.go
```diff
@@ -17,8 +17,8 @@ import (
...
-	opentracing "github.com/opentracing/opentracing-go"
```
jcvenegas: Same comment, do this changed needed ?

virtcontainers/sandbox.go
caoruidong: Yes. I will discard this.

virtcontainers/network.go
caoruidong: That will be same with field name and makes an error.

virtcontainers/sandbox.go
```diff
@@ -1105,7 +1105,32 @@ func (s *Sandbox) AddInterface(inf *grpc.Interface) (*grpc.Interface, error) {
 		return nil, err
 	}
 
-	endpoint, err := createVirtualNetworkEndpoint(len(s.networkNS.Endpoints), inf.Name, s.config.NetworkConfig.InterworkingModel)
+	var endpoint Endpoint
```
bergwolf: This chunk of code is copied from part of `createEndpointsFromScan`. Please add a new function for it in network.go and just call the new function here.

caoruidong: Done. There is a closure.

jcvenegas:
@devimc  @bergwolf  take a look again :)

sboeuf:
@bergwolf @amshinde PTAL

Issue #769, opened 2018-09-21
RFC: change 9pfs mount option to cache=mmap


bergwolf:
# Summary

I have done a some compatibility tests for kata 9pfs and I think we should change 9pfs mount option from default `cache=none` to `cache=mmap`. It does not give worse POSIX compliance (compared to cache=none) but allows us to run `dnf install` and `mariadb:latest` images.

We still need the 9pfs open-delete-fstat workaround. Even though it gives worse fstest results but it works around some important common usages otherwise we'll fail both `apt update` and `mariadb:latest`.

|setup docker/overlayfs |9pfs cache option|guest kernel|fstest failures| apt update| dnf install|launch mariadb|
|:-:|:-:|:-:|:-:|:-:|:-:|:-:|
|runc| N/A| N/A|2/8789|Y|Y|Y|
| kata| cache=none| upstream 4.14.57| 8/8789| N| N| N|
| kata| cahce=mmap| upstream 4.14.57| 8/8789| N| Y| N|
| kata| cache=none| 9p patched 4.14.67.11| 30/8789| Y| N| N|
| kata| cache=mmap| 9p patched 4.14.67.11| 30/8789| Y| Y| Y|

## 0. test setup
0.0 check storage driver
```
$docker info|grep Storage
Storage Driver: overlay2
```

0.1 clone pjdfstest:
`git clone git@github.com:pjd/pjdfstest.git`

0.2 run docker:
`docker run -it -v /path/to/pjdfstest:/fstest ubuntu`

0.3 run kata:
`docker run -it --runtime kata -v /path/to/pjdfstest:/fstest ubuntu`

0.4 run `apt update` test:
`docker run --runtime kata -it --rm ubuntu apt update`

0.5 run `dnf install -y strace` test:
`docker run --runtime kata -it --rm fedora dnf install -y strace`

0.6 run mariadb test:
`docker run --runtime kata --rm -it --env MYSQL_ALLOW_EMPTY_PASSWORD=yes mariadb`

## 1. docker with overlayfs
```
Test Summary Report
-------------------
/pjdfstest/tests/symlink/03.t        (Wstat: 0 Tests: 6 Failed: 2)
  Failed tests:  1-2
Files=232, Tests=8789, 200 wallclock secs ( 2.18 usr  0.82 sys + 41.07 cusr 15.38 csys = 59.45 CPU)
Result: FAIL
```
[...]

PR #770, opened 2018-09-21
agent: change 9pfs mount option to cache=mmap

bergwolf:
It does not give better pjdfstest results but allows us to pass
ubuntu `apt update`, fedora `dnf install`, and also launch `mariadb:latest`.

Fixes: #769

cc @sboeuf @WeiZhang555 @gnawux @egernst

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -56,7 +56,7 @@ var (
 	kata9pDevType        = "9p"
 	kataBlkDevType       = "blk"
 	kataSCSIDevType      = "scsi"
-	sharedDir9pOptions   = []string{"trans=virtio,version=9p2000.L", "nodev"}
+	sharedDir9pOptions   = []string{"trans=virtio,version=9p2000.L,cache=mmap", "nodev"}
 	shmDir               = "shm"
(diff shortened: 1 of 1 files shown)


gnawux:
+1

1. `mmap` does help some in some daily case: `apt`, `dnf`, `mariadb`
2. `mmap` does not make the result of `fstest` worse than `none`

So I support the change.

jodh-intel:
Thanks for raising @bergwolf! I think we'd better wait for @grahamwhaley to comment on this as I believe he also looked at these settings a while back...

grahamwhaley:
Indeed, thanks for the research and numbers. I need to go study what mmap mode means for 9pfs. I need to check if it affects the cache mode at all, and if we see that as a risk etc.

jcvenegas:
Given it fixes key functionalities lets merge it before next release @jodh-intel  @grahamwhaley .

grahamwhaley:
Travis on jollies? We have enough acks - you want to push this through @jodh-intel ?

jodh-intel:
lgtm

/me looks left and right and then leans nonchalantly on the merge button...

PR #781, opened 2018-09-24
vc: qemu: Add config option to choose entropy source.

jcvenegas:
Today we use `/dev/random` as entropy source.
This is a blocking entropy source, when the
host is getting a low amount of entropy the
Kata VM startup takes longer because block
trying to get entropy from the source.

This change allow to change the entropy source.
For example: urandom
Fixes: #702

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,6 +136,8 @@ DEFMEMSLOTS := 10
 DEFBRIDGES := 1
 #Default network model
 DEFNETWORKMODEL := macvtap
+#Default entropy source
+DEFENTROPYSOURCE := /dev/urandom
 
 DEFDISABLEBLOCK := false
 DEFBLOCKSTORAGEDRIVER := virtio-scsi
@@ -222,6 +224,8 @@ USER_VARS += DEFENABLEDEBUG
 USER_VARS += DEFDISABLENESTINGCHECKS
 USER_VARS += DEFMSIZE9P
 USER_VARS += DEFHOTPLUGVFIOONROOTBUS
+USER_VARS += DEFENTROPYSOURCE
+
 
 V              = @
 Q              = $(V:1=)
@@ -323,6 +327,7 @@ const defaultEnableDebug bool = $(DEFENABLEDEBUG)
 const defaultDisableNestingChecks bool = $(DEFDISABLENESTINGCHECKS)
 const defaultMsize9p uint32 = $(DEFMSIZE9P)
 const defaultHotplugVFIOOnRootBus bool = $(DEFHOTPLUGVFIOONROOTBUS)
+const defaultEntropySource = "$(DEFENTROPYSOURCE)"
 
 // Default config file used by stateless systems.
 var defaultRuntimeConfiguration = "$(CONFIG_PATH)"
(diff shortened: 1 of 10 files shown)


virtcontainers/qemu_arch_base.go
```diff
@@ -525,7 +525,8 @@ func (q *qemuArchBase) appendVFIODevice(devices []govmmQemu.Device, vfioDev conf
 func (q *qemuArchBase) appendRNGDevice(devices []govmmQemu.Device, rngDev config.RNGDev) []govmmQemu.Device {
 	devices = append(devices,
 		govmmQemu.RngDevice{
-			ID: rngDev.ID,
+			ID:       rngDev.ID,
+			Filename: "/dev/urandom",
```
sboeuf: @jcvenegas looks good, but don't you think we should make this configurable instead of hardcoding in govmm ?
We should have the opportunity to select /dev/urandom from Kata code, WDYT?

jcvenegas: Agree, updating.

WeiZhang555:
The code looks good, but I have a question. Do we need a configuration option? I mean, in which situation should we choose /dev/random against /dev/urandom ? Will be /dev/urandom always a prefered option?

sboeuf:
@laijs @WeiZhang555 I have asked the question about `/dev/urandom` on the kata-dev mailing list. Please give more details about the potential issue related to using `/dev/urandom`

jodh-intel:
Would it be worth atleast logging the value of `/proc/sys/kernel/random/entropy_avail` on container create as a debug aid?

I suspect we can't reliably determine how much entropy we'd actually need. But if we could, if the system entropy falls below that threshold, we could make that a hard error.

cli/kata-env.go
```diff
@@ -319,6 +320,7 @@ func getHypervisorInfo(config oci.RuntimeConfig) HypervisorInfo {
 		Msize9p:           config.HypervisorConfig.Msize9p,
 		UseVSock:          config.HypervisorConfig.UseVSock,
 		MemorySlots:       config.HypervisorConfig.MemSlots,
+		EntropySource:     config.HypervisorConfig.EntropySource,
```
jodh-intel: - should this be displayed in `kata-env`?
- should we do any validation on this before handing to the hypervisor? (aka does it exist, is it a char device?)
  Looking at govmm, it also doesn't check this but that may well be by design since presumably qemu will :)
  https://github.com/intel/govmm/blob/master/qemu/qemu.go#L1035..L1042

/cc @markdryan

Makefile
```diff
@@ -136,6 +136,8 @@ DEFMEMSLOTS := 10
 DEFBRIDGES := 1
 #Default network model
 DEFNETWORKMODEL := macvtap
+#Default entropy source
+DEFENTROPYSOURCE := /dev/random
```
egernst: should the default be /dev/urandom?

sboeuf: Oh sorry, I'd like `urandom` to be the default here, and we should add some comments related to `/dev/random`, explaining this can be used but with some limitations.

cli/kata-env.go
jcvenegas: @jodh-intel qemu complains about it,  `OCI runtime create failed: qemu-system-x86_64: -object rng-random,id=rng0,filename=/tmp/r: Could not open '/tmp/r': No such file or directory: unknown`

bergwolf:
Per Ted's recommendation on the mailing list, we are proceeding the patch with just seeding virtio-rng with `/dev/urandom`[1]. Let's remove the `wip` label and merge it since it has already got enough ACKs.

[1] http://lists.katacontainers.io/pipermail/kata-dev/2018-September/000439.html

Issue #791, opened 2018-09-27
Support memory hotplugged removable


BetaXOi:
Since kata support hotplug memory by update([#624](https://github.com/kata-containers/runtime/pull/624)),it's easy to increase memory but not support to reduce.
I modify some code to support remove the memory hotplugged.

But the feature maybe conflict with udev rules in VM(the kata container image build by `osbuilder`):
- centos
There is a memory udev rules(/usr/lib/udev/rules.d/40-redhat.rules) in VM already, it will cause the memory hotplugged maybe mapped in ZONE_DMA32 instead of ZONE_Movable.
- clearlinux
There is no any memory udev rules in VM, the memory hotplugged will be mapped in ZONE_Movable, so you can remove the memory safly in the case.
- other
no test yet 

I have no good solution how to fix the conflict, suggestions anyone?

jcvenegas:
@BetaXOi hey thanks for create the issue I am trying to follow a different aprouch here using `virtio-balloon` see #790

jcvenegas:
@BetaXOi correct me if I am wrong but, one thing that is difficult is that even if we request to the OS, sometimes is not possible to remove the slot because memory is fragmented and if it is possible this may be slow because the kernel has to move pages to only one slot before  allow hotplug  the device.

jcvenegas:
@BetaXOi  Take a look to #790

jcvenegas:
@BetaXOi interesting! I will read it tomorrow morning, seems that we will need some measures to compare this the two options. I was worried that memory can not being hot removed (I've tried was slow and sometimes not possible)but in theory your PR https://github.com/kata-containers/agent/pull/386 will help to this. I will test your patches tomorrow. If you find additional benefits  please feel free to add it. 

Something my first idea was combination of both hotplug when we needed more memory and balloon to reduce the memory. But now seeeing that is possible to hot remove I want to evaluate both.

PR #792, opened 2018-09-27
Support the memory hotplugged remove

BetaXOi:
Fix: #791


amshinde:
@BetaXOi Can you fix your commit message to include a subsystem and a description.
Please refer to the Contribution guidelines here:
https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#patch-format

cc @jcvenegas Can you take a look, since you have looking at memory hotplug/unplug.

bergwolf:
I think we've decided to look at balloon and virtio-mem first.

jcvenegas:
@BetaXOi as first step we are going to balloon + hotplug , I still would like to evaluate this approach in order to remove memory slots, but first I want to get numbers about the time it takes to remove an slot
under an stressed system. Lets work on that to get a clear about idea about it :).   But also as @bergwolf mentioned, for longer term solution we are evaluating to virtio-mem.

raravena80:
@BetaXOi ping, any updates?

Issue #790, opened 2018-09-26
Use memory ballooning when memory is reduced


jcvenegas:
# Description of problem

When the memory is updated to be reduced today update a memory cgroup inside of the VM this
will prevent the process keep using more memory but if the VM already took the memory before reduce it we never take back memory to the host. 

# Expected result

Ask the VM to return  reduced memory from the VM.

To do this we should start the VM with balloon enabled.
When the memory is updated (via memory update or delete container) request via QMP.

# Actual result

The memory keeps as part of the Guest. 
---
\cc @clarecch  @linzichang @bergwolf

sboeuf:
@jcvenegas is the `virtio-balloon` capable of keeping a specific size? I'm asking because the risk I see using ballooning here is that we will rely on the host to make choices about the size of the balloon for our several VM(aka containers).
And if we think about the case where all guests need the full amount of memory from their own VM, the host will adjust the balloon for some VMs, meaning some other will run out of memory.
I'm worried that ballooning will generate some bad QoS in such cases.

jcvenegas:
@sboeuf  the idea is to send a qmp command  to reduce guest memory here
`qemu.hotplugMemory(remove, memory_bytes)`
https://github.com/kata-containers/runtime/blob/master/virtcontainers/qemu.go#L1057

sboeuf:
@jcvenegas yes as I discussed online, this makes sense ;)

clarecch:
@jcvenegas What's the difference between return memory and hot remove memory?

grahamwhaley:
@clarecch I my understanding:
- hot remove memory removes the memory (DIMM/slot) from the guest memory map. When you do a `free` in the guest you will see the 'total' go down. You are effectively 'physically unplugging' memory in the guest machine. Once the memory is removed, all the physical host side memory backing up that DIMM/slot should be handed back to the host as 'free'.
- return memory (ballooning) finds the empty/un-used memory in the guest (memory that no process is using), and unmaps the physical memory on the host side that is 'backing' that memory.  When you do 'free' in the Guest, you will see no difference - but, the memory listed in the 'free' column may now not have any physical host memory backing it. The benefit here is that memory that VMs are not actually using can be passed back to the host, and used by other VMs. Running 'free' on the Host, you hould see the 'free' column number increase.

Both of these are tricky for different reasons :-)

clarecch:
Thanks for your explanation@grahamwhaley , very fascinating.

[...]

Issue #788, opened 2018-09-26
Support elastic memory hotplug


linzichang:
Since we make memory update and memory limit working base on memory hotplug(see #624 #786), and memory hotplug need to be memblock align, which is defined by linux kernel and can be read from `/sys/devices/system/memory/block_size_bytes`. But users may not let memory to be memblock align, and actually, they shouldn't know what is "memblock". Now we make every update or limit to be forcing on memblock align to make hotplug success. It isn't good enough and waste memory. So we need a more elastic way, just like https://github.com/kata-containers/runtime/pull/624#issuecomment-419656853 mentions.

PR #793, opened 2018-09-27
update: Return memory to host on memory update.

jcvenegas:
If memory update will reduce memory we want to get memory
back to the host and restrict the guest to use more memory.
```bash
$ docker exec -ti kata-1 free -h
              total        used        free      shared  buff/cache   available
Mem:           5.9G         23M        5.8G        8.2M         18M        5.8G
Swap:            0B          0B          0B
$ docker update --memory 2G kata-1
kata-1
$ docker exec -ti kata-1 free -h
              total        used        free      shared  buff/cache   available
Mem:           1.9G         56M        1.8G        8.2M         18M        1.8G
Swap:            0B          0B          0B
```

Fixes: #790 

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -68,10 +68,6 @@ type QemuState struct {
 	UUID                 string
 	HotplugVFIOOnRootBus bool
 	VirtiofsdPid         int
-	// Flag to know if a VFIODev is assigned used by features that are not
-	// compatible with it, e.g. resize the balloon to return memory back is
-	// not recommended.
-	VFIODeviceCounter int
 }
 
 // qemu is an Hypervisor interface implementation for the Linux qemu hypervisor.
@@ -1134,15 +1130,11 @@ func (q *qemu) hotplugVFIODevice(device *config.VFIODev, op operation) (err erro
 	if err != nil {
 		return err
 	}
-	if q.state.VFIODeviceCounter < 0 {
-		return fmt.Errorf("VFIODeviceCounter is < 0 (%d) ", q.state.VFIODeviceCounter)
-	}
 
 	devID := device.ID
 
 	if op == addDevice {
 
-		q.state.VFIODeviceCounter++
 		// When HasVFIODevice is set balloon size is set to maximal memory
 		q.updateMemoryBalloon(0)
 		// In case HotplugVFIOOnRootBus is true, devices are hotplugged on the root bus
@@ -1188,10 +1180,6 @@ func (q *qemu) hotplugVFIODevice(device *config.VFIODev, op operation) (err erro
 		if err := q.qmpMonitorCh.qmp.ExecuteDeviceDel(q.qmpMonitorCh.ctx, devID); err != nil {
 			return err
 		}
-		if q.state.VFIODeviceCounter <= 0 {
-			return fmt.Errorf("VFIODeviceCounter is <= 0 (%d) ", q.state.VFIODeviceCounter)
-		}
-		q.state.VFIODeviceCounter--
 	}
 
 	return nil
@@ -1933,25 +1921,12 @@ func (q *qemu) resizeVCPUs(reqVCPUs uint32) (currentVCPUs uint32, newVCPUs uint3
 
 func (q *qemu) updateMemoryBalloon(sizeMB uint32) error {
 
-	if q.state.VFIODeviceCounter > 0 {
-		// When VFIO devices are used we dont want to get pages back to
-		// the host. Lets increase the balloon to a large amount of
-		// memory:
-		// - The guest memmory is not affected
+	// From qemu 3.1 it is safe to use balloon and VFIO devices
+	// where "Inhibit ballooning" is implemented.
 
-		// TODO: This is not required for qemu +v3.1.0 where "Inhibit
-		// ballooning" is implemented.
-
-		// Reference:
-		// https://www.ibm.com/support/knowledgecenter/en/SSZJY4_3.1.0/liabp/liabpmemoryballooning.htm
-		// https://bugs.launchpad.net/qemu/+bug/1762707
-		hostMB, err := q.hostMemMB()
-		if err != nil {
-			return err
-		}
-		sizeMB = uint32(hostMB)
-		q.Logger().WithField("memory-mb", sizeMB).Warnf("VM uses VFIO, balloon size will be total host memory")
-	}
+	// Reference:
+	// https://www.ibm.com/support/knowledgecenter/en/SSZJY4_3.1.0/liabp/liabpmemoryballooning.htm
+	// https://bugs.launchpad.net/qemu/+bug/1762707
 
 	if q.qmpMonitorCh.qmp == nil {
 		return fmt.Errorf("qmp channel is nil")
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -68,6 +68,10 @@ type QemuState struct {
 	UUID                 string
 	HotplugVFIOOnRootBus bool
 	VirtiofsdPid         int
+	// Flag to know if a VFIODev is assigned used by features that are not
+	// compatible with it, e.g. resize the balloon to return memory back is
+	// not recommended.
+	VFIODeviceCounter int
 }
 
 // qemu is an Hypervisor interface implementation for the Linux qemu hypervisor.
@@ -109,6 +113,7 @@ const (
 	rngID                    = "rng0"
 	vsockKernelOption        = "agent.use_vsock"
 	fallbackFileBackedMemDir = "/dev/shm"
+	balloonID                = "balloon0"
 )
 
 var qemuMajorVersion int
@@ -594,6 +599,11 @@ func (q *qemu) createSandbox(ctx context.Context, id string, networkNS NetworkNa
 		Filename: q.config.EntropySource,
 	}
 	qemuConfig.Devices, err = q.arch.appendRNGDevice(qemuConfig.Devices, rngDev)
+	if err != nil {
+		return err
+	}
+
+	qemuConfig.Devices, err = q.arch.appendBalloonDevice(qemuConfig.Devices, balloonID)
 	if err != nil {
 		return err
 	}
@@ -1124,10 +1134,17 @@ func (q *qemu) hotplugVFIODevice(device *config.VFIODev, op operation) (err erro
 	if err != nil {
 		return err
 	}
+	if q.state.VFIODeviceCounter < 0 {
+		return fmt.Errorf("VFIODeviceCounter is < 0 (%d) ", q.state.VFIODeviceCounter)
+	}
 
 	devID := device.ID
 
 	if op == addDevice {
+
+		q.state.VFIODeviceCounter++
+		// When HasVFIODevice is set balloon size is set to maximal memory
+		q.updateMemoryBalloon(0)
 		// In case HotplugVFIOOnRootBus is true, devices are hotplugged on the root bus
 		// for pc machine type instead of bridge. This is useful for devices that require
 		// a large PCI BAR which is a currently a limitation with PCI bridges.
@@ -1171,6 +1188,10 @@ func (q *qemu) hotplugVFIODevice(device *config.VFIODev, op operation) (err erro
 		if err := q.qmpMonitorCh.qmp.ExecuteDeviceDel(q.qmpMonitorCh.ctx, devID); err != nil {
 			return err
 		}
+		if q.state.VFIODeviceCounter <= 0 {
+			return fmt.Errorf("VFIODeviceCounter is <= 0 (%d) ", q.state.VFIODeviceCounter)
+		}
+		q.state.VFIODeviceCounter--
 	}
 
 	return nil
@@ -1693,11 +1714,20 @@ func (q *qemu) disconnect() {
 func (q *qemu) resizeMemory(reqMemMB uint32, memoryBlockSizeMB uint32, probe bool) (uint32, memoryDevice, error) {
 
 	currentMemory := q.config.MemorySize + uint32(q.state.HotpluggedMemory)
+
 	err := q.qmpSetup()
 	if err != nil {
 		return 0, memoryDevice{}, err
 	}
 	var addMemDevice memoryDevice
+
+	// The memory balloon update has to be done in all cases: This is
+	// because currentMemory is the amount hotplugged + initial memory  but
+	// If the balloon size was reduced previously in the guest and now is
+	// increased again, the guest memory will not be updated.
+	if err := q.updateMemoryBalloon(reqMemMB); err != nil {
+		return 0, memoryDevice{}, err
+	}
 	switch {
 	case currentMemory < reqMemMB:
 		//hotplug
@@ -1899,6 +1929,36 @@ func (q *qemu) resizeVCPUs(reqVCPUs uint32) (currentVCPUs uint32, newVCPUs uint3
 		newVCPUs -= vCPUsRemoved
 	}
 	return currentVCPUs, newVCPUs, nil
+}
+
+func (q *qemu) updateMemoryBalloon(sizeMB uint32) error {
+
+	if q.state.VFIODeviceCounter > 0 {
+		// When VFIO devices are used we dont want to get pages back to
+		// the host. Lets increase the balloon to a large amount of
+		// memory:
+		// - The guest memmory is not affected
+
+		// TODO: This is not required for qemu +v3.1.0 where "Inhibit
+		// ballooning" is implemented.
+
+		// Reference:
+		// https://www.ibm.com/support/knowledgecenter/en/SSZJY4_3.1.0/liabp/liabpmemoryballooning.htm
+		// https://bugs.launchpad.net/qemu/+bug/1762707
+		hostMB, err := q.hostMemMB()
+		if err != nil {
+			return err
+		}
+		sizeMB = uint32(hostMB)
+		q.Logger().WithField("memory-mb", sizeMB).Warnf("VM uses VFIO, balloon size will be total host memory")
+	}
+
+	if q.qmpMonitorCh.qmp == nil {
+		return fmt.Errorf("qmp channel is nil")
+	}
+
+	q.Logger().WithField("memory-balloon-size-mb", sizeMB).Debugf("Updating memory balloon size")
+	return q.qmpMonitorCh.qmp.ExecuteBalloon(q.qmpMonitorCh.ctx, uint64(sizeMB)<<utils.MibToBytesShift)
 }
 
 func (q *qemu) cleanup() error {
--- a/virtcontainers/qemu_arch_base.go
+++ b/virtcontainers/qemu_arch_base.go
@@ -112,6 +112,9 @@ type qemuArch interface {
 	// addBridge adds a new Bridge to the list of Bridges
 	addBridge(types.Bridge)
 
+	// appendBalloonDevice appens a Balloon device to devices
+	appendBalloonDevice(devices []govmmQemu.Device, balloonID string) ([]govmmQemu.Device, error)
+
 	// handleImagePath handles the Hypervisor Config image path
 	handleImagePath(config HypervisorConfig)
 
@@ -684,4 +687,17 @@ func (q *qemuArchBase) setBridges(bridges []types.Bridge) {
 
 func (q *qemuArchBase) addBridge(b types.Bridge) {
 	q.Bridges = append(q.Bridges, b)
+}
+
+func (q *qemuArchBase) appendBalloonDevice(devices []govmmQemu.Device, balloonID string) ([]govmmQemu.Device, error) {
+	if balloonID == "" {
+		return devices, fmt.Errorf("Balloon ID is empty")
+	}
+	devices = append(devices,
+		govmmQemu.BalloonDevice{
+			ID: balloonID,
+		},
+	)
+
+	return devices, nil
 }
--- a/virtcontainers/qemu_arch_base_test.go
+++ b/virtcontainers/qemu_arch_base_test.go
@@ -10,6 +10,7 @@ import (
 	"io/ioutil"
 	"net"
 	"path/filepath"
+	"reflect"
 	"testing"
 
 	govmmQemu "github.com/intel/govmm/qemu"
@@ -561,4 +562,38 @@ func TestQemuArchBaseAppendNetwork(t *testing.T) {
 	devices, err = qemuArchBase.appendNetwork(devices, macvtapEp)
 	assert.NoError(err)
 	assert.Equal(expectedOut, devices)
+}
+
+func TestQemuArchBaseAppendBalloonDevice(t *testing.T) {
+	expectedDevices := []govmmQemu.Device{
+		govmmQemu.BalloonDevice{
+			ID: balloonID,
+		},
+	}
+	type args struct {
+		devices   []govmmQemu.Device
+		balloonID string
+	}
+	tests := []struct {
+		name    string
+		args    args
+		want    []govmmQemu.Device
+		wantErr bool
+	}{
+		{"Empty ID", args{[]govmmQemu.Device{}, ""}, []govmmQemu.Device{}, true},
+		{"Use valid ID", args{[]govmmQemu.Device{}, balloonID}, expectedDevices, false},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			q := &qemuArchBase{}
+			got, err := q.appendBalloonDevice(tt.args.devices, tt.args.balloonID)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("Got  %v, want %v", got, tt.want)
+			}
+		})
+	}
 }
--- a/virtcontainers/qemu_test.go
+++ b/virtcontainers/qemu_test.go
@@ -567,3 +567,28 @@ func TestQemuGetpids(t *testing.T) {
 	assert.True(pids[0] == 100)
 	assert.True(pids[1] == 200)
 }
+
+func TestQemuUpdateMemoryBalloon(t *testing.T) {
+	type fields struct {
+		qmp *govmmQemu.QMP
+	}
+
+	tests := []struct {
+		name    string
+		fields  fields
+		sizeMB  uint32
+		wantErr bool
+	}{
+		{"qmp nil", fields{}, 0, true},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			q := &qemu{}
+			q.qmpMonitorCh.qmp = tt.fields.qmp
+			if err := q.updateMemoryBalloon(tt.sizeMB); (err != nil) != tt.wantErr {
+				t.Errorf("error = %v, wanted = %v", err, tt.wantErr)
+			}
+		})
+	}
+}


jcvenegas:
So I am getting an strange behavior when hotplug memory just by adding `-balloon virtio`. When I hotplug memory the kernel total memory is not near what I expect. It is increased but not as I would like. 
```
[jcvenega@jcvenega-nuc runtime]$ docker run -dti -m 256M --runtime kata-runtime --name kata-1 stress-ng
ec02de8650364902004e3a73047ea8219d077d8fe0b321f2c3c8440263fad725
[jcvenega@jcvenega-nuc runtime]$ docker exec -ti kata-1  free -h
              total        used        free      shared  buff/cache   available
Mem:           234M         20M        201M        1.7M         12M        204M
Swap:            0B          0B          0B
[jcvenega@jcvenega-nuc runtime]$ docker update --memory 4G --memory-swap 4G kata-1
kata-1
[jcvenega@jcvenega-nuc runtime]$ docker exec -ti kata-1  free -h
              total        used        free      shared  buff/cache   available
Mem:           746M         81M        652M        1.7M         12M        647M
Swap:            0B          0B          0B
```

jcvenegas:
update: I just tried in a VM with the kata kernel and same qemu binary and worked I'll debug in the kata guest OS.

jcvenegas:
found the issue, will send the fix

jodh-intel:
nice.

lgtm

jcvenegas:
depends on https://github.com/intel/govmm/pull/55

jcvenegas:
@jcvenegas @sboeuf I think this is ready to a new review.  I think this is almost the final version, just waiting to merge https://github.com/intel/govmm/pull/55 and I will create a PR to test this on tests repo.

sboeuf:
@jcvenegas Ok I'll take a look later today!

virtcontainers/container.go
```diff
@@ -1310,10 +1310,21 @@ func (c *Container) updateMemoryResources(oldResources ContainerResources, newRe
...
 	if oldMemMB == newMemMB {
-		c.Logger().WithFields(logrus.Fields{
-			"old-mem": fmt.Sprintf("%dMB", oldMemMB),
-			"new-mem": fmt.Sprintf("%dMB", newMemMB),
-		}).Debug("the actual number of Mem will not be modified")
+		c.Logger().Debug("the current memory will not be modified")
+		addMemDevice := &memoryDevice{
+			sizeMB: 0,
+		}
+		data, err := c.sandbox.hypervisor.hotplugAddDevice(addMemDevice, memoryDev)
```
sboeuf: Why are we trying to hotplug more memory here? We're in the case where `oldMemMB == newMemMB`, and as mentioned by the debug log `the current memory will not be modified`.

jcvenegas:
@sboeuf take a look, I think now is more explicit the the balloon usage. @linzichang @clarecch please take a look this PR is planning to move part of the logic on how we manage memory. The last to commits  do that.

clarecch:
Why should we move hotplug logic from container to sandbox? Is there any reason?

clarecch:
Hi @jcvenegas , most of this PR looks nice, but two doubts：
1.  Why should we move hotplug logic from container to sandbox? I think `c.sandbox.XXX` is the same as `s.XXX`.
2.  I am glad to see the thought of https://github.com/kata-containers/runtime/pull/624#issuecomment-419656853 is almost realized, as @linzichang and me are too busy to handle the https://github.com/kata-containers/runtime/issues/788 issue. Just a question, do we need `vm_reserved_memory ` memtions in https://github.com/kata-containers/runtime/pull/624#issuecomment-419656853 to keep the vm running normally when sandbox memory is all used by container? With `vm_reserved_memory `, we can still hotplug memory from hostos as long as hostos has enough free memory , and the guestos will not panic even if all sandbox memory is used by container.

virtcontainers/qemu.go
```diff
@@ -1191,6 +1229,42 @@ func (q *qemu) disconnect() {
 	q.qmpShutdown()
 }
 
+func (q *qemu) resizeMemory(reqMemMB uint32, memoryBlockSizeMB uint32) (uint32, error) {
+	currentMemory := q.config.MemorySize + uint32(q.state.HotpluggedMemory)
+	err := q.qmpSetup()
+	if err != nil {
+		return 0, err
+	}
+	if err := q.updateMemoryBalloon(reqMemMB); err != nil {
```
clarecch: Should `updateMemoryBalloon` be called after memory is hotplugged? For instance if we have 1G memory in vm now, we balloon 2G memory which exceed the total memory of vm, before we hotplug another 1G to vm. Is this feasible?

clarecch: As I know ballooning is effective only when the scope of memory resizing do not exceed the `VM max memory`. But I am not sure whether ballooning is constrained to the `VM max memory` originates from the VM’s creation or the `VM max memory` when we hotplug more memory to VM. Please take a look at this.

virtcontainers/sandbox.go
```diff
@@ -1362,14 +1363,68 @@ func (s *Sandbox) EnterContainer(containerID string, cmd Cmd) (VCContainer, *Pro
...
+		// Set and save container's config VCPUs field only
+		c.config.Resources.VCPUs = newResources.VCPUs
+		if err := c.storeContainer(); err != nil {
+			return err
+		}
+	}
+
+	// Memory is not updated if memory limit not set
+	if newResources.MemMB != 0 {
+		c.config.Resources.MemMB = newResources.MemMB
```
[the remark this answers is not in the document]
jcvenegas: @clarecch  I need to test this,  
from my perspective we still need to update the cgroup 
   - the container should update its cgroup
   - The sandbox manage the memory from the guest.

I expect that cgroup limit does not depends on the guest physical memory, so we can set the cgroup with more memory than the memory that is available in the VM.

A exmaple that is valid is when:
1. have a sandbox/pod/VM  with all (almost ) the memory of the host added. 
`create sandbox, create container -memory all the host memory`
2. Then we add another container to the Sandbox/pod/VM without any memory cgroup. 
3. After we update the cgroup of the new container to increase memory.
   a. The sandbox wont increase the memory but the container will share the memory with other containers and the cgroup will be limited.

virtcontainers/qemu.go
jcvenegas: > Should updateMemoryBalloon be called after memory is hotplugged? For instance if we have 1G memory in vm now, we balloon 2G memory which exceed the total memory of vm, before we hotplug another 1G to vm. Is this feasible?

Yes feasible, it is feasible, you can increase the ballon with more memory than you actually have in the VM, in that case the memory VM will be the `min(ballon, guest physical memory)`.

> As I know ballooning is effective only when the scope of memory resizing do not exceed the VM max memory.But I am not sure whether ballooning is constrained to the VM max memory originates from the VM’s creation or the VM max memory when we hotplug more memory to VM. Please take a look at this.

From my testing, if the balloon is limited, the max memory is not. So we are safe to hotplug more memory and the balloon should not be affected. The slot enabling is independent  in the of the balloon mechanism. We can change the order if you want.

jcvenegas:
@clarecch 

> Hi @jcvenegas , most of this PR looks nice, but two doubts：
>  Why should we move hotplug logic from container to sandbox? I think c.sandbox.XXX is the same as s.XXX.

Yes, actually it also works this is more a logical change, talking with @sboeuf,  We seen that the container should not be aware of the sandabox resources, but its resources (only namespaces and cgroups). Because this PR is adding another resource management here: `balloon`. So would be more clean to let the sandbox do this. To not make this PR bigger, I am planning to move the logic of vcpus in a follow up PR.

>  I am glad to see the thought of #624 (comment) is almost realized, as @linzichang and me are too busy to handle the #788 issue. Just a question, do we need vm_reserved_memory memtions in #624 (comment) to keep the vm running normally when sandbox memory is all used by container? With vm_reserved_memory, we can still hotplug memory from hostos as long as hostos has enough free memory , and the guestos will not panic even if all sandbox memory is used by container.

Yes, I thinks we still need need vm_reserved_memory 1G or 2G is fine for two purposes keep the sandbox (mostly for the agent, systemd and kernel ). This is something I am not adding because was not sure if needed to add when #786 is merged.

virtcontainers/qemu.go
clarecch: > From my testing, if the balloon is limited, the max memory is not. So we are safe to hotplug more memory and the balloon should not be affected. The slot enabling is independent in the of the balloon mechanism. We can change the order if you want.

For instance, if we have a VM with 1G memory, then we want to hotplug another 1G memory, we will balloon 2G and then hotplug 1G memory with qemu cmd. It will success when we balloon 2G, but in fact it will be 1G according to `min(ballon, guest physical memory)`. Finally, we will get a VM with 1G balloon, 2G physical memory. So I think we should change the order. 

But I am afraid although we hotplug 1G memory then balloon 2G, the balloon will still be 1G, as I am not sure whether ballooning is constrained to the VM max memory originates from the VM’s creation or the VM max memory when we hotplug more memory to VM.

virtcontainers/sandbox.go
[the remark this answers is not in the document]
clarecch: Thanks for explanation.

clarecch:
>Yes, I think we still need need vm_reserved_memory 1G or 2G is fine for two purposes keep the sandbox (mostly for the agent, systemd and kernel ). This is something I am not adding because was not sure if needed to add when #786 is merged.

@jcvenegas I think it would be like this. After both #786 and #793 merged, if we `docker run --m 2G`, we will setup a VM with 2G memory and then hotplug 2G memory. With #793, after judge 2G=2G we will not hotplug any memory, then we set container memory limit cgroup to be 2G. If the container use all 2G memory, there will be no enough memory remaining for system.

clarecch:
When the PR is ready, we can delete this TODO https://github.com/kata-containers/runtime/blob/master/virtcontainers/container.go#L1300 and this issue can be closed https://github.com/kata-containers/runtime/issues/788

sboeuf:
@jcvenegas thanks for the new comments in the code, the code looks good to me!
I expect you have tested this properly to make sure we don't have any regressions regarding the different use cases related to CPU and memory, right?

clarecch:
Thanks @jcvenegas , this code LGTM

sboeuf:
@jcvenegas just added `do-not-merge` label as I've realized you need to properly vendor govmm, and not simply modify the files.

virtcontainers/qemu.go
```diff
@@ -1051,33 +1058,64 @@ func (q *qemu) hotplugMemory(memDev *memoryDevice, op operation) (int, error) {
...
 
-	// We do not support memory hot unplug.
-	if op == removeDevice {
-		// Dont fail for now, until we fix it.
-		// We return that we only unplugged 0
-		q.Logger().Warn("hot-remove VM memory not supported")
-		return 0, nil
-	}
-
+	q.Logger().Debugf("hotplug %dMB mem", memDev.sizeMB)
```
jodh-intel: Nit: This would be better as:

```suggestion
	q.Logger().WithField("hotplug-memory-mb", memDev.sizeMB).Debug("hotplugging memory")
```

virtcontainers/qemu.go
```diff
@@ -1051,33 +1058,64 @@ func (q *qemu) hotplugMemory(memDev *memoryDevice, op operation) (int, error) {
...
-
-	// calculate current memory
 	currentMemory := int(q.config.MemorySize) + q.state.HotpluggedMemory
 
-	// Don't exceed the maximum amount of memory
-	if currentMemory+memDev.sizeMB > int(maxMem) {
-		return 0, fmt.Errorf("Unable to hotplug %d MiB memory, the SB has %d MiB and the maximum amount is %d MiB",
-			memDev.sizeMB, currentMemory, q.config.MemorySize)
+	if memDev.sizeMB == 0 {
+		q.Logger().WithFields(
```
jodh-intel: Nit: this could be simplified to:

```suggestion
		q.Logger().WithField("hotplug", "memory").Debug("updating balloon as hotplug not required")
```

virtcontainers/qemu.go
```diff
@@ -1101,6 +1134,11 @@ func (q *qemu) hotplugAddMemory(memDev *memoryDevice) (int, error) {
 	return memDev.sizeMB, q.storage.storeHypervisorState(q.id, q.state)
 }
 
+func (q *qemu) updateMemoryBalloon(sizeMB uint32) error {
+	q.Logger().Debugf("Update balloon %dMB memory", sizeMB)
```
jodh-intel: This could be:

```suggestion
	q.Logger().WithField("memory-balloon-size-mb", sizeMB).Debugf("Updating memory balloon size")
```

virtcontainers/sandbox.go
```diff
@@ -1364,14 +1365,70 @@ func (s *Sandbox) EnterContainer(containerID string, cmd Cmd) (VCContainer, *Pro
...
 }
 
+func (s *Sandbox) updateMemoryResources() error {
+	// the hypervisor.MemorySize is the amount of memory reserved for
+	// the VM and contaniners without memory limit
+	sandboxMemoryMB := s.hypervisor.hypervisorConfig().MemorySize
+	for _, c := range s.containers {
+		sandboxMemoryMB += c.config.Resources.MemMB
+	}
+	s.Logger().Debugf("Request to hypervisor to update memory: %dMB", sandboxMemoryMB)
```
jodh-intel: Again, could you make the memory amount a log field?

virtcontainers/container.go
```diff
@@ -1242,6 +1226,7 @@ func (c *Container) removeResources() error {
 }
 
 func (c *Container) updateVCPUResources(oldResources, newResources ContainerResources) error {
+	//FIXME: This should be managed by the sandbox not the container.
```
devimc: please file an issue for this

virtcontainers/qemu.go
```diff
@@ -1191,6 +1229,76 @@ func (q *qemu) disconnect() {
...
+		}
+		memoryRemoved, ok := data.(int)
+		if !ok {
+			return currentMemory, fmt.Errorf("Could not get the memory removed, got %+v", data)
+		}
+		currentMemory -= uint32(memoryRemoved)
+	}
+
+	// Update balloon in any memory update case with the size requested by the caller.
+	if err := q.updateMemoryBalloon(reqMemMB); err != nil {
```
devimc: I guess you forgot to add `case currentMemory == reqMemMB` , right?

jcvenegas: This include all the cases. The hotplug or hot-remove is handled in the `switch` statement and as a separated case the ballloon is always updated with the size of the memory requested by the caller..

jcvenegas:
vendor test failing , depends on https://github.com/kata-containers/tests/pull/821

sboeuf:
@jcvenegas a topic we discussed with @mcastelino @egernst @amshinde was the ability for Qemu to use balloon with VFIO.
Could you please test VFIO with your PR before we decide to move forward and merge it?

More details about the test, you should verify that after you passed a VFIO device, you can resize the memory to a smaller amount than initially allowed.

jcvenegas:
I confirm it is working I tested resizing the balloon with a NIC card and a sound card.

raravena80:
@jcvenegas ping 😄

jcvenegas:
rebased  testing again.
/test

clarecch:
> Could you please test VFIO with your PR before we decide to move forward and merge it?
More details about the test, you should verify that after you passed a VFIO device, you can resize the memory to a smaller amount than initially allowed.

Hi, @sboeuf, is this what you worry about?
```For guests that use PCI pass-through (VFIO), you must disable the memory ballooning feature by changing the memballoon clause to model='none'. This is because the guest balloon driver with a host under memory pressure can make the overall memory situation in the host more worse if PCI pass-through is in use.```

Reference: https://www.ibm.com/support/knowledgecenter/en/SSZJY4_3.1.0/liabp/liabpmemoryballooning.htm

**Comment by jcvenegas** [comment]
@clarecch thanks for the reference, here another 
https://bugs.launchpad.net/qemu/+bug/1762707

So in this case we can check if a device is assigned to the VM before the runtime tries to remove memory from the host.

**Comment by sboeuf** [comment]
@clarecch yes that's the potential issue that you pointed out here.
@jcvenegas maybe a good thing to return an error if the user tries to resize the container memory but a VFIO device is attached to the VM.

File: virtcontainers/sandbox.go [hunk 10]
```diff
@@ -1386,7 +1392,30 @@ func (s *Sandbox) UpdateContainer(containerID string, resources specs.LinuxResou
 		return err
 	}
 
-	return c.update(resources)
+	newResources := ContainerResources{
+		VCPUs: uint32(utils.ConstraintsToVCPUs(*resources.CPU.Quota, *resources.CPU.Period)),
+		// do page align to memory, as cgroup memory.limit_in_bytes will be aligned to page when effect.
+		// TODO use GetGuestDetails to get the guest OS page size.
+		MemByte: (*resources.Memory.Limit >> 12) << 12,
```
jcvenegas: @clarecch  should I move this to `calcHotplugMemMiBSize()` ?

clarecch: I don't think we have need

virtcontainers/sandbox.go
```diff
@@ -1780,3 +1805,43 @@ func (s *Sandbox) AddDevice(info config.DeviceInfo) (api.Device, error) {
...
+		vcpusAdded := newCPUs - oldCPUs
+		if err := s.agent.onlineCPUMem(vcpusAdded, true); err != nil {
+			return err
+		}
+	}
+	s.Logger().Debugf("Sandbox CPUs: %d", newCPUs)
+
+	// Update Memory
+	s.Logger().WithField("memory-sandbox-size-byte", sandboxMemoryByte).Debugf("Request to hypervisor to update memory")
+	newMemory, err := s.hypervisor.resizeMemory(uint32(sandboxMemoryByte>>utils.MibToBytesShift), s.state.GuestMemoryBlockSizeMB)
```
clarecch: I think there is some problem. The reason we change the unit of c.config.Resources.Mem from MiB to Byte is to prevent the 4095B < memory < 1MiB from being lost and ignore 1B < memory < 4096B.

There is a example:
docker run 128M, we will hotplug 128M memory, and then,
docker update 128M+1B, we will do nothing,
docker update 100M+4096B, we will hotplug another 128M memory after calculate by `calcHotplugMemMiBSize`

By simply do 20 right shift to sandboxMemoryByte, we will lose memory less than 1M.

clarecch: Of course, in other architecture it may be not 4096B

jcvenegas: Sorry for the late response, I was on vacations, the only option I see is to move most of the logic bytes to avoid being converting in all the places.

sboeuf:
@egernst splitting make sense :+1:

egernst:
oh, and my apologies for being so late to this PR, @jcvenegas

jcvenegas:
that is fine I'll open another PR and depend on that for this PR.

raravena80:
@jcvenegas ping, from your weekly Kata herder.

raravena80:
@jcvenegas any updates? are we eventually merging?

jodh-intel:
@jcvenegas - are you planning to update this?

jcvenegas:
I will rebase this today .

jodh-intel:
Ping @jcvenegas.

raravena80:
@jcvenegas ping

virtcontainers/qemu.go
```diff
@@ -1553,6 +1567,23 @@ func (q *qemu) resizeVCPUs(reqVCPUs uint32) (currentVCPUs uint32, newVCPUs uint3
...
+
+	if q.state.hasVFIODevice {
+		//Reference:
+		//https: //www.ibm.com/support/knowledgecenter/en/SSZJY4_3.1.0/liabp/liabpmemoryballooning.htm
+		hostMB, err := q.hostMemMB()
+		if err != nil {
+			return err
+		}
+		sizeMB = uint32(hostMB)
+		q.Logger().Warnf("VM uses VFIO, balloon size will be total host memory (%d MB)", sizeMB)
```
devimc: q.Logger().WithField("mem", sizeMB).Warn("VM uses VFIO, balloon size will be total host memory")

jcvenegas:
ping @kata-containers/runtime

sboeuf:
@egernst PTAL so that we can move forward with this

virtcontainers/qemu.go
bergwolf: If `currentMemory == reqMemMB`, `resizeMemory` should be noop. Please add a check at the beginning of the function to make sure we do nothing in such case (which means we do not even setup the qmp connection).

grahamwhaley:
Looks like @bergwolf has some questions - @bergwolf , if you'd like your questions to block this merge, can you add a 'request changes' via the review dialog so we don't accidentally merge without addressing feedback. thx!

jodh-intel:
Ping @jcvenegas.

virtcontainers/qemu.go
```diff
@@ -54,6 +54,10 @@ type QemuState struct {
 	HotpluggedMemory     int
 	UUID                 string
 	HotplugVFIOOnRootBus bool
+	// Flag to know if a VFIODev is assigned used by features that are not
+	// compatible with it, e.g. resize the balloon to return memory back is
+	// not recommended.
+	hasVFIODevice bool
```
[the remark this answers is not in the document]
jcvenegas: You are right! fixed

virtcontainers/qemu.go
```diff
@@ -1553,6 +1567,23 @@ func (q *qemu) resizeVCPUs(reqVCPUs uint32) (currentVCPUs uint32, newVCPUs uint3
 	return currentVCPUs, newVCPUs, nil
 }
 
+func (q *qemu) updateMemoryBalloon(sizeMB uint32) error {
+
+	if q.state.hasVFIODevice {
+		// Reference:
+		// https://www.ibm.com/support/knowledgecenter/en/SSZJY4_3.1.0/liabp/liabpmemoryballooning.htm
+		hostMB, err := q.hostMemMB()
```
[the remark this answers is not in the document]
jcvenegas: @bergwolf the idea is to set the balloon to very large size, this is to avoid memory is returned back to the host on VFIO case, this is patched in qemu +`3.1.0` , I added comments in the code. 
\cc @awilliam


@bergwolf  For `guest maximum dimm` I need to do a qmp call that will give me inforamation in bytes, while we dont move all the code logic to use only bytes I dont want to add more code to convert from MB to Bytes and and vice versa. 

Also, I added comments about why I dont add a `currentMemory == reqMemMB` validation.

jodh-intel:
Ping @jcvenegas.

virtcontainers/qemu.go
```diff
@@ -1567,6 +1587,32 @@ func (q *qemu) resizeVCPUs(reqVCPUs uint32) (currentVCPUs uint32, newVCPUs uint3
 	return currentVCPUs, newVCPUs, nil
 }
 
+func (q *qemu) updateMemoryBalloon(sizeMB uint32) error {
+
+	if q.state.HasVFIODevice {
+		// When VFIO devices are used we dont want to get pages back to
+		// the host. Lets increase the balloon to a large amount of
```
amshinde: @jcvenegas I havent looked a whole lot into memory balooning, but basic question here...if we dont want to give pages to the host, then shouldnt we be deflating the balloon here, so that no memory is returned to the host instead of increasing the baloon size?

virtcontainers/qemu.go
```diff
@@ -1567,6 +1587,32 @@ func (q *qemu) resizeVCPUs(reqVCPUs uint32) (currentVCPUs uint32, newVCPUs uint3
...
+
+		// Reference:
+		// https://www.ibm.com/support/knowledgecenter/en/SSZJY4_3.1.0/liabp/liabpmemoryballooning.htm
+		// https://bugs.launchpad.net/qemu/+bug/1762707
+		hostMB, err := q.hostMemMB()
+		if err != nil {
+			return err
+		}
+		sizeMB = uint32(hostMB)
+		q.Logger().WithField("memory-mb", sizeMB).Warnf("VM uses VFIO, balloon size will be total host memory")
```
amshinde: Again, I would like to see results of how things work when mdev devices are assigned. According to the bug that you referenced here, mdev device may or may not pin memory depending on the vendor driver. We should be testing all these scenarios as well.

amshinde: You also mentioned that the `Inhibit Balooning` feature is available in 3.1.0.
Considering this plus various other features, we should really be updating qemu version that we package.
cc @egernst

jodh-intel:
@egernst, @bergwolf - please could you guys re-review?

virtcontainers/qemu.go
```diff
@@ -879,6 +886,10 @@ func (q *qemu) hotplugVFIODevice(device *config.VFIODev, op operation) error {
 	devID := device.ID
 
 	if op == addDevice {
+
+		q.state.HasVFIODevice = true
```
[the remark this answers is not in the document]
jcvenegas: @amshinde  @Weichen81 take a look I used a device counter to verify the amount of devices.

virtcontainers/qemu.go
jcvenegas: @amshinde increase the balloon size will increase the amount of memory available in the guest.  If we reduce it the pages will be given back to the host.  So ideally is never  get pages back to the host when we use VFIO.

virtcontainers/qemu.go
jcvenegas: @amshinde sounds good! I'll contact you to define a set of test to verify.  Also will ping @awilliam
to review this logic and verify if is  better just move to qemu 3.1+

jcvenegas: @amshinde  yeah this sounds a more safe approach, we can set as pre-requisite before we land this PR.

virtcontainers/qemu.go
amshinde: Dug a little bit into the actual qmp command
https://github.com/coreos/qemu/blob/master/qmp-commands.hx#L1830

Looks like this command actually sets the size of the VM and not the balloon!
So we need to update the docomentation here: 
https://github.com/intel/govmm/blob/78d079db6d1f3e32e3ed7578a54baa6257b058a7/qemu/qmp.go#L1353

Also, the comments here, also need to be updated to mention, that we are actually deflating the balloon and not inflating it.

virtcontainers/qemu.go
awilliam: In QEMU (3.1+) the balloon inhibitor is activated for any vfio device, whether it's a physical or mdev device.  While mdev devices are theoretically more compatible with ballooning, there is no guarantee that the vendor driver isn't sloppy at tracking pages actively used by the device in the guest, which could make them available to the balloon driver in the guest and result in the same sort of mismatches that we're trying to avoid.  It's safer to ignore that mdev device might be more balloon friendly for now, we'll need to develop new interfaces to expose if they ever truly are.

virtcontainers/qemu.go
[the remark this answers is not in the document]
awilliam: It looks a bit strange to me as well, it's relying on an assigned device VM not supporting memory over-commit and thus never having a footprint larger than host memory.  But an mdev VM doesn't necessarily pin all of memory and therefore could be over-committed to some extent.  What happens if you have such a VM, with a fully deflated balloon, and then call this?  Seems it could actually cause a balloon inflation.  The QEMU balloon inhibitor takes the approach that we do nothing when an inhibitor is added.  Anything that gets pinned by vfio will be re-allocated in the VM address space by the act of that pinning, it's only a matter that the guest balloon driver is possibly still holding onto pages, keeping the balloon inflated, which are now backed in the VM address space anyway.  This also maintains the appearance of ballooning within the guest, guest available memory changes with the balloon, only the memory isn't freed back to the host.  The important part is that the inhibitor prevents further inflation of the balloon, which can zap pages from the VM which are already pinned by vfio.  If you want to make sure the guest balloon driver is releasing all pages when a vfio device is added, perhaps figure out the data type used for the QMP interface and pass the maximum value it supports rather than creating this dependency on host memory size.

virtcontainers/qemu.go
jodh-intel: /cc @markdryan.

virtcontainers/qemu.go
jcvenegas: @awilliam thank you, so I think the best a option is to move to qemu 3.1 first

raravena80:
@jcvenegas ping, any updates?

raravena80:
@jcvenegas ping, any updates? Thx.

grahamwhaley:
Branch is now conflicted (again). @jcvenegas - can you give us an update on your plan with this at least - are you actively pursuing and still hoping to land?

raravena80:
@jcvenegas any updates? Thx!

caoruidong:
ping @jcvenegas

raravena80:
@jcvenegas ping.

raravena80:
ping @jcvenegas

caoruidong:
ping @jcvenegas Shall we close it?

jcvenegas:
I'll try to get this PR merged this year, @bergwolf @egernst  please take a look would be nice to have it merged for 1.9

jcvenegas:
@devimc @egernst @bergwolf ping  place take a look

virtcontainers/qemu.go
```diff
@@ -1128,6 +1134,9 @@ func (q *qemu) hotplugVFIODevice(device *config.VFIODev, op operation) (err erro
 	devID := device.ID
 
 	if op == addDevice {
+
+		// When HasVFIODevice is set balloon size is set to maximal memory
```
devimc: and where is `HasVFIODevice` set ?

raravena80:
@jcvenegas any update on this ticket?

jcvenegas:
@raravena80  thanks for ping on this I was waiting for feedback from @egernst and @bergwolf  if still we want to enable this, the PR is functional but this PR does not have a lot of priority today.

raravena80:
@jcvenegas any updates?

Your weekly Kata herder.

jodh-intel:
Hmm, this is a pain - the only failing CI is the nemu one. But we no longer support that so we *could* land this. But GitHub won't let us. Thoughts @jcvenegas?

@chavafg - btw, as nemu is dead, do we still need the nemu CI jobs?

grahamwhaley:
a re-submit as a new PR might be the easiest thing to do.
Good thought on the CI - if we have not done it already.

chavafg:
@jodh-intel we already removed the nemu CI for master branch. 
We still have it running for stable branches (1.8.x)... Although not sure if we still support that branch. Maybe we can now remove it definitely. 

The nemu register above was from an old execution, we can ignore it, or as @grahamwhaley comments, the PR can be re-pushed.

jcvenegas:
@jodh-intel its being a long time of this PR I think make sense to ask to @kata-containers/architecture-committee  if we want it or should I close it.

raravena80:
@chavafg @jcvenegas any updates on this PR? Thx

virtcontainers/qemu.go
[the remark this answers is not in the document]
egernst: (yikes, 2 years later... )

In the context of Kata, I don't think we'll be in a position of memory over-commit for a pod/container. If there aren't any other side effects, I'm not sure we should be _that_ concerned, other than it looking a bit odd!

Issue #801, opened 2018-09-29
(replace this text with the list of steps you followed)


jcvenegas:
```
$ docker run -ti -m 256M --runtime kata-runtime --rm --name kata-1 ubuntu sh
$ docker exec -ti kata-1 free -h
              total        used        free      shared  buff/cache   available
Mem:           234M         21M        200M        1.7M         12M        203M
Swap:            0B          0B          0B
$ docker update --memory 5G --memory-swap 5G kata-1
$ docker exec -ti kata-1 free -h
              total        used        free      shared  buff/cache   available
Mem:           5.0G         97M        4.9G        1.7M         12M        4.8G
Swap:            0B          0B          0B
$ docker update --memory 2G kata-1
$ docker exec -ti kata-1 free -h
              total        used        free      shared  buff/cache   available
Mem:           5.0G         97M        4.9G        1.7M         12M        4.8G
Swap:            0B          0B          0B
$ docker update --memory 3G kata-1
$ docker exec -ti kata-1 free -h
              total        used        free      shared  buff/cache   available
Mem:           6.0G        115M        5.9G        1.7M         12M        5.8G
Swap:            0B          0B          0B

```

Note2: This requries https://github.com/kata-containers/agent/pull/392 to see all memory updated.

Not more memory added to the VM if alredy meets the required memory.

More memory is addded

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #802, opened 2018-09-29
memory: update: Update memory with the memory removed.

jcvenegas:
If we reduced the memory of a container, its
cgroup in the VM was updated properly. But we assumed
that the memory was also removed from the VM.

Then when it is added more memory again we added more than
needed.

Fixes: #801


virtcontainers/container.go
```diff
@@ -1325,16 +1329,37 @@ func (c *Container) updateMemoryResources(oldResources, newResources ContainerRe
...
+			return fmt.Errorf("Could not get the memory added, got %+v", data)
+		}
+		newResources.MemMB = oldMemMB + uint32(memoryAdded)
 		if err := c.sandbox.agent.onlineCPUMem(0, false); err != nil {
 			return err
 		}
 	}
 	// hot remove memory unsupported
+	if oldMemMB > newMemMB {
+		// Memory is not updated was not updaetd
```
jodh-intel: Typo, but also could you make this comment a little clearer?

sboeuf: @jcvenegas please address this comment. I agree we need a clearer comment here.

grahamwhaley:
I'll mark as DNM until we have the comment fixed, but will fire off the tests so we can progress...

jcvenegas:
ready to merge

PR #805, opened 2018-10-02
virtcontainers: qemu: Add proper support for virt machine type

### LINKED ISSUE #804: virtcontainers: NEMU virt machine type not fully supported
We need to fix the hypervisor implementation for `qemu` if we want to be able to use NEMU `virt` machine type with Kata Containers.

sboeuf:
The virt machine type provided by the NEMU project needs to be
supported the same way we support pc and q35 machine types.

First, this patch takes care of adding the hotpluggable block device
capability to this machine type, this way when using devicemapper, we
prevent the code from falling back on using 9pfs instead of SCSI.

It also add one or several bridges to this machine type, as the code
is tightly coupled to the fact that a bridge is required for PCI
hotplug.

At last, it changes the name of the PCI host bridge (main bus), to
use "pcie.0". The default set up from pc machine type "pci.0" is not
suitable for this machine type.

Fixes #804

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -1194,7 +1194,8 @@ func (q *qemu) disconnect() {
 // genericAppendBridges appends to devices the given bridges
 func genericAppendBridges(devices []govmmQemu.Device, bridges []Bridge, machineType string) []govmmQemu.Device {
 	bus := defaultPCBridgeBus
-	if machineType == QemuQ35 {
+	switch machineType {
+	case QemuQ35, QemuVirt:
 		bus = defaultBridgeBus
 	}
 
@@ -1227,13 +1228,14 @@ func genericBridges(number uint32, machineType string) []Bridge {
 	var bt bridgeType
 
 	switch machineType {
-
 	case QemuQ35:
 		// currently only pci bridges are supported
 		// qemu-2.10 will introduce pcie bridges
 		fallthrough
 	case QemuPC:
 		bt = pciBridge
+	case QemuVirt:
+		bt = pcieBridge
 	case QemuPseries:
 		bt = pciBridge
 	default:
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -123,11 +123,11 @@
   revision = "3520598351bb3500a49ae9563f5539666ae0a27c"
 
 [[projects]]
-  digest = "1:034b5648db9f53a2b6f4f84925ee031cbeaba69daa10d95e74e8242707d7a5b0"
+  digest = "1:c63a5bf4f3fd94ae838ce53e3492c0a467c40d09cada9301d228df7164ba4e55"
   name = "github.com/intel/govmm"
   packages = ["qemu"]
   pruneopts = "NUT"
-  revision = "9905ae92c5915c07abeb669eaa4d7f7408834b51"
+  revision = "f03df80fc3dc52f65ed6c7d12806279b44185d32"
 
 [[projects]]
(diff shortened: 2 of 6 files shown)


virtcontainers/qemu.go
```diff
@@ -1225,12 +1226,11 @@ func genericBridges(number uint32, machineType string) []Bridge {
 	var bt bridgeType
 
 	switch machineType {
-
 	case QemuQ35:
 		// currently only pci bridges are supported
 		// qemu-2.10 will introduce pcie bridges
 		fallthrough
-	case QemuPC:
+	case QemuPC, QemuVirt:
```
mcastelino: @sboeuf you should be using a -device pcie-pci-bridge along the lines of 

`-device pcie-pci-bridge,id=<id>,bus=pcie.0`

What does this end up being in this case? pci-bridge?

What is the b.Type for virt?

https://github.com/kata-containers/runtime/pull/805/commits/ecb55ec33de135e84bf3a0408e885acf2a70c33e#diff-f02dbf94518fc0fe68a41a78d181b390R1202

virtcontainers/qemu.go
```diff
@@ -1225,12 +1226,11 @@ func genericBridges(number uint32, machineType string) []Bridge {
 	var bt bridgeType
 
 	switch machineType {
-
 	case QemuQ35:
 		// currently only pci bridges are supported
 		// qemu-2.10 will introduce pcie bridges
```
mcastelino: @devimc what did you mean by 
`
		// currently only pci bridges are supported
		// qemu-2.10 will introduce pcie bridges			// qemu-2.10 will introduce pcie bridges
`

were you referring to `name "pcie-pci-bridge", bus PCI`.

If so @sboeuf as virt is based on latest qemu we should switch to pcie-pci-bridge for virt as we know it will not show up with older QEMU versions.

virtcontainers/qemu.go
sboeuf: @mcastelino this is part of discussions that we need to have in order to improve things. But I thought that for now, we should at least get this working the same way it works with `pc`, don't you think?

This ends up doing adding a simple `-device pci-bridge` in this case.

mcastelino: @sboeuf as we use the latest QEMU for virt we should be using pcie-pci-bridge for now. And for Q35 based on the QEMU version used that too should use pcie-pci-bridge, but that is for later.

virtcontainers/qemu.go
sboeuf: @mcastelino I totally agree with you here, but again I was not trying to fix q35, but more trying to match what was already in place. IMO, we should have a follow up PR updating q35 and virt accordingly.
But your call, we can make this happen directly from this PR.

sboeuf:
@mcastelino I'll modify this PR once https://github.com/intel/govmm/pull/53 lands.

sboeuf:
@mcastelino I've updated including the patch from govmm.

grahamwhaley:
CIs not happy. The 16.04 run looks like an OBS signing issue?:
```
Install Kata Containers OBS repository
gpg: no valid OpenPGP data found.
```

and F27 seems to be an OBS timeout.
/cc @marcov @jcvenegas for any input.

jodh-intel:
lgtm

jodh-intel:
Oooh - /me just finds https://status.opensuse.org/ :)

marcov:
download.opensuse.org  is up again

/test

PR #811, opened 2018-10-09
Refactor network.go

amshinde:
Split network.go into separate endpoint implementations.


virtcontainers/bridgedmacvlan_endpoint.go
```diff
@@ -0,0 +1,116 @@
...
+
+// NetworkPair returns the network pair of the endpoint.
+func (endpoint *BridgedMacvlanEndpoint) NetworkPair() *NetworkInterfacePair {
+	return &endpoint.NetPair
+}
+
+// Attach for virtual endpoint bridges the network pair and adds the
+// tap interface of the network pair to the hypervisor.
+func (endpoint *BridgedMacvlanEndpoint) Attach(h hypervisor) error {
+	networkLogger().Info("Attaching macvlan endpoint")
```
jodh-intel: For consistency, this should be:

```
networkLogger().WithField("endpoint-type", "bridged-macvlan").Info("Attaching endpoint")
```

But as mentioned recently, it would be good if we could factor out all these log calls for `Attach()` and `Detach()` and maybe get the caller to call `Type()` on the endpoint and log that in a single location (maybe in `add()` and `remove()` in `virtcontainers/default_network.go`?)

amshinde:
@sboeuf PTAL

virtcontainers/bridgedmacvlan_endpoint.go
amshinde: @jodh-intel Missed that from our previous conversation! I have fixed this now, by getting rid of all those calls and logging at a single location.

jodh-intel:
Thanks @amshinde.

lgtm

virtcontainers/qemu.go
```diff
@@ -833,7 +833,7 @@ func (q *qemu) hotplugVFIODevice(device *config.VFIODev, op operation) error {
 	return nil
 }
 
-func (q *qemu) hotplugMacvtap(drive *VirtualEndpoint) error {
+func (q *qemu) hotplugMacvtap(drive *VethEndpoint) error {
```
caoruidong: This function is for macvtap

amshinde: @caoruidong Yes, but the underlying interface is the Veth interface. Macvtap is used as the bridging mechanism. Answering your previous question, the difference is when you get a macvtap interface from the network plugin that you can directly pass to the VM vs an interface such as a veth that you get from the plugin wherein you use a macvtap or a bridge+tap for bridging the interface to the VM.

caoruidong: Oh, so the new MacvtapEndpoint is without veth pair? Got it.

caoruidong:
Why not create a sub-directory under virtcontainers?

jodh-intel:
/me grinds his teeth at CI failures - 2 nodes missing VT-x and a network timeout. Re-spinnning...

/cc @chavafg, @mnaser.

bergwolf:
Thanks @amshinde ! It makes the code a lot more clear.
LGTM!

devimc:
lgtm

PR #813, opened 2018-10-11
qemu: Disable the default romfile used by virtio-pci

### LINKED ISSUE #812: qemu: Disable the default romfile used by virtio-pci
Based on https://github.com/intel/govmm/pull/56, we should be able to provide an empty ROM file so that we don't bring unneeded dependency (`efi-virtio.rom` in this case) when we use Qemu.

sboeuf:
As we try to make sure we don't pull unneeded dependency when using
QEMU or NEMU as the hypervisor, and because SeaBIOS and OVMF firmware
already handle what's done by the default efi-virtio.rom binary, this
commit gets rid of this dependency by providing a default empty one.

Fixes #812

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -123,11 +123,11 @@
   revision = "3520598351bb3500a49ae9563f5539666ae0a27c"
 
 [[projects]]
-  digest = "1:c63a5bf4f3fd94ae838ce53e3492c0a467c40d09cada9301d228df7164ba4e55"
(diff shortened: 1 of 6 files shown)


bergwolf:
`open /sys/module/kvm_intel/parameters/unrestricted_guest: no such file or directory`
Maybe we should consider enable tcg when kvm is not supported?

caoruidong:
I think it is CI environment problem.

jodh-intel:
lgtm

devimc:
lgtm

PR #819, opened 2018-10-15
kata-check: do not require nested vt

bergwolf:
We do not really require nested VT to run kata containers. Let's not
depend on it being there.

Let's see if it fixes the recent nested vt related CI issues.

### DIFF
--- a/cli/kata-check.go
+++ b/cli/kata-check.go
@@ -291,7 +291,10 @@ var kataCheckCLICommand = cli.Command{
 		span, _ := trace(ctx, "kata-check")
 		defer span.Finish()
 
-		setCPUtype()
+		err = setCPUtype()
+		if err != nil {
+			return err
+		}
 
 		details := vmContainerCapableDetails{
 			cpuInfoFile:           procCPUInfo,
--- a/cli/kata-check.go
+++ b/cli/kata-check.go
@@ -60,7 +60,7 @@ const (
 var (
 	procCPUInfo  = "/proc/cpuinfo"
 	sysModuleDir = "/sys/module"
-	modInfoCmd   = "modinfo"
+	modProbeCmd  = "modprobe"
 )
 
 // variables rather than consts to allow tests to modify them
@@ -120,8 +120,9 @@ func haveKernelModule(module string) bool {
 		return true
 	}
 
-	// Now, check if the module is unloaded, but available
-	cmd := exec.Command(modInfoCmd, module)
+	// Now, check if the module is unloaded, but available.
+	// And modprobe it if so.
+	cmd := exec.Command(modProbeCmd, module)
 	err := cmd.Run()
 	return err == nil
 }
--- a/cli/kata-check_amd64.go
+++ b/cli/kata-check_amd64.go
@@ -67,7 +67,6 @@ func setCPUtype() {
 			"kvm_intel": {
 				desc: "Intel KVM",
 				parameters: map[string]string{
-					"nested": "Y",
 					// "VMX Unrestricted mode support". This is used
 					// as a heuristic to determine if the system is
 					// "new enough" to run a Kata Container
@@ -97,9 +96,6 @@ func setCPUtype() {
 			},
 			"kvm_amd": {
 				desc: "AMD KVM",
-				parameters: map[string]string{
-					"nested": "1",
-				},
 			},
 			"vhost": {
 				desc: msgKernelVirtio,
--- a/cli/kata-check_amd64.go
+++ b/cli/kata-check_amd64.go
@@ -6,9 +6,12 @@
 package main
 
 import (
+	"fmt"
 	"github.com/sirupsen/logrus"
 	"io/ioutil"
 	"strings"
+
+	vc "github.com/kata-containers/runtime/virtcontainers"
 )
 
 const (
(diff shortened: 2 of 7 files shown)


caoruidong:
If we run in a VM, do we still need this check to print a waring/error?

bergwolf:
@caoruidong No, the option is to allow guest to be able to use kvm in nested mode. kvm is working on the host w/o it so we can still launch kata containers inside a vm. I think some cloud vendors might want to disable it in the host kernel code but still allow kvm-intel to be present, which is likely what we saw recently in CI.

caoruidong:
@bergwolf Got it. What we want to check is the host environment but we cannot check it from guest. So we don't need this.
LGTM

bergwolf:
@marcov 
> for users wanting to run kata with in a VM with kvm, it would be useful to check if nested kvm is setup properly.

Nested kvm is only needed when users want to run nested guests within their own guests, which in kata containers case, translates into running nested guests within kata containers. I'm not sure that is a common use case that we want to support. E.g., we do not enable `CONFIG_KVM` in kata containers default kernel right now. I agree that there might be people wanting to run kata containers within kata containers, but that feels like a very uncommon use case and such customized expert users should be able to check and enable proper host and guest kernel configs on their own.

marcov:
@bergwolf I was referring to users running kata-containers runtime inside a VM that is managed  e.g. with libvirt.

If nesting mode is not enabled on the host, then kata-check reports an error when checking "Virtualization Support". In this case, it would be nice to have an hint that nesting should be enabled on the host.

marcov:
> If nesting mode is not enabled on the host, then kata-check reports an error when checking "Virtualization Support". In this case, it would be nice to have an hint that nesting should be enabled on the host.

However, this should not be handled by this PR.
lgtm

grahamwhaley:
CI machine kvm enablement should now be fixed... let's try that test again...
/test

bergwolf:
Of the latest five CIs tests, four of them failed `kata-check` due to missing vhost kernel module. The remaining one failed on boot time and memory foot print test:
```
19:06:23 time="2018-10-16T11:06:23Z" level=warning msg="Failed Maxval (8.140000 < 8.470350) for [boot-times]"
19:06:23 time="2018-10-16T11:06:23Z" level=warning msg="Check for [boot-times] failed [Failed]"
19:06:23 time="2018-10-16T11:06:23Z" level=warning msg=" with [[*F* boot-times 90.0% 114.5% 110.0% 20.0% 110.8% 116.4% 5.1% 1.3% 20]]"
19:06:23 time="2018-10-16T11:06:23Z" level=warning msg="Failed Minval (123134.400000 > 119967.900000) for [memory-footprint]"
19:06:23 time="2018-10-16T11:06:23Z" level=warning msg="Check for [memory-footprint] failed [Failed]"
19:06:23 time="2018-10-16T11:06:23Z" level=warning msg=" with [[*F* memory-footprint 92.3% 89.9% 107.7% 15.4% 89.9% 89.9% 0.0% 0.0% 1]]"
```

grahamwhaley:
@bergwolf metrics boot time is a known issue I need to look at - I have a feeling something changed or got merged recently - we have had a string of those.
metrics memory footprint is a new one, so we will have to keep an eye on that one.

chavafg:
I added `modinfo vhost` and `modinfo vhost_net` commands to the beginning of the jenkins job. As you can see in:
http://jenkins.katacontainers.io/job/kata-containers-runtime-ubuntu-16-04-PR/977/console
at the very beginning of the log, both modules are present. 
```
filename:       /lib/modules/4.4.0-128-generic/kernel/drivers/vhost/vhost.ko
description:    Host kernel accelerator for virtio
author:         Michael S. Tsirkin
license:        GPL v2
version:        0.0.1
srcversion:     4B2C406DD45508EBDDC1500
depends:        
retpoline:      Y
intree:         Y
vermagic:       4.4.0-128-generic SMP mod_unload modversions retpoline 
parm:           max_mem_regions:Maximum number of memory regions in memory map. (default: 64) (ushort)
filename:       /lib/modules/4.4.0-128-generic/kernel/drivers/vhost/vhost_net.ko
alias:          devname:vhost-net
alias:          char-major-10-238
description:    Host kernel accelerator for virtio net
author:         Michael S. Tsirkin
license:        GPL v2
version:        0.0.1
srcversion:     6C8BD745CFC0DA445E95381
depends:        vhost,macvtap
retpoline:      Y
intree:         Y
vermagic:       4.4.0-128-generic SMP mod_unload modversions retpoline 
parm:           experimental_zcopytx:Enable Zero Copy TX; 1 -Enable; 0 - Disable (int)
```

But then it fails when running the `kata-check` tool.  Found that the CI is running the tool without sudo, which is required by  `modprobe`.
https://github.com/kata-containers/tests/blob/master/.ci/install_runtime.sh#L39

I'll open a PR in `tests` repo to run it with the correct permissions. 

And maybe in this PR we can add more information to the log about what went wrong, so a user knows if it was a problem with permissions.

raravena80:
@bergwolf ping.

bergwolf:
https://github.com/kata-containers/tests/pull/823 is merged. Let's restart CI.

sboeuf:
@bergwolf @chavafg 
Just need some clarifications here. Do we have nested virtualization checks part of the CI now?
Correct me if I'm wrong but you want to remove the nested check from Kata itself, and move it to the responsibilty of the environment running Kata tests?

sboeuf:
@chavafg @bergwolf should we simply close this PR?

bergwolf:
@sboeuf The idea is that nested kvm is only required on L1 hypervisor if we want make the guest a L2 hypervisor. I'm trying to reduce the requirement because we do not need nested kvm on the host to run kata containers. One possible use case of nested kvm is to run kata containers inside kata containers. But I don't think we support that. And even if we do, we do not need to fail `kata-check` on platforms that do not support nested kvm.

sboeuf:
@bergwolf makes sense!
LGTM

grahamwhaley:
We can ignore the metrics CI (I could hand-trigger just that if we really needed, but I don't think it relates to this PR), and the codecov reduction is trivial - so, I think we could merge this. Maybe one of @marcov @chavafg @jodh-intel would like to confirm and force merge?

PR #823, opened 2018-10-16
create/run: Make bundle path default to cwd

### LINKED ISSUE #821: cannot run kata-runtime standalone  with default path for bundle
# Description of problem

 https://github.com/kata-containers/runtime/blob/master/cli/run.go#L34

We should set a default to the current directory.  Today this doesn't work.

```
cli.StringFlag{
			Name:  "bundle, b",
			Value: "",
			Usage: `path to the root of the bundle directory, defaults to the current directory`,
		},
```

To reproduce:
```
# create the top most bundle directory
mkdir /mycontainer
cd /mycontainer

# create the rootfs directory
mkdir rootfs

# export busybox via Docker into the rootfs directory
docker export $(docker create busybox) | tar -C rootfs -xvf -

runc spec

# run as root
cd /mycontainer
kata-runtime run mycontainerid

### PR DESCRIPTION
**Pull request by jodh-intel**
The bundle path was documented as defaulting to the current directory
but was not being set to that value if not explicitly specified.

Fixes #821.

### DIFF
--- a/cli/create.go
+++ b/cli/create.go
@@ -92,6 +92,36 @@ var createCLICommand = cli.Command{
 // Use a variable to allow tests to modify its value
 var getKernelParamsFunc = getKernelParams
 
+func handleFactory(ctx context.Context, runtimeConfig oci.RuntimeConfig) {
+	if !runtimeConfig.FactoryConfig.Template {
+		return
+	}
+
+	factoryConfig := vf.Config{
+		Template: true,
+		VMConfig: vc.VMConfig{
+			HypervisorType:   runtimeConfig.HypervisorType,
+			HypervisorConfig: runtimeConfig.HypervisorConfig,
+			AgentType:        runtimeConfig.AgentType,
+			AgentConfig:      runtimeConfig.AgentConfig,
+		},
+	}
+
+	kataLog.WithField("factory", factoryConfig).Info("load vm factory")
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: cli/create.go [hunk 1]
```diff
@@ -92,6 +92,36 @@ var createCLICommand = cli.Command{
 // Use a variable to allow tests to modify its value
 var getKernelParamsFunc = getKernelParams
 
+func handleFactory(ctx context.Context, runtimeConfig oci.RuntimeConfig) {
+	if runtimeConfig.FactoryConfig.Template == false {
```
marcov: Why not! 
:) I mean, why not `!` instead of `== false`

cli/create.go
```diff
@@ -92,6 +92,36 @@ var createCLICommand = cli.Command{
...
+
+	f, err := vf.NewFactory(ctx, factoryConfig, true)
+	if err != nil {
+		kataLog.WithError(err).Warn("load vm factory failed, about to create new one")
+		f, err = vf.NewFactory(ctx, factoryConfig, false)
+		if err != nil {
+			kataLog.WithError(err).Warn("create vm factory failed")
+		}
+	}
+	if err == nil {
```
marcov: Maybe better to return early at  line 117 above to avoid this extra check?

jodh-intel:
Hi @marcov - good points! Branch updated.

Issue #571, opened 2018-08-10
[feature request] mount blockdevices in the guest VM instead of passing them through as volume


hex2a:
# Description of problem
When specifying a block device as volume, it is passed as block device, however in many cases a user might want the filesystem on the device to be mounted.

An implementation similar to runq [0] would also allow to specify the KVM cache mode.
[0] https://github.com/gotoz/runq#storage

# Expected result
block device is mounted inside the VM, and exposed to the container as filesystem/bind mount.

# Actual result
block device is passed through as block device

bergwolf:
The main question to support this is how we can integrate it with upper layers in the stack.

1. For docker, I think we already support mounting block device as a volume. E.g., you can do
`docker run -d -v /dev/sdx:/mnt/data busybox`. This introduces incompatible behavior w.r.t. docker's. And it would require `disable_block_device_use = false` in the kata containers' configuration.

2. For kubernetes w/ containerd or cri-o, what we get from upper layer is always a mounted hostpath for any volume. So we cannot pass the underlying block device to the guest, otherwise we'll face potential kernel crash or even silent data corruption.

3. For kubernetes w/ frakti, there are a few block devices (cinder rbd, gce pd and ceph rbd) that can be passed as block device directly to the guest. And these are implemented via special flexvolume driver that bypasses mounting the device on the host and translates the volume information into a block device based volume that frakti can understand.

WeiZhang555:
I think mounting the block device for user inside Guest could be a useful use case, I was also thinking about this for long time. The only concern is that it's not fully compatible with `Mount` semantics, and maybe we will have other use cases that requires a "-v /dev/sda:/dev/sda"

hex2a:
@WeiZhang555 
I added the `runq` approach for reference.

if it is passed as `-v /dev/sda:/dev/sda`, it is passed as block device, if it should be mounted its specified it like this:
```
--volume <image  name>:/dev/disk/<cache type>/<filesystem type>/<mountpoint>
```

WeiZhang555:
@hex2a this is a bit hacky and rely too much on the guest mountpoint pattern, I don't think it's a good way

amshinde:
@WeiZhang555 @bergwolf What if we mount the volume if the destination mount point is any other path besides "/dev/*" ?
I realized that this will be useful for workloads(esp database workloads) that use volumes, 9p is an issue and the Entrypoint depends on the volume to be present.

WeiZhang555:
[...]

PR #828, opened 2018-10-17
newContainer: Not attach device if it is a CDROM

### LINKED ISSUE #829: "docker run --privileged" got "docker: Error response from daemon: OCI runtime create failed: QMP command failed: unknown."
10月 17 15:28:30 teawater-virtual-machine kata-runtime[91986]: time="2018-10-17T15:28:30.891596459+08:00" level=info msg="Attaching block device" arch=amd64 command=create container=5edec94fbf4076667aa8ee546f10cde98d6ef62f3097c2a3889ee316ea0fd0e9 device=/dev/sr0 name=kata-runtime pid=91986 source=virtcontainers subsystem=device
10月 17 15:28:30 teawater-virtual-machine kata-runtime[91986]: time="2018-10-17T15:28:30.891779981+08:00" level=info msg="{\"arguments\":{\"driver\":\"raw\",\"file\":{\"driver\":\"file\",\"filename\":\"/dev/sr0\"},\"node-name\":\"drive-e6670677eccb9cab\"},\"execute\":\"blockdev-add\"}" arch=amd64 command=create container=5edec94fbf4076667aa8ee546f10cde98d6ef62f3097c2a3889ee316ea0fd0e9 name=kata-runtime pid=91986 source=virtcontainers subsystem=qmp
10月 17 15:28:30 teawater-virtual-machine kata-runtime[91986]: time="2018-10-17T15:28:30.977805252+08:00" level=info msg="{\"error\": {\"class\": \"GenericError\", \"desc\": \"Could not open '/dev/sr0': No medium found\"}}" arch=amd64 command=create container=5edec94fbf4076667aa8ee546f10cde98d6ef62f3097c2a3889ee316ea0fd0e9 name=kata-runtime pid=91986 source=virtcontainers subsystem=qmp
10月 17 15:28:30 teawater-virtual-machine kata-runtime[91986]: time="2018-10-17T15:28:30.978272497+08:00" level=info msg="Unplugging block device" arch=amd64 command=create container=5edec94fbf4076667aa8ee546f10cde98d6ef62f3097c2a3889ee316ea0fd0e9 device=/dev/dm-0 name=kata-runtime pid=91986 source=virtcontainers subsystem=device

teawater:
Got "docker: Error response from daemon: OCI runtime create failed:
QMP command failed: unknown." when "docker run --privileged" with kata.
In qemu part, it got:
"Could not open '/dev/sr0': Read-only file system"
or
"No medium found"
The cause is qemu need open block device to get its status.
But /dev/sr0 is a CDROM that cannot be opened.

This patch let newContainer doesn't attach device if it is a CDROM
to handle the issue.

Fixes #829

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -28,6 +28,25 @@ import (
 	"github.com/kata-containers/runtime/virtcontainers/utils"
 )
 
+// https://github.com/torvalds/linux/blob/master/include/uapi/linux/major.h
+// This file has definitions for major device numbers.
+var cdromMajors = map[int64]string{
+	11: "SCSI_CDROM_MAJOR",
+	15: "CDU31A_CDROM_MAJOR",
+	16: "GOLDSTAR_CDROM_MAJOR",
+	17: "OPTICS_CDROM_MAJOR",
+	18: "SANYO_CDROM_MAJOR",
+	20: "MITSUMI_X_CDROM_MAJOR",
+	23: "MITSUMI_CDROM_MAJOR",
+	24: "CDU535_CDROM_MAJOR",
+	25: "MATSUSHITA_CDROM_MAJOR",
+	26: "MATSUSHITA_CDROM2_MAJOR",
+	27: "MATSUSHITA_CDROM3_MAJOR",
+	28: "MATSUSHITA_CDROM4_MAJOR",
+	29: "AZTECH_CDROM_MAJOR",
+	32: "CM206_CDROM_MAJOR",
+}
+
 // Process gathers data related to a container process.
 type Process struct {
 	// Token is the process execution context ID. It must be
@@ -522,6 +541,20 @@ func (c *Container) unmountHostMounts() error {
 	return nil
 }
 
+func filterDevices(sandbox *Sandbox, c *Container, devices []ContainerDevice) (ret []ContainerDevice) {
+	for _, dev := range devices {
+		major, _ := sandbox.devManager.GetDeviceByID(dev.ID).GetMajorMinor()
+		if _, ok := cdromMajors[major]; ok {
+			c.Logger().WithFields(logrus.Fields{
+				"device": dev.ContainerPath,
+			}).Info("Not attach device because it is a CDROM")
+			continue
+		}
+		ret = append(ret, dev)
+	}
+	return
+}
+
 // newContainer creates a Container structure from a sandbox and a container configuration.
 func newContainer(sandbox *Sandbox, contConfig ContainerConfig) (*Container, error) {
 	span, _ := sandbox.trace("newContainer")
@@ -609,11 +642,12 @@ func newContainer(sandbox *Sandbox, contConfig ContainerConfig) (*Container, err
 				return &Container{}, err
 			}
 
-			c.devices = append(c.devices, ContainerDevice{
+			storedDevices = append(storedDevices, ContainerDevice{
 				ID:            dev.DeviceID(),
 				ContainerPath: info.ContainerPath,
 			})
 		}
+		c.devices = filterDevices(sandbox, c, storedDevices)
 	}
 
 	return c, nil


jodh-intel:
Hi @teawater - thanks for raising. I'm a little confused though - we don't support `--privileged`:

- https://github.com/kata-containers/documentation/blob/master/Limitations.md#docker-run---privileged

lifupan:
Hi @jodh-intel, can we give a container right to access the devices in the guest using --privileged ? Since 
when add a block device to container using "-v /dev/sdbX:/dev/sdbX", kata only passthrough this
block device into container,  but it cannot be mounted since container doesn't has the right.

raravena80:
@teawater ping, any updates?

teawater:
I didn't see any good reason that kata doesn't support "--privileged".
Maybe it is not a safe way.  But "--privileged" is a very simple way to help the user access block device inside the container at least.  (BTW, we have a long way to go with access block device.  I think we need a configurable way.)

amshinde:
@teawater Kata does support some notion of privileged. The container has access to all devices inside the guest, we pass as many devices as we can from the host and the container should run with all the capabilities that are associated with privileged mode, basically all capabilities.
 I am interested in understanding what is the exact error you got with qmp first, rather than ignoring the error altogether.
@lifupan You should be able to mount a device in privileged mode, please file a bug if you are seeing otherwise. I have tried this in the past simply by passing --cap-add SYS_ADMIN to the container.

teawater:
@amshinde This is a qemu issue.  qemu need open block device to get its status.
But /dev/sr0 is a cdrom.
With disk inside, qemu will get:
"Could not open '/dev/sr0': Read-only file system\"
Without disk inside, qemu will get:
No medium found
So I think ignore the error is a suitable way to handle it.

WeiZhang555:
Ignoring the "AttachError" may cover some real issue, so I think it's not safe.

@jodh-intel We do actually support `--privileged`, except the `/dev/sr0` issue found by @teawater , I've met same issue before :-) , without cdrom, `--privileged` can work in "most" cases :-)

@lifupan 
>Hi @jodh-intel, can we give a container right to access the devices in the guest using --privileged ? Since when add a block device to container using "-v /dev/sdbX:/dev/sdbX", kata only passthrough this
block device into container, but it cannot be mounted since container doesn't has the right.

This is a good question, we have a feature request discussing this, see 
https://github.com/kata-containers/runtime/issues/571

In a word, I think you should first considering `--cap-add SYS_ADMIN` as @amshinde suggested, `--privileged` may not be safe though it's convenient.

We'd better find a more robust way to handle `/dev/sr0` error than ignoring the `AttachError`

teawater:
I make a patch that not attach device if it is a CDROM.

virtcontainers/container.go
```diff
@@ -28,6 +28,23 @@ import (
 	"github.com/kata-containers/runtime/virtcontainers/utils"
 )
 
+var cDROMMajors = map[int64]string{
```
WeiZhang555: The name looks weird, how about cdromMajors?

Also can you add some comments with reference links to this major definition? Then we can easily check if these are all :-)

teawater:
@WeiZhang555 made a new version according to your comments.

WeiZhang555:
LGTM

/test

bergwolf:
Is CDROM the only device type we cannot handle right now? What about floppy disks? There are also other types of devices like `/dev/kvm`, `/dev/autofs` etc. that can be confusing to users if they represent the host ones. I wonder if we want to use a white list instead.

jshachm:
+1 @bergwolf   
Now that we begin to deal with the special device, white list is better.

teawater:
I agree with a device list.
And what about let it can be config by the config file?

amshinde:
@bergwolf Today we ignore all devices except block and vfio.

virtcontainers/container.go
```diff
@@ -609,12 +640,13 @@ func newContainer(sandbox *Sandbox, contConfig ContainerConfig) (*Container, err
...
 			}
 
-			c.devices = append(c.devices, ContainerDevice{
+			storedDevices = append(storedDevices, ContainerDevice{
 				ID:            dev.DeviceID(),
 				ContainerPath: info.ContainerPath,
 			})
 		}
 	}
+	c.devices = filterDevices(sandbox, c, storedDevices)
```
sboeuf: The filtering should not happen every time the `container` instance is created in memory. It only needs to happen at `create`, meaning this filtering function should only be called from the case `err != nil` here.

sboeuf:
@grahamwhaley is the failure on metrics-16-04 expected?
If that's the case, and if this is okay to ignore it, please go ahead and merge this ;)

grahamwhaley:
Yep - metrics CI is coming back up today - just found one more 'cleanup' issue to fix now we are on baremetal build slaves. so, ignore metrics here right now, and i'll merge...

Issue #844, opened 2018-10-19
factory fails when kernel parameter config is empty


bergwolf:
When kernel parameter is set to empty, vm factory fails the vm config comparison but look closely the two config values are actually the same.

[...]

PR #845, opened 2018-10-19
factory: use customised deep compare

bergwolf:
reflect.DeepEqual would return false when comparing nil map/slice with
empty map/slice. We would want to return success in such case, since it
is possible for upper layers to send these kind of configs.

Fixes: #844

### DIFF
--- a/virtcontainers/factory/factory.go
+++ b/virtcontainers/factory/factory.go
@@ -100,6 +100,71 @@ func resetHypervisorConfig(config *vc.VMConfig) {
 	config.ProxyConfig = vc.ProxyConfig{}
 }
 
+func compareStruct(foo, bar reflect.Value) bool {
+	for i := 0; i < foo.NumField(); i++ {
+		if !deepCompareValue(foo.Field(i), bar.Field(i)) {
+			return false
+		}
+	}
+
+	return true
+}
+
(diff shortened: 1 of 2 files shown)


virtcontainers/factory/factory.go
```diff
@@ -100,6 +100,71 @@ func resetHypervisorConfig(config *vc.VMConfig) {
 	config.ProxyConfig = vc.ProxyConfig{}
 }
 
+func compareStruct(foo, bar reflect.Value) bool {
+	for i := 0; i < foo.NumField(); i++ {
```
jodh-intel: Missing check on field counts for `foo` and `bar`? They might compare as the same but if `bar` has extra fields that would be incorrect surely?

bergwolf: They have to pass `foo.Type() == bar.Type()` so they must be of the same struct type.

jodh-intel: Ah - good point!

jodh-intel:
lgtm

WeiZhang555:
LGTM

CI retriggered, let's see if there's any real issue.

jodh-intel:
Still seeing network timeouts.

/test

bergwolf:
CI green now.

PR #847, opened 2018-10-19
Support mount, uts and ipc namespaces in the host

### LINKED ISSUE #160: Enable mount namespace
_From @amshinde on May 6, 2017 1:47_

A new mount namespace needs to be created before performing mounts

_Copied from original issue: containers/virtcontainers#236_

devimc:
Unlike other runtimes, we have several components like qemu, kata-shim and
kata-proxy that should be placed in new namespace to improve the isolation and
security. The problem is that we need persistent namespaces and in the case of
the mount namespace (aka headache namespace ;)), it MUST BE created before
starting golang execution and in a new mount point with a propagation different
to shared, even worse, to make it persistent the new namespace
(/proc/PID/ns/mnt) MUST be 'bind mounted' by a process that doesn't know of the
existence of it. To get more information about this, see
github.com/karelzak/util-linux/commit/c84f2590dfdb8570beeb731e0105f8fe597443d1

With these new functions now is possible to join, create, remove and make
persistent namespaces. Next diagram can help us to understand how it works.

```
 ============================================================================
| Host's namespace                                    | New namespace!       |
|                                                     |                      |
| ---------------------------------------------------------------------------|
| C (no threads, no goroutines, NO GOLANG!)           |                      |
|                                                     |                      |
|  ---------   ---------------                        |                      |
| | Runtime |-|listen children|--|fork-setns-unshare|-|-------.              |
|  ---------   ---------------\                       |       |              |
|      |               |       \                      |       |              |
|    fork              |        Wait children &       |       |              |
|      |               |             exit             |       |              |
|----------------------------------------------------------------------------|
| Golang               |                              |       |              |
|      |        join,create,make,remove               |       |              |
|  ---------    persistent namespace                  |   ---------          |
| | Runtime |          |                              |  | Runtime |         |
|  ---------        --------                          |   ---------          |
|      |           |-Create |                         |       |              |
|  parse jsons     |-Start  |                         | Create,start,etc.    |
|  and cmdline     |-Delete |                         |   Containers         |
|       \          |-etc    |                         |       |              |
|        \         |--------                          |      exit            |
|         --subcmd-'                                  |                      |
|          \                                          |                      |
|           - exit                                    |                      |
============================================================================
```

fixes #160

Depends-on: github.com/kata-containers/tests#946

### DIFF
--- a/cli/create.go
+++ b/cli/create.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"syscall"
 
 	vc "github.com/kata-containers/runtime/virtcontainers"
 	vf "github.com/kata-containers/runtime/virtcontainers/factory"
@@ -85,6 +86,8 @@ var createCLICommand = cli.Command{
 
 // Use a variable to allow tests to modify its value
 var getKernelParamsFunc = getKernelParams
+
+var mountRootfsFunc = mountRootfs
 
 func handleFactory(ctx context.Context, runtimeConfig oci.RuntimeConfig) {
 	if !runtimeConfig.FactoryConfig.Template {
@@ -247,6 +250,10 @@ func setKernelParams(containerID string, runtimeConfig *oci.RuntimeConfig) error
 
 func createSandbox(ctx context.Context, ociSpec oci.CompatOCISpec, runtimeConfig oci.RuntimeConfig,
 	containerID, bundlePath, console, consoleSocket string, disableOutput, systemdCgroup bool) (vc.Process, error) {
+	if err := newPersistentNamespaces(containerID, "", ociSpec.Linux.Namespaces); err != nil {
+		return vc.Process{}, err
+	}
+
 	span, ctx := trace(ctx, "createSandbox")
 	defer span.Finish()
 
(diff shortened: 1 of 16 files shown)


sboeuf:
@devimc thanks for the PR, but please be more verbose about the problem statement, and the reasons behind the choices you made. It is a pretty huge PR affecting main components, and it needs to be clear about the `why` and `how`.

jshachm:
waiting for the doc~ @devimc  and maybe we need to take `mount ns leak` into consideration

sboeuf:
@bergwolf @WeiZhang555 @egernst PTAL as this brings some heavy architectural changes. It's very interesting and complex as Golang is really not friendly with namespaces.

@devimc are you planning on presenting this PR to the Kata Arch Committee? I feel this needs to be understood and acknowledged by AC members before we can move forward.

devimc:
@sboeuf  @WeiZhang555  yes I'd like to present this in the next Arch meeting

sboeuf:
@devimc 
> yes I'd like to present this in the next Arch meeting

good! I think we'll have the next AC meeting on November 19th (unless skipped for low attendance), but I have good hope it'll be maintained.

@WeiZhang555 
>I have a question in mind, with this feature, can we add an extention: if two containers share same namespace, put them into same VM(POD) ?

Yes definitely, the `parent-runtime` process can spawn the `child-runtime` process into any namespace, even if existing already.

bergwolf:
Another interesting thing to see is how it copes with containerd-shim-v2. /cc @lifupan @gnawux

gnawux:
Not a bad idea fundamentally, but I think we'd better keep the shim-v2 and the cli are equivalent in features. Let's discuss this in the next meeting.

By the way, @devimc should you change the `utc` in PR title to `uts`?

devimc:
@kata-containers/runtime  @bergwolf @gnawux @sboeuf  please take a look

amshinde:
@devimc This needs a rebase. 
We talked about having this in the architecture committee, I know I missed the last one. Was this discussed at all?

raravena80:
@devimc any updates?

devimc:
@raravena80 ehmm nop, I'd like to present this in the kata arch meeting

jodh-intel:
Hi @devimc - Thanks for raising! As @sboeuf has noted, this is a complex one. It does kind of feel like this issue should be handled by fixing golang rather than us having to jump through "hoops of fire" as it were ;) On that topic, is anyone aware of a golang github issue regarding the namespace issues? I took a quick look but didn't find anything useful (ironically, searching on google lead me back to the Kata github project! ;)

Awesome ASCII art btw, but it would be helpful if you could provide a step-by-step explanation of how this works to allow us all to "digest" this PR more quickly.

caoruidong:
What issue? The namespace problem fixed in golang 1.10? https://github.com/golang/go/issues/20676 #148

jodh-intel:
@caoruidong - I was referring to @sboeuf's comment:

> It's very interesting and complex as Golang is really not friendly with namespaces.

devimc:
@jodh-intel https://github.com/golang/go/issues/8676 `Status changed to WontFix`

jodh-intel:
@devimc - This has been open for a while now (and the branch is now conflicted).

I think there is a general feeling that although this PR solves a problem, it introduces more complexity that we are comfortable with, unless there is a *very* good reason for it. It also introduces C code. I understand why that is, but again, I think we'd rather not use C unless there is a super-compelling reason for it due to the potential security implications. It would also require us to setup new static analysis, etc, etc.

Hence, please could you / @amshinde provide further details on why we need this with, ideally with some examples. Trying to trace back through the issues leads to the initial issue... containing no details :-((

I think you did present this at the Architecture Committee meeting? Please could you add details here of what the feedback was. If there wasn't any, or wasn't enough, I think it might be worth you raising this topic again in the AC meeting. Now that folk have had time to consider the impact of adding this change we might get more feedback.

PR #851, opened 2018-10-23
vfio: Change the way the driver is fetched

### LINKED ISSUE #612: Virtio network interfaces failed to be bound back to the driver
# Description of problem

In case of physical network interface, the driver would be virtio_net if using a vm as the host (nested virtualized). After the teardown of container, kata tried to bind interfaces back to virtio_net, but there was no path of /sys/bus/pci/drivers/virtio_net/ (actually it should be /sys/bus/virtio/drivers/virtio_net/). There are two ways fixing that:

1. Whenever using a virtio_ driver, we try to use  /sys/bus/pci/drivers/virtio-pci/{bind, unbind} to manipulate interfaces and drivers
2. Whenever using a virtio_ driver, we try to replace bus type from pci to virtio, that is, /sys/bus/virtio/drivers/virtio_net/{bind, unbind}

amshinde:
Instead of using ethtool for getting the driver for network
devices, use sysfs instead. This is because in case of virtio
devices, ethtool returns virtio-net instead of virtio-pci for
virtio network devices. We need to bind/unbind from virtio-pci
driver in case of virtio-net devices.

Fixes #612

### DIFF
--- a/virtcontainers/physical_endpoint.go
+++ b/virtcontainers/physical_endpoint.go
@@ -8,6 +8,7 @@ package virtcontainers
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


jodh-intel:
lgtm

amshinde:
@sboeuf We need `ethtool` to check if the network interface has a PCI address and thereby a physical device.

PR #852, opened 2018-10-24
Add support for ipvlan network driver

amshinde:
Ipvlan is supported using tc filtering rules to redirect traffic between the ipvlan interface and tap interface.

### DIFF
--- a/virtcontainers/ipvlan_endpoint.go
+++ b/virtcontainers/ipvlan_endpoint.go
@@ -1,0 +1,120 @@
+// Copyright (c) 2018 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package virtcontainers
+
+import (
+	"fmt"
+
+	"github.com/containernetworking/plugins/pkg/ns"
+)
+
+// IPVlanEndpoint represents a ipvlan endpoint that is bridged to the VM
+type IPVlanEndpoint struct {
+	NetPair            NetworkInterfacePair
+	EndpointProperties NetworkInfo
+	EndpointType       EndpointType
+	PCIAddr            string
+}
+
+func createIPVlanNetworkEndpoint(idx int, ifName string) (*IPVlanEndpoint, error) {
+	if idx < 0 {
+		return &IPVlanEndpoint{}, fmt.Errorf("invalid network endpoint index: %d", idx)
+	}
+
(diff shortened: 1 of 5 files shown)


amshinde:
Needs to be merged after the PR for tc-filter.

jodh-intel:
Is this related to #827? It would help if you could add a full PR reference if that's the case to make it easier to track what is going on :)

amshinde:
cc @mcastelino

virtcontainers/ipvlan_endpoint.go
```diff
@@ -0,0 +1,120 @@
...
+
+package virtcontainers
+
+import (
+	"fmt"
+
+	"github.com/containernetworking/plugins/pkg/ns"
+)
+
+// IPVlanEndpoint represents a macvlan endpoint that is bridged to the VM
```
mcastelino: @amshinde copy paste error here

`IPVlanEndpoint represents a macvlan endpoint`

virtcontainers/ipvlan_endpoint.go
```diff
@@ -0,0 +1,120 @@
...
+	EndpointType       EndpointType
+	PCIAddr            string
+}
+
+func createIPVlanNetworkEndpoint(idx int, ifName string) (*IPVlanEndpoint, error) {
+	if idx < 0 {
+		return &IPVlanEndpoint{}, fmt.Errorf("invalid network endpoint index: %d", idx)
+	}
+
+	// Use tc filtering for ipvlan, since the other inter networking models will
```
mcastelino: @amshinde are there any other cases you know of where macvlan or bridge does not work. Ideally you need to have a check that sees if the current interworkingModel works for this type of interface. This function can return what interworkingModel will support it, if the user selected one does not. This will make the code generic.

amshinde: @mcastelino Yes, I need to handle this for macvlan as well. I agree this code needs to be generic.
I think it will make sense to have an interface for each endpoint that advertises the internetworking models it supports. I will add that for every endpoint so that we can make a decision to fallback to a model that works for an endpoint. 
I'll do this in a separate PR.

amshinde:
@jodh-intel I shall add a reference going forward :) 
Now that #827 is merged, this is ready for review :) PTAL.

bergwolf:
LGTM!

PR #858, opened 2018-10-25
vendor: Rely on new agent package types

### LINKED ISSUE #856: vendor: Update agent vendoring relying on pkg/types
Some agent types definition that were generic enough to be reused everywhere, have been split from the initial grpc package.

This prevents from importing the entire protobuf package through the grpc one, and prevents binaries such as kata-netmon to stay in sync with the types definitions.

### LINKED ISSUE #857: netmon: Rely on the new agent pkg/types
This will help getting rid of the burden of maintaining the sync between the types defined by the agent and the ones defined by netmon.

sboeuf:
Now that the agent has split some generic structures definitions from the gRPC protocol itself, this pull request takes care of the revendoring of this change from the runtime.
Also, because it allows binary importing this path to avoid overhead due to the whole protocol definition, the binary `kata-netmon` can now rely on those types instead of duplicating them.

Fixes #856 
Fixes #857


sboeuf:
Depends on https://github.com/kata-containers/agent/pull/400

jodh-intel:
lgtm

sboeuf:
Just updated the vendoring now that agent#400 has been merged. Let's wait for the CI and merge that ;)

sboeuf:
This PR is ready and got all approvals it needs. Github seems to be stuck not reporting the true status of `pullapprove` here.
Let's merge it.

PR #872, opened 2018-10-30
enable default network endpoints hotplug for vm factory

bergwolf:
For non-factory case, default network endpoints are cold plugged. But for vm factory, we need to hotplug them after vm is assigned to the new sandbox.

### DIFF
--- a/virtcontainers/network.go
+++ b/virtcontainers/network.go
@@ -1430,9 +1430,9 @@ type network interface {
 	run(networkNSPath string, cb func() error) error
 
 	// add adds all needed interfaces inside the network namespace.
-	add(sandbox *Sandbox) error
+	add(sandbox *Sandbox, hotplug bool) error
 
 	// remove unbridges and deletes TAP interfaces. It also removes virtual network
 	// interfaces and deletes the network namespace.
-	remove(sandbox *Sandbox) error
+	remove(sandbox *Sandbox, hotunplug bool) error
 }
--- a/virtcontainers/network.go
+++ b/virtcontainers/network.go
@@ -1194,6 +1194,7 @@ func generateInterfacesAndRoutes(networkNS NetworkNamespace) ([]*types.Interface
 			Name:        endpoint.Name(),
 			Mtu:         uint64(endpoint.Properties().Iface.MTU),
 			HwAddr:      endpoint.HardwareAddr(),
+			PciAddr:     endpoint.PciAddr(),
 		}
 
 		ifaces = append(ifaces, &ifc)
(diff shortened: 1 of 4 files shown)


bergwolf:
The testing PR is also green kata-containers/tests/pull/875

virtcontainers/network.go
```diff
@@ -1430,9 +1430,9 @@ type network interface {
 	run(networkNSPath string, cb func() error) error
 
 	// add adds all needed interfaces inside the network namespace.
-	add(sandbox *Sandbox) error
+	add(sandbox *Sandbox, hotattach bool) error
```
sboeuf: s/`hotattach`/`hotplug`/

virtcontainers/network.go
```diff
@@ -1430,9 +1430,9 @@ type network interface {
...
 
 	// remove unbridges and deletes TAP interfaces. It also removes virtual network
 	// interfaces and deletes the network namespace.
-	remove(sandbox *Sandbox) error
+	remove(sandbox *Sandbox, hotdetach bool) error
```
sboeuf: s/`hotdetach`/`hotplug`/

bergwolf: you mean `hotunplug` right?

bergwolf:
@sboeuf I don't think vm factory should make any difference for users w.r.t. network setup. Right now if users call `docker run --runtime kata-runtime busybox`, they won't have any network setup in the guest because we intentionally skip network setup for vm factory (due to cold plug).

With this PR, vm factory looks just like the non-factory case from user point of view. Users get network connectivity if netns is setup with docker prestart hooks and conventional CNI plugins. They can still skip prestart hooks and use `kata-runtime network` subcommand or enlightened CNI plugins to add network devices. It doesn't affect user behavior whether vm factory is in use or not.

virtcontainers/network.go
sboeuf: yes sorry

sboeuf:
@bergwolf 
>With this PR, vm factory looks just like the non-factory case from user point of view. Users get network connectivity if netns is setup with docker prestart hooks and conventional CNI plugins. They can still skip prestart hooks and use kata-runtime network subcommand or enlightened CNI plugins to add network devices. It doesn't affect user behavior whether vm factory is in use or not.

I understand but that means for a factory case, we're also making the network creation a default. I'm fine with that, I simply thought you wanted more flexibility for the factory use case.

This code LGTM if this is the way you wanna go :)

bergwolf:
Thanks @sboeuf . Yes, that's what I wanted to do. If users don't want the default network, they can set `disable_new_netns = true`. vm factory is not supposed to control the network configuration.

ping @kata-containers/runtime PTAL.

lifupan:
LGTM

PR #874, opened 2018-10-31
[WIP][RFC]persist: baseline persist data format

### LINKED ISSUE #803: [RFC] baseline persist data for live-upgrade

>1) kata-runtime

These files including:
```
├── agent.json
│   ├── devices.json
│   ├── mounts.json
│   ├── process.json
│   └── state.json
├── devices.json
├── hypervisor.json
├── lock
├── network.json
├── proxy.sock
└── state.json
```

e.g.

```
type DeviceInfo struct {
    HostPath string

    ContainerPath string

    // p - FIFO
    // More info in mknod(1).
    DevType string

    Major int64
    Minor int64

    FileMode os.FileMode

    // id of the device owner.
    UID uint32

    // id of the device group.
    GID uint32

    // device was hotplugged.
    Hotplugged bool

    ID string

}
```

WeiZhang555:
Fixes #803

The disk persist data should be "versioned" and baselined, any modification in
persist data should be considered potential break of backward compatibility.

TODO LIST:
- [x] network endpoints data isn't formalized
- [x] config.json from `/var/lib/vc/sbs/<sid>/`
- [ ] `/var/run/kata-containers` can be re-organized or merged with `/var/run/vc`

### DIFF
--- a/virtcontainers/persist/api/device.go
+++ b/virtcontainers/persist/api/device.go
@@ -1,0 +1,100 @@
+// Copyright (c) 2016 Intel Corporation
+// Copyright (c) 2018 Huawei Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package persistapi
+
+// ============= sandbox level resources =============
+
+// BlockDrive represents a block storage drive which may be used in case the storage
+// driver has an underlying block storage device.
+type BlockDrive struct {
+	// File is the path to the disk-image/device which will be used with this drive
+	File string
+
+	// Format of the drive
+	Format string
+
+	// ID is used to identify this drive in the hypervisor options.
+	ID string
+
+	// Index assigned to the drive. In case of virtio-scsi, this is used as SCSI LUN index
+	Index int
+
+	// PCIAddr is the PCI address used to identify the slot at which the drive is attached.
+	PCIAddr string
+
+	// SCSI Address of the block device, in case the device is attached using SCSI driver
+	// SCSI address is in the format SCSI-Id:LUN
+	SCSIAddr string
+
+	// VirtPath at which the device appears inside the VM, outside of the container mount namespace
+	VirtPath string
+}
+
+// VFIODev represents a VFIO drive used for hotplugging
+type VFIODev struct {
+	// ID is used to identify this drive in the hypervisor options.
+	ID string
+
+	// Type of VFIO device
+	Type string
+
+	// BDF (Bus:Device.Function) of the PCI address
+	BDF string
+
+	// Sysfsdev of VFIO mediated device
+	SysfsDev string
+}
+
+// VhostUserDeviceAttrs represents data shared by most vhost-user devices
+type VhostUserDeviceAttrs struct {
+	DevID      string
+	SocketPath string
+	Type       string
+
+	// MacAddress is only meaningful for vhost user net device
+	MacAddress string
+}
+
+// DeviceState is sandbox level resource which represents host devices
+// plugged to hypervisor, one Device can be shared among containers in POD
+// Refs: virtcontainers/device/drivers/generic.go:GenericDevice
+type DeviceState struct {
+	ID string
+
+	// Type is used to specify driver type
+	// Refs: virtcontainers/device/config/config.go:DeviceType
+	Type string
+
+	RefCount    uint
+	AttachCount uint
+
+	// Type of device: c, b, u or p
+	// c , u - character(unbuffered)
+	// p - FIFO
+	// b - block(buffered) special file
+	// More info in mknod(1).
+	DevType string
+
+	// Major, minor numbers for device.
+	Major int64
+	Minor int64
+
+	// DriverOptions is specific options for each device driver
+	// for example, for BlockDevice, we can set DriverOptions["blockDriver"]="virtio-blk"
+	DriverOptions map[string]string
+
+	// ============ device driver specific data ===========
+	// BlockDrive is specific for block device driver
+	BlockDrive *BlockDrive `json:",omitempty"`
+
+	// VFIODev is specific VFIO device driver
+	VFIODevs []*VFIODev `json:",omitempty"`
+
+	// VhostUserDeviceAttrs is specific for vhost-user device driver
+	VhostUserDev *VhostUserDeviceAttrs `json:",omitempty"`
+	// ============ end device driver specific data ===========
+}
--- a/virtcontainers/persist/api/config.go
+++ b/virtcontainers/persist/api/config.go
@@ -1,0 +1,195 @@
+// Copyright (c) 2016 Intel Corporation
+// Copyright (c) 2018 Huawei Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package persistapi
+
+// Param is a key/value representation for hypervisor and kernel parameters.
+type Param struct {
+	Key   string
+	Value string
+}
+
+// HypervisorConfig saves configurations of sandbox hypervisor
+type HypervisorConfig struct {
+	// NumVCPUs specifies default number of vCPUs for the VM.
+	NumVCPUs uint32
+
+	//DefaultMaxVCPUs specifies the maximum number of vCPUs for the VM.
+	DefaultMaxVCPUs uint32
+
+	// DefaultMem specifies default memory size in MiB for the VM.
+	MemorySize uint32
+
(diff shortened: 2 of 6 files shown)


WeiZhang555:
@amshinde Can you help me fix the network endpoint persist data? It was content from "network.json", currently it saves quite a lot of data which contains some unnecessary part, e.g. "EndpointProperties".

https://github.com/kata-containers/runtime/pull/874/files#diff-38d341573c08cdfaf94b6efd91f2ed8dR13

I think you should be right person who is most familiar with "*_endpoints.go" @amshinde

bergwolf:
@WeiZhang555 Nice! I think we should also clarify a few things in order to roll out the v1 persistent data format:
1. The target of backward compatibility is to let newer runtime be able to read and process persistent data of old runtime
2. When removing a field, we should keep obsolete fields for a few releases even if they are not used, so that the json package can still unpack the old format 
3. When adding a new field, it should have always have a default working value so that it still works when new runtime reads from old persistent data

Also since the PR is an incompatible change by its own, it should be merged as a new release and we can take the opportunity to revisit existing fields to see if we want to drop or re-organize any of them.

WeiZhang555:
@bergwolf I agree with every point :smile: 
So far we haven't got a conclusion about our compabitility policy, https://github.com/kata-containers/documentation/pull/196 can give us some basic reference but it's not merged yet.

WeiZhang555:
>How do you see the project using the version field? I wonder if most of the presence/unexpected fields can be handled by marshalling of the json itself?

Regarding your last question in comment:

It's for some breaking change. e.g.

v1:
```BlockIndex int```

Then we find that int isn't right , we actually need is a map, then

v2:
```BlockIndex int  // legacy
BlockIndexMap map[int]bool```

And new handling code should be :

```if v1 {
  // this container is created by old runtime, go to legacy handling logic
  // handle BlockIndex
} else if v2 {
  // moder runtime, handling modern fields
  // handle BlockIndexMap
}```

**Comment by WeiZhang555** [comment]
@sboeuf 

>  We have two choices, either the runtime v2 knows how to fallback and use BlockIndex, or it graciously fails saying that BlockIndexMap is empty.

First one is better. The example I gave is not very good, but real world can be very complicated, in my example, we can detect different versions from `BlockIndex` or `BlockIndexMap` without help of `PersistVersion`, but I would prefer we keep it in case we need a very clear version check.

Another example, one day we want to add a restriction:

"containers in same POD must be created&started by same runtime version"

To achieve this, a `PersistVersion` is very necessary. Without this, we cannot find a reliable way to detect  "different" runtime version.

> For instance, from the structure Bridge, all we need is to store the ID. This means that we should not define the entire structure Bridge as part of this package. Instead, we should rework the existing codebase to split Bridge:

1. In this example, is `ID` the only required field? `DeviceAddr` saves device pci address, in future, if we want to support flexible hot plug/unplug, we need to know free pci slot, so I think maybe `DeviceAddr` need to persist too. Same consideration to other fields, I think we need to take care of them carefully.

2. ` This would make virtcontainers a consumer of this persistapi package too ` This isn't necessary. `virtcontainers` is internal implementation, it can do anything with any structure, what we required is a stable fundamental persist data, and virtcontainer implementation change shouldn't influence persist data format. So I would prefer a converter, as I suggested to virtcontainers <-> agent GRPC structure translation.

**Comment by WeiZhang555** [comment]
#883  is a demo to show how to make use of this package laterly.

**Comment by sboeuf** [comment]
@WeiZhang555 

>/var/run/kata-containers can be re-organized or merged with /var/run/vc

I don't think this should be done as we store in `/var/run/kata-containers` what is specific to the `kata-runtime` cli, whilst the `/var/run/vc` path is used by the Kata API.

>To achieve this, a PersistVersion is very necessary. Without this, we cannot find a reliable way to detect "different" runtime version.

I am fine with that, the example you mentioned sounds reasonable.

>In this example, is ID the only required field? DeviceAddr saves device pci address, in future, if we want to support flexible hot plug/unplug, we need to know free pci slot, so I think maybe DeviceAddr need to persist too. Same consideration to other fields, I think we need to take care of them carefully.

No, no I was simply using this as an example, but if we think DeviceAddr needs to be stored, then it should be part of the `BridgeState` structure too.

>This would make virtcontainers a consumer of this persistapi package too This isn't required but it's OK. virtcontainers is internal implementation, it can do anything with any structure, what we required is a stable fundamental persist data, and virtcontainer implementation change shouldn't influence persist data format.

Yes, virtcontainers would rely on this package (persistapi), and the structures such as `Bridge`, `Container`,... would still be defined by virtcontainers. But each of these structures would include (or embed?) the corresponding `****State` structure defined by `persistapi` package.

So, it'd be nice to identify only what's strictly needed to be stored, so that `persistapi` package would define only `BridgeState`, `ContainerState`, `SandboxState`, and so on.

**Comment by amshinde** [comment]
@WeiZhang555 Sorry for the delay, was away for a couple of days.
In case of endpoints, if we want to go with minimal data that needs to stored, I think the following fields should be sufficient:

id - id used to pass the netdev option to qemu
index - to deduce the tap Name(if tap/macvtap interface is used) and the bridge Name if bridging is used.
ifName - the interface name 

With this, we would need translation logic to go from this data to the actual endpoint type, since the endpoint structures are themselves composed of structures. (I think some of the endpoint structures needs to be revisited and can be simplified further.)

**Comment by caoruidong** [comment]
> @amshinde
> 
> > id - id used to pass the netdev option to qemu
> 
> Nit: we should be careful not to mention QEMU specific things as part of the description of this persist API.

How about we make a rule to generate id used by qemu?

**Comment by WeiZhang555** [comment]
I think `id` should be also persisted since we'll need it later for removing from qemu.

File: virtcontainers/persist/api/device.go [hunk 1]
```diff
@@ -0,0 +1,103 @@
...
+	// Type of device: c, b, u or p
+	// c , u - character(unbuffered)
+	// p - FIFO
+	// b - block(buffered) special file
+	// More info in mknod(1).
+	DevType string
+
+	// Major, minor numbers for device.
+	Major int64
+	Minor int64
```
bergwolf: With `HostPath`, do we still need `DevType`/`Major/Minor`?

WeiZhang555: I think Major/Minor is more important since this is compliant with OCI spec: https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#devices

Maybe we should remove the `HostPath` ?

bergwolf: I think it depends on how we plan on using the information. Right now we use it to reconstruct the device slice in memory. And we are not using `HostPath` directly, -- `createDevice()` looks for host path with `Major/Minor` pair instead. So yes, I agree we should remove `HostPath`.

sboeuf:
@WeiZhang555 
>I think id should be also persisted since we'll need it later for removing from qemu.

Yes I agree. My point was to avoid mentioning QEMU from the networking structures since it's supposed to be independent from the hypervisor.

virtcontainers/persist/api/device.go
WeiZhang555: Addressed

WeiZhang555:
ping @kata-containers/runtime

jodh-intel:
Hi @WeiZhang555 - Thanks for raising! I think it would be interesting if you took this RFC a little further so we can see how this API might work. For example, you've created ` virtcontainers/persist/api/`, but this would need to link up to `virtcontainers/resource_storage.go` somehow I think?

As noted on https://github.com/kata-containers/kata-containers/issues/25, I think it is best not to rely on the version number and have the code which reads state off the disk for an X-1 version of Kata "fill in the gaps" to set defaults where it is sensible to do so (if it can't, that would be an error of course). One reason for doing this being that unless we are really careful, we may forget to increase `CurPersistVersion` when we change the on-disk API files. But that doesn't matter if the file parsing code can handle missing data itself.

jodh-intel:
Also, as we start to restructure the codebase, what impact will this have on the persistence API design? See for example https://github.com/kata-containers/runtime/pull/1096.

jodh-intel:
@WeiZhang555 - ah! I've just found #874 ;)

jodh-intel:
Err, I mean I just found #883 ;)

raravena80:
@WeiZhang555 ping. Any updates?

WeiZhang555:
Everying is moved to #883 with implementation codes. 
So closing this, thanks everyone for review, please move to #883 for more detailed info :-)

PR #887, opened 2018-11-06
virtcontainers: make sandbox manage VM resources

jcvenegas:
This PR moves the logic to manage VM resources from `container` to `sandbox`. This will allow do global calculations of all the containers running in a sandbox.

The resources that now manage the sandbox are:
- cpus
- memory

The container still update its resources in this case cgroups.


virtcontainers/sandbox.go
```diff
@@ -1404,7 +1410,30 @@ func (s *Sandbox) UpdateContainer(containerID string, resources specs.LinuxResou
 		return err
 	}
 
-	return c.update(resources)
+	newResources := ContainerResources{
+		VCPUs: uint32(utils.ConstraintsToVCPUs(*resources.CPU.Quota, *resources.CPU.Period)),
```
egernst: You are effectively mapping to vCPU at the container level, here, which I do not think we want.

We should calculate in terms of CPU shares/quota at the sandbox level, and only after an adjustment is made should we translate to vCPU.  That way two containers asking for 0.2 CPU each will result in 1vCPU for, not 2.

virtcontainers/qemu.go
```diff
@@ -1341,3 +1431,45 @@ func (q *qemu) getThreadIDs() (*threadIDs, error) {
...
+		return mem, nil
+	}
+
+	// TODO: hot add memory aligned to memory section should be more properly. See https://github.com/kata-containers/runtime/pull/624#issuecomment-419656853
+	return uint32(math.Ceil(float64(mem)/float64(memorySectionSizeMB))) * memorySectionSizeMB, nil
+}
+
+func (q *qemu) resizeVCPUs(reqVCPUs uint32) (currentVCPUs uint32, newVCPUs uint32, err error) {
+
+	currentVCPUs = q.config.NumVCPUs + uint32(len(q.state.HotpluggedVCPUs))
```
egernst: Is NumVCPUs the default, and hot plugged is what we've added?  Should the default be included in the CPU calculation here, since, AFAIU, it shouldn't be constrained/part of the workload?

virtcontainers/qemu.go
```diff
@@ -1341,3 +1431,45 @@ func (q *qemu) getThreadIDs() (*threadIDs, error) {
...
+func calcHotplugMemMiBSize(mem uint32, memorySectionSizeMB uint32) (uint32, error) {
+	if memorySectionSizeMB == 0 {
+		return mem, nil
+	}
+
+	// TODO: hot add memory aligned to memory section should be more properly. See https://github.com/kata-containers/runtime/pull/624#issuecomment-419656853
+	return uint32(math.Ceil(float64(mem)/float64(memorySectionSizeMB))) * memorySectionSizeMB, nil
+}
+
+func (q *qemu) resizeVCPUs(reqVCPUs uint32) (currentVCPUs uint32, newVCPUs uint32, err error) {
```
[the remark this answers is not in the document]
egernst: Though, I wonder if this should be done really in sandbox.go instead of qemu.go?

virtcontainers/container.go
```diff
@@ -646,15 +646,7 @@ func (c *Container) checkBlockDeviceSupport() bool {
...
-func createContainer(sandbox *Sandbox, contConfig ContainerConfig) (c *Container, err error) {
-	if sandbox == nil {
-		return nil, errNeedSandbox
-	}
-
-	c, err = newContainer(sandbox, contConfig)
-	if err != nil {
-		return
-	}
+func (c *Container) createContainer() (err error) {
```
sboeuf: The prototype should be:
```go
func (c *Container) create() (err error)
```
No need to repeat `Container` here.

virtcontainers/sandbox.go
```diff
@@ -1283,15 +1284,19 @@ func (s *Sandbox) CreateContainer(contConfig ContainerConfig) (VCContainer, erro
...
+	// Sandbox is reponsable to update VM resources needed by Containers
+	s.updateResources()
 	if err != nil {
 		return nil, err
 	}
 
-	// Update sandbox config.
-	s.config.Containers = append(s.config.Containers, contConfig)
-	err = s.storage.storeSandboxResource(s.id, configFileType, *(s.config))
+	err = c.createContainer()
```
sboeuf: I feel like we should call this before we update the VM resources.

virtcontainers/sandbox.go
```diff
@@ -1283,15 +1284,19 @@ func (s *Sandbox) CreateContainer(contConfig ContainerConfig) (VCContainer, erro
...
-	err = c.storeContainer()
+	// Sandbox is reponsable to update VM resources needed by Containers
+	s.updateResources()
 	if err != nil {
 		return nil, err
 	}
 
-	// Update sandbox config.
-	s.config.Containers = append(s.config.Containers, contConfig)
-	err = s.storage.storeSandboxResource(s.id, configFileType, *(s.config))
```
sboeuf: Why did you remove those two lines? We still want to store the resources from a sandbox perspective, right?

virtcontainers/container.go
```diff
@@ -1185,191 +1152,3 @@ func (c *Container) detachDevices() error {
 	}
 	return nil
 }
-
-func (c *Container) addResources() error {
```
sboeuf: Please squash this commit with the next one. We want in the same commit the part where you remove from container to apply onto sandbox.

virtcontainers/sandbox.go
jcvenegas: If we call it before the update VM resources, the resources wont exist at container creation, for some scenarios we should ensure they exist before create the container. For example, if the container cgroup has a cpuset information.
```diff
$ git diff
diff --git a/virtcontainers/sandbox.go b/virtcontainers/sandbox.go
index 85cbdcc..280ff58 100644
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -1284,13 +1284,13 @@ func (s *Sandbox) CreateContainer(contConfig ContainerConfig) (VCContainer, erro                                               
                return nil, err
        }

-       // Sandbox is reponsable to update VM resources needed by Containers                                                                           
-       s.updateResources()
+       err = c.create()
        if err != nil {
                return nil, err
        }

-       err = c.create()
+       // Sandbox is reponsable to update VM resources needed by Containers                                                                           
+       s.updateResources()
        if err != nil {
                return nil, err
        }
```
Create a container with cpuset data, the agent will try to apply cgroups but vcpus resources are not created yet.
```bash
$  docker run -ti --runtime kata-runtime --rm --name kata-1 --cpus --cpuset-cpus="0-2" stress-ng  nproc                 
invalid argument "--cpuset-cpus=0-2" for "--cpus" flag: failed to parse --cpuset-cpus=0-2 as a rational number                                         
See 'docker run --help'.
$  docker run -ti --runtime kata-runtime --rm --name kata-1 --cpus 4 --cpuset-cpus="0-2" stress-ng  nproc               
docker: Error response from daemon: OCI runtime create failed: rpc error: code = Internal desc = Could not run process: container_linux.go:348: starting container process caused "process_linux.go:279: applying cgroup configuration for process caused \"failed to write 0-2 to cpuset.cpus: write /sys/fs/cgroup/cpuset/docker/051d8204fbb1d4585fff45b70fb8de303745070f39e9c267725abcd5a1dd2a71/cpuset.cpus: invalid argument\"": unknown.                         
```
With the current code this work

```bash
$ git reset --hard
$ make & make install
=$  docker run -ti --runtime kata-runtime --rm --name kata-1 --cpus 4 --cpuset-cpus="0-2" stress-ng  nproc               
3
```

virtcontainers/hypervisor.go
```diff
@@ -574,6 +574,8 @@ type hypervisor interface {
 	addDevice(devInfo interface{}, devType deviceType) error
 	hotplugAddDevice(devInfo interface{}, devType deviceType) (interface{}, error)
 	hotplugRemoveDevice(devInfo interface{}, devType deviceType) (interface{}, error)
+	resizeMemory(memMB uint32, memoryBlockSizeMB uint32) (uint32, error)
+	resizeVCPUs(vcpus uint32) (uint32, uint32, error)
```
bergwolf: These two overlap with `hotplugAddDevice` and `hotplugRemoveDevice`. Any reason those two cannot be used? I'm asking because if we keep the helper functions in sandbox.go, they can be reused by different hypervisor implementations.

virtcontainers/sandbox.go
bergwolf: They are moved to api.go but @jcvenegas please keep them in sandbox.go so that api.go doesn't need to call any storage related functions (as of now).

virtcontainers/hypervisor.go
egernst: agreed that these should really be at sandbox level

jcvenegas: @bergwolf  the all idea is for the VM memory managment. For #793 will introduce manage the memory in with two methods balloon and hotplug. For longer term related with qemu implementation there is the option to use `virtio-mem` (`qmp` interface has a command called `memory-device-resize`).

 So when balloon is added (this PR is a depedency  of it), we think we dont want expose the balloon logic in the hypervisor interface or  hide it as part of hotplug/unplug memory calls. So the resizeMemory is looking to ask the hypervisor implementation manage the memory  as best as it could.  
Same for any other hypervisors that may have a variant of this methods (static memory, balloon, hotplug or something similar to virtio-mem), when we add them we will decide what is the best method for it but keep the hypervisor interface for vc simple. 

I think for resizeVCPUs() we could live without it but, wanted to give a similar method to the interface.  On vcpu managment only option today is hotplug.

jcvenegas: @bergwolf @sboeuf @egernst  is this point OK for you?

sboeuf: @jcvenegas Yes I agree with you as this is something we did talk about offline.

@bergwolf @egernst the problem with `hotplug` methods is that they assume we will hotplug some CPUs or some memory. But it's really up to the hypervisor implementation to decide what he's gonna do when the sandbox require a specific amount of memory or CPUs.
As mentioned by @jcvenegas, this will allow for a clean abstraction, as the only thing that matters for the sandbox is to make sure it'll request what it needs. The implementation details of using hotplug or balloning, or anything else to perform what is asked, is simply left to the hypervisor implementation.

virtcontainers/qemu.go
```diff
@@ -1074,33 +1075,61 @@ func (q *qemu) hotplugMemory(memDev *memoryDevice, op operation) (int, error) {
...
 		return 0, err
 	}
 
-	maxMem, err := q.hostMemMB()
-	if err != nil {
-		return 0, err
+	currentMemory := int(q.config.MemorySize) + q.state.HotpluggedMemory
+
+	if memDev.sizeMB == 0 {
+		q.Logger().WithField("hotplug", "memory").Debug("hotplug is not required")
```
devimc: please create a new log with `hotplug` field

```go
memLog := q.Logger().WithField("hotplug", "memory")
```

virtcontainers/qemu.go
```diff
@@ -1074,33 +1075,61 @@ func (q *qemu) hotplugMemory(memDev *memoryDevice, op operation) (int, error) {
...
+		return 0, nil
 	}
 
-	// calculate current memory
-	currentMemory := int(q.config.MemorySize) + q.state.HotpluggedMemory
+	switch op {
+	case removeDevice:
+		q.Logger().WithFields(
+			logrus.Fields{
```
devimc:
```go
memLog.WithField("operation", "remove").Debugf("Requested to remove memory: %d MB", memDev.sizeMB)
```

virtcontainers/qemu.go
```diff
@@ -1074,33 +1075,61 @@ func (q *qemu) hotplugMemory(memDev *memoryDevice, op operation) (int, error) {
...
+				"hotplug":   "memory",
+				"operation": "remove",
+			},
+		).Debugf("Requested to remove memory: %d MB", memDev.sizeMB)
+		// Dont fail for now, until we fix it.
+		// We return that we only unplugged 0
+		q.Logger().Warn("hot-remove VM memory not supported")
+		return 0, nil
+	case addDevice:
+		q.Logger().WithFields(
```
devimc:
```go
memLog.WithField("operation", "add").Debugf("Requested to remove memory: %d MB", memDev.sizeMB)
```

virtcontainers/qemu.go
```diff
@@ -1074,33 +1075,61 @@ func (q *qemu) hotplugMemory(memDev *memoryDevice, op operation) (int, error) {
...
-	currentMemory := int(q.config.MemorySize) + q.state.HotpluggedMemory
+	switch op {
+	case removeDevice:
+		q.Logger().WithFields(
+			logrus.Fields{
+				"hotplug":   "memory",
+				"operation": "remove",
+			},
+		).Debugf("Requested to remove memory: %d MB", memDev.sizeMB)
+		// Dont fail for now, until we fix it.
```
devimc: is there an issue for this ?

virtcontainers/qemu.go
```diff
@@ -1074,33 +1075,61 @@ func (q *qemu) hotplugMemory(memDev *memoryDevice, op operation) (int, error) {
...
+			return 0, err
+		}
 
-	// Don't exceed the maximum amount of memory
-	if currentMemory+memDev.sizeMB > int(maxMem) {
-		return 0, fmt.Errorf("Unable to hotplug %d MiB memory, the SB has %d MiB and the maximum amount is %d MiB",
-			memDev.sizeMB, currentMemory, q.config.MemorySize)
+		// Don't exceed the maximum amount of memory
+		if currentMemory+memDev.sizeMB > int(maxMem) {
+			// Fixme: return a typed error
```
devimc: I guess you'll fix it in this pr

virtcontainers/sandbox.go
```diff
@@ -1798,3 +1828,63 @@ func (s *Sandbox) AddDevice(info config.DeviceInfo) (api.Device, error) {
...
+		vcpusAdded := newCPUs - oldCPUs
+		if err := s.agent.onlineCPUMem(vcpusAdded, true); err != nil {
+			return err
+		}
+	}
+	s.Logger().Debugf("Sandbox CPUs: %d", newCPUs)
+
+	// Update Memory
+	s.Logger().WithField("memory-sandbox-size-byte", sandboxMemoryByte).Debugf("Request to hypervisor to update memory")
+	newMemory, err := s.hypervisor.resizeMemory(uint32(sandboxMemoryByte>>utils.MibToBytesShift), s.state.GuestMemoryBlockSizeMB)
```
clarecch: As we have discuss in https://github.com/kata-containers/runtime/pull/793#discussion_r228384791 
By simply do 20 right shift to sandboxMemoryByte, we will lose memory less than 1M.
For instance, if we `docker run 128M` we will get a vm of 2G+128M memory. 
Then we `docker update 128M+4096`, we will do nothing with `MibToBytesShift` right shift here. But the memory limit cgroup of container will be 128M+4096, so it will exceed 128MB memory. 
WDYT.

clarecch: Oh, it's reasonable. Thanks for your answer.

virtcontainers/hypervisor.go
bergwolf: Fair enough. I agree we can have a resource descriptive interface. Thanks @sboeuf @jcvenegas .

virtcontainers/qemu.go
```diff
@@ -1237,6 +1261,72 @@ func (q *qemu) disconnect() {
...
+		addMemMB := currentMemory - reqMemMB
+		memHotunplugMB, err := calcHotplugMemMiBSize(addMemMB, memoryBlockSizeMB)
+		if err != nil {
+			return currentMemory, err
+		}
+
+		addMemDevice := &memoryDevice{
+			sizeMB: int(memHotunplugMB),
+		}
+		data, err := q.hotplugRemoveDevice(addMemDevice, memoryDev)
```
bergwolf: Since we don't support hotunplug memory right now, I would prefer we do not subtract the memory from total memory and we are correct currentMemory at least. A FIXME might be better here if we do not want to fail `hotplugRemoveDevice` as of now. I understand you kept the old code as is, so maybe a follow github issue to fix the behavior might be better. Otherwise if someone does `update` frequently can end up with a vm with much more memory than actually being used.

bergwolf:
LGTM. I think the PR is ready to be merged. Thanks @jcvenegas ! We still see metrics failure. @grahamwhaley any suggestions? Is it OK to just ignore metrics for now?

virtcontainers/qemu.go
sboeuf: There is nothing to worry about here since the code path from `hotplugRemoveDevice()` leads to `hotplugDevice()` which leads to `hotplugMemory()`. And here is the code from `hotplugMemory`:
```go
        if memDev.sizeMB < 0 {
		return 0, fmt.Errorf("cannot hotplug negative size (%d) memory", memDev.sizeMB)
	}

	// We do not support memory hot unplug.
	if op == removeDevice {
		// Dont fail for now, until we fix it.
		// We return that we only unplugged 0
		q.Logger().Warn("hot-remove VM memory not supported")
		return 0, nil
	}

	err := q.qmpSetup()
	if err != nil {
		return 0, err
	}
```

So the data returned will be `0` in case of removal, hence the amount of memory returned is gonna be the `currentMemory`.
Am I missing something?

virtcontainers/sandbox.go
sboeuf: @jcvenegas makes sense to do that in separate PR for govmm, thanks!

virtcontainers/container.go
jcvenegas: fixed

virtcontainers/qemu.go
jcvenegas: fixed

virtcontainers/qemu.go
jcvenegas: I am not fixing in this PR most of our errors are not typed today, would be good to have a work for that.

virtcontainers/qemu.go
jcvenegas: I added an extra validation to verify this is always 0, I hope this make clear that we expect 0 at this moment.

jcvenegas:
initrd job failed, not sure if is a known issue? 
```
Update CPU constraints Update CPU set 
  cpuset should be equal to 0-3
  /tmp/jenkins/workspace/kata-containers-runtime-ubuntu-16-04-PR-initrd/go/src/github.com/kata-containers/tests/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:46
Running command '/usr/bin/docker [docker run --cidfile /tmp/cid988537153/2jhAz8kKRWnuDm0AWqACaGL07wAJw0 --runtime kata-runtime --rm --cpus=4 --name 2jhAz8kKRWnuDm0AWqACaGL07wAJw0 -dt debian bash]'
Killing process timeout reached '60' seconds
Running command '/usr/bin/docker [docker rm -f 2jhAz8kKRWnuDm0AWqACaGL07wAJw0]'
Killing process timeout reached '60' seconds

• Failure [120.006 seconds]
Update CPU constraints
/tmp/jenkins/workspace/kata-containers-runtime-ubuntu-16-04-PR-initrd/go/src/github.com/kata-containers/tests/integration/docker/cpu_test.go:319
  Update CPU set
  /tmp/jenkins/workspace/kata-containers-runtime-ubuntu-16-04-PR-initrd/go/src/github.com/kata-containers/tests/vendor/github.com/onsi/ginkgo/extensions/table/table.go:92
    cpuset should be equal to 0-3 [It]
    /tmp/jenkins/workspace/kata-containers-runtime-ubuntu-16-04-PR-initrd/go/src/github.com/kata-containers/tests/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:46

    Expected
        <int>: -1
    to be zero-valued
```

sboeuf:
@jcvenegas did it fail because of your changes from this morning? Because the CI was all green before your last modifications, right?

jcvenegas:
@sboeuf it pass the last time, let me add some logging to see if this is caused by the PR.
initrd shows 6 errors:
```

[Fail] Update CPU constraints Update CPU set [It] cpuset should be equal to 0-3 
/tmp/jenkins/workspace/kata-containers-runtime-ubuntu-16-04-PR-initrd/go/src/github.com/kata-containers/tests/integration/docker/cpu_test.go:376

[Fail] Update CPU constraints Update CPU set [It] cpuset should be equal to 0,-2,3 
/tmp/jenkins/workspace/kata-containers-runtime-ubuntu-16-04-PR-initrd/go/src/github.com/kata-containers/tests/integration/docker/cpu_test.go:376

[Fail] docker volume remove bind-mount source before container exits [It] should exit cleanly without leaking process 
/tmp/jenkins/workspace/kata-containers-runtime-ubuntu-16-04-PR-initrd/go/src/github.com/kata-containers/tests/integration/docker/volume_test.go:175

[Fail] load [AfterEach] load with docker load a container should load image 
/tmp/jenkins/workspace/kata-containers-runtime-ubuntu-16-04-PR-initrd/go/src/github.com/kata-containers/tests/integration/docker/load_test.go:26

[Fail] docker privileges [AfterEach] check no-new-privileges flag should display the correct uid 
/tmp/jenkins/workspace/kata-containers-runtime-ubuntu-16-04-PR-initrd/go/src/github.com/kata-containers/tests/integration/docker/privileges_test.go:34

[Fail] build [AfterEach] build with docker docker build env vars should display env vars 
/tmp/jenkins/workspace/kata-containers-runtime-ubuntu-16-04-PR-initrd/go/src/github.com/kata-containers/tests/integration/docker/build_test.go:31

```

and ubuntu failed with a cri-o test 
```
not ok 32 ctr with non-root user has no effective capabilities
# (in test file ctr.bats, line 1091)
# time="2018-11-13T18:41:35Z" level=error msg="File descriptor 3 (pipe:[430845]) leaked on pvdisplay invocation. Parent PID 18313: /tmp/jenkins/workspace/kata-con
# File descriptor 6 (/dev/mapper/control) leaked on pvdisplay invocation. Parent PID 18313: /tmp/jenkins/workspace/kata-con
#   Failed to find physical volume "/dev/vdb".
# " error="exit status 5" 
```

virtcontainers/kata_agent.go
```diff
@@ -1252,8 +1252,15 @@ func (k *kataAgent) resumeContainer(sandbox *Sandbox, c Container) error {
...
+	wait := true
+	k.Logger().WithFields(
+		logrus.Fields{
+			"cpus":    cpus,
+			"cpuOnly": cpuOnly,
+			"Wait":    wait,
+		}).Debugf("request to  agent to online cpus and memory")
 	req := &grpc.OnlineCPUMemRequest{
-		Wait:    false,
+		Wait:    wait,
```
devimc: boot time could be slower with this change, have you measured `docker run --cpus=8 -ti busybox true` with `Wait: true` and `Wait: false`?

grahamwhaley: I'll throw  in the obligatory reference to https://github.com/kata-containers/tests/tree/master/metrics/report where you can use the grabdata (and tell it to only measure boot time if you want) to do a before/after comparison.... ;-)

jcvenegas: @devimc I suspect this could happend while we increase the amount of vcpus hotplugged to the VM.

So I was wondering if cpuset should be deferred as well. Let me try with server with more cpus

sboeuf: Any results regarding this @jcvenegas?
Also, what's the rationale behind this change?

devimc: @jcvenegas the problem with `Wait: true` is that the agent waits for the CPUs and the runtime has to wait for the agent, IMO resources should be connected in background, in the case of CPUs I haven't seen any problem with this behaviour, but I guess you detected something with the memory.

sboeuf: @jcvenegas we should not keep some racy code. So I would suggest that you use `wait = true` but run this in a goroutine that you can synchronize with at some point. And by that, I mean that we should allow parallel stuff to happen while the CPU gets onlined, but have to wait before to update cpuset.
Please open a separate issue and PR for that.

In the meantime, and because this bug is not introduced by this PR, I don't see why we could not merge it.

grahamwhaley:
Just for ref, metrics CI is running again, but I need to track down some failures. pls ignore metrics CI right now - but, if you have any concerns over possible perf impact, please use https://github.com/kata-containers/tests/tree/master/metrics/report to do a hand-compare of before/after results..

virtcontainers/sandbox.go
```diff
@@ -1798,3 +1828,63 @@ func (s *Sandbox) AddDevice(info config.DeviceInfo) (api.Device, error) {
...
+		}
+		if cpu := c.Resources.CPU; cpu != nil {
+			if cpu.Period != nil && cpu.Quota != nil {
+				*sumResources.CPU.Period += *cpu.Period
+				*sumResources.CPU.Quota += *cpu.Quota
+			}
+		}
+	}
+
+	sandboxVCPUs := uint32(utils.ConstraintsToVCPUs(*sumResources.CPU.Quota, *sumResources.CPU.Period))
```
woshijpf: Use `utils.ConstraintsToVCPUs()` to calculate VCPUS is wrong. I have made the following test:
```bash
# 1. start a sandbox with default number of hypervisor cpu
# docker run -tid --runtime kata-runtime --annotation io.kubernetes.docker.type=podsandbox --network=none euleros bash
8589b666d0ff3d3211b3b3ebd2f8387e881d6d1068ca6518b1be6917e55a7208

# exec `lscpu` command in the container to view CPU numbers in the sandbox
 # docker exec -it 8589b666d0ff lscpu
Architecture:          x86_64 
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                1
On-line CPU(s) list:   0
Thread(s) per core:    1
Core(s) per socket:    1
Socket(s):             1
Vendor ID:             GenuineIntel

# 2. start a new container A with `--cpus 1.1` and join into the previous sandbox 
# docker run -tid --runtime kata-runtime --annotation io.kubernetes.docker.type=container --annotation io.kubernetes.sandbox.id=8589b666d0ff3d3211b3b3ebd2f8387e881d6d1068ca6518b1be6917e55a7208 --cpus 1.1 --network=none euleros bash
26984866a88783450b2d94cb0688db57fc939f4fd4d52416d68d30fad84ffd25

#  exec `lscpu` command in the container to view CPU numbers in the sandbox
 # docker exec -it 8589b666d0ff lscpu
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                3
On-line CPU(s) list:   0-2

# 3. start a new container B with `--cpus 0.8` and join into the previous sandbox
docker run -tid --runtime kata-runtime --annotation io.kubernetes.docker.type=container --annotation io.kubernetes.sandbox.id=8589b666d0ff3d3211b3b3ebd2f8387e881d6d1068ca6518b1be6917e55a7208 --cpus 0.8 --network=none euleros bash
fdd5c1a10429ee77b8984911061ff62c6d479dfc131da30ea0193e9a614c0166

#  exec `lscpu` command in the container to view CPU numbers in the sandbox
# docker exec -it fdd5 lscpu
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                2
On-line CPU(s) list:   0,1
Thread(s) per core:    1
Core(s) per socket:    1
Socket(s):             2
Vendor ID:             GenuineIntel
```
After  I create container B in step 3, I get the CPU numbers is **2**, however expected value is **3**，I think we should propose a new method to calculate VCPU numbers.

jcvenegas: Hey thanks for the the comment nice catch! We should add this as a test and agree need to fix ConstraintsToVCPUs as needed.

Nice way to test pods in docker  I am am trying the annotation flag  but not worked for me what docker version are you using?

jcvenegas:
CI should be fixed by https://github.com/kata-containers/tests/pull/1007

caoruidong:
Great job! Can this be merged?

jcvenegas:
@caoruidong  @sboeuf  yes, ready to merge

clarecch:
Nice @jcvenegas

virtcontainers/sandbox.go
jshachm: @jcvenegas 
Hah this is a internal version we are using。
You can patch docker with https://github.com/moby/moby/pull/37289.

Issue #894, opened 2018-11-09
Hotplugged nic's MAC is not bound back


caoruidong:
# Description of problem

When using hotplug network, if we add a nic and then delete it. Its MAC isn't bound back.

# Expected result

Add and delete a nic several times without modifying nic's MAC in description.

# Actual result

Not able to add the nic again. Must modify nic's MAC in description.

---

I'm not sure know if this is a bug. So I raise a PR.

PR #895, opened 2018-11-09
network: support hotplug a nic several times

caoruidong:
Bind back the nic's MAC in HotDetach. So we don't need to modify
nic's MAC in description to hotplug it again.

Fixes: #894


sboeuf:
@caoruidong looks good, let's wait for @amshinde's input.

PR #900, opened 2018-11-10
golang.mk: Check and install yq before use it

### LINKED ISSUE #899: make: yq: Command not found
make
make: yq: Command not found
golang.mk:12: *** "ERROR: cannot determine minimum golang version".  Stop.

teawater:
golang.mk call yq to get golang_version_min but some environments do
not install it.
This patch check and install yq before use it to handle the issue.

Fixes #899

### DIFF
--- a/golang.mk
+++ b/golang.mk
@@ -6,7 +6,14 @@
 # Check that the system golang version is within the required version range
 # for this project.
 
-golang_version_min=$(shell yq r versions.yaml languages.golang.version)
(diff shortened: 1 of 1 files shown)


grahamwhaley:
/test
but, I think this will pass the tests as I suspect the build machines already have `yq` installed.
I guess you already tested this locally by hand @teawater :-)  When I added the check code I thought I didn't add this as calling install_yq might not work without some infra - so confirmation that this works as expected would be great :-)
If confirmed then L G T M

caoruidong:
lgtm. If environment is set up by kata repos like CI, it will have yq installed. This is usefully when manually set up machine.

PR #904, opened 2018-11-12
sandbox: Extend sandbox API

### LINKED ISSUE #903: sandbox: Extend sandbox API
In order to support use cases such as `kata-containerd-shim-v2`, where we have a long running process holding the `Sandbox` pointer, there is no reason to call into the stateless functions, as they would recreate a new `Sandbox` pointer and the corresponding ones for containers.

Instead, we should allow such case to use the sandbox API directly, avoiding to waste time in re-initialization and duplication.

sboeuf:
In order to support use cases such as `kata-containerd-shim-v2`, where we have a long running process holding the `Sandbox` pointer, there is no reason to call into the stateless functions, as they would recreate a new `Sandbox` pointer and the corresponding ones for containers.

Instead, this PR allows such case to use the sandbox API directly, avoiding to waste time in re-initialization and duplication.

Fixes #903

### DIFF
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -468,12 +468,7 @@ func StartContainer(ctx context.Context, sandboxID, containerID string) (VCConta
 	}
 	defer s.releaseStatelessSandbox()
 
-	c, err := s.StartContainer(containerID)
-	if err != nil {
-		return nil, err
-	}
-
-	return c, nil
+	return s.StartContainer(containerID)
 }
 
 // StopContainer is the virtcontainers container stopping entry point.
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -784,17 +784,11 @@ func togglePauseContainer(ctx context.Context, sandboxID, containerID string, pa
 	}
 	defer s.releaseStatelessSandbox()
 
-	// Fetch the container.
-	c, err := s.findContainer(containerID)
-	if err != nil {
-		return err
-	}
-
 	if pause {
-		return c.pause()
+		return s.PauseContainer(containerID)
 	}
 
-	return c.resume()
+	return s.ResumeContainer(containerID)
 }
 
 // PauseContainer is the virtcontainers container pause entry point.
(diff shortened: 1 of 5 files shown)


sboeuf:
@lifupan @bergwolf 
I'd like to get this current PR merged before https://github.com/kata-containers/runtime/pull/572, to be able to rely on it, as this would allow for a more efficient use of Kata API from `kata-containerd-shim-v2`.
WDYT?

grahamwhaley:
Hi @sboeuf - no, it should be up now - at least we think we'd fixed the last jenkins config niggle. But, that build has some very strange failure mode:

```
00:25:02 docker pull'd: ubuntu
00:25:02  run 1
00:25:23 docker: Error response from daemon: OCI runtime create failed: Failed to check if grpc server is working: rpc error: code = Unavailable desc = transport is closing: unknown.
00:25:23 No kernel last line
00:25:23 ERROR: No kernel last line
```

It could be this PR, it could be spurious, or it is possible the bare metal slave has somehow gotten corrupt :-(
let's try a retest here and see what we see - and then we can go look at the slave and the slave build history if this persists.

/test

sboeuf:
@grahamwhaley unfortunately same error after re-triggering the metrics CI.

sboeuf:
@grahamwhaley moreover, this PR does not modify the performance of the overall code, it simply exports a few functions. So it should not affect metrics CI.

bergwolf:
@sboeuf I agree we should add sandbox APIs to support sandbox operations whenever there is a valid sandbox pointer.
> as they would recreate a new Sandbox pointer and the corresponding ones for containers.

The claim may not be true because if `shim-v2` is setting `Stateful` to true in the sandbox config, `releaseStatelessSandbox` would do nothing and `api.go` can also reuse the previous created sandbox pointer as it is saved in the global sandbox list.

That said, the PR still makes a lot of sense to increase the coverage of the sandbox APIs. Thanks and of course it
LGTM!

lifupan:
Yeah, those changes will make kata's sandbox APIs much more rich and consistent. 
LGTM

lifupan:
Hi @sboeuf, I'm waiting for this PR to be merged and then I can update https://github.com/kata-containers/runtime/pull/572 accordingly.

sboeuf:
As mentioned by @grahamwhaley, this PR is not causing the metrics CI failures, let's move forward and merge it.

sboeuf:
@lifupan here you go, you can rebase #572 on top of this PR now that it's been merged.

Issue #908, opened 2018-11-15
Type of Statfs_t struct on s390x is uint32 not int64


alicefr:
In go [Type](https://github.com/golang/go/blob/7a6ccece877232fc58872c85fdf0718629b83d91/src/syscall/ztypes_linux_amd64.go#L120) of the struct Statfs_t is `int64` for `amd64`, `arm` and `ppc64le`. However, on s390 the [Type](https://github.com/golang/go/blob/7a6ccece877232fc58872c85fdf0718629b83d91/src/syscall/ztypes_linux_s390x.go#L118) is `uint32`. In [oci/cli.go](https://github.com/kata-containers/runtime/blob/0911331974fbb82b75d0ecbe21d4eac292cfd5cd/cli/oci.go#L126), you convert directly the cgroupFsType to `int64`. For this reason, on s390 the build fails:
```sh
$ make 
# github.com/kata-containers/runtime/cli
./oci.go:251:17: invalid operation: statFs.Type != int64(cgroupFsType) (mismatched types uint32 and int64)
Makefile:328: recipe for target '/root/go/src/github.com/kata-containers/runtime/kata-runtime' failed
make: *** [/root/go/src/github.com/kata-containers/runtime/kata-runtime] Error 2
```

PR #909, opened 2018-11-15
build: introduction of archConvertStatFs function

alicefr:
Fixes: #908

Type of StatFs is not always declared as int64 for all the architecture(e.g s390x).
The function archConvertStatFs could be reimplemented for other architecture
to correctly convert the StatFs.Type.


alicefr:
@sboeuf thanks! I changed the commit message :)

Issue #914, opened 2018-11-16
clear the Mac address for vfio NIC passthrough


hejianet:
# Description of problem

    When passes a vf to qemu, the guest kernel driver(eg. mlx5) might
    set the mac address. Thus without this patch, the updateInterface
    in kata agent side will be failed and vfio-pci device can't be found by guest

# Expected result

The sriov vf can be successfully passed to guest kernel and container can find the vf NIC by lspci

# Actual result

the vfio-pci can't be passed to guest by qemu command-line

---

hejianet:
The fix includes 2 parts:
1. runtime side: set the mac address to null
2. agent side check the link name if the mac address is null

egernst:
I think in best practices the NIC should be configured with VFs created and appropriate settings made (mac spoofing or setting MACs for each VF). If this is skipped you’ll run unto this issue. I don’t think we should clear the MAC, as we’d lose the desired value which may have been explicitly set, right?

hejianet:
@egernst Thanks for the comments
Even if you have MACs for vf from the host's point of view, but when passing the vf to the guest, some guest NIC drivers (e.g. mlx5 will set the randome MACs again). Please refer to mlx5e_set_netdev_dev_addr in en_main.c of kernel src code.

sboeuf:
IIUC we want to make sure the interface created inside the VM after the NIC has been passed through, will have the MAC address that was assigned to the device on the host, right?

If that's the case, we need to rely on b/d/f from the PCI device, so that we can find it inside the guest and apply the setup of the MAC address in this case.

@amshinde correct me if I'm wrong but I guess the logic inside the agent is really tied to the fact that we always expect a TAP interface that we setup with the right MAC address previously.

This use case is different and is all about making sure that the agent actually updates the MAC address in case the VFIO device is a NIC.

amshinde:
@hejianet I have added my comments on your PR in the agent. If the driver is changing the mac address (which seems like incorrect behaviour on the driver side), you can instead rely on the PCI address. Make sure that the runtime passed the PCI address in case of physical interfaces(don't think that is being done currently)

PR #915, opened 2018-11-16
vfio: clear the Mac address for SRIOV NIC passthrough

hejianet:
When passes a vf to qemu, the guest kernel driver(eg. mlx5) might
set the mac address. Thus without this patch, the updateInterface
in kata agent side will be failed and the vfio-pci can't be passed
to qemu command line

The fix includes 2 parts:
runtime side: set the mac address to null
agent side check the link name if the mac address is null

Fixes: #914
Signed-off-by: Jia He <justin.he@arm.com>


caoruidong:
lgtm

devimc:
cc @amshinde @mcastelino

sboeuf:
I have added do not merge label here because based on @amshinde comments and my comments here: https://github.com/kata-containers/agent/pull/414 we think removing the MAC address is not the right approach here.

Issue #921, opened 2018-11-19
cli: Fix console for big endian architecture


alicefr:
The function [ioctl](https://github.com/kata-containers/runtime/blob/c0551de09355ebb67f73ca4b6b6ff2643688c784/cli/console.go#L94) can lead to a big endian issue. The same construct was present in the `containerd/console` package and it has been fixed with this [PR](https://github.com/containerd/console/pull/20)

PR #922, opened 2018-11-19
cli: Fix console for big endian architecture

alicefr:
The function ioctl can lead to a big endian bug.
Issue already solved in containerd/console:
https://github.com/containerd/console/pull/20/commits/dbd69c59b81ec72258da35a335605e4b3ea17e6a

Fixes: #921

### DIFF
--- a/cli/console.go
+++ b/cli/console.go
@@ -91,27 +91,23 @@ func (c *Console) Close() error {
 	return nil
 }
 
-func ioctl(fd uintptr, flag, data uintptr) error {
-	if _, _, err := unix.Syscall(unix.SYS_IOCTL, fd, flag, data); err != 0 {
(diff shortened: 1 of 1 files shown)


jodh-intel:
Ouch!

lgtm

For reference (from https://golang.org/pkg/unsafe/#Pointer with added emphasis):

> (4) Conversion of a Pointer to a uintptr when calling syscall.Syscall.
>
> The Syscall functions in package syscall pass their uintptr arguments directly to the operating system, which then may, depending on the details of the call, reinterpret some of them as pointers. That is, the system call implementation is implicitly converting certain arguments back from uintptr to pointer.
> 
> **If a pointer argument must be converted to uintptr for use as an argument, that conversion must appear in the call expression itself**.

Issue #924, opened 2018-11-19
virtcontainers: Fix remove hotplug vfio devices bug


woshijpf:
# Description of problem
1. start a new  sandbox and start a new container A in it.
2. start a new container B in the sandbox as same as step 1 started, but with `--device <vfio-device-path>` to hotplug a vfio device(I use GPU card as vfio device) from host to sandbox, and then you can find a pci device in container B
3. container B exits
4. start a new container C in the sandbox as same as step 1 started, and hotplug the same vfio device same as step 2. However, container C fails to start, and you  will get the error return message.

# Analysis
After some work to debug runtime code, I find the cause of this problem is that `vfioDevs` struct member in `struct VFIODevice` begin with **lowercase**, which can't been written into sandbox device.json file in `c.sandbox.storeSandboxDevices()` function.

woshijpf:
I will raise a PR later, because I meet some problem while push code to github.

PR #925, opened 2018-11-20
virtcontainers: fix sandbox store struct VFIODevice bug

woshijpf:
Since struct VFIODevice needed to be stored into disk by storeSandboxDevices() function,
however struct VFIODevice has a field named "vfioDevs", which is named begin with lower-case,
so it can't be written into file by json.Marshal.And this bug will will cause hotplug vfio
device can not been removed correctly while container exits.

Fixes: #924

### DIFF
--- a/virtcontainers/device/drivers/vfio.go
+++ b/virtcontainers/device/drivers/vfio.go
@@ -32,7 +32,7 @@ const (
 // to be used by the Virtual Machine.
 type VFIODevice struct {
 	*GenericDevice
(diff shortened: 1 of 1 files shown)


caoruidong:
Nice catch! /test

jshachm:
LGTM

[Approved with PullApprove](https://pullapprove.com/kata-containers/runtimes/pull-request/925/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=kata-containers/runtimes)

Issue #939, opened 2018-11-24
govmmQemu.LaunchQemu() would get error of "waitid: no child processes" for shimv2


lifupan:
# Description of problem

containerd-shimv2 will have its own children reapers which would reap the
qemu's process before LaunchQemu's cmd.Run to reap it. For this case, the govmmQemu.LaunchQemu()  will get an error contains substring of "no child processes", thus it's needed to ignore this error and deal it as success.
# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #940, opened 2018-11-24
virtcontainers: fix the error of missing reap qemu process

lifupan:
When govmmQemu.LaunchQemu() launching a qemu process using
os/exec.Cmd.Run(), its child qemu process would be reaped
by other reapers, in this case Cmd.Wait() will get an error
of "waitid: no child processes", thus it's better to ignore
this error since it has launch qemu successfully.

Fixes: #939

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -552,6 +552,14 @@ func (q *qemu) startSandbox() error {
 	var strErr string
 	strErr, err = govmmQemu.LaunchQemu(q.qemuConfig, newQMPLogger())
 	if err != nil {
+		// Some programs such as containerd-shimv2 may have its own children reapers which has reaped the
+		// qemu's process before cmd.Wait in govmmQemu.LaunchQemu() to reap it. For this case, govmmQemu.LaunchQemu()
+		// will get an error contains substring of "no child processes", thus it's needed
+		// to ignore this error and deal it as success.
+		if strings.Contains(err.Error(), "no child processes") {
+			err = nil
+			return nil
+		}
 		return fmt.Errorf("%s", strErr)
 	}
 


jodh-intel:
@lifupan - I'd rather we didn't match on a string error message as that may change (for example i18n). 

@markdryan - any thoughts on this?

markdryan:
@jodh-intel What I don't understand about this PR is, who is reaping the processes?  My feeling is that if you're expecting someone else to reap the process you should be calling Command.Start and not Command.Run.  But if you were to do this, I don't understand how you'd know whether the process launched correctly and indeed when it had launched correctly and could be communicated with via QMP, without introducing a race.  Does clearcontainers used --daemonize.  If so, then it needs to call Wait and it needs to reap the qemu process as this is the mechanism that is used to avoid a race when communicating with the new VM over QMP.

@lifupan Can you clarify what is reaping the QEMU processes?  Does it really need to?

lifupan:
Hi @markdryan, the containerd-shimv2 for kata will launch a children process reaper which will reap all of the children process. This reaper is necessary since this shimv2 will run as a daemon and it should reap all of the children processes. Take qemu for example, when launching a qemu process, it will create several sub processes, but the govmmQemu.LaunchQemu() will only reap/wait the top child process, the other child processes  were also needs to be reaped by shimv2's reaper.

markdryan:
But if someone else reaps the top level qemu process how does startSandbox know whether the process successfully launched or not?  It can ignore the "no child processes" error, but when it receives this error, how can it know that process has really launched?  There may have been some real error that prevented it from launching that startSandbox would not be able to detect.

lifupan:
Hi, @markdryan , the error string of "no child processes" is specific to wait4() syscall return error number "10", which means as the string said "no child processes", if startSandbox really failed, the syscall will return other error numbers which would be translate to other error string instead of strings contains  "no child processes".  @jodh-intel, since the startSandbox will only return the error string instead of the errno, I think there is no other choice except test the strings.  For the map between errno and error string, it comes from go's syscall file https://golang.org/src/syscall/zerrors_linux_amd64.go for amd64 arch, I think the error string will not be changed very often.

jodh-intel:
>  I think the error string will not be changed very often.

I'm sure you are right, but one day it will change and this will break. For example, what if hypothetically golang v1.12 decides to introduce a way to make those error string display in the users native locale? That would break this code.

I'd much rather we found a solution that didn't involve string matching as this change -- although I'm sure it works today -- is fragile imho.

markdryan:
@lifupan What I'm concerned about is a potential race.  Perhaps I misunderstand how all this works, but if something else is reaping the child processes presumably this something else, the containerd-shimv2 process, is calling wait.  Is this correct?  Is it then not the case that we have we have two threads in the same process calling some version of wait?  If this is true, isn't there a race condition?  Here's the specific scenario I'm worried about.

The go-routine in which startSandbox runs launches the QEMU process and calls wait.  Wait is also running in some other containerd-shimv2 go-routine.  The QEMU launch fails for some reason.  The exit code is reaped by the containerd-shimv2 go-routine.  The wait call in the startSandbox go-routine receives a "no child processes" which we then ignore.  StartSandbox thinks the QEMU process has been launched correctly, when in fact the launch has failed.  Can this scenario happen?

lifupan:
> @lifupan What I'm concerned about is a potential race. Perhaps I misunderstand how all this works, but if something else is reaping the child processes presumably this something else, the containerd-shimv2 process, is calling wait. Is this correct? Is it then not the case that we have we have two threads in the same process calling some version of wait? If this is true, 

Yes, this is true.
But the following scenario will not happen.

> 
> The go-routine in which startSandbox runs launches the QEMU process and calls wait. Wait is also running in some other containerd-shimv2 go-routine. The QEMU launch fails for some reason. The exit code is reaped by the containerd-shimv2 go-routine. The wait call in the startSandbox go-routine receives a "no child processes" which we then ignore. StartSandbox thinks the QEMU process has been launched correctly, when in fact the launch has failed. Can this scenario happen?

Actually the launch qemu of  startSandbox() is composed by two steps: the start() and wait().
Once the start() executed successfully, I think the qemu process has been launched successfully. If 
start() failed, it will return an error directly and the wait() will be skipped. In this case, it will never return an error contains the string of "no child process".

sboeuf:
@lifupan @markdryan what about having a function `LaunchQemuNoWait()` that is not using `Run()`, but instead would use `Start()`. This would return a `cmd` in order to let the caller`Wait()`. This way, we know we can ignore only the error related to reaping the process.
Should we also identify that we're in a long running process (`containerd-shim-v2`) since it would help not ignoring the error for other use cases.

markdryan:
> @lifupan @markdryan what about having a function LaunchQemuNoWait() that is not using Run(), but instead would use Start(). This would return a cmd in order to let the callerWait(). This way, we know we can ignore only the error related to reaping the process

I suggested something along this lines in the original govmm PR (https://github.com/intel/govmm/pull/66), but now I don't think this will solve the issue.  I could very well by misunderstanding how this process reaping is going to work but it seems to me that there may be an architectural issue that needs to be solved here.  If something else is reaping qemu processes, how can startSandbox reliably and consistently detect that its attempt to launch a VM has really succeeded?

jodh-intel:
Wait a second, let's summarise what is happening here...

- Runtime calls `govmmQemu.LaunchQemu()`.
- `govmmQemu.LaunchQemu()` eventually calls `cmd.Run()`.
- `cmd.Run()` calls https://golang.org/pkg/os/#StartProcess.
- `cmd.Run()` calls `c.Start()` and then calls `c.Wait()`, returning the result of the wait to the caller.
- `cmd.Start()` calls https://golang.org/pkg/os/#StartProcess which creates a new process (single fork+exec - does *not* double fork fwics).

As I undestand it, this PR is to fix the scenario where the containerd-shim-v2 somehow manages to wait on the child *before* `c.Wait()` does. But that's impossible [1]: since the parent of the waited for qemu process is still running (it has to be - it's calling `c.Wait()` :), there is no way anything else can wait for it.

containerd-shim-v2 might register a subreaper but that can only wait on the child process *if the qemu child processes parent exits*. But that isn't happening - it's calling `c.Wait()`.

---
[1] - Quoting from `prctl(2)`:

> A  subreaper  fulfills  the role of init(1) for its descendant processes.  When a process becomes orphaned (i.e., its immediate parent terminates) then that process will be reparented to the nearest still living ancestor subreaper.

bergwolf:
@jodh-intel 
> But that's impossible [1]: since the parent of the waited for qemu process is still running (it has to be - it's calling c.Wait() :), there is no way anything else can wait for it.

That is possible with Golang, unfortunately. The same chunk of code run by a goroutine can be scheduled to different OS threads. So the thread calling `cmd.Wait()` might not be the one calling `cmd.Start()`, unless we lock the thread in the first place. So another way of solving the problem is to use a dedicated goroutine  and call `LockOSThread` to spawn the qemu process. /cc @lifupan

markdryan:
> cmd.Start() calls https://golang.org/pkg/os/#StartProcess which creates a new process (single fork+exec - does not double fork fwics).

@jodh-intel There is a double fork here as you're using daemonize (https://github.com/kata-containers/runtime/blob/2528b62fe881388cdc3f50fca4ff793e44699dc1/virtcontainers/qemu.go#L423)?  So qemu process started by LaunchCustomQemu forks itself after some checks and then exits.  The wait in LaunchQEMU catches the exit of the parent, but the VM continues to run in the child, providing there was no error.  I'm not sure that is relevant to your argument, but I wanted to point this out.

My assumption, which @lifupan confirmed above, was that there were two different go-routines, and hence threads, calling Wait.  I'm not sure what happens if you do this but presumably you can't predict which Wait will reap the process.

lifupan:
> > @lifupan @markdryan what about having a function LaunchQemuNoWait() that is not using Run(), but instead would use Start(). This would return a cmd in order to let the callerWait(). This way, we know we can ignore only the error related to reaping the process
> 
> I suggested something along these lines in the original govmm PR ([intel/govmm#66](https://github.com/intel/govmm/pull/66)), but now I don't think this will solve the issue. I could very well be misunderstanding how this process reaping is going to work, but it seems to me that there may be an architectural issue that needs to be solved here. If something else is reaping qemu processes, how can startSandbox reliably and consistently detect that its attempt to launch a VM has really succeeded?

I agree this is an architecture issue, since the containerd-shimv2 for kata combines the shim's function of children reaper and launching a daemon process of qemu into one program. It seems the reasonable fix is deferring launching the reaper goroutine in shimv2 until  start qemu process successfully. But this 
looks very weird since it'll move some shim's function into kata project.

 this will get containerd project 


> Wait a second, let's summarise what is happening here...
> 
> * Runtime calls `govmmQemu.LaunchQemu()`.
> * `govmmQemu.LaunchQemu()` eventually calls `cmd.Run()`.
> * `cmd.Run()` calls https://golang.org/pkg/os/#StartProcess.
> * `cmd.Run()` calls `c.Start()` and then calls `c.Wait()`, returning the result of the wait to the caller.
> * `cmd.Start()` calls https://golang.org/pkg/os/#StartProcess which creates a new process (single fork+exec - does _not_ double fork fwics).
> 
> As I undestand it, this PR is to fix the scenario where the containerd-shim-v2 somehow manages to wait on the child _before_ `c.Wait()` does. But that's impossible [1]: since the parent of the waited for qemu process is still running (it has to be - it's calling `c.Wait()` :), 

Hi @jodh-intel  As far as I understood the parent of the waited for qemu process is the shimv2 process itself, and there is another reaper goroutine running in shimv2 too. Thus there is a race wait between this c.Wait() and reaper goroutine.  That's this PR tried to fix.

It seems the best fix is:
Defer the shimv2's reaper goroutine start after launching qume process successfully.  Thus can avoid this wait race condition. But this will get containerd project involved and we should add the reaper function into kata's virtruntime pkg. 


Another compromised  method is just as @sboeuf said, we just launching qemu as no wait, since even we run cmd.Wait() successfully, I think we cannot make sure 100%  the qemu daemon had run successfully.

jodh-intel:
> That is possible with Golang, unfortunately.

OMG! :sob: 

Another solution then: make either the runtime or govmm register a sub-reaper to stop containerd-shim-v2's subreaper grabbing the child!

bergwolf:
@lifupan @sboeuf 
> we just launching qemu as no wait, since even we run cmd.Wait() successfully, I think we cannot make sure 100% the qemu daemon had run successfully.

Eventually we rely on qmp connection to make sure qemu process is up and functioning. However, the expectation we want from `cmd.Run()` is the existence of the qmp socket. If we just call `cmd.Start()`, we'll have to retry with a timeout when connecting to the qmp socket.

lifupan:
> @jodh-intel
> 
> > But that's impossible [1]: since the parent of the waited for qemu process is still running (it has to be - it's calling c.Wait() :), there is no way anything else can wait for it.
> 
> That is possible with Golang, unfortunately. The same chunk of code run by a goroutine can be scheduled to different OS threads. So the thread calling `cmd.Wait()` might not be the one calling `cmd.Start()`, unless we lock the thread in the first place. So another way of solving the problem is to use a dedicated goroutine and call `LockOSThread` to spawn the qemu process. /cc @lifupan

@bergwolf  I think LockOSThread  cannot fix this issue. Just as @markdryan  said, two goroutine/threads  calls waiting, and we cannot predict which one will reap the child process.

So it seems the reasonable fix is deferring  to launch the shimv2's reaper goroutine after launching qemu successfully. This will involve the modification in  containerd's project.

bergwolf:
@lifupan The problem with the `LockOSThread` solution is that qemu does fork twice with `--daemonize`.

What about just checking the qmp socket and connection?

Also, assuming the code path you were talking about is the `StartShim` function, can you still reproduce the failure with `GOMAXPROCS` environment variable set to 1 when starting the containerd shimv2 binary? I'm wondering if it is a bug in containerd instead as it sets the current thread a subreaper yet still runs with two OS threads by default. Or you have verify my guess with the following change:
```
diff --git a/runtime/v2/shim/shim.go b/runtime/v2/shim/shim.go
index 39484c19..595db84a 100644
--- a/runtime/v2/shim/shim.go
+++ b/runtime/v2/shim/shim.go
@@ -97,7 +97,7 @@ func setRuntime() {
        if os.Getenv("GOMAXPROCS") == "" {
                // If GOMAXPROCS hasn't been set, we default to a value of 2 to reduce
                // the number of Go stacks present in the shim.
-               runtime.GOMAXPROCS(2)
+               runtime.GOMAXPROCS(1)
        }
 }

```

lifupan:
> @lifupan The problem with the `LockOSThread` solution is that qemu does fork twice with `--daemonize`.
> 
> What about just checking the qmp socket and connection?
> 
> Also, assuming the code path you were talking about is the `StartShim` function, can you still reproduce the failure with `GOMAXPROCS` environment variable set to 1 when starting the containerd shimv2 binary? I'm wondering if it is a bug in containerd instead as it sets the current thread a subreaper yet still runs with two OS threads by default. Or you have verify my guess with the following change:
> 
> ```
> diff --git a/runtime/v2/shim/shim.go b/runtime/v2/shim/shim.go
> index 39484c19..595db84a 100644
> --- a/runtime/v2/shim/shim.go
> +++ b/runtime/v2/shim/shim.go
> @@ -97,7 +97,7 @@ func setRuntime() {
>         if os.Getenv("GOMAXPROCS") == "" {
>                 // If GOMAXPROCS hasn't been set, we default to a value of 2 to reduce
>                 // the number of Go stacks present in the shim.
> -               runtime.GOMAXPROCS(2)
> +               runtime.GOMAXPROCS(1)
>         }
>  }
> ```

Yes, even set maxprocs as 1, I can still reproduce this issue. @bergwolf I'm confused why you think only one thread will prevent this issue?

bergwolf:
> Yes, even set maxprocs as 1, I can still reproduce this issue. @bergwolf I'm confused why you think only one thread will prevent this issue?

Because I thought the failure comes from reparenting rather than being waited in two places. Clearly I'm still missing the root cause of the issue. In `StartShim` code path, what part of containerd is handling SIGCHLD?

lifupan:
> > Yes, even set maxprocs as 1, I can still reproduce this issue. @bergwolf I'm confused why you think only one thread will prevent this issue?
> 
> Because I thought the failure comes from reparenting rather than being waited in two places. Clearly I'm still missing the root cause of the issue. In `StartShim` code path, what part of containerd is handling SIGCHLD?

Not in StartShim(), instead it's in shim.Run() which located in containerd's runtime v2 shim pkg,
and we cannot stop it to reap the child process there since it will depend this to publish events to containerd daemon.

lifupan:
@jodh-intel @markdryan @sboeuf @bergwolf I think this patch isn't the perfect fix, but it's the reasonable method that I had figured out by now.  I know @markdryan concerns scenario is:
the cmd.Start() executes successfully, but due to some reason the qemu daemon failed to run and
exited with some error, then this error status was reaped by the reaper in shimv2 instead of the cmd.Wait(). Thus the cmd.Wait() got error of "no child processes" and this error was skipped by this
fix and dealt with launchQemu() successfully. 

For this case, I think we can let it go on since the following code path is qemu.waitSandbox() which will 
wait for the qemu socket available, but it will get timeout and return an error. 

So for this scenario, without this fix, it will return an error in launchQemu() instead of waitSandbox(), so the key difference is where it returned an error.  I don't think it's a big deal for this.

bergwolf:
@lifupan and I discussed several alternatives and we ended up concluding that this PR is the best approach right now. In future we might ask containerd to fold the reaper to runc shimv2 instead of putting it in the generic code. But for now let's work around it.

LGTM

jodh-intel:
Can we first explore my comment here: https://github.com/kata-containers/runtime/pull/940#issuecomment-442029602?

Since we already have `libcontainer` vendored, we can make use of [`SetSubreaper()`](https://github.com/opencontainers/runc/blob/master/libcontainer/system/linux.go#L142), exactly as [`containerd` does](https://github.com/containerd/containerd/blob/master/cmd/containerd-shim/shim_linux.go#L39). Since subreapers "nest", our subreaper will then be guaranteed to be called for our qemu child. We can then unset the child subreaper, which will revert child processes to being reaped by "the next subreaper up" (containerd's).

lifupan:
> Can we first explore my comment here: [#940 (comment)](https://github.com/kata-containers/runtime/pull/940#issuecomment-442029602)?
> 
> Since we already have `libcontainer` vendored, we can make use of [`SetSubreaper()`](https://github.com/opencontainers/runc/blob/master/libcontainer/system/linux.go#L142), exactly as [`containerd` does](https://github.com/containerd/containerd/blob/master/cmd/containerd-shim/shim_linux.go#L39). Since subreapers "nest", our subreaper will then be guaranteed to be called for our qemu child. We can then unset the child subreaper, which will revert child processes to being reaped by "the next subreaper up" (containerd's).

@jodh-intel  For shimv2, the kata runtime has been combined into shimv2 and formed as one single process, and this process has been set as subreaper,  so there is no meaning to set subreaper again in kata runtime.

jodh-intel:
Ugh - that is a complete pita :cry:

I'm really not in favour of this fix. However, since it seems we have no other options that support all modes of operation, reluctantly...

lgtm

sboeuf:
@lifupan @bergwolf @markdryan @jodh-intel 
There's something I don't get here. We could definitely know (based on if we're running with containerd-shim-v2 or not) if we need to reap the qemu process or not. Based on this, we should be able to provide `govmm` a flag to ask to wait for its child or not.
We would not do a Cmd.Wait() if we knew that an external subreaper was taking care of this, right?

lifupan:
> @lifupan @bergwolf @markdryan @jodh-intel
> There's something I don't get here. We could definitely know (based on if we're running with containerd-shim-v2 or not) if we need to reap the qemu process or not. Based on this, we should be able to provide `govmm` a flag to ask to wait for its child or not.
> We would not do a Cmd.Wait() if we knew that an external subreaper was taking care of this, right?

@sboeuf 
Actually we don't know it's running in a kata cli or shimv2 in govmmQemu.LaunchQemu(), since the VC API hasn't provided any flags to indicate it's called by cli or shimv2, of course we can add it, but I don't think it's reasonable to do it.

lifupan:
@sboeuf @bergwolf   Make launch a qemu with a cmd.Start without cmd.Wait in govmm is also a workaround, do you guys think it's a deal to make a workaround so much complex instead of prefer to
this simple one?

sboeuf:
@bergwolf 
>What about using SandboxConfig.Stateful to indicate that?

Oh yes of course you're right! The point of `stateful` is to indicate that we have a long running process that can take care of reaping.

@lifupan 
>@sboeuf @bergwolf Make launch a qemu with a cmd.Start without cmd.Wait in govmm is also a workaround, do you guys think it's a deal to make a workaround so much complex instead of prefer to
this simple one?

I think relying on `Stateful` to start a Qemu process without waiting for it since we're not trying to reap the child process is a correct approach, and it does not look too complex to me. We just need to define a parameter in LaunchQemu() to indicate we don't need to Wait() the process (needs to happen in govmm). And we need to make sure the hypervisor implementation (Qemu in this case) gets the information about Stateful.

WDYT guys?

lifupan:
@sboeuf @jodh-intel @bergwolf 
I had just figured out another way to fix this issue, please have a look on https://github.com/kata-containers/runtime/pull/964, this way can resolve this issue ultimately and had involved any changes in kata runtime and govmm.

lifupan:
Close this PR and I had opened another one: https://github.com/kata-containers/runtime/pull/1004

PR #950, opened 2018-11-28

### LINKED ISSUE #926: Moving to Docker 18.09 does not delete the VM and proxy when the container terminates

sboeuf:

Fixes #926

### DIFF
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -337,10 +337,11 @@ func StatusSandbox(ctx context.Context, sandboxID string) (SandboxStatus, error)
 		return SandboxStatus{}, errNeedSandboxID
 	}
 
-	lockFile, err := rLockSandbox(sandboxID)
+	lockFile, err := rwLockSandbox(sandboxID)
 	if err != nil {
 		return SandboxStatus{}, err
 	}
+	defer unlockSandbox(lockFile)
 
 	s, err := fetchSandbox(ctx, sandboxID)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


sboeuf:
@jodh-intel @devimc @amshinde @chavafg PTAL, this PR fixes the issue seen with Docker 18.09. Other than that, we don't need to modify anything else, the fact that containerd removed the call to `kill --all` is not a big deal, but we were returning the wrong container status, which was not appreciated by containerd, and eventually `delete` was never called.

egernst:
Suggested commit message edit:
```
When our runtime is asked for the container status, we also handle the scenario where the container is stopped if the shim process for that container on the host has terminated.

In the current implementation, we retrieve the container status before stopping the container, causing a wrong status to be returned.  The wait for the original go-routine's completion was done in a defer within the caller of statusContainers(), resulting in the statusContainer()'s values to return the pre-stopped value. 
 
This bug is first observed when updating to docker v18.09/containerd v1.2.0. With the current implementation, containerd-shim receives the TaskExit when it detects kata-shim is terminating. When checking the container state, however, it does not get the expected "stopped" value.

The following commit resolves the described issue by simplifying the locking used around the status container calls. Originally StatusContainer would request a read lock.  If we needed to update the container status in statusContainer, we'd start a go-routine which would request a read-write lock, waiting for the original read lock to be released.  Can't imagine a bug could linger in this logic.  We now just request a read-write lock in the caller (StatusContainer), skipping the need for a separate go-routine and defer.  This greatly simplifies the logic, and removes the original bug.
```

sboeuf:
@egernst commit message updated! Thanks!

jodh-intel:
@sboeuf - nice!!

Can you find a way to add a unit test for this change though? Given the importance of this fix, it would be good to know we have a test to assert the correct behaviour.

jodh-intel:
@sboeuf - my bad - s/unit test/any type of test/ ;)

jodh-intel:
ftr, I can confirm this PR fixes the issue with Docker 18.09 (tested on ubuntu bionic). The CI will be testing using docker 18.06.1 (from https://github.com/kata-containers/runtime/blob/master/versions.yaml#L154) so we can say that it works with both versions.

sboeuf:
@jodh-intel yes that's a good idea, I think we can write a simple integration test using `kata-runtime state` after we run a simple workload that naturally exits after a few seconds. The result should be `stopped` directly, where before it would have returned `running`.
Reference: https://github.com/kata-containers/tests/issues/956

jodh-intel:
Thanks @sboeuf - let's land this to allow you to write that test... ;)

sboeuf:
Thanks @jodh-intel! I'll write this today!

PR #964, opened 2018-12-03
shimv2: fix the error of reaping qemu process mistakenly

lifupan:
For kata shimv2, it doesn't need to reap it's container processes
since all of them are running in VM, thus there is no need to
set it as a subreaper and also no need to take care of the SIGCHLD
singal.

So this commit will try to unset the subreaper and ignore the SIGCHLD
signal. Since ignoring the SIGCHLD signal will break the containerd
shim's events.Publisher publish method, thus here re-write a publish
function to publish the events to containerd.

Fixes: #939

### DIFF
--- a/cli/containerd-shim-kata-v2/main.go
+++ b/cli/containerd-shim-kata-v2/main.go
@@ -10,6 +10,11 @@ import (
 	"github.com/kata-containers/runtime/containerd-shim-v2"
 )
 
+func shimConfig(config *shim.Config){
+	config.NoReaper = true
+	config.NoSubreaper = true
+}
+
 func main() {
-	shim.Run("io.containerd.kata.v2", containerdshim.New)
+	shim.Run("io.containerd.kata.v2", containerdshim.New, shimConfig)
 }
--- a/vendor/github.com/containerd/containerd/runtime/v2/shim/shim.go
+++ b/vendor/github.com/containerd/containerd/runtime/v2/shim/shim.go
@@ -53,6 +53,17 @@ type Shim interface {
 	StartShim(ctx context.Context, id, containerdBinary, containerdAddress string) (string, error)
 }
 
+// BinaryOpts allows the configuration of a shims binary setup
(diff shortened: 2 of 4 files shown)


containerd-shim-v2/service.go
```diff
@@ -202,9 +215,35 @@ func (s *service) StartShim(ctx context.Context, id, containerdBinary, container
...
+	// thus it's needed to get the struct's
+	// fields value from an interface value
+	// using reflection.
+	elem := reflect.ValueOf(l).Elem()
+	typeOfT := elem.Type()
+
+	for i := 0; i < elem.NumField(); i++ {
+		f := elem.Field(i)
+		switch typeOfT.Field(i).Name {
+		case "address":
```
jodh-intel: This might be a bit "too clever" ;) Since those struct elements are private, they could change and should be able to change without breaking any clients. But we would be broken here if they changed if I'm understanding this correctly?

jodh-intel: It might be worth raising a PR on containerd shim to see if they would be prepared to make these fields public.

lifupan: Yes, to fix this issue gently there must be some adjustments in containerd side, by now I think we can workaround it in our side first and then we communicate with containerd community on how to fix it ultimately.

sboeuf: I'd really appreciate if we could start the conversation with containerd community right away to get a hint at how they feel regarding this proposal. I don't want to merge to Kata Containers a workaround that might never be properly fixed if containerd community does not agree on these changes.
Please start the discussion with them by adding @jodh-intel and myself in copy!

amshinde: Agree. Please open an issue with containerd for this @lifupan.

sboeuf:
thanks @lifupan

lifupan:
close this PR, since I had opened another one: https://github.com/kata-containers/runtime/pull/1004

PR #977, opened 2018-12-04

### LINKED ISSUE #926: Moving to Docker 18.09 does not delete the VM and proxy when the container terminates

sboeuf:

Fixes #926

### DIFF
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -338,10 +338,11 @@ func StatusSandbox(ctx context.Context, sandboxID string) (SandboxStatus, error)
 		return SandboxStatus{}, errNeedSandboxID
 	}
 
-	lockFile, err := rLockSandbox(sandboxID)
+	lockFile, err := rwLockSandbox(sandboxID)
 	if err != nil {
 		return SandboxStatus{}, err
 	}
+	defer unlockSandbox(lockFile)
 
 	s, err := fetchSandbox(ctx, sandboxID)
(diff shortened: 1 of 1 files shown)


jodh-intel:
lgtm

/retest

jcvenegas:
lgtm

jodh-intel:
Ping @kata-containers/runtime

raravena80:
Ping @sboeuf any updates? thx

grahamwhaley:
@sboeuf - this still alive I presume? I guess you are busy on other stuff - just fmi, any eta on getting back to this. I'm loathed to re-trigger the CIs if we know it will be reworked...

sboeuf:
@grahamwhaley is it still relevant? Because the patch is already in master, but this PR is specific to backport it to stable 1.3.

grahamwhaley:
ah, I didn't spot the v1.3 branch @sboeuf right, I'm pretty sure this will be stale then.
speaking of which then, let me check what our docs say about current supported versions - I suspect we are support v1.4 and v1.5, but need to check the defined workflow and timelines etc.
/cc @egernst @gnawux

sboeuf:
@jodh-intel @egernst should we close this PR as we're trying to backport to `stable-1.3` here? I'm asking because we're currently backporting to stable-1.4 and stable-1.5, but not anymore to stable-1.3.

jcvenegas:
@sboeuf +1 to close it .

PR #978, opened 2018-12-04

### LINKED ISSUE #926: Moving to Docker 18.09 does not delete the VM and proxy when the container terminates

sboeuf:

Fixes #926

### DIFF
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -337,10 +337,11 @@ func StatusSandbox(ctx context.Context, sandboxID string) (SandboxStatus, error)
 		return SandboxStatus{}, errNeedSandboxID
 	}
 
-	lockFile, err := rLockSandbox(sandboxID)
+	lockFile, err := rwLockSandbox(sandboxID)
 	if err != nil {
 		return SandboxStatus{}, err
 	}
+	defer unlockSandbox(lockFile)
 
 	s, err := fetchSandbox(ctx, sandboxID)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


amshinde:
@sboeuf Any reason why this was opened when you have #977 open?

sboeuf:
@amshinde #977 is on `stable-1.3` and #978 is on `stable-1.4` ;)

jodh-intel:
lgtm

jcvenegas:
Seems that centos failing with a timeout, restarting but I've seen this same error on different PRs now
```
Running command '/usr/local/bin/kata-runtime [kata-runtime --log=/tmp/bundle938849770/log run --console= --pid-file=/tmp/bundle938849770/pid AsJvY9fDD1MTRdIX9Urd]'
[kata-runtime --log=/tmp/bundle938849770/log run --console= --pid-file=/tmp/bundle938849770/pid AsJvY9fDD1MTRdIX9Urd]
Timeout: 60 seconds
Exit Code: 1
```

jcvenegas:
restart again

```
HEAD is now at fa540c8... Merge pull request #1913 from runcom/fix-devmapper-112
M	Makefile
/tmp/jenkins/workspace/kata-containers-runtime-centos-7-4-PR/go/src/github.com/kata-containers/tests
/tmp/jenkins/workspace/kata-containers-runtime-centos-7-4-PR/go/src/github.com/kubernetes-sigs/cri-o/test /tmp/jenkins/workspace/kata-containers-runtime-centos-7-4-PR/go/src/github.com/kata-containers/tests
Ensure crio service is stopped before running the tests
Ensure docker service is stopped before running the tests
helpers.bash: line 128: /tmp/jenkins/workspace/kata-containers-runtime-centos-7-4-PR/go/src/github.com/kubernetes-sigs/cri-o/test/copyimg/copyimg: No such file or directory
Error pulling quay.io/crio/redis
make: *** [crio] Error 1
```

PR #982, opened 2018-12-06
Network: use device name in the host to determine nic's type

### LINKED ISSUE #981: Potential bug when determining nic's type in AddInterface
# Description of problem

We reuse the logic of createEndpointsFromScan() in the AddInterface() to determine a nic's type. While originally it scans a network ns, finds a link and generate NetworkInfo from link. It will use link's name to determine whether it is a physical/vhost_user nic.
Now physical and vhost_user endpoints hotplug are not supported yet. But in AddInterface(), we generate NetworkInfo from cli args. The name field is actually what we want to set into the VM. So it is never existed in the host. This may be a potential bug. I think we should use device field instead?

caoruidong:
When adding interface, we should use device field in API to determine a
nic's type, not the name we set into the VM.

Fixes #981

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -1101,7 +1101,7 @@ func (s *Sandbox) generateNetInfo(inf *types.Interface) (NetworkInfo, error) {
 	return NetworkInfo{
 		Iface: NetlinkIface{
 			LinkAttrs: netlink.LinkAttrs{
-				Name:         inf.Name,
+				Name:         inf.Device,
 				HardwareAddr: hw,
 				MTU:          int(inf.Mtu),
 			},
--- a/virtcontainers/sandbox_test.go
+++ b/virtcontainers/sandbox_test.go
@@ -18,12 +18,13 @@ import (
 	"syscall"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
-
 	"github.com/kata-containers/runtime/virtcontainers/device/config"
 	"github.com/kata-containers/runtime/virtcontainers/device/drivers"
 	"github.com/kata-containers/runtime/virtcontainers/device/manager"
 	"github.com/kata-containers/runtime/virtcontainers/pkg/annotations"
+	"github.com/kata-containers/runtime/virtcontainers/pkg/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/vishvananda/netlink"
 	"golang.org/x/sys/unix"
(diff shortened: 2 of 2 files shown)


bergwolf:
@caoruidong Does it fix the issue by just setting `Iface.Name` to `inf.Device` in `generateNetInfo` instead? Unlike `types.Interface`, `NetworkInfo` is the runtime internal data structure where the `Iface.Name` field refers to the host network device. /cc @amshinde

virtcontainers/sandbox.go
```diff
@@ -1111,9 +1112,42 @@ func (s *Sandbox) generateNetInfo(inf *types.Interface) (NetworkInfo, error) {
...
-	netInfo, err := s.generateNetInfo(inf)
+	var netInfo NetworkInfo
+
+	netlinkHandle, err := getNetlinkHandle(s.config.NetworkConfig.NetNSPath)
+	if err != nil {
+		return nil, err
+	}
+	defer netlinkHandle.Delete()
+
+	link, err := netlinkHandle.LinkByName(inf.Device)
```
jodh-intel: I'm wondering if we should add a check for `inf != nil`?

caoruidong:
@bergwolf I observe that Device field is actually not used ether in runtime or agent. So I'm not very clear about its semantic meaning. I think your advice is a good one. But Iface.Name may be used somewhere else. I will go through the whole process to check.

bergwolf:
LGTM. Thanks @caoruidong !

jodh-intel:
@caoruidong - this change looks harmless, but I don't understand it as "it looks wrong" :(

Also, it's these sorts of tiny changes which can turn into nasty bugs. This PR might be fixing a bug, but I'd feel more comfortable if there was atleast one unit test with this change :)

@amshinde, @mcastelino, @sboeuf - could you clarify some of this?

caoruidong:
@jodh-intel Yes, I agree. I suspect it is a potential bug when hot attaching a physical nic. Though this feature has not been merged. I'll see if I can add one test

raravena80:
@caoruidong any updates on this PR?

caoruidong:
https://mvdan.cc/xurls/cmd/xurls is broken again

caoruidong:
This PR is not necessary. I just want to record a potential issue.
Now the device field is never used except logging.  I think we need to clarify the purpose of every field when hotplug different types of nic like physical and veth.

Issue #666, opened 2018-08-30
Add support for s390x


ydjainopensource:
This issue will track Z support for Kata.

jodh-intel:
Not sure why I got that one. I've re-assigned to Yash :)

PR #667, opened 2018-08-30
[WIP] Runtime : Add support for Z

ydjainopensource:
Initial support for Z. Does not pass some unit tests

Fixes #666

Signed-off-by: Yash Jain <ydjainopensource@gmail.com>

Issue #659, opened 2018-08-29
vhost-user devices not available on Z


ydjainopensource:
# Description of problem

Qemu on s390x does not support vhost-user devices. We need to find a workaround to bypass this limitation. 

any thoughts?

/cc @markdryan @rbradford @mcastelino @WeiZhang555 @jcvenegas @amshinde @devimc @jodh-intel

jodh-intel:
First question: why doesn't it support this feature?

borntraeger:
Several reasons:
the s390 machine uses the ccw bus instead of the pci bus to drive virtio.  So we would need
 vhost-user-ccw backends, which are not wired up. This would be a smaller change, but there is another issue: vhost-user requires shared memory between qemu and another process. Right now shared memory (such as  -object memory-backend-file or hugetlbfs) is not supported either with QEMU/kvm on s390x.

jodh-intel:
Hi @borntraeger - thanks for the information. Is adding support for `memory-backend-file` planned?

jodh-intel:
/cc @markdryan, @rbradford.

Related: https://github.com/intel/govmm.

alicefr:
@jodh-intel would it be possible to avoid `VhostUserEndpointType`? We can avoid to use huge page by avoiding to set the `#enable_hugepages = true` in the `configuration.toml`. It isn't complete clear to me in which case the vhost-user devices are required. Do they need explicitly to be configured in the `devices.json` .

jcvenegas:
@ydjainopensource @alicefr the VhostUserEndpointType endpoint today I think is only used for dpdk use cases. If that is the case the VhostUserEndpointType should not be used at all. @egernst is that correct?

amshinde:
@alicefr You will hit `VhostUserEndpointType` code path, only if you are using a CNM/CNI network plugin that explicitly passes a vhost-user endpoint.  You do not need to do any other configuration. So, we can simply list this as a limitation on s390x.

egernst:
Yes, @jcvenegas.

jcvenegas:
agree with @amshinde it should be documented as a limitation.

ydjainopensource:
Sounds sane to me. We will document this as a limitation.

alicefr:
[...]

PR #990, opened 2018-12-07
s390x: add support for s390x

alicefr:
The PR adds the support for s390x.

In the case of CCW devices, the vhost-user devices are not supported.
See #659. An error message is thrown if they tried to be used.

Memory hotplug is not supported on s390 yet and an error message is thrown.

The VirtioNetPCI has been changed to VirtioNet. The generalization
allows to set the VirtioNet to the correct CCW device for s390x.

A lot of the elements in this PR have been taken from https://github.com/kata-containers/runtime/pull/667

Fixes: #666

Co-authored-by: Yash D Jain ydjainopensource@gmail.com
Signed-off-by: Alice Frosi <afrosi@de.ibm.>

### DIFF
--- a/cli/kata-check_data_s390x_test.go
+++ b/cli/kata-check_data_s390x_test.go
@@ -1,0 +1,21 @@
+// Copyright (c) 2018 IBM
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package main
+
+const testCPUInfoTemplate = `
+vendor_id       : IBM/S390
+# processors    : 4
+bogomips per cpu: 20325.00
+max thread id   : 0
+features	: esan3 zarch stfle msa ldisp eimm dfp edat etf3eh highgprs te vx sie 
+cache0          : level=1 type=Data scope=Private size=128K line_size=256 associativity=8
+cache1          : level=1 type=Instruction scope=Private size=96K line_size=256 associativity=6
+cache2          : level=2 type=Data scope=Private size=2048K line_size=256 associativity=8
+cache3          : level=2 type=Instruction scope=Private size=2048K line_size=256 associativity=8
+cache4          : level=3 type=Unified scope=Shared size=65536K line_size=256 associativity=16
+cache5          : level=4 type=Unified scope=Shared size=491520K line_size=256 associativity=30
+processor 0: version = FF,  identification = FFFFFF,  machine = 2964
+`
--- a/cli/kata-check_s390x.go
+++ b/cli/kata-check_s390x.go
@@ -1,0 +1,124 @@
+// Copyright (c) 2018 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package main
+
+import (
+	"fmt"
+	"github.com/sirupsen/logrus"
+	"strings"
+)
+
+const (
+	cpuFlagsTag        = genericCPUFlagsTag
+	archCPUVendorField = genericCPUVendorField
+	// On s390x the cpu model is indicated by the field machine.
+	// Example:
+	// processor 0: version = FF,  identification = 3FEC87,  machine = 2964
+	archCPUModelField = "machine"
+)
+
+// archRequiredCPUFlags maps a CPU flag value to search for and a
+// human-readable description of that value.
+var archRequiredCPUFlags = map[string]string{}
+
+// archRequiredCPUAttribs maps a CPU (non-CPU flag) attribute value to search for
+// and a human-readable description of that value.
+var archRequiredCPUAttribs = map[string]string{}
+
+// archRequiredKernelModules maps a required module name to a human-readable
+// description of the modules functionality and an optional list of
+// required module parameters.
+var archRequiredKernelModules = map[string]kernelModule{
+	"kvm": {
+		desc: "Kernel-based Virtual Machine",
+	},
+}
+
+func setCPUtype() error {
+	return nil
+}
+
+// kvmIsUsable determines if it will be possible to create a full virtual machine
+// by creating a minimal VM and then deleting it.
+func kvmIsUsable() error {
+	return genericKvmIsUsable()
+}
+
+func archHostCanCreateVMContainer() error {
+	return kvmIsUsable()
+}
+
+// hostIsVMContainerCapable checks to see if the host is theoretically capable
+// of creating a VM container.
+func hostIsVMContainerCapable(details vmContainerCapableDetails) error {
+
+	_, err := getCPUInfo(details.cpuInfoFile)
+	if err != nil {
+		return err
+	}
+
+	count, err := checkKernelModules(details.requiredKernelModules, archKernelParamHandler)
+	if err != nil {
+		return err
+	}
+
+	if count == 0 {
+		return nil
+	}
+
+	return fmt.Errorf("ERROR: %s", failMessage)
+
+}
+
+func archKernelParamHandler(onVMM bool, fields logrus.Fields, msg string) bool {
+	return genericArchKernelParamHandler(onVMM, fields, msg)
+}
+
+// getS390xCPUDetails returns the cpu information
+func getS390xCPUDetails() (vendor, model string, err error) {
+	prefixModel := "processor"
+	cpuinfo, err := getCPUInfo(procCPUInfo)
+	if err != nil {
+		return "", "", err
+	}
+
+	lines := strings.Split(cpuinfo, "\n")
+
+	for _, line := range lines {
+		if archCPUVendorField != "" {
+			if strings.HasPrefix(line, archCPUVendorField) {
+				fields := strings.Split(line, ":")
+				if len(fields) > 1 {
+					vendor = strings.TrimSpace(fields[1])
+				}
+			}
+		}
+		if archCPUModelField != "" {
+			if strings.HasPrefix(line, prefixModel) {
+				fields := strings.Split(strings.TrimSpace(line), ",")
+				cpuModel := strings.Split(fields[2], "=")
+				model = strings.TrimSpace(cpuModel[1])
+			}
+		}
+	}
+
+	if vendor == "" {
+		return "", "", fmt.Errorf("cannot find vendor field in file %v", procCPUInfo)
+	}
+
+	if model == "" {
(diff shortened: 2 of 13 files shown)


cli/kata-check_data_s390x_test.go
```diff
@@ -0,0 +1,21 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package main
+
+const testCPUInfoTemplate = `
+vendor_id       : IBM/S390
+# processors    : 4
+bogomips per cpu: 20325.00
```
jodh-intel: Wow :)

cli/kata-check_s390x_test.go
```diff
@@ -0,0 +1,285 @@
...
+		result := archKernelParamHandler(d.onVMM, d.fields, d.msg)
+		if d.expectIgnore {
+			assert.True(result, "test %d (%+v)", i, d)
+		} else {
+			assert.False(result, "test %d (%+v)", i, d)
+		}
+	}
+}
+
+type TestDataa struct {
```
jodh-intel: This seems to have an extra "a" at the end. But also appears to be unused?

cli/utils_s390x.go
```diff
@@ -0,0 +1,12 @@
+// +build s390x
```
jodh-intel: This is technically redundant as the filename already ends in the build architecture.

virtcontainers/qemu_s390x.go
```diff
@@ -0,0 +1,123 @@
...
+		Type:    QemuCCWVirtio,
+		Options: defaultQemuMachineOptions,
+	},
+}
+
+// MaxQemuVCPUs returns the maximum number of vCPUs supported
+func MaxQemuVCPUs() uint32 {
+	// Max number of virtual Cpu defined in qemu. See
+	// https://github.com/qemu/qemu/blob/80422b00196a7af4c6efb628fae0ad8b644e98af/target/s390x/cpu.h#L55
+	// #define S390_MAX_CPUS 248
```
jodh-intel: It's a shame this can't be queried by calling qemu (via govmm) directly. But clearly not a blocker as other architectures also hard-code a value.

Aside: For the future, I do wonder if we could push the hard-coded arch-specific values down into govmm rather than having them in the runtime though - wdyt @markdryan?

cli/utils_s390x.go
alicefr: you're right. I remove it

cli/kata-check_s390x_test.go
alicefr: I did the same as [`amd64`](https://github.com/kata-containers/runtime/blob/a323a87b591f2e38efb74b5b371e66f230fd188e/cli/kata-check_amd64_test.go#L487) and [`ppc64le`](https://github.com/kata-containers/runtime/blob/a323a87b591f2e38efb74b5b371e66f230fd188e/cli/kata-check_ppc64le_test.go#L211). I forgot to changed to the function `genericTestGetCPUDetails`. Now I fixed it

jodh-intel: Thanks!

virtcontainers/qemu_s390x.go
jodh-intel: Sure - this was more a question for @markdryan ;)

jodh-intel:
Ping @kata-containers/runtime! ;)

jodh-intel:
Travis is complaining due to a static analysis issue:

```
virtcontainers/qemu.go:907:9:warning: if block ends with a return statement, so drop this else and outdent its block (golint)
```

virtcontainers/qemu.go
```diff
@@ -892,9 +892,13 @@ func (q *qemu) hotplugNetDevice(endpoint Endpoint, op operation) error {
...
-		if err = q.qmpMonitorCh.qmp.ExecuteNetPCIDeviceAdd(q.qmpMonitorCh.ctx, tap.Name, devID, endpoint.HardwareAddr(), addr, bridge.ID, romFile, int(q.config.NumVCPUs), q.arch.runNested()); err != nil {
-			return err
+
+		if machine, _ := q.getQemuMachine(); machine.Type == QemuCCWVirtio {
+			err = q.qmpMonitorCh.qmp.ExecuteNetCCWDeviceAdd(q.qmpMonitorCh.ctx, tap.Name, devID, endpoint.HardwareAddr(), addr, bridge.ID, int(q.config.NumVCPUs))
+		} else {
+			err = q.qmpMonitorCh.qmp.ExecuteNetPCIDeviceAdd(q.qmpMonitorCh.ctx, tap.Name, devID, endpoint.HardwareAddr(), addr, bridge.ID, romFile, int(q.config.NumVCPUs), q.arch.runNested())
 		}
+		return err
 	} else {
```
alicefr: @jodh-intel I can remove this else to simplify the code structure. It's actually redundant

jodh-intel: Thanks!

alicefr:
@jodh-intel now the changes in `virtcontainers/qemu.go` are a little bit more, but it simplify the code structure. I moved the switch out of the if statement. It was the same for both if and else statement and removed the else because the if statement has a `return err`

jodh-intel:
Thanks @alicefr.

virtcontainers/qemu.go
```diff
@@ -892,33 +893,27 @@ func (q *qemu) hotplugNetDevice(endpoint Endpoint, op operation) error {
...
-		case TapEndpointType:
-			drive := endpoint.(*TapEndpoint)
-			tap = drive.TapInterface
-		default:
-			return fmt.Errorf("this endpoint is not supported")
-		}
 
-		if err := q.removeDeviceFromBridge(tap.ID); err != nil {
-			return err
+		if machine, _ := q.getQemuMachine(); machine.Type == QemuCCWVirtio {
```
devimc:
```go
	if machine, _ := q.getQemuMachine(); machine.Type == QemuCCWVirtio {
		return q.qmpMonitorCh.qmp.ExecuteNetCCWDeviceAdd(q.qmpMonitorCh.ctx, tap.Name, devID, endpoint.HardwareAddr(), addr, bridge.ID, int(q.config.NumVCPUs))
	}
	return q.qmpMonitorCh.qmp.ExecuteNetPCIDeviceAdd(q.qmpMonitorCh.ctx, tap.Name, devID, endpoint.HardwareAddr(), addr, bridge.ID, romFile, int(q.config.NumVCPUs), q.arch.runNested())
```

I think this looks better, @alicefr what do you think?

virtcontainers/qemu_s390x.go
```diff
@@ -0,0 +1,123 @@
...
+
+	devices = append(devices, console)
+
+	return devices
+}
+
+// appendVhostUserDevice throws an error if vhost devices are tried to be used.
+// See issue https://github.com/kata-containers/runtime/issues/659
+func (q *qemuS390x) appendVhostUserDevice(devices []govmmQemu.Device, attr config.VhostUserDeviceAttrs) []govmmQemu.Device {
+	logrus.Fatalln("No vhost-user devices supported on s390x")
```
devimc: please change `appendVhostUserDevice` in `qemuArchBase` to return `([]govmmQemu.Device, error)`

virtcontainers/qemu.go
alicefr: I changed the code as you suggested

virtcontainers/qemu_s390x.go
alicefr: I introduced the `error` and changed in the palce where the function is called

virtcontainers/qemu.go
```diff
@@ -892,33 +893,25 @@ func (q *qemu) hotplugNetDevice(endpoint Endpoint, op operation) error {
...
-			return err
-		}
-		if err := q.qmpMonitorCh.qmp.ExecuteNetdevDel(q.qmpMonitorCh.ctx, tap.Name); err != nil {
-			return err
-		}
+	if err := q.removeDeviceFromBridge(tap.ID); err != nil {
+		return err
+	}
+
+	devID := "virtio-" + tap.ID
```
jcvenegas: Nit: Now this string is repeated wonder the value make it a constant.

jodh-intel:
@bergwolf, @caoruidong, @grahamwhaley, @devimc - ptal

virtcontainers/qemu.go
alicefr: @sboeuf you're right, I can add a check

cli/kata-check_s390x.go
```diff
@@ -0,0 +1,125 @@
...
+				model = strings.TrimSpace(cpuModel[1])
+			}
+		}
+	}
+
+	if vendor == "" {
+		return "", "", fmt.Errorf("cannot find vendor field in file %v", procCPUInfo)
+	}
+
+	// model name is optional
```
caoruidong: Is it optional?

virtcontainers/qemu.go
```diff
@@ -891,34 +893,28 @@ func (q *qemu) hotplugNetDevice(endpoint Endpoint, op operation) error {
...
-			return err
-		}
-		if err := q.qmpMonitorCh.qmp.ExecuteNetdevDel(q.qmpMonitorCh.ctx, tap.Name); err != nil {
-			return err
-		}
+	if err := q.qmpMonitorCh.qmp.ExecuteDeviceDel(q.qmpMonitorCh.ctx, devID); err != nil {
+		return err
+	}
+	if err := q.qmpMonitorCh.qmp.ExecuteNetdevDel(q.qmpMonitorCh.ctx, tap.Name); err != nil {
+		return err
```
caoruidong: Thanks for refactoring!

cli/kata-check_s390x.go
alicefr: you're right, it's not optional. I forgot to remove the comment. Before the model was not correctly parsed

alicefr:
I introduced the error check for the `getQemuMachine()` and moved the `devID := "virtio-" + tap.ID
` with the switch outside the if statement. I removed the model optional comment

jodh-intel:
\o/ Thanks again @alicefr! :smile:

Issue #995, opened 2018-12-09
Shimv2 run container failed when factory/template enabled


lifupan:
# Description of problem

When agent is configured as longLive, the VM's agent created
by factory will not close it's client once it connected, thus
the sandbox's agent cannot re-connect successfully.

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #996, opened 2018-12-09
virtcontainers: share the agent's client between factory's VM and san…

lifupan:
…dbox

When agent is configured as longLive, the VM's agent created
by factory will not close it's client once it connected, thus
the sandbox's agent cannot re-connect successfully.

Sharing the agent's client between VM's agent and sandbox
can fix this issue.

Fixes: #995


virtcontainers/agent.go
```diff
@@ -152,6 +152,12 @@ type agent interface {
 	// get agent url
 	getAgentURL() (string, error)
 
+	// get the agent's client
+	getClient() interface{}
+
+	// set to use an existing agent's client
+	setClient(client interface{}) error
```
bergwolf: What about `reuseAgent(a agent) error` API and internally assert agent is of the right type and reuse agent.client internally. Then we don't expose `interface{}` and agent's internal knowledge around.

lifupan: @bergwolf Updated!

bergwolf:
LGTM, thanks!

jodh-intel:
lgtm

PR #1004, opened 2018-12-11
shimv2: fix the error of reaping qemu process mistakenly

lifupan:
For kata shimv2, it doesn't need to reap it's container processes
since all of them are running in VM, thus there is no need to
set it as a subreaper and also no need to take care of the SIGCHLD
singal.


Ace-Tang:
Shim v2 do not need to reap the children process since kata-shim will do the reaper, is it ?

sboeuf:
@Ace-Tang in case of shim-v2, there's no need for external reaping since the only process that is run by kata-runtime is gonna be the hypervisor, and this one is reaped by kata-runtime already.

lifupan:
> Shim v2 do not need to reap the children process since kata-shim will do the reaper, is it ?

No, actually it's kata agent. Since the containers processes are running in VM, in which kata agent will do the processes reaping and shimv2 only forward the exit status from kata agent to containerd.

bergwolf:
LGTM

bergwolf:
metrics is unstable recently. Let's merge it since the change doesn't affect metrics anyway.

Issue #1008, opened 2018-12-11
sudo make install fail


teawater:
sudo make install
INFO: yq was not found, installing it
make: /bin/yq: Command not found
golang.mk:20: *** "ERROR: cannot determine minimum golang version".  Stop.

grahamwhaley:
hi @teawater This might be because your `GOPATH` is not set?
If so, we should check that it is (even though technically now `golang` itself does not need a `GOPATH` to be set), as we rely on the value of `GOPATH` in many of our files and scripts to build and test kata containers.

jodh-intel:
Yep - I tend to install using the following (as do the CI scripts):

```sh
$  make clean && make && sudo -E PATH=$PATH make install
```

PR #1015, opened 2018-12-12
make: Add "GOPATH not set"

teawater:
Most of the projects, they can be built with "make".  After that,
"sudo make install" can install the application.
It is not work for kata-runtime because kata-runtime's make must work
with golang in the environment that default sudo cannot supply it.
But "make install" doesn't need golang.

So add "GOPATH not set" to handle the issue.

Fixes: #1008

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,26 +12,32 @@ for file in /etc/os-release /usr/lib/os-release; do \
     fi \
 done)
 
-GOARCH=$(shell go env GOARCH)
 HOST_ARCH=$(shell arch)
-SKIP_GO_VERSION_CHECK=
+ifeq ($(GOPATH),)
+    SKIP_GO_VERSION_CHECK=y
+else
+    SKIP_GO_VERSION_CHECK=
+endif
 
 ifeq ($(SKIP_GO_VERSION_CHECK),)
     include golang.mk
 endif
 
-ifeq ($(ARCH),)
-    ARCH = $(GOARCH)
+ifneq ($(GOPATH),)
+    GOARCH=$(shell go env GOARCH)
+    ifeq ($(ARCH),)
+        ARCH = $(GOARCH)
+    endif
+
+    ARCH_DIR = arch
+    ARCH_FILE_SUFFIX = -options.mk
+    ARCH_FILE = $(ARCH_DIR)/$(ARCH)$(ARCH_FILE_SUFFIX)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -14,24 +14,30 @@ done)
...
 ifeq ($(SKIP_GO_VERSION_CHECK),)
     include golang.mk
 endif
 
-ifeq ($(ARCH),)
-    ARCH = $(GOARCH)
-endif
+ifneq ($(GOPATH),)
+    ifeq ($(ARCH),)
+        ARCH = $(GOARCH)
```
jcvenegas: you may want to move GOARCH=$(shell go env GOARCH) here

Makefile
```diff
@@ -14,24 +14,30 @@ done)
...
+    endif
 
-ARCH_DIR = arch
-ARCH_FILE_SUFFIX = -options.mk
-ARCH_FILE = $(ARCH_DIR)/$(ARCH)$(ARCH_FILE_SUFFIX)
-ARCH_FILES = $(wildcard arch/*$(ARCH_FILE_SUFFIX))
-ALL_ARCHES = $(patsubst $(ARCH_DIR)/%$(ARCH_FILE_SUFFIX),%,$(ARCH_FILES))
+    ARCH_DIR = arch
```
jcvenegas: none of the following variables are dependent of GOPATH

Makefile
```diff
@@ -498,6 +505,10 @@ show-summary: show-header
 	@printf "• golang:\n"
 	@printf "\t"
 	@go version
+else
+	@printf "• No golang mode:\n"
```
grahamwhaley: suggest a reword here:
```
    @printf "• GOPATH not set:\n"
    @printf "\tCan only install prebuilt binaries\n"
```

Makefile
teawater: If they are moved out from "ifneq ($(GOPATH),)" will get following error without GOPATH:
sudo make install
[sudo] password for teawater:
Makefile:39: arch/-options.mk: No such file or directory
make: *** No rule to make target 'arch/-options.mk'.  Stop.

lifupan:
LGTM

gnawux:
LGTM

Actually, I found this issue during the kubecon demo rehearsal, and thanks @teawater for solving it.

> @jcvenegas 
> but agree probably wont need to block if non of the GOLANG variables are not set.

Yes, this is my point.

PR #1026, opened 2018-12-14
Adding: Firecracker Client API Support

mcastelino:
Adding the firecracker client API support.  This is in preparation of firecracker support in Kata containers.
This includes the swagger definition, auto generated code and the vendoring of packages required by this code.

This does not include the actual implementation of the hypervisor interface which will be done in subsequent commit.


Gopkg.toml
```diff
@@ -70,6 +70,19 @@
  name = "github.com/containerd/containerd"
  revision = "f05672357f56f26751a521175c5a96fc21fa8603"
 
+[[constraint]]
+ name = "github.com/go-openapi/runtime"
```
sboeuf: Nit: This is a TOML file, so I think it requires you to use 2 spaces to indent the entries under `constraints`.

Gopkg.toml
```diff
@@ -70,6 +70,19 @@
...
+ revision = "aadb2cc7b8862fdab251bf60a8b9d2a951c9746c"
+
+
```
sboeuf: Nit: Please remove the extra blank line.

Gopkg.toml
mcastelino: Looks like we did the same for containerd too. Let fix both to make it consistent

bergwolf:
What's the relationship between this package and the FC client offered by aws team (https://github.com/firecracker-microvm/firecracker-go-sdk)? Is the official one missing some key functionality for us to use?

WeiZhang555:
>What's the relationship between this package and the FC client offered by aws team (https://github.com/firecracker-microvm/firecracker-go-sdk)? Is the official one missing some key functionality for us to use?

I have same question with @bergwolf , can we just vendor firecracker client and re-use it?

jodh-intel:
We could just exclude the licence checks from the auto-generated `virtcontainers/pkg/fireclient` package by updating:

- https://github.com/kata-containers/tests/blob/master/.ci/static-checks.sh#L355

mcastelino:
Closing this in favor of https://github.com/kata-containers/runtime/pull/1044
This directly vendors the firecracker-go-dsk

PR #1028, opened 2018-12-14
network: Don't assume multiple queues support by default

sboeuf:
This pull request reorganizes slightly the network code to be able to check the hypervisor capabilities. A new hypervisor capabilities called `multiQueueSupport` has been added, and can be used to determine if the network should be setup with or without multiple queues.

### DIFF
--- a/virtcontainers/network.go
+++ b/virtcontainers/network.go
@@ -505,7 +505,12 @@ func getLinkByName(netHandle *netlink.Handle, name string, expectedLink netlink.
 func xConnectVMNetwork(endpoint Endpoint, h hypervisor) error {
 	netPair := endpoint.NetworkPair()
 
-	queues := int(h.hypervisorConfig().NumVCPUs)
+	queues := 0
+	caps := h.capabilities()
+	if caps.isMultiQueueSupported() {
+		queues = int(h.hypervisorConfig().NumVCPUs)
+	}
+
 	disableVhostNet := h.hypervisorConfig().DisableVhostNet
 
 	if netPair.NetInterworkingModel == NetXConnectDefaultModel {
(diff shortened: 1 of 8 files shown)


mcastelino:
LGTM

virtcontainers/network.go
```diff
@@ -501,30 +501,50 @@ func getLinkByName(netHandle *netlink.Handle, name string, expectedLink netlink.
...
-// The endpoint type should dictate how the connection needs to be made
-func xconnectVMNetwork(endpoint Endpoint, connect bool, numCPUs uint32, disableVhostNet bool) error {
+// The endpoint type should dictate how the connection needs to happen.
+func xConnectVMNetwork(endpoint Endpoint, h hypervisor) error {
 	netPair := endpoint.NetworkPair()
 
+	queues := 0
+	caps := h.capabilities()
+	if caps.isMultiQueueSupported() {
+		queues = int(h.hypervisorConfig().NumVCPUs)
```
egernst: I still don't care for this default, but its a reasonable starting point, as it matches current logic.  We should be doing this only up to a certain number (4?)

sboeuf: Oh yeah I get it, but this PR should not modify the current behavior IMO, that should be part of a PR fixing this, wdyt?

sboeuf:
Let me fix the CI right now!

bergwolf:
LGTM. `jenkins-ci-ubuntu-16-04-initrd` failed on kata-containers/tests/issues/1010, restarting.

grahamwhaley:
@egernst - rather than hard code the queue cap limit (to 4 or whatever), add a toml config option to allow that to be set, and default that cap to 4 maybe.

Issue #1035, opened 2018-12-17
The log's output in katautils.LoadConfiguration() will break shimv2


lifupan:
# Description of problem

containerd would like to get the shimv2's socket
address from shimv2's stdout, thus it's better
to discard the log's output before shimv2 init
it's logger.

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #1036, opened 2018-12-17
shimv2: discard log before shim init log output

lifupan:
containerd would like to get the shim's socket
address from shimv2's stdout, thus it's better
to discard the log's output in stdio before shimv2 init
it's logger and add a hook to log into syslog at the same time.

Fixes: #1035

### DIFF
--- a/containerd-shim-v2/service.go
+++ b/containerd-shim-v2/service.go
@@ -8,6 +8,7 @@ package containerdshim
 import (
 	"context"
 	"encoding/json"
+	"io/ioutil"
 	"os"
 	sysexec "os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


containerd-shim-v2/service.go
```diff
@@ -56,6 +57,10 @@ var vci vc.VC = &vc.VCImpl{}
 // New returns a new shim service that can be used via GRPC
 func New(ctx context.Context, id string, publisher events.Publisher) (cdshim.Shim, error) {
 	logger := logrus.WithField("ID", id)
+	// Discard the log before shim init its log output. Otherwise
+	// it will output into stdio, from which containerd would like
+	// to get the shim's socket address.
+	logrus.SetOutput(ioutil.Discard)
```
jodh-intel: Ouch - yes I've seen problems like this before. But won't we lose potentially useful information here? Is there a way to suppress the log output until after containerd gets the socket address? Alternatively, can we redirect this log output somewhere else?

lifupan: Hi @jodh-intel 
So how about passing "ignoreLogging=false" to LoadConfiguration() in which will add hook logging into syslog?

jodh-intel: Sounds good. That would make the shim work more like the current runtime. It would also mean we can make use of https://github.com/kata-containers/tests/tree/master/cmd/log-parser for the shim.

lifupan:
ping @kata-containers/community

gnawux:
LGTM

jodh-intel:
@bergwolf - sure! ;)

lgtm

PR #1041, opened 2018-12-17
Copy files from host to guest

devimc:


virtcontainers/capabilities.go
```diff
@@ -47,3 +49,25 @@ func (caps *capabilities) isMultiQueueSupported() bool {
 func (caps *capabilities) setMultiQueueSupport() {
 	caps.flags |= multiQueueSupport
 }
+
+func (caps *capabilities) isHotplugSupported() bool {
```
sboeuf: Please drop this, as we don't need it.

virtcontainers/capabilities.go
```diff
@@ -9,6 +9,8 @@ const (
 	blockDeviceSupport = 1 << iota
 	blockDeviceHotplugSupport
 	multiQueueSupport
+	hotplugUnsupported
```
sboeuf: Please drop this as we don't need it.

virtcontainers/capabilities.go
```diff
@@ -47,3 +49,25 @@ func (caps *capabilities) isMultiQueueSupported() bool {
...
 }
+
+func (caps *capabilities) isHotplugSupported() bool {
+	if caps.flags&hotplugUnsupported == 0 {
+		return true
+	}
+	return false
+}
+
+func (caps *capabilities) setHotplugUnsupported() {
```
sboeuf: Please drop this as we don't need it.

virtcontainers/kata_agent.go
```diff
@@ -252,6 +253,11 @@ func (k *kataAgent) configure(h hypervisor, id, sharePath string, builtin bool,
 		k.proxyBuiltIn = true
 	}
 
+	caps := h.capabilities()
+	if !caps.is9pSupported() {
```
sboeuf: Please add some comments here, making clear that if 9p is not supported, we don't want to create any shared directory, which is why we're returning right here.

virtcontainers/kata_agent.go
```diff
@@ -252,6 +253,11 @@ func (k *kataAgent) configure(h hypervisor, id, sharePath string, builtin bool,
...
+		return nil
+	}
+
 	// Adding the shared volume.
 	// This volume contains all bind mounted container bundles.
```
sboeuf: Please extend the comment here, saying that this code path should only be executed when 9p is supported.

virtcontainers/kata_agent.go
```diff
@@ -1003,6 +1054,14 @@ func (k *kataAgent) createContainer(sandbox *Sandbox, c *Container) (p *Process,
 
 	k.handleShm(grpcSpec, sandbox)
 
+	caps := sandbox.hypervisor.capabilities()
+	if !caps.is9pSupported() {
```
sboeuf: I would do this check and the file copy from `c.mountSharedDirMounts()` instead. This means we centralize everything in the `container.go` implementation. But that also means that you need to define `copyFile()` as part of the agent interface, which makes sense to me as you're asking to copy a file to the guest through the agent support.

egernst:
I am having mixed-thoughts regarding the capability flag for the hypervisor.  I would like this to be .toml based in addition, perhaps.  WDYT? @mcastelino @sboeuf @devimc

sboeuf:
@egernst 
>I am having mixed-thoughts regarding the capability flag for the hypervisor. I would like this to be .toml based in addition, perhaps. WDYT? @mcastelino @sboeuf @devimc @sameo

Mid-term, definitely. If we think about real use cases, then we might need to cover the case where we want to use QEMU without 9pfs (even if 9p is supported), which means the internal decision should be based on hypervisor capabilities AND input from user (configuration.toml basically).

Long term, maybe we simply want to get rid of 9p, and we could call this capability `filesystemSharingSupport`, which would be covered by `virtio-fs` in case of QEMU, and which would be false in case of FC for instance.

virtcontainers/capabilities.go
```diff
@@ -47,3 +49,25 @@ func (caps *capabilities) isMultiQueueSupported() bool {
...
+	caps.flags |= hotplugUnsupported
+}
+
+func (caps *capabilities) is9pSupported() bool {
```
mcastelino: @egernst instead of calling this is9pSupported, should this not be something more generic. 9p is one way to share files between host and the VM?

virtcontainers/capabilities.go
```diff
@@ -9,6 +9,8 @@ const (
 	blockDeviceSupport = 1 << iota
 	blockDeviceHotplugSupport
 	multiQueueSupport
+	hotplugUnsupported
+	plan9FSUnsupported
```
mcastelino: @egernst instead of calling this is9pSupported, should this not be something more generic. 9p is one way to share files between host and the VM?

virtcontainers/kata_agent.go
```diff
@@ -1003,6 +1056,14 @@ func (k *kataAgent) createContainer(sandbox *Sandbox, c *Container) (p *Process,
...
+		// 9p is not supported, files must be copied
+		if err := k.copyFiles(c.mounts, newMounts); err != nil {
```
mcastelino: @egernst does this mean even files coming in from tmpfs, secrets all of them get copied? What happens to volumes.

virtcontainers/kata_agent.go
```diff
@@ -612,22 +618,27 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
...
+	if caps.is9pSupported() {
+		sharedDir9pOptions = append(sharedDir9pOptions, fmt.Sprintf("msize=%d", sandbox.config.HypervisorConfig.Msize9p))
 
-	storages := []*grpc.Storage{sharedVolume}
+		// We mount the shared directory in a predefined location
+		// in the guest.
+		// This is where at least some of the host config files
+		// (resolv.conf, etc...) and potentially all container
+		// rootfs will reside.
+		sharedVolume := &grpc.Storage{
```
mcastelino: @egernst what is the else case here. Do we lose support for volumes?

mcastelino:
> Long term, maybe we simply want to get rid of 9p, and we could call this capability `filesystemSharingSupport`, which would be covered by `virtio-fs` in case of QEMU, and which would be false in case of FC for instance.

Yes, this will decouple it from the specific implementation.

devimc:
Depends-on https://github.com/kata-containers/agent/pull/433

virtcontainers/capabilities.go
sboeuf: Yeah we could call it `filesystemSharingSupport`. @egernst @devimc ?

sboeuf:
@devimc https://github.com/kata-containers/agent/pull/433 got merged! Could you please update this PR :)

virtcontainers/container.go
```diff
@@ -477,17 +477,27 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
...
-
-		if err := bindMount(c.ctx, m.Source, mountDest, false); err != nil {
-			return nil, err
+		guestSource := filepath.Join(guestSharedDir, filename)
+
+		// copy file to contaier's rootfs if filesystem sharing is not supported, otherwise
+		// bind mount it in the shared directory.
+		caps := c.sandbox.hypervisor.capabilities()
+		if !caps.isFsSharingSupported() {
+			// filesystem sharing is not supported, files must be copied
```
jodh-intel: How about converting this comment into a log call?

virtcontainers/hyperstart_agent.go
```diff
@@ -1005,3 +1005,8 @@ func (h *hyper) setGuestDateTime(time.Time) error {
 	// hyperstart-agent does not support setGuestDateTime
 	return nil
 }
+
+func (h *hyper) copyFile(src, dst string) error {
+	// hyperstart-agent does not support copyFile
+	return nil
```
jodh-intel: Shouldn't this be...

```suggestion
	return errors.New("hyperstart-agent does not support copyFile")
```

virtcontainers/kata_agent.go
```diff
@@ -1708,3 +1726,39 @@ func (k *kataAgent) convertToRoutes(aRoutes []*aTypes.Route) (routes []*types.Ro
 
 	return routes
 }
+
+func (k *kataAgent) copyFile(src, dst string) error {
+	var st unix.Stat_t
+
+	err := unix.Stat(src, &st)
+	if err != nil {
+		return fmt.Errorf("Could get file %s information", src)
```
jodh-intel:
```suggestion
		return fmt.Errorf("Could not get file %s information: %v", src, err)
```

virtcontainers/kata_agent.go
```diff
@@ -1708,3 +1726,39 @@ func (k *kataAgent) convertToRoutes(aRoutes []*aTypes.Route) (routes []*types.Ro
...
+	}
+
+	b, err := ioutil.ReadFile(src)
+	if err != nil {
+		return fmt.Errorf("Could not read file %s", src)
```
jodh-intel: Again, it would help to include the error so the user can see why the read failed.

virtcontainers/hyperstart_agent.go
devimc: based on other implementations https://github.com/kata-containers/runtime/blob/master/virtcontainers/hyperstart_agent.go#L999-L1007

I don't think so

virtcontainers/kata_agent.go
devimc: good catch! thanks

virtcontainers/kata_agent.go
devimc: yes, thanks :smile:

devimc:
@egernst  changes applied, thanks

devimc:
@jodh-intel changes applied, thanks

devimc:
@egernst  ok, sounds good

virtcontainers/kata_agent.go
```diff
@@ -63,6 +65,7 @@ var (
 	shmDir               = "shm"
 	kataEphemeralDevType = "ephemeral"
 	ephemeralPath        = filepath.Join(kataGuestSandboxDir, kataEphemeralDevType)
+	grpcMaxDataSize      = int64(1024 * 1024)
```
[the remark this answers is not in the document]
jodh-intel: Good question! @devimc - is this just a restriction added to avoid a DoS or something?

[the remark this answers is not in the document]
sameo: In other words: Is this related to any gRPC limitations?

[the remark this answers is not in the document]
devimc: according with this https://godoc.org/google.golang.org/grpc#MaxRecvMsgSize the default size 4MB , I can use it but if the value changes to 1000MB probably we'll face DoS attacks

PR #1047, opened 2018-12-18
virtcontainers: fix sandbox startup sequence

### LINKED ISSUE #1020: Create container failed when Guest OS boot slowly
# Description of problem
I got error message below when use my custom Qemu version
``` bash
$ docker run --rm --name test -ti alpine uname -a                
docker: Error response from daemon: oci runtime error: Failed to check if grpc server is working: rpc error: code = Unavailabl
e desc = transport is closing.
```
There is a similar issue #979, but i replace qemu instead of kernel.

It's easy to reproduce even use original qemu-lite:
- modify `kernel_params = "boot_delay=100"` in `configuration.toml`
- create container again

# Reason
The `runtime` does not wait Guest OS kernel prepare `agent.channel.0`(the frontend of virtserialport for kata-agent).
[the correct sequence](https://www.plantuml.com/plantuml/svg/pLTDRniX43sNNp5o3fLMsYZjqOgiA5RJGnF5IQHUeXxmxiH6jc7Digdpxmlx1GkiRULKJglvU0oFny3Ub8fALUssX7mjDumNLD8TBBIDPQoWN2redZE-WkzNnFflnQ6pVxiYP2aECFb55ONdYYksmma1FGgB0kzCganmHHb7MGRkal9SemHskaAk0ciXnU4ZiAu_2fJlh1IIO8Fc5BOmAM5PaursfaNZ76GzuQuAu_VQEepxMM6f9d1nprPhTRkjCCXvV25VmZdhbTLpGWr6o6iwysYQGYQHAdnfsBcJOZSN_8sj2CqKUzSUdrT2XWOuXJcAP5OHOjLL-n0cj8UnE-PbNwTfD1romHlISlHG7CCPwsmrD8LQqhyVxD9STRNV9i4Afis9BiYH5RJ3cY1DPw9MnHJkQSMpjT5D7zY3x5rhLu9XPAo6A2kHKleFCmEahnSIyp-pjg92eXOUGc8KTrcAR9FqyLnem2s-gTvYHf9OSIQ9vrhIRAFrwjYDUWsGLyLCQtwgPIsKwH3r2GafqSV2XNzywV9g6c23akDs4EAamiNTtQqU0_yGfX4P9FLQBhCrvHhlyYe9Kk9CdC683NcJMeb3QaOfEaM46SwUUzvnJxDojxdqLFOcA8r-CepJEcXCndFimFobx5FdfTFcZmDcbSAdXyMS5dJ9jamnB6CfvfWqEH9L9Jde7FrP5eANUF9GENMUSwR6EfVVN1ei1wVhCR7ErBQzGO-9DCncYxpuHN_NHJ86MVKjrcQ5nytUcTseBKq3Sw_1hf61jHsJIs4lrc55XZMtf2DzFYBkS5upt2cjshkwSqQtqLuNJTVr-SV5Q0MagEgs9N4_Q5pTY6vLjiPiSyOkXMINka4yunxc1k4vCkUfee_Lsymgp5PuL7LxCKfxT9FYCkrMuDxgxae-JSaOu7zBpVYJ9xpR0vHZhQEvpERs0UJd7djicJ2REZhxwtKQl1NzsE0PHcxqPkdVXty0)

[the incorrent sequence](https://www.plantuml.com/plantuml/svg/xLTDRniX43sNNu6SWwMLDabxj28hZTNqq4Hn4gcNg0UyEx6Hrx1csTJ-zuNzWWNMTfCUEwVLCFCOXiU3lIua4RBSP0YDYpLbEH5aW-VAHnEQ4oRd94qfM-8lbyYCt_1Tw_zyYT22x_3e1v44FvLCqWsC45QM6nJyHeLCE9E4CX25DrmGbYeKpq-MmAJdpGNVxJtlQf-3UAC55mXgDAkmkKxnongjrM2SruEzh4VOb7xyLddxSTzBAEG8dztHvKhUP2LuEG_Fw9C_PxMoQavSWI7q4a-TDarm8e18UAwxSoluPiRPArqYaaZwfaRSlYBKT-1Zc8D8UXK-LbMr2w53EnYpOqxsLHm7qt7AMTsq51mKot72EXiEJN15wP_tPcaskvfldM08q-G45sJH2ZTcN3YEfxnYnGJVaP8bAysRBh03sThUjWJTaQ4QWbq9bD9zw1bm_78c8FqrRIhA1IZY0OuquyuBdgoZBfvn1PZ1g-mysgB8a3EAdA45ITQAhvPVirS3ELLC5USdYjPSQeMeJeZV4dKiRFY7vpQlQWDUWs2GzKAiL7nsUtkZh3UkxM9iW6gQuQXQpNco8amXdbz6NbAu5oVqn8Iy2iN5VdC6c_Gtwx5stsdOy4ZT0LjckgPsBbne9YKGRh2TExweOWuTdj-4lklij3pzjeEabF1uFv-Hd2neHYM58fIYZqoT8q2MWc6LepwBdBC2ZXumgyvJpjUGYhb2KsDPEAtUX5IgaS2UtWJ4izua9txUtNk1Z5vMTQCrMVxHC_TdQvMd5ZFxIcn53HlV8SBavfBjLwIxPfTqGFC3zFRdrMPFQVZUrfrIaWLrBvYkwdCFZEOA5hoi94p0jYTYhGMtAbb1iZvXb_nc5qegdc2BPnhXAJ3diQ8FrTk9dDQQhygyrIWL8pI3a8RP6eWFQDsA57eYl01WZVwcU0_zIeA2M4DppHHd9rkgQ0koOXmaKQeQ8FZUyWpSD_-lYNznJPWEUzBLijf-zTdiVlSn6G9ixIF7vGDy3DBno9FOV_jvA8Ukavh3C_D0TdCFFSPrc4aTdFtbAlR-9TnOxvcEhjLcgT_BFm00)

Maybe we need add event notify for `kata-agent`, and `runtime` wait the event passively instead of send `CheckRequest` actively。

BetaXOi:
Wait the event of agent open the virt serial port frontend before
send request to agent.

Fixes #1020

Signed-off-by: NingBo <ning.bo9@zte.com.cn>

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -584,6 +584,33 @@ func (k *kataAgent) setProxy(sandbox *Sandbox, proxy proxy, pid int, url string)
 	return nil
 }
 
+func (k *kataAgent) checkAgent(sandbox *Sandbox) error {
+	savedCtx := k.ctx
+	ctx, cancel := context.WithTimeout(k.ctx, checkRequestTimeout)
+	k.ctx = ctx
+	defer func() {
+		cancel()
+		k.ctx = savedCtx
+	}()
+
+	// wait agent is running
+	switch k.vmSocket.(type) {
+	case Socket:
+		if err := sandbox.hypervisor.waitAgent(ctx); err != nil {
+			k.Logger().WithError(err).Error("Failed to check if agent is running")
+			return err
+		}
+	default:
+	}
+
+	// check grpc server is serving
+	if err := k.check(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
 	span, _ := k.trace("startSandbox")
 	defer span.Finish()
@@ -604,8 +631,7 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
 		hostname = hostname[:maxHostnameLen]
 	}
 
-	// check grpc server is serving
-	if err = k.check(); err != nil {
+	if err = k.checkAgent(sandbox); err != nil {
 		return err
 	}
 
@@ -1474,8 +1500,6 @@ type reqFunc func(context.Context, interface{}, ...golangGrpc.CallOption) (inter
 func (k *kataAgent) installReqFunc(c *kataclient.AgentClient) {
 	k.reqHandlers = make(map[string]reqFunc)
 	k.reqHandlers["grpc.CheckRequest"] = func(ctx context.Context, req interface{}, opts ...golangGrpc.CallOption) (interface{}, error) {
-		ctx, cancel := context.WithTimeout(ctx, checkRequestTimeout)
-		defer cancel()
 		return k.client.Check(ctx, req.(*grpc.CheckRequest), opts...)
 	}
 	k.reqHandlers["grpc.ExecProcessRequest"] = func(ctx context.Context, req interface{}, opts ...golangGrpc.CallOption) (interface{}, error) {
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -7,6 +7,7 @@ package virtcontainers
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"math"
 	"os"
@@ -61,6 +62,7 @@ type qemu struct {
 	config HypervisorConfig
 
 	qmpMonitorCh qmpChannel
+	qmpEventCh   chan govmmQemu.QMPEvent
 
 	qemuConfig govmmQemu.Config
 
@@ -567,7 +569,12 @@ func (q *qemu) waitSandbox(timeout int) error {
 		return fmt.Errorf("Invalid timeout %ds", timeout)
 	}
 
-	cfg := govmmQemu.QMPConfig{Logger: newQMPLogger()}
+	// The buffer should be enough to receive events during the boot of the VM
+	eventCh := make(chan govmmQemu.QMPEvent, 1024)
+	cfg := govmmQemu.QMPConfig{
+		EventCh: eventCh,
+		Logger:  newQMPLogger(),
+	}
 
 	var qmp *govmmQemu.QMP
 	var disconnectCh chan struct{}
@@ -592,7 +599,7 @@ func (q *qemu) waitSandbox(timeout int) error {
 	}
 	q.qmpMonitorCh.qmp = qmp
 	q.qmpMonitorCh.disconn = disconnectCh
-	defer q.qmpShutdown()
+	q.qmpEventCh = eventCh
 
 	qemuMajorVersion = ver.Major
 	qemuMinorVersion = ver.Minor
@@ -1488,4 +1495,45 @@ func (q *qemu) cleanup() error {
 	q.fds = []*os.File{}
 
 	return nil
+}
+
+func (q *qemu) waitAgent(ctx context.Context) error {
+	type eventData struct {
+		ID   string `json:"id,omitempty"`
+		Open bool   `json:"open,omitempty"`
+	}
+
+	if q.qmpEventCh == nil {
+		q.Logger().Warnf("Skip waiting agent because QMP event channel has not been initialized")
+		return nil
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("timeout")
+		case event := <-q.qmpEventCh:
+			// see https://github.com/kata-containers/qemu/blob/qemu-lite-2.11.0/qapi/char.json#L520
+			if event.Name == "VSERPORT_CHANGE" {
+				eventDataJSON, err := json.Marshal(event.Data)
+				if err != nil {
+					q.Logger().Errorf("Marshal failed: %v", err)
+					continue
+				}
+
+				var data eventData
+				if err := json.Unmarshal(eventDataJSON, &data); err != nil {
+					q.Logger().Errorf("Unmarshal failed: %v", err)
+					continue
+				}
+
+				if data.ID == defaultKataDeviceID {
+					if data.Open {
+						q.Logger().Info("Wait agent event done")
+						return nil
+					}
+				}
+			}
+		}
+	}
 }
--- a/virtcontainers/fc.go
+++ b/virtcontainers/fc.go
@@ -709,3 +709,7 @@ func (fc *firecracker) getThreadIDs() (*threadIDs, error) {
 func (fc *firecracker) cleanup() error {
 	return nil
 }
+
+func (fc *firecracker) waitAgent(ctx context.Context) error {
+	return nil
+}
--- a/virtcontainers/hypervisor.go
+++ b/virtcontainers/hypervisor.go
@@ -607,4 +607,5 @@ type hypervisor interface {
 	hypervisorConfig() HypervisorConfig
 	getThreadIDs() (*threadIDs, error)
 	cleanup() error
+	waitAgent(ctx context.Context) error
 }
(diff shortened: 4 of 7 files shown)


virtcontainers/qemu.go
```diff
@@ -559,6 +559,28 @@ func (q *qemu) startSandbox() error {
...
 }
 
+func (q *qemu) waitAgent(eventCh chan govmmQemu.QMPEvent, timeout time.Duration) error {
+	for {
+		select {
+		case <-time.After(timeout):
+			return fmt.Errorf("Failed to wait agent (timeout %fs)", timeout.Seconds())
+		case ev := <-eventCh:
+			q.Logger().Infof("Receive event %s", ev)
+			if ev.Name == "VSERPORT_CHANGE" {
```
ydjainopensource: Have you considered adding this constant to govmm and calling it from there?

devimc: @BetaXOi great! I like this, but I agree with @ydjainopensource this event should be a constant

virtcontainers/qemu.go
```diff
@@ -610,7 +636,8 @@ func (q *qemu) waitSandbox(timeout int) error {
 		return err
 	}
 
-	return nil
+	timeoutDuration := time.Duration(timeout)*time.Second - time.Now().Sub(timeStart)
+	return q.waitAgent(eventCh, timeoutDuration)
```
sboeuf: Hi @ningbo9, I'm not sure about the issue you're trying to solve here. I have checked the description of `boot_delay` which says `Milliseconds to delay each printk during boot.`. And because every printk gets delayed, adding a simple `boot_delay=1` bumps by 3s the container boot time. If I do `boot_delay=2`, I get a bump around 6s. So clearly putting `boot_delay=100` (~ 5min) will exceed/reach any timeout we might have inside Kata Containers right now.

I know you took `boot_delay` as an example to reproduce the behavior you were seeing when moving to a slower hypervisor, but I'm wondering if we even want to support such cases where it takes ages to the VM to be created, hence the serial port to also be created?

bergwolf:
I'd like to have a single point to wait for the agent to be up, which is the agent's check() request ATM. I'm not sure the waiting in the PR makes sense because we don't just send any request through the serial port right after launching qemu. Instead of waiting, we would use the time to do other things at the runtime and also possibly launch the proxy daemon. So the patch will be adding unneeded delays for no obvious gains IMHO.

So at best if you do want to wait for the event, I think it should be done in agent.go `startSandbox()` after launching the proxy and before sending `check()` request.

BetaXOi:
> I'd like to have a single point to wait for the agent to be up, which is the agent's check() request ATM. I'm not sure the waiting in the PR makes sense because we don't just send any request through the serial port right after launching qemu. Instead of waiting, we would use the time to do other things at the runtime and also possibly launch the proxy daemon. So the patch will be adding unneeded delays for no obvious gains IMHO.
> 
> So at best if you do want to wait for the event, I think it should be done in agent.go `startSandbox()` after launching the proxy and before sending `check()` request. That should also fix #1020.

I think you are right, waiting event in `waitSandbox` maybe not suitable, i will update commit later.

BetaXOi:
> The patch is not completely wrong, but seems incomplete to me. Problem is, I don't know if we can consider the agent ready because the serial port is opened. And also what about the vsock use case?

@sboeuf the function name `waitAgent` and waiting event in `waitSandbox` maybe not suitable, i just want to wait the serial port has appeared already, otherwise frontend will not receive any data send to backend(send request via UNIX socket `kata-proxy` listened ).
My initial thought was that the serial port must has appeared when Guest OS start agent service, as @bergwolf said, this really does lead to long time than before.

virtcontainers/qemu.go
BetaXOi: Thank you, i will change this in next commit.

BetaXOi:
@sboeuf @bergwolf I have updated the commit, please help review.

virtcontainers/qemu.go
```diff
@@ -1476,3 +1483,39 @@ func (q *qemu) resizeVCPUs(reqVCPUs uint32) (currentVCPUs uint32, newVCPUs uint3
 	}
 	return currentVCPUs, newVCPUs, nil
 }
+
+func (q *qemu) waitAgent(timeout time.Duration) error {
```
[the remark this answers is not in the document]
bergwolf: And it worth noting with a comment that it can only be called after `waitSandbox` since otherwise we'll crash on nil `qmpEventCh`.

virtcontainers/kata_agent.go
```diff
@@ -593,6 +594,12 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
 		hostname = hostname[:maxHostnameLen]
 	}
 
+	// wait agent is running
+	if err = sandbox.hypervisor.waitAgent(waitAgentTimeout); err != nil {
```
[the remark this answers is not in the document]
BetaXOi: I have update commit again, combine it all toghther in `checkAgent` function

virtcontainers/qemu.go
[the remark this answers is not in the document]
BetaXOi: Check `vmSocket` type now, and call `waitAgent` for `Socket` type only.
Check if the `qmpEventCh` is valid in `waitAgent` now.

caoruidong:
@BetaXOi Please could you rebase?

BetaXOi:
@caoruidong @bergwolf Rebased on master.

virtcontainers/kata_agent.go
```diff
@@ -584,6 +584,33 @@ func (k *kataAgent) setProxy(sandbox *Sandbox, proxy proxy, pid int, url string)
...
+	defer func() {
+		cancel()
+		k.ctx = savedCtx
+	}()
+
+	// wait agent is running
+	switch k.vmSocket.(type) {
+	case Socket:
+		if err := sandbox.hypervisor.waitAgent(ctx); err != nil {
+			k.Logger().WithError(err).Error("Failed to check if agent is runing")
```
caoruidong: typo runing

BetaXOi: fixed

virtcontainers/kata_agent.go
```diff
@@ -584,6 +584,33 @@ func (k *kataAgent) setProxy(sandbox *Sandbox, proxy proxy, pid int, url string)
...
+	k.ctx = ctx
+	defer func() {
+		cancel()
+		k.ctx = savedCtx
+	}()
+
+	// wait agent is running
+	switch k.vmSocket.(type) {
+	case Socket:
+		if err := sandbox.hypervisor.waitAgent(ctx); err != nil {
```
sboeuf: By waiting for the agent only in case of `Socket`, which means it does not apply for VSOCK, we assume the agent knows about the hypervisor implementation. This is not okay, because that's the whole point about the hypervisor interface (to provide some abstraction). The whole timeout handling + decision to return immediately if the type is `Socket` should be handled by the hypervisor implementation directly.
@bergwolf WDYT?

jodh-intel: Any thoughts on this @bergwolf?

raravena80:
@BetaXOi ping, any updates? Thx!

virtcontainers/kata_agent.go
jodh-intel: It's worth pointing out here that the grpc client (used by `k.check()` below) will keep dialing the grpc server running in the agent until it gets a response.

grahamwhaley:
Hi @BetaXOi - please let us know if you will be revisting and fixing this PR, or it may have to be closed or 'put up for adoption' :-)
As a minimum it now needs a rebase as the master branch code has moved on...

PR #1052, opened 2018-12-18
virtcontainers: network: Use multiqueue flag only when appropriate

### LINKED ISSUE #1051: network: Don't use multiqueue flag if multiqueue is not supported
The multiqueue flag associated with the TUNTAP network device cannot be used if the number of queues indicates 0. When 0, this means the multiqueue is not supported, and we cannot use the according flag.

sboeuf:
The multiqueue flag associated with the TUNTAP network device cannot
be used if the number of queues indicates 0. When 0, this means the
multiqueue is not supported, and we cannot use the according flag.

Fixes #1051


mcastelino:
LGTM

Sorry I missed this in the initial review.

bergwolf:
LGTM

Issue #1053, opened 2018-12-19
vsock: Close vsock fd before returning context


mcastelino:
# Description of problem

The vsock fd opened when trying to find a free CID prevents firecracker from opening that vsock fd when launching the VM. QEMU can consume the vhost fd directly. By holding on to the vhost fd we prevent any race condition.

For firecracker we need to find a race free way to hold on to the CID without holding on to the vsock.

mcastelino:
@bergwolf @devimc moving the conversation here from https://github.com/kata-containers/runtime/pull/1054 to the issue.

One way to fix this issue is for kata to have some sort of systemlevel lock when we manage CIDs.

i.e. if we hold on to the lock till the VM consumes the CID. 

The CID lookup code should wait on the system level lock before it allocates new CIDs.
The lock is taken by the runtime in the small window between the runtime closing the fd and the firecracker VMM latching on to the CID.

devimc:
> One way to fix this issue is for kata to have some sort of systemlevel lock when we manage CIDs.

you won't see race conditions between kata-runtime processes, but there are  other processes running in the host (i.e xpra) that can request CIDs

bergwolf:
@mcastelino Another way is to extend FC API to accept an fd option when configuring vsock. We can launch FC similar to how we launch qemu and pass the vsock vhost fd to it. Then we tell FC which vhost fd owns the cid and it can just reuse it instead of opening the vsock vhost device again.

Then the current workflow works for FC as well as QEMU.

bergwolf:
> Another way is to extend FC API to accept an fd option when configuring vsock. 

@mcastelino Let's see how FC responds it -- https://github.com/firecracker-microvm/firecracker/issues/782

PR #1054, opened 2018-12-19
vsock potential bug: Close vsock fd

mcastelino:
The vsock fd opened when trying to find a free CID prevents
firecracker from opening that vsock fd.

Currently there is no other consumer of this vsockfd close it
and return nil.

Fixes: #1053

### DIFF
--- a/virtcontainers/utils/utils_linux.go
+++ b/virtcontainers/utils/utils_linux.go
@@ -76,14 +76,16 @@ func FindContextID() (*os.File, uint64, error) {
 	// Looking for the first available context ID.
 	for cid := contextID; cid <= maxUInt; cid++ {
 		if err := ioctlFunc(vsockFd.Fd(), ioctlVhostVsockSetGuestCid, cid); err == nil {
-			return vsockFd, cid, nil
+			vsockFd.Close()
+			return nil, cid, nil
 		}
 	}
 
 	// Last chance to get a free context ID.
 	for cid := contextID - 1; cid >= firstContextID; cid-- {
 		if err := ioctlFunc(vsockFd.Fd(), ioctlVhostVsockSetGuestCid, cid); err == nil {
-			return vsockFd, cid, nil
+			vsockFd.Close()
+			return nil, cid, nil
 		}
 	}
 


mcastelino:
@devimc is there a consumer for this fd?

virtcontainers/utils/utils_linux.go
```diff
@@ -76,14 +76,16 @@ func FindContextID() (*os.File, uint64, error) {
 	// Looking for the first available context ID.
 	for cid := contextID; cid <= maxUInt; cid++ {
 		if err := ioctlFunc(vsockFd.Fd(), ioctlVhostVsockSetGuestCid, cid); err == nil {
-			return vsockFd, cid, nil
+			vsockFd.Close()
```
lifupan: Is it better combine those three fd close into a defer function?

mcastelino: Yes. I still need to see in the QEMU code if it uses the fds. I may need to close the fd's at the point where I hand it off to firecracker to prevent a race condition. However closing it will trigger a race.

bergwolf:
@mcastelino The FC API doesn't ask for the vsock vhostfd, while the kata vsock cid generator relies on passing it to the hypervisor to make sure the generated cid is not reused. For simplicity, you can just close the vhostfd in fc.go right before sending it to the FC server, where the same guest cid will be taken by FC. There is a small race window that the cid might be taken by others between closing the vhostfd and FC taking the cid, but we can live with it for now and it is the best we can do without FC API to inherent the vsock vhostfd.

A more elegant way of solving the problem is to introduce VHOST_VSOCK_SET_GUEST_CID_ANY ioctl that would assign any available cid to the opened vsock vhostfd. Then we can get rid of the user space vsock cid generator. @stefanha is there any reason kernel doesn't want to assign the guest cid on its own?

mcastelino:
> @mcastelino The FC API doesn't ask for the vsock vhostfd, while the kata vsock cid generator relies on passing it to the hypervisor to make sure the generated cid is not reused. For simplicity, you can just close the vhostfd in fc.go right before sending it to the FC server, where the same guest cid will be taken by FC. There is a small race window that the cid might be taken by others between closing the vhostfd and FC taking the cid, but we can live with it for now and it is the best we can do without FC API to inherent the vsock vhostfd.

@bergwolf another option would be for firecracker to provide an option where it picks the CID. 

In a later change which I am testing https://github.com/mcastelino/runtime-1/commit/de4b201da31b6ceb6d52048f81268d7bcf4df5af I can reduce the race window a little further.

We can reduce it even further by closing it just prior to instance start but that is probably the best we can do for now.

I will look at the firecracker code to see if we can make it choose the CID and hold on to it, which seems to be best course of action.

mcastelino:
> @mcastelino The vsock vhostfd is passed to qemu and is supposed to be closed by the qemu process. That said, we might have leaked it on error path since we don't have a cleanup function for it.

Yes. Let me add issue to track the potential leak of the vsock fd.

Issue #1055, opened 2018-12-19
It will take much more time to launch kata when both vsock and template enabled


lifupan:
# Description of problem

vsock conflicts with factory, when both of them are enabled, kata will try
to create a new vm template, then use this template to launch another qemu
process. But this template is useless when try to launch other kata containers.
Thus it's better to disable template when vsock enabled to save a lot of boot time.

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #1056, opened 2018-12-19
katautils: check config factory/template and vsock

lifupan:
Vsock conflicts with factory, when both of them are enabled,
kata will try to create a new vm template which is useless,
thus it's better to disable template to save a lot of time.

Fixes: #1055

### DIFF
--- a/pkg/katautils/config.go
+++ b/pkg/katautils/config.go
@@ -737,8 +737,14 @@ func checkNetNsConfig(config oci.RuntimeConfig) error {
 
 // checkFactoryConfig ensures the VM factory configuration is valid.
 func checkFactoryConfig(config oci.RuntimeConfig) error {
-	if config.FactoryConfig.Template && config.HypervisorConfig.InitrdPath == "" {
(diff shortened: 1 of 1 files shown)


pkg/katautils/config.go
```diff
@@ -705,6 +705,14 @@ func updateConfig(configPath string, tomlConf tomlConfig, config *oci.RuntimeCon
 		return err
 	}
 
+	// vsock conflicts with factory, when both of them are enabled, kata will try
+	// to create a new vm template which is useless, thus it's better to disable
+	// template to save a lot of time.
+	if config.HypervisorConfig.UseVSock && config.FactoryConfig.Template {
+		kataUtilsLogger.Warn("vsock conflicts with factory, configure to not use it")
+		config.FactoryConfig.Template = false
```
bergwolf: Should we just post an error and bail out here? Just so that we don't make users think template+vsock is working.

bergwolf:
LGTM! Thanks @lifupan !

jodh-intel:
lgtm

PR #1058, opened 2018-12-19
sandbox: cleanup sandbox if creation failed

bergwolf:
This includes cleaning up the sandbox on disk resources, and closing fds opened when preparing the hypervisor.

### DIFF
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -74,6 +74,13 @@ func createSandboxFromConfig(ctx context.Context, sandboxConfig SandboxConfig, f
 		return nil, err
 	}
 
+	// cleanup sandbox resources in case of any failure
+	defer func() {
+		if err != nil {
+			s.Delete()
+		}
+	}()
+
 	// Create the sandbox network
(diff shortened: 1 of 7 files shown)


virtcontainers/api.go
```diff
@@ -74,6 +74,13 @@ func createSandboxFromConfig(ctx context.Context, sandboxConfig SandboxConfig, f
 		return nil, err
 	}
 
+	// cleanup sandbox resources in case of any failure
+	defer func() {
+		if err != nil {
+			s.Delete()
```
lifupan: I'm afraid is it reasonable to do sandbox.Delete() if it failed in fetching an existing sandbox instead of creating a new one.

bergwolf: This is the create sandbox API.

sameo:
Can we squash the 2 commits in one? I guess the second one fixes a build issue introduced by the firt one...

bergwolf:
@sameo The first one builds without issues. The second one was spotted when testing the first one by hand. Anyway, I've squashed them into one since they kind of fix the same problem.

PR #1063, opened 2018-12-19
virtcontainers: cgroups: Don't error if no thread ID

### LINKED ISSUE #1062: cgroups: Don't error if no thread IDs available
In case the hypervisor implementation does not return any thread ID, this should not issue any error since there's simply nothing to constraint.

sboeuf:
In case the hypervisor implementation does not return any thread
ID, this should not issue any error since there is simply nothing
to constraint.

Fixes #1062

### DIFF
--- a/virtcontainers/cgroups.go
+++ b/virtcontainers/cgroups.go
@@ -100,8 +100,14 @@ func (s *Sandbox) applyCPUCgroup(rc *specs.LinuxResources) error {
 	// when new container joins, new CPU could be hotplugged, so we
 	// have to query fresh vcpu info from hypervisor for every time.
 	tids, err := s.hypervisor.getThreadIDs()
-	if err != nil || tids == nil {
+	if err != nil {
 		return fmt.Errorf("failed to get thread ids from hypervisor: %v", err)
+	}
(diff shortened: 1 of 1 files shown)


virtcontainers/cgroups.go
```diff
@@ -100,9 +100,15 @@ func (s *Sandbox) applyCPUCgroup(rc *specs.LinuxResources) error {
 	// when new container joins, new CPU could be hotplugged, so we
 	// have to query fresh vcpu info from hypervisor for every time.
 	tids, err := s.hypervisor.getThreadIDs()
-	if err != nil || tids == nil {
+	if err != nil {
 		return fmt.Errorf("failed to get thread ids from hypervisor: %v", err)
 	}
+	if tids == nil {
+		// If there's no tid returned from the hypervisor, this is not
+		// a bug. It simply means there is nothing to constraint, hence
```
egernst: s/constraint/constrain

Looks good otherwise (do same change in commit message)

sboeuf: ok!

sboeuf: updated!

Issue #1059, opened 2018-12-19
virtcontainers code base is monolithic


sameo:
## Description of problem

virtcontainers was built and designed around the idea of defining virtualized container modules and use them to configure a hardware virtualization based container runtime. Those modules were:

- Hypervisors
- Shims
- Agents
- Proxies
- Storage
- Networking

As a result, one would expect the virtcontainers code to be logically split into different internal packages implementing those modules. Instead, it is mostly flat and monolithic. Most of those modules are living under the `virtcontainers` package.

### Architecture flexibility and cleanliness
Living under one single package allows us to build hard dependencies between those modules. We can avoid building clean and agnostic APIs between those modules because we're part of the same golang package.
For example, there is no reason why the hypervisor code should depend on some state storage (`resourceStorage`) definitions. Hypervisor implementations could call into a separated state storage API if they feel they need to, but the hypervisor interface should not depend on it.

In general, we want to make the virtcontainers simpler and more flexible. While enabling the Firecracker hypervisor, we realized how difficult it could be to support a VMM that does not follow the QEMU lifecycle. Forcing ourselves into cleanly separating the code into independent packages would provide the virtcontainers code with more flexibility for supporting those use cases.

### Code maintenance
Building independent internal packages would make the code easier to maintain. It would break the number of dependencies between different modules and would allow for a more fine grained maintenance between modules and their test cases.

This would make the code also easier to approach. Getting back to it after roughly a 6 months break showed how difficult it could be for a first comer to understand and contribute to this piece of software. Modularizing it will hopefully make the virtcontainers learning curve a little less steep.

### Complexity reduction
As our support matrix has grown over the past months, the code complexity has grown as well. Splitting it into separate modules should make it less complex and better prepared for the 2.0 overall simplification.

## Proposed plan

This is a long term task, and at the same time virtcontainers is a stable code base. We want to make iterative changes to it and add move code into their own modules, one module at a time.

The plan below will get updated as we move forward.
I propose that our first iteration should create the following internal packages. We will add issues to track to each of those items.

* [x] State storage: Rework the current `resourceStorage` interface into its own virtcontainers state manager. (#1099) PR #1066 
* [x] Types: Group shared internal types into their own single package. (#1095)
* [x] Hypervisor assets: Kernel, initrd, rootfs, etc
* [ ] Hypervisor: The hypervisor interface (#1229) PR #1208
* [ ] Networking  endpoints: The `Endpoint` interface (#1208)
* [ ] Agent
[...]

PR #1066, opened 2018-12-20
Virtcontainers store

sameo:
## Description and goals

This is a rewrite of the `resource_storage` and its `filesystem_resource_storage` backend, with the following goals:

* Provide a neutral store API for virtcontainers. The API and the physical backend should be decoupled.
* Provide a `filesystem` backend implementation and make it the default one for virtcontainers.
* Provide a separate `store` package for the core virtcontainers code to consume. Since several components are depending on a store API, this will break these dependencies and allow us to move #1059 further away.

## Shortcomings

There are a **lot** of places in the virtcontainers code where we assume that our store backend will be a local filesystem one. This is practically just right but ideally we should not make any of those assumptions. This PR does not completely remove those assumptions, but the `store` package top level API is neutral enough so that we could eventually remove them from the core virtcontainers code.


egernst:
Happy new year @sameo 

I'm looking forward to seeing more dialogue on this PR w.r.t. the store itself.

Does it make sense to create a seperate PR for the commits which make use of the 'internal types' pkg?  That should be ready to go as is?  Would like to at least get that in for 1.5 (scheduled week of 1/23).  WDYT?

sameo:
@egernst Bonne Année a toi aussi!

I will extract the `types` package out of this PR and create a separate one today.

sameo:
> @egernst Bonne Année a toi aussi!
> 
> I will extract the `types` package out of this PR and create a separate one today.

@egernst See #1096

sameo:
@egernst @sboeuf @jodh-intel @bergwolf @WeiZhang555 
So I can start docker containers cleanly with this new storage implementation.
CI is probably going to shout at me, but I think this is ready to be reviewed.

sameo:
Unit tests pass locally...

virtcontainers/store/filesystem.go
```diff
@@ -42,13 +128,53 @@ func (f *filesystem) trace(name string) (opentracing.Span, context.Context) {
...
+
+	// Root directory
+	if err := os.MkdirAll(f.path, DirMode); err != nil {
+		return err
+	}
+
+	// Lock
+	lockPath := filepath.Join(f.path, LockFile)
+
+	lockFile, err := os.Create(lockPath)
```
sboeuf: Just trying to follow here. We're creating a lock file for every store, and we have one store per type of root path such as `/var/run/` or `var/lib`, right?

I'm just concerned that we might create too many lock files if we cannot control that too many stores have been created.

virtcontainers/store/backend.go
```diff
@@ -48,4 +48,5 @@ type backend interface {
 	delete()
 	load(item Item, data interface{}) error
 	store(item Item, data interface{}) error
+	raw(id string) (string, error)
```
sboeuf: Could you please highlight in the commit message and through some comments in the code, some concrete use cases for this new API.

virtcontainers/store/backend.go
```diff
@@ -49,4 +49,6 @@ type backend interface {
 	load(item Item, data interface{}) error
 	store(item Item, data interface{}) error
 	raw(id string) (string, error)
+	lock(item Item, exclusive bool) (string, error)
```
sboeuf: I don't get why we need those new locking API. If I followed until then, we have any `Load()` performing a read only lock, and `Store()` performing a read/write lock, which looks like it should be enough. Also, from the Kata API perspective, we have the locking of the pod, meaning we cannot access two containers of the same pod concurrently.

Ohh maybe I understand now, are you trying to do the pod locking through this API? And you're doing this because the lock file has to be created somewhere... It's a tricky one because it's not natural to have the locking being handled through the storage, but I guess that's how it was kind of working already.

Unless we share the global path for pods (in `types` maybe), and we could create a separate `lock/` package.

virtcontainers/store/vc.go
```diff
@@ -126,6 +126,21 @@ func (s *VCStore) Raw(id string) (string, error) {
...
 }
 
+// Lock takes an exclusive lock on the virtcontainers state Lock item.
+func (s *VCStore) Lock() (string, error) {
+	return s.state.ItemLock(Lock, false)
+}
+
+// RLock takes a shared lock on the virtcontainers state Lock item.
+func (s *VCStore) RLock() (string, error) {
+	return s.state.ItemLock(Lock, true)
```
sboeuf: Should be
```go
return s.state.ItemLock(Lock, false)
```

virtcontainers/store/vc.go
```diff
@@ -126,6 +126,21 @@ func (s *VCStore) Raw(id string) (string, error) {
 	return s.state.Raw(id)
 }
 
+// Lock takes an exclusive lock on the virtcontainers state Lock item.
+func (s *VCStore) Lock() (string, error) {
+	return s.state.ItemLock(Lock, false)
```
sboeuf: Should be
```go
return s.state.ItemLock(Lock, true)
```

virtcontainers/filesystem_resource_storage.go
```diff
@@ -1,873 +0,0 @@
...
-		return err
-	}
-
-	var tempDevices []api.Device
-	for _, d := range typedDevices {
-		l := fs.Logger().WithField("device-type", d.Type)
-		l.Info("Device type found")
-
-		switch d.Type {
-		case string(config.DeviceVFIO):
```
sboeuf: Where have you moved specific fetch implementation related to devices? I could not find it in `virtcontainers/store/filesystem.go`

virtcontainers/store/manager.go
```diff
@@ -0,0 +1,154 @@
...
+	"context"
+	"net/url"
+	"sync"
+
+	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/sirupsen/logrus"
+)
+
+// Item represents a virtcontainers items that will be managed through the store.
+type Item uint8
```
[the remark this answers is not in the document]
sameo: I want it to be a specific type. Are you suggesting this should be a string instead?

virtcontainers/store/manager.go
```diff
@@ -132,18 +132,29 @@ func (m *manager) findStore(root string) *Store {
 
 // New will create a new virtcontainers Store.
```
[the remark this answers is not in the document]
sameo: Fixed.

virtcontainers/store/backend.go
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2018 Intel Corporation
```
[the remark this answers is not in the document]
sameo: Fixed.

virtcontainers/store/filesystem.go
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) 2018 Intel Corporation
```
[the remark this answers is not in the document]
sameo: Fixed.

virtcontainers/store/backend.go
sameo: I did. Please let me know if this looks better.

virtcontainers/store/vc.go
sameo: Ooops. Fixed, thanks.

virtcontainers/store/vc.go
sameo: Fixed as well, thanks.

virtcontainers/filesystem_resource_storage.go
sameo: Right, I'm fixing that now...

sameo: Fixed now.

sameo: I need to add a lot more unit tests to `vc_test.go`

virtcontainers/store/manager.go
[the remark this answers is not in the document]
sboeuf: No, no, I was just thinking this could be a simple `uint`, but doesn't matter.

virtcontainers/store/manager.go
```diff
@@ -150,6 +152,18 @@ func New(ctx context.Context, root string) (*Store, error) {
...
 
+	backend, err := newBackend(s.scheme)
+	if err != nil {
+		return nil, err
+	}
+
+	s.backend = backend
+
+	// Create new backend
+	if err := s.backend.new(ctx, s.path, s.host); err != nil {
```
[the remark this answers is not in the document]
sboeuf: @sameo did you see this comment?

virtcontainers/store/backend.go
sboeuf: Yes it helps a lot to understand :)
IMO, you should even mention
```
// For example the Firecracker code uses this API to create temp
// files under the sandbox state root path, and uses them as block
// driver backend placeholder.
```
from the commit message.

virtcontainers/store/backend.go
sboeuf: @sameo did you miss this comment?

WeiZhang555:
Hi @sameo , I love the idea, and actually I was trying to do same thing as you in https://github.com/kata-containers/runtime/pull/883 in a different way.

I checked your implementation, one thing I think we can improve is, we don't need to specify "item" in Store implementation.

The internal API of PR #883 is:
```
// PersistDriver is interface describing operations to save/restore persist data
	type PersistDriver interface {
		Dump() error
		Restore(sid string) error
		Destroy() error
		GetStates() (*SandboxState, map[string]ContainerState, error)
		RegisterHook(name string, f SetFunc)
	}
```

First, we should have a baselined structure "SandboxState" and "ContainerState", these two structure are related to kata ABI compatibility.

Second, container internal data should convert to "SandboxState" and "ContainerState" , then call the `Dump()` to persist in a chosen way(e.g. local fs), use "Dump(sandboxID)" to restore. use `Destroy()` to remove from persist storage.  "SandboxState" and "ContainerState" are only interfaces between kata VC implementation and persist storage drivers.

Your implementation keeps "item" (state, device etc) and didn't remove them. I think this makes the storage implementation still too complicated.

sameo:
Ok, I have unit tests *really* passing now locally.

sameo:
> Hi @sameo , I love the idea, and actually I was trying to do same thing as you in #883 in a different way.

Cool, thanks

> I checked your implementation, one thing I think we can improve is, we don't need to specify "item" in Store implementation.

Do you mean that I should have implemented a set of APIs for each item we want to load/restore?

```Go
func (s *store) LoadState() (types.State, error) {}
func (s *store) StoreState(state types.State) error {}
func (s *store) LoadConfig() (types.Config, error) {}
func (s *store) StoreConfig(types.Config) error {}
[...]
```

FWIW, I started doing this, see for example vc.go `Load/StoreDevices`, but I can only do that for exported types (I don't want to include a `virtcontainers` dependency to `store`). So this `Item`-less API will grow as we export virtcontainers types into the `types` package.

WeiZhang555:
@sameo 

```
func (s *store) LoadState() (types.State, error) {}
func (s *store) StoreState(state types.State) error {}
func (s *store) LoadConfig() (types.Config, error) {}
func (s *store) StoreConfig(types.Config) error {}
[...]
```

Nope, I mean we can make it much more simpler. I suggest we use this kind of API

```
func (s *store) Load() (types.Everything, error) {}
func (s *store) Store(types.Everything) error {}
```

No `Item` at all.

amshinde:
@sameo Looks like there are merge conflicts that need to be resolved here. Can you take a look.
The CI is failing too.

virtcontainers/store/manager.go
```diff
@@ -0,0 +1,268 @@
...
+type manager struct {
+	sync.RWMutex
+	stores map[string]*Store
+}
+
+var stores = &manager{stores: make(map[string]*Store)}
+
+func (m *manager) addStore(s *Store) (err error) {
+	if s == nil {
+		return nil
```
jodh-intel: Shouldn't this be an error scenario?

virtcontainers/store/manager.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	m.Lock()
+	defer m.Unlock()
+
+	if m.stores[s.root] != nil {
```
jodh-intel: What if `s.root == ""`?

virtcontainers/store/manager.go
```diff
@@ -0,0 +1,268 @@
...
+
+	return ""
+}
+
+// Store is an opaque structure representing a virtcontainers Store.
+type Store struct {
+	sync.RWMutex
+	ctx context.Context
+
+	root   string
```
jodh-intel: This seems to be the mandatory key for a store. But why not just call it `name`?

virtcontainers/store/manager.go
```diff
@@ -0,0 +1,268 @@
...
+	scheme string
```
jodh-intel: I'm not clear on what these remaining fields are for? If they are optional / for the backend to decide how they should be used and what they mean, why can't they be removed and put into each `backend` as required?

virtcontainers/store/manager.go
```diff
@@ -0,0 +1,268 @@
...
+	defer m.Unlock()
+
+	delete(m.stores, root)
+}
+
+func (m *manager) findStore(root string) *Store {
+	m.RLock()
+	defer m.RUnlock()
+
+	if s := m.stores[root]; s != nil {
```
jodh-intel: Can't this be just?:

```go
return m.stores[root]
```

virtcontainers/store/backend.go
```diff
@@ -0,0 +1,58 @@
...
+	case filesystemBackend:
+		return &filesystem{}, nil
+	}
+
+	return nil, fmt.Errorf("Unsupported scheme %s", scheme)
+}
+
+type backend interface {
+	new(ctx context.Context, path string, host string) error
+	delete()
```
jodh-intel: I think this should return an `error` as although the FS backend may not need it, other backends might.

virtcontainers/store/filesystem.go
```diff
@@ -0,0 +1,310 @@
...
+	f.rawPath = filepath.Join(f.path, "raw")
+	f.lockTokens = make(map[string]*os.File)
+
+	f.logger().Debugf("New filesystem store backend for %s", path)
+
+	return f.initialize()
+}
+
+func (f *filesystem) delete() {
+	os.RemoveAll(f.path)
```
jodh-intel: This call return an error, so why not return it?

virtcontainers/store/filesystem.go
```diff
@@ -0,0 +1,310 @@
+// Copyright (c) 2019 Intel Corporation
```
jodh-intel: It would be clearer if this was called `virtcontainers/store/filesystem_backend.go`.

sameo:
@amshinde I will fix both very soon, thanks.

virtcontainers/store/filesystem.go
```diff
@@ -0,0 +1,310 @@
...
+		file, err := os.Create(filePath)
+		if err != nil {
+			return "", err
+		}
+
+		return filesystemScheme + "://" + file.Name(), nil
+	}
+
+	// Generate a random temporary file.
+	file, err := ioutil.TempFile(f.rawPath, "raw-")
```
jodh-intel: `file` not closed.

sameo:
@WeiZhang555 This PR does not intend to change the Kata storage layout and simplify it the way you did it with your PR. I think your PR is needed but since it breaks backward compatibility, it will only be mergeable with 2.0. Instead my PR aims at moving the whole storage code into its own package without breaking backward compatibility.
We need this because several virtcontainers components rely on the storage code and can't be moved into their own package before the storage one is.

Once this PR is merged, we can work on simplifying and improving the storage layout from the new storage package.

virtcontainers/store/filesystem.go
jodh-intel: Maybe `fs_backend.go` for brevity.

WeiZhang555:
@sameo I think that makes a lot of sense. Please move forward and let's merge it when it's good.

sameo:
All, I will address the pending comments and fix the CI early next week. I have prepared an additional PR on top of that one to show how we could start moving large pieces of code into their own packages.
See for example the hypervisor package creation at #1208

virtcontainers/store/filesystem.go
sameo: Done.

virtcontainers/store/filesystem.go
sameo: Renamed to `filesystem_backend.go`, thanks for the suggestion.

virtcontainers/store/filesystem.go
sameo: Renamed to `filesystem_backend.go`, thanks.

virtcontainers/store/backend.go
sameo: You're right, I changed that too.

virtcontainers/store/manager.go
sameo: Indeed, fixed.

virtcontainers/store/manager.go
sameo: I renamed `root` to `url`.

virtcontainers/store/manager.go
sameo: It is now, thanks.

virtcontainers/store/manager.go
sameo: This is an error as well, fixed. And unit tested.

sameo:
Basic CI passes, let's retest.

sameo:
I just fixed a network NS leak caused by a typo in the Store conversion patch.

sameo:
@sboeuf @WeiZhang555 @bergwolf @jodh-intel The CI passes now.

WeiZhang555:
LGTM

Let's retest it again before it can be merged.

/retest

sameo:
@WeiZhang555 The fc and ARM CI are broken, as expected :-/

PR #1069, opened 2018-12-21
mounts: Ignore existing mounts if they cannot be honored

### LINKED ISSUE #1068: mounts: Ignore existing mounts if they cannot be honored
In case we use an hypervisor that cannot support filesystem sharing, we copy files over to the VM rootfs through the gRPC protocol.
This is a nice workaround, but it only works with regular files, which means no device file, no socket file, no directory, etc... can be sent this way.
That's why we should simply accept the limitation in this use case, and ignore any non-regular file.

sboeuf:
In case we use an hypervisor that cannot support filesystem sharing,
we copy files over to the VM rootfs through the gRPC protocol. This
is a nice workaround, but it only works with regular files, which
means no device file, no socket file, no directory, etc... can be
sent this way.

This is a limitation that we accept here, by simply ignoring those
non-regular files.

Fixes #1068

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -431,10 +431,10 @@ func (c *Container) createContainersDirs() error {
 	return nil
 }
 
-func (c *Container) shareFiles(m Mount, idx int, hostSharedDir, guestSharedDir string) (string, error) {
+func (c *Container) shareFiles(m Mount, idx int, hostSharedDir, guestSharedDir string) (string, bool, error) {
 	randBytes, err := utils.GenerateRandomBytes(8)
 	if err != nil {
-		return "", err
+		return "", false, err
 	}
 
 	filename := fmt.Sprintf("%s-%s-%s", c.id, hex.EncodeToString(randBytes), filepath.Base(m.Destination))
@@ -445,20 +445,35 @@ func (c *Container) shareFiles(m Mount, idx int, hostSharedDir, guestSharedDir s
 	caps := c.sandbox.hypervisor.capabilities()
 	if !caps.isFsSharingSupported() {
 		c.Logger().Debug("filesystem sharing is not supported, files will be copied")
+
+		fileInfo, err := os.Stat(m.Source)
+		if err != nil {
+			return "", false, err
+		}
+
+		// Ignore the mount if this is not a regular file (excludes
+		// directory, socket, device, ...) as it cannot be handled by
+		// a simple copy. But this should not be treated as an error,
+		// only as a limitation.
+		if !fileInfo.Mode().IsRegular() {
+			c.Logger().WithField("ignored-file", m.Source).Debug("Ignoring non-regular file as FS sharing not supported")
+			return "", true, nil
+		}
+
 		if err := c.sandbox.agent.copyFile(m.Source, guestDest); err != nil {
-			return "", err
+			return "", false, err
 		}
 	} else {
 		// These mounts are created in the shared dir
 		mountDest := filepath.Join(hostSharedDir, c.sandbox.id, filename)
 		if err := bindMount(c.ctx, m.Source, mountDest, false); err != nil {
-			return "", err
+			return "", false, err
 		}
 		// Save HostPath mount value into the mount list of the container.
 		c.mounts[idx].HostPath = mountDest
 	}
 
-	return guestDest, nil
+	return guestDest, false, nil
 }
 
 // mountSharedDirMounts handles bind-mounts by bindmounting to the host shared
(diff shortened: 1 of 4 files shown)


sboeuf:
/cc @mcastelino @egernst @bergwolf 
That's the fix to get k8s running with Kata+FC :smiley:

virtcontainers/container.go
```diff
@@ -445,29 +445,45 @@ func (c *Container) shareFiles(m Mount, idx int, hostSharedDir, guestSharedDir s
...
+		if err != nil {
+			return "", false, err
+		}
+
+		// Ignore the mount if this is not a regular file (excludes
+		// directory, socket, device, ...) as it cannot be handled by
+		// a simple copy. But this should not be treated as an error,
+		// only as a limitation.
+		if !fileInfo.Mode().IsRegular() {
+			c.Logger().Debug("Ignore (%s), fs sharing not supported and not a regular file", m.Source)
```
jodh-intel: Nit: this and other log calls could be converted into structured format:

```go
c.Logger().WithField("ignored-file", m.Source).Debug("Ignoring non-regular file as FS sharing not supported")
```

bergwolf: And it is causing CI failure:
```
# github.com/kata-containers/runtime/virtcontainers
virtcontainers/container.go:459: Entry.Debug call has possible formatting directive %s
FAIL	github.com/kata-containers/runtime/virtcontainers [build failed]
Makefile:421: recipe for target 'go-test' failed
```

egernst: I fixed it for him and repushed.

virtcontainers/kata_agent.go
```diff
@@ -719,6 +719,30 @@ func (k *kataAgent) replaceOCIMountSource(spec *specs.Spec, guestMounts []Mount)
...
 }
 
+func (k *kataAgent) removeIgnoredOCIMount(spec *specs.Spec, ignoredMounts []Mount) error {
+	var mounts []specs.Mount
+
+	for _, m := range spec.Mounts {
+		found := false
+		for _, ignoredMount := range ignoredMounts {
+			if ignoredMount.Source == m.Source {
+				k.Logger().Debugf("Removing OCI mount (%s)", m.Source)
```
jodh-intel: Nit: this one could be converted to a structured format too.

egernst: ooooooh.  ok, fixed!

jodh-intel: Ta! ;)

virtcontainers/container.go
```diff
@@ -431,10 +431,10 @@ func (c *Container) createContainersDirs() error {
 	return nil
 }
 
-func (c *Container) shareFiles(m Mount, idx int, hostSharedDir, guestSharedDir string) (string, error) {
+func (c *Container) shareFiles(m Mount, idx int, hostSharedDir, guestSharedDir string) (string, bool, error) {
```
mcastelino: @sboeuf it would be good to document what the return values actually mean or give them expressive names.

virtcontainers/container.go
```diff
@@ -445,29 +445,45 @@ func (c *Container) shareFiles(m Mount, idx int, hostSharedDir, guestSharedDir s
...
+
+		// Ignore the mount if this is not a regular file (excludes
+		// directory, socket, device, ...) as it cannot be handled by
+		// a simple copy. But this should not be treated as an error,
+		// only as a limitation.
+		if !fileInfo.Mode().IsRegular() {
+			c.Logger().WithField("ignored-file", m.Source).Debug("Ignoring non-regular file as FS sharing not supported")
+			return "", true, nil
+		}
+
```
mcastelino: @sboeuf we need to monitor these files on the host side to ensure that if they get updated we push the changes in. Maybe raise an issue

virtcontainers/container.go
```diff
@@ -445,29 +445,45 @@ func (c *Container) shareFiles(m Mount, idx int, hostSharedDir, guestSharedDir s
...
+	return guestDest, false, nil
 }
 
 // mountSharedDirMounts handles bind-mounts by bindmounting to the host shared
 // directory which is mounted through 9pfs in the VM.
 // It also updates the container mount list with the HostPath info, and store
 // container mounts to the storage. This way, we will have the HostPath info
 // available when we will need to unmount those mounts.
-func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) ([]Mount, error) {
+func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) ([]Mount, []Mount, error) {
```
mcastelino: The comments need to be updated to reflect the new logic and that the ignored mounts actually means.
Also we special case /dev/shm today.

virtcontainers/kata_agent.go
```diff
@@ -1769,6 +1798,12 @@ func (k *kataAgent) copyFile(src, dst string) error {
 		Gid:      int32(st.Gid),
 	}
 
+	// Handle the special case where the file is empty
```
mcastelino: @sboeuf why was this needed?

sboeuf: Because the file was not copied at all when the size was 0, because of the logic below.

virtcontainers/container.go
egernst: (open issue, and mark it as known limitation)

raravena80:
@sboeuf  Q. Will this work with Kubernetes `ConfigMaps`?

mcastelino:
> @sboeuf Q. Will this work with Kubernetes `ConfigMaps`?
> 
> Meaning the `ConfigMap` will be copied to the firecracker VM?

@raravena80 it will not. I am opening an issue with details the limitations.
Today we copy files only in the case of pure files.

Configmap, Secrets. Tokens etc as shared as directories with keys set to file names and values the contents. We need to enhance our logic to handle these as a next step.

I am putting together a proposal for the same on special casing to handle a few critical features.

mcastelino:
@raravena80 https://github.com/kata-containers/runtime/issues/1071

PR #1075, opened 2018-12-26
fix: fail to do clean without named return

Ace-Tang:
```
defer func() {
    if err != nil {

    }
}
```
when returned variable is not named, if err is re-defined in if
condition, like `if err := s.getAndStoreGuestDetails(); err != nil`
in `createSandboxFromConfig()`, defer function can not catch error.
So it better to make returned err named.
Other change in virtcontainers/sandbox.go and virtcontainers/vm.go
is not must, but make more sense.

### DIFF
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -62,11 +62,9 @@ func CreateSandbox(ctx context.Context, sandboxConfig SandboxConfig, factory Fac
 	return s, err
 }
 
-func createSandboxFromConfig(ctx context.Context, sandboxConfig SandboxConfig, factory Factory) (*Sandbox, error) {
+func createSandboxFromConfig(ctx context.Context, sandboxConfig SandboxConfig, factory Factory) (_ *Sandbox, err error) {
 	span, ctx := trace(ctx, "createSandboxFromConfig")
 	defer span.Finish()
-
-	var err error
 
 	// Create the sandbox.
 	s, err := createSandbox(ctx, sandboxConfig, factory)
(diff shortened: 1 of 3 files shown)


caoruidong:
Nice catch! This is a real bug. But we need to replace := with = in `err := s.getAndStoreGuestDetails()`, otherwise we still can not catch the error.

Ace-Tang:
@caoruidong , I use named return, so `if err:= foo()` will not effect, defer still can catch error

caoruidong:
I see your point. Seems https://github.com/kata-containers/runtime/pull/1058 has fixed this. We do not have `if err := s.getAndStoreGuestDetails(); err != nil` now

Ace-Tang:
I see, actually, I want to change all returned err to named, since you do not know who will add `if err := foo()` this later, so I think change the function return is a solution.

caoruidong:
If use := in if block, the err will still be a new local variable and not affect the named return value, see https://play.golang.org/p/vQX70XYpefF

Ace-Tang:
> If use := in if block, the err will still be a new local variable and not affect the named return value, see https://play.golang.org/p/vQX70XYpefF

@caoruidong , the example get error, 
```
if ww:=d();true{
		fmt.Println(ww)
	}

	defer func(){
		fmt.Println(ww)
	}()
	return
```
`ww` is a new variable, it lifecycle is in if block, if you want it take effect to the returned variable, you should return in the if block, or in last return, there is no new ww here, ww = 1 only exist in if block.
I make some fix with https://play.golang.org/p/mOSs7BSp-II

caoruidong:
Oh yes, in this format the user must return value by name. LGTM

jodh-intel:
Closing. Feel free to comment/re-open if required...

PR #1077, opened 2018-12-27
firecracker: Network API update

mcastelino:
The Network API no longer supports the state field. Using the
state field causes the API call to fail.

Note: The firecracker binary needs to be updated to match the API
as the older versions will expect the state field.


caoruidong:
lgtm and /test

virtcontainers/fc.go
```diff
@@ -525,7 +525,6 @@ func (fc *firecracker) fcAddNetDevice(endpoint Endpoint) error {
 		GuestMac:          endpoint.HardwareAddr(),
 		IfaceID:           &ifaceID,
 		HostDevName:       endpoint.NetworkPair().TapInterface.TAPIface.Name,
-		State:             "Attached",
```
bergwolf: I guess you need to update FC vendor for it as well?

mcastelino: @bergwolf Makes sense, even though that API change was already present before and I missed it as we were using an older firecracker binary. Revendored to match 0.12.0.

bergwolf:
Thanks @mcastelino !
LGTM.

PR #1085, opened 2019-01-03
cli: allow to kill a stopped container and sandbox

bergwolf:
cri containerd calls kill on stopped sandbox and if we
fail the call, it can cause `cri stopp` command to fail
too.

Also add a few logs to track container state transitioning.

Fixes: #1084

### DIFF
--- a/cli/kill.go
+++ b/cli/kill.go
@@ -129,13 +129,15 @@ func kill(ctx context.Context, containerID, signal string, all bool) error {
 		return err
 	}
 
-	// container MUST be created, running or paused
-	if status.State.State != vc.StateReady && status.State.State != vc.StateRunning && status.State.State != vc.StatePaused {
-		return fmt.Errorf("Container %s not ready, running or paused, cannot send a signal", containerID)
-	}
+	kataLog.WithField("signal", signal).WithField("container state", status.State.State).Info("kill")
 
-	if err := vci.KillContainer(ctx, sandboxID, containerID, signum, all); err != nil {
-		return err
+	// container MUST be created, running or paused
+	if status.State.State == vc.StateReady || status.State.State == vc.StateRunning || status.State.State == vc.StatePaused {
+		if err := vci.KillContainer(ctx, sandboxID, containerID, signum, all); err != nil {
+			return err
+		}
+	} else if !all {
+		return fmt.Errorf("container not running")
 	}
 
 	if signum != syscall.SIGKILL && signum != syscall.SIGTERM {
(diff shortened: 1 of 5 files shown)


cli/kill.go
```diff
@@ -129,13 +129,13 @@ func kill(ctx context.Context, containerID, signal string, all bool) error {
...
-		return fmt.Errorf("Container %s not ready, running or paused, cannot send a signal", containerID)
-	}
+	kataLog.WithField("signal", signal).WithField("container state", status.State.State).Info("kill")
 
-	if err := vci.KillContainer(ctx, sandboxID, containerID, signum, all); err != nil {
-		return err
+	// container MUST be created, running or paused
+	if status.State.State == vc.StateReady || status.State.State == vc.StateRunning || status.State.State == vc.StatePaused {
+		if err := vci.KillContainer(ctx, sandboxID, containerID, signum, all); err != nil {
+			return err
```
lifupan: If the sandbox/container has stopped, will the following  vci.StopSandbox/vci.StopContainer return another err?

bergwolf: No, it's ignored in sandbox.stop()

lifupan:
It seems there's something wrong with the CI. 
LGTM!

bergwolf:
@sboeuf Good point! After taking a closer look, it is because of a tiny difference with how we handle `kill -a` compared with runc. The `-a` option is not documented in OCI runtime spec. As a result we have to follow runc behavior and do not fail on container status when `-a` is specified, see https://github.com/opencontainers/runc/blob/master/libcontainer/container_linux.go#L378

I've updated the PR to include above change. PTAL.

egernst:
nit: should we have split into two commits?  One commit for the change in behavior and then a second commit for adding more useful logs?

virtcontainers/sandbox.go
```diff
@@ -1569,6 +1569,11 @@ func (s *Sandbox) Stop() error {
 	span, _ := s.trace("stop")
 	defer span.Finish()
 
+	if s.state.State == StateStopped {
+		s.Logger().Info("sandbox already stopped")
+		return nil
```
jodh-intel: How about adding a new unit tests for this scenario, asserting `nil` is returned when state is stopped?

cli/kill.go
```diff
@@ -129,13 +129,15 @@ func kill(ctx context.Context, containerID, signal string, all bool) error {
...
+		}
+	} else if !all {
+		return fmt.Errorf("container not running")
```
jodh-intel: Could you add a unit test for this scenario (and the one above in the `if`)?

sboeuf:
@jodh-intel unit test has been added, if this is okay for you, please merge the PR!

jodh-intel:
Thanks @bergwolf!

lgtm

Issue #1089, opened 2019-01-05
ppc64le: kata-runtime kata-env fails due to missing model field from returned /proc/cpuinfo


nitkon:
# Description of problem

kata-runtime kata-env

# Expected result

Details of the running environment

# Actual result
cannot find model field in file /proc/cpuinfo

---

PR #1090, opened 2019-01-05
kata-runtime: Return correct kata-env on ppc64le

nitkon:
The contents of /proc/cpuinfo were
trimmed and hence the "model" field could
not be parsed despite being a field in
/proc/cpuinfo. Fix this issue.

Fixes: #1089

### DIFF
--- a/cli/kata-check_ppc64le.go
+++ b/cli/kata-check_ppc64le.go
@@ -97,12 +97,61 @@ func archKernelParamHandler(onVMM bool, fields logrus.Fields, msg string) bool {
 	return genericArchKernelParamHandler(onVMM, fields, msg)
 }
 
-func getCPUDetails() (vendor, model string, err error) {
-	vendor, model, err = genericGetCPUDetails()
-	if err == nil {
-		model = "POWER8"
+func getPPC64leCPUInfo(cpuInfoFile string) (string, error) {
+	text, err := katautils.GetFileContents(cpuInfoFile)
+	if err != nil {
+		return "", err
 	}
-	return vendor, model, err
+
+	if len(strings.TrimSpace(text)) == 0 {
+		return "", fmt.Errorf("Cannot determine CPU details")
+	}
+
+	return text, nil
+}
+
+func getCPUDetails() (vendor, model string, err error) {
+
+	if vendor, model, err := genericGetCPUDetails(); err == nil {
+		return vendor, model, nil
+	}
+
+	cpuinfo, err := getPPC64leCPUInfo(procCPUInfo)
+	if err != nil {
+		return "", "", err
+	}
+
+	lines := strings.Split(cpuinfo, "\n")
+
+	for _, line := range lines {
+		if archCPUVendorField != "" {
+			if strings.HasPrefix(line, archCPUVendorField) {
+				fields := strings.Split(line, ":")
+				if len(fields) > 1 {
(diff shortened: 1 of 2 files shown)


cli/kata-check.go
```diff
@@ -76,14 +76,11 @@ func getCPUInfo(cpuInfoFile string) (string, error) {
 		return "", err
 	}
 
-	cpus := strings.SplitAfter(text, "\n\n")
-
-	trimmed := strings.TrimSpace(cpus[0])
-	if trimmed == "" {
+	if len(strings.TrimSpace(text)) == 0 {
```
jodh-intel: This function is now returning *all* details for *all* cpus fwics. This is to accommodate PPC64le which has an unusual `/proc/cpuinfo` - it contains an extra paragraph at the end which includes the `model` details:

- https://github.com/kata-containers/runtime/blob/master/cli/kata-check_data_ppc64le_test.go#L111

I guess there is no harm *per se* in returning all the duplicated information but it still feels a bit odd to me. Assuming this works as expected on other platforms, please can you update the comment for the function as it is now incorrect as it states:

> // getCPUInfo returns details of **the first CPU** read from the specified cpuinfo file

Hi @Pennyzct and @alicefr - Happy 2019! :smile: Please could you take a look from the ARM64 and os390x perspectives please?

jodh-intel:
I forgot to say, how about adding a ppc64le-specific unit-test which either actually runs a real `/proc/cpuinfo` check to ensure the expected `model` value is returned by the function, or which creates a fake `/proc/cpuinfo` containing `model` and again checks the returned value?

cli/kata-check.go
Pennyzct: Hi~ @jodh-intel, Happy 2019! :) ARM64 share the same structure as amd64, all duplicated info. Do we need to make this func arch-dependent??

grahamwhaley:
You may need to make the `cpuinfo` funcs arch dependant - iirc there is no standard for the info that comes out of `/proc/cpuinfo` across architectures (btdt)

cli/kata-check.go
jodh-intel: Yep - I agree with both you and @grahamwhaley @Pennyzct - it would be much clearer to have arch-specific functions for this.

jodh-intel:
@nitkon - I'll add the dnm label for now so you can add a `getPPC64leCPUDetails()` function in a similar manner to `getS390xCPUDetails()`...

nitkon:
ping for review ...

cli/kata-check_ppc64le.go
```diff
@@ -97,12 +97,62 @@ func archKernelParamHandler(onVMM bool, fields logrus.Fields, msg string) bool {
...
+			}
+		}
+	}
+
+	if archCPUVendorField != "" && vendor == "" {
+		return "", "", fmt.Errorf("cannot find vendor field in file %v", procCPUInfo)
+	}
+
+	// model name is optional
+	if archCPUModelField != "" && model == "" {
```
caoruidong: Is it optional?

nitkon: Nope. Thanks! Fixed.

jodh-intel:
Thanks @nitkon.

lgtm

nitkon:
All green. Good to merge. @jodh-intel @caoruidong

Issue #1110, opened 2019-01-10
update supported docker version


chavafg:
update docker to v18.09-ce

sboeuf:
@chavafg did they fix issues with devicemapper?

PR #1112, opened 2019-01-10
versions: update supported docker version to 18.09

chavafg:
Update docker to v18.09-ce

Fixes: #1110.

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -176,7 +176,7 @@ externals:
     description: "Moby project container manager"
     notes: "Docker Swarm requires an older version of Docker."
     url: "https://github.com/moby/moby"
-    version: "v18.06-ce"
+    version: "v18.09-ce"
     meta:
       swarm-version: "1.12.1"
 


devimc:
that means no more devicemapper?

egernst:
I think we want to hold off until we have a block based snapshotter plug-in in place

sboeuf:
Yes agreed that we cannot afford to move to 18.09 since they broke devicemapper, and since we don't have a backup plan yet.

chavafg:
hmmm, right, but that would mean we wont be able to move to fedora 29, which only provides docker 18.09, unless we hack it to use the 18.06 from the fedora 28 repo.

jodh-intel:
@egernst - one for the architecture committee meeting maybe?

raravena80:
@chavafg any updates? Thx. Are we still waiting for the snapshotter plugin?

chavafg:
yes, still waiting on that, I can close this in the meantime...

Issue #1114, opened 2019-01-11
ppc64le: TestCCCheckCLIFunction fails on IBM P9 system


nitkon:
# Description of problem
```
=== RUN   TestCCCheckCLIFunction
--- FAIL: TestCCCheckCLIFunction (0.08s)
    assertions.go:239: 
                          
	Error Trace:	kata-check_ppc64le_test.go:42
        
	            	
			kata-check_ppc64le_test.go:97
        
	Error:      	Received unexpected error:
        
	            	SMT is not Off. System is not capable of running Kata Containers
    assertions.go:239: 
                          
	Error Trace:	kata-check_ppc64le_test.go:112
        
	Error:      	Received unexpected error:
        
	            	SMT is not Off. System is not capable of running Kata Containers
    assertions.go:239: 
                          
	Error Trace:	kata-check_ppc64le_test.go:118
        
	Error:      	Should be true
    assertions.go:239: 
                          
	Error Trace:	kata-check_ppc64le_test.go:118
        
	Error:      	Should be true
```

# Expected result
PASS: TestCCCheckCLIFunction 

# Actual result

FAIL: TestCCCheckCLIFunction 

---

PR #1115, opened 2019-01-11
runtime: Fix TestCCCheckCLIFunction on P9 system

nitkon:
IBM P9 systems can support virtualization with SMT
on. Make sure TestCCCheckCLIFunction passes on P8
(works only with SMT=off) and P9 as well.

Fixes: #1114

### DIFF
--- a/cli/kata-check_data_ppc64le_test.go
+++ b/cli/kata-check_data_ppc64le_test.go
@@ -5,110 +5,20 @@
 
 package main
 
-const testCPUInfoTemplate = `
-processor	: 0
-cpu		: POWER8E (raw), altivec supported
-clock		: 3690.000000MHz
-revision	: 2.1 (pvr 004b 0201)
+import (
+	"github.com/sirupsen/logrus"
+	"io/ioutil"
+)
 
-processor	: 8
-cpu		: POWER8E (raw), altivec supported
-clock		: 3690.000000MHz
-revision	: 2.1 (pvr 004b 0201)
+var testCPUInfoTemplate = setTestCPUInfoTemplate()
 
-processor	: 16
-cpu		: POWER8E (raw), altivec supported
-clock		: 2360.000000MHz
-revision	: 2.1 (pvr 004b 0201)
+func setTestCPUInfoTemplate() string {
 
-processor	: 24
-cpu		: POWER8E (raw), altivec supported
-clock		: 2061.000000MHz
-revision	: 2.1 (pvr 004b 0201)
+	var kataLog *logrus.Entry
+	content, err := ioutil.ReadFile("/proc/cpuinfo")
 
-processor	: 32
-cpu		: POWER8E (raw), altivec supported
-clock		: 3690.000000MHz
-revision	: 2.1 (pvr 004b 0201)
-
-processor	: 40
-cpu		: POWER8E (raw), altivec supported
-clock		: 3690.000000MHz
-revision	: 2.1 (pvr 004b 0201)
-
-processor	: 48
-cpu		: POWER8E (raw), altivec supported
-clock		: 3690.000000MHz
-revision	: 2.1 (pvr 004b 0201)
-
-processor	: 56
-cpu		: POWER8E (raw), altivec supported
-clock		: 3690.000000MHz
-revision	: 2.1 (pvr 004b 0201)
-
-processor	: 64
-cpu		: POWER8E (raw), altivec supported
-clock		: 3690.000000MHz
-revision	: 2.1 (pvr 004b 0201)
-
-processor	: 72
-cpu		: POWER8E (raw), altivec supported
-clock		: 3059.000000MHz
-revision	: 2.1 (pvr 004b 0201)
-
-processor	: 80
-cpu		: POWER8E (raw), altivec supported
-clock		: 2693.000000MHz
-revision	: 2.1 (pvr 004b 0201)
-
-processor	: 88
-cpu		: POWER8E (raw), altivec supported
-clock		: 2061.000000MHz
-revision	: 2.1 (pvr 004b 0201)
-
-processor	: 96
-cpu		: POWER8E (raw), altivec supported
-clock		: 3690.000000MHz
-revision	: 2.1 (pvr 004b 0201)
-
-processor	: 104
-cpu		: POWER8E (raw), altivec supported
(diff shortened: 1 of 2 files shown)


jodh-intel:
Your comment about P8/P9 makes me wonder if this check needs to be improved:

- https://github.com/kata-containers/runtime/blob/master/cli/kata-check_ppc64le.go#L72..L76

Should it be checking for all P versions `<= 8` ?

Also, I wonder if it would make sense to refactor the test code a bit to keep both versions of `cli/kata-check_data_ppc64le_test.go` and dtrt depending on the system under test?

nitkon:
/retest
Ping for review... @jodh-intel @devimc

cli/kata-check_data_ppc64le_test.go
```diff
@@ -112,3 +118,210 @@ model		: 8247-22L
...
+	if err != nil {
+		kataLog.WithError(err).Error("failed to read file")
+		return cpuTypeUnknown
+	}
+	str := string(content)
+	if strings.Contains(str, archP8) {
+		return cpuTypeP8
+	} else if strings.Contains(str, archP9) {
+		return cpuTypeP9
+	} else {
```
devimc: this else is not needed, just `return cpuTypeUnknown`

cli/kata-check_data_ppc64le_test.go
```diff
@@ -112,3 +118,210 @@ model		: 8247-22L
...
+	cpuTypeP9      = 0
+	cpuTypeUnknown = -1
+	archP8         = "POWER8"
+	archP9         = "POWER9"
+)
+
+var testCPUInfoTemplate = setTestCPUInfoTemplate()
+
+func setTestCPUInfoTemplate() string {
+
```
devimc: please trim this function

cli/kata-check_data_ppc64le_test.go
```diff
@@ -112,3 +118,210 @@ model		: 8247-22L
...
+	var kataLog *logrus.Entry
+
+	content, err := ioutil.ReadFile("/proc/cpuinfo")
+	if err != nil {
+		kataLog.WithError(err).Error("failed to read file")
+		return cpuTypeUnknown
+	}
+	str := string(content)
+	if strings.Contains(str, archP8) {
+		return cpuTypeP8
```
jodh-intel: I was wondering if it would be more "future proof" to use a regex (something like `cpu\s+:\s*POWER([0-9]+)`) to extract the numeric suffix, convert it to an integer, then just have some code that says something like:

```go
if powerVersion <= 8 {
    // ...
} else {
  // ...
}

**Comment by nitkon** [comment]
I think the 2 failures are not PR related. Ping for review... @jodh-intel @devimc

**Comment by nitkon** [comment]
CI is happy. Ping @jodh-intel @grahamwhaley

File: cli/kata-check_data_ppc64le_test.go [hunk 4]
```diff
@@ -112,3 +119,206 @@ model		: 8247-22L
...
+
+timebase	: 512000000
+platform	: PowerNV
+model		: 9006-22P
+machine		: PowerNV 9006-22P
+firmware	: OPAL
+MMU		: Radix
+`
+
+var testCPUInfoTemplate = setTestCPUInfoTemplate()
```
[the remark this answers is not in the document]
nitkon: @grahamwhaley : Updated my PR. Tell me if this makes more sense. :-)

grahamwhaley:
Ah, I see, you set it up to be used by the generic test - as an arch specific setting.
I get it now. I'd still like an ack from @jodh-intel if we think this is the right approach - reading from `/proc/cpuinfo` on the test machine feels like it will only test that machine, whereas coding fragments from both a series 8 and 9 machine would maybe be able to test both.
I'm OK either way - it's your `arch` you are testing ;-)

cli/kata-check_data_ppc64le_test.go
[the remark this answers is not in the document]
nitkon: @grahamwhaley :  `testCPUInfoTemplate` is used everytime `makeCPUInfoFile` is called which in turn is called by many test cases when testing anything against /proc/cpuinfo. 
https://github.com/kata-containers/runtime/blob/e06c8aafdc2b10787f4fb5b843f592a5d02490e8/cli/kata-check_test.go#L121

nitkon:
@grahamwhaley: Also we are testing it against that test machines smt value which if on P8 checks if SMT is off and P9 checks nothing as SMT on/off does not matter. Its also helpful when we do `kata-runtime kata-check`.
https://github.com/kata-containers/runtime/blob/b7ebc276aa1fd3bf9d56cbd58eb8b77619fc171d/cli/kata-check_ppc64le.go#L161

nitkon:
ping @jodh-intel for review...

nitkon:
polite ping @jodh-intel ...

cli/kata-check_ppc64le.go
```diff
@@ -69,7 +71,14 @@ func hostIsVMContainerCapable(details vmContainerCapableDetails) error {
 		return err
 	}
 
-	if strings.Contains(text, "POWER8") {
+	ae := regexp.MustCompile("[0-9]")
```
jodh-intel: Shouldn't this be `[0-9]+`?

cli/kata-check_ppc64le.go
```diff
@@ -69,7 +71,14 @@ func hostIsVMContainerCapable(details vmContainerCapableDetails) error {
...
+	re := regexp.MustCompile("POWER[0-9]")
+	powerProcessor, err := strconv.Atoi(ae.FindString(re.FindString(text)))
+	if err != nil {
+		kataLog.WithError(err).Error("Failed to find Power Processor number from /proc/cpuinfo")
```
jodh-intel: Nit: Don't hard-code filename:

```go
kataLog.WithError(err).Error("Failed to find Power Processor number from %q", details.cpuInfoFile)
```

cli/kata-check_ppc64le.go
nitkon: @jodh-intel : Makes sense

cli/kata-check_ppc64le.go
nitkon: @jodh-intel : Done ...

jodh-intel:
lgtm

Issue #1083, opened 2019-01-02
Kata containers cannot support static QoS class.


mcastelino:
# Description of problem

When a cluster is setup with `cpuManagerPolicy: static` for example with a kubeadm.yaml

```
apiVersion: kubeadm.k8s.io/v1alpha3
kind: InitConfiguration
nodeRegistration:
  criSocket: /var/run/crio/crio.sock
---
apiVersion: kubelet.config.k8s.io/v1beta1
kind: KubeletConfiguration
featureGates:
  RuntimeClass: true
cpuManagerPolicy: static
systemReserved:
  cpu: 500m
  memory: 256M
kubeReserved:
  cpu: 500m
  memory: 256M
---
apiVersion: kubeadm.k8s.io/v1alpha3
kind: ClusterConfiguration
networking:
  dnsDomain: cluster.local
  podSubnet: 10.244.0.0/16
  serviceSubnet: 10.96.0.0/12
apiServerExtraArgs:
  feature-gates: RuntimeClass=true
```

POD creation fails for pods with the runtime class set to Kata with the error

```
Internal PreStartContainer hook failed: updating resources for container "php-apache" failed: Container php-apache is not running
```

# Expected result

POD creation should work.

# Actual result

POD creation fails

mcastelino:
https://github.com/clearlinux/cloud-native-setup/tree/master/clr-k8s-examples can be used to expose this error with 

```
kubectl apply -f tests/test-deploy-kata.yaml
```

krsna1729:
@mcastelino @egernst duplicate of #878 ?

egernst:
Yes.  @mcastelino ok if we close this and work off of #878?

egernst:
Well -- I actually didn't see it *fail* before - I just saw it not honored.  @mcastelino do you have any logs in journal for this?  /cc @jcvenegas

mythi:
I've ran into this before. My log is:

```
 Error: failed to create containerd task: OCI runtime create failed: rpc error: code = Internal desc = Could not run process: container_linux.go:348: starting container process caused "process_linux.go:279: applying cgroup configuration for process caused \"failed to write 0-47 to cpuset.cpus: write /sys/fs/cgroup/cpuset/kubepods/burstable/pod25e103bb-0f80-11e9-b6c0-c81f66f6420d/haproxy-ingress/cpuset.cpus: invalid argument\"": unknown
```

krsna1729:
[...]

PR #1126, opened 2019-01-14
update: allow do update on ready.

jcvenegas:
Before, we would only allow for a container-update command
to proceed if the container was in the running state. So
long as the container is created, this should be allowed.

This was found using the `static` policy for Kubernetes CPU
manager[1]. Where the `update` command is called after the
`create` runtime command (when the container state is `ready`).

[1] https://github.com/kubernetes/community/blob/95a4a1/contributors/design-proposals/node/cpu-manager.md#example-scenarios-and-interactions

Fixes: #1083

### DIFF
--- a/cli/update.go
+++ b/cli/update.go
@@ -165,8 +165,8 @@ other options are ignored.
 		span.SetTag("sandbox", sandboxID)
 
 		// container MUST be running
-		if status.State.State != types.StateRunning {
-			return fmt.Errorf("Container %s is not running", containerID)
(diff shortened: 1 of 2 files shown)


amshinde:
lgtm
Nit: @jcvenegas I would also include some more explanation in the commit message describing the scenario you are trying to solve, namely receiving an update command for a ready container in case of static cpu-manager in kubernetes.

jcvenegas:
@amshinde  added explanation about the use case to solve  in the commit message and PR description for future references.

Issue #1131, opened 2019-01-15
containerd-shim-v2: Wait() should support multiple callers


sboeuf:
I don't think the `containerd-shim-v2` API specifies that we should only support one single call of the `Wait()` function. Instead, it'd feel more natural to be able to have several callers being able to invoke `Wait()` and to get the exit code when the process returns.

PR #1133, opened 2019-01-16
shimv2: multi wait support for wait service

lifupan:
Support multi wait on one container/exec process.

Fixes: #1131

### DIFF
--- a/containerd-shim-v2/service.go
+++ b/containerd-shim-v2/service.go
@@ -728,12 +728,20 @@ func (s *service) Wait(ctx context.Context, r *taskAPI.WaitRequest) (*taskAPI.Wa
 	//wait for container
 	if r.ExecID == "" {
 		ret = <-c.exitCh
+
+		// refill the exitCh with the container process's exit code in case
+		// there were other waits on this process.
+		c.exitCh <- ret
 	} else { //wait for exec
 		execs, err := c.getExec(r.ExecID)
 		if err != nil {
 			return nil, err
 		}
 		ret = <-execs.exitCh
+
+		// refill the exitCh with the exec process's exit code in case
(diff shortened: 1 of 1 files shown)


containerd-shim-v2/service.go
```diff
@@ -728,12 +728,20 @@ func (s *service) Wait(ctx context.Context, r *taskAPI.WaitRequest) (*taskAPI.Wa
 	//wait for container
 	if r.ExecID == "" {
 		ret = <-c.exitCh
+
+		// refill the exitCh with the container process's exit code in case
+		// there were other waits on this process.
+		c.exitCh <- ret
```
sboeuf: Is it the appropriate solution here? What happens in the case the caller invoke `Wait()` from two different threads, but the process is still running? It feels to me this is not gonna go well.
Maybe we should add some larger locking on this function if we want to avoid any potential race condition here.

lifupan: Hi @sboeuf, If more than one Wait() are invoked when the process is still running, all of them will hang on this exitCh channel, until the process exits and then they will be serialized to get the return code.

sboeuf: you're right!

sboeuf:
@bergwolf @amshinde PTAL

bergwolf:
LGTM!

[Approved with PullApprove](https://pullapprove.com/kata-containers/runtimes/pull-request/1133/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=kata-containers/runtimes)

bergwolf:
`jenkins-ci-fedora-vsocks` seems stuck that jenkins didn't even create any link here. Let's ignore it!

Issue #1130, opened 2019-01-15
containerd-shim-v2: Use UnmarshalAny to properly retrieve resources


sboeuf:
Here https://github.com/kata-containers/runtime/blob/master/containerd-shim-v2/service.go#L703-L706, we assume the type being unmarshalled has been serialized through JSON encoding, and also we assume it is of type `specs.LinuxResources`.
Instead, we should use `UnmarshalAny()` function from `github.com/containerd/typeurl` package, which is the right way to proceed on a type `protobuf.Any`.

lifupan:
Hi @sboeuf, for the update request to kata shimv2, we are sure that this protobuf.Any is a json Marshaled specs.LinuxResources, so I think it's ok for this assume here.  Please see 
https://github.com/containerd/containerd/blob/96f306ee4cb2b3e881408658602c5d831995c7ad/task_opts.go#L205.

sboeuf:
@lifupan I agree it's not a huge deal, but it would be cleaner not to take containerd implementation as reference here. Since the API uses a protobuf.Any, I think we should use `typeurl.UnmarshalAny()`.
You could imagine that CRI-O and containerd use different structures to define resources, and our implementation should be able to handle both IMO (even if that's not the case in practice).

lifupan:
@sboeuf I agree it'll be much more clear to use typeurl.UnmarshalAny() here. I'll send a PR later to address it.

sboeuf:
Great, and while you're at it, you should also fix the `Exec()` here: https://github.com/kata-containers/runtime/blob/master/containerd-shim-v2/exec.go#L74-L78

PR #1134, opened 2019-01-16
shimv2: use the UnmarshalAny() to unmarshal Protobuf.Any

lifupan:
It'll be much clear to unmarshal Protobuf.Any using
UnmarshalAny().

Fixes: #1130

### DIFF
--- a/containerd-shim-v2/exec.go
+++ b/containerd-shim-v2/exec.go
@@ -6,13 +6,13 @@
 package containerdshim
 
 import (
-	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
 
 	"github.com/containerd/containerd/api/types/task"
 	"github.com/containerd/containerd/errdefs"
+	"github.com/containerd/typeurl"
 	googleProtobuf "github.com/gogo/protobuf/types"
 	"github.com/kata-containers/runtime/virtcontainers/types"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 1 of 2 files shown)


containerd-shim-v2/exec.go
```diff
@@ -73,9 +73,11 @@ func newExec(c *container, stdin, stdout, stderr string, terminal bool, jspec *g
 
 	// process exec request
 	var spec specs.Process
-	if err := json.Unmarshal(jspec.Value, &spec); err != nil {
+	v, err := typeurl.UnmarshalAny(jspec)
+	if err != nil {
 		return nil, err
 	}
+	spec = *v.(*specs.Process)
```
sboeuf: Please add a switch case to check for the type, and if this is not the expected type, let's throw an error.

containerd-shim-v2/service.go
```diff
@@ -701,11 +701,13 @@ func (s *service) Update(ctx context.Context, r *taskAPI.UpdateTaskRequest) (*pt
 	defer s.Unlock()
 
 	var resources specs.LinuxResources
-	if err := json.Unmarshal(r.Resources.Value, &resources); err != nil {
-		return empty, err
+	v, err := typeurl.UnmarshalAny(r.Resources)
+	if err != nil {
+		return nil, err
 	}
+	resources = *v.(*specs.LinuxResources)
```
sboeuf: Please add a switch case to check for the type, and if this is not the expected type, let's throw an error.

containerd-shim-v2/exec.go
lifupan: @sboeuf updated.

containerd-shim-v2/service.go
lifupan: updated.

sboeuf:
@egernst PTAL

bergwolf:
LGTM

bergwolf:
`jenkins-ci-fedora-vsocks` failed on network issue. Restarting...

sboeuf:
`jenkins-ci-fedora-vsocks` does not work at all, as it fails after about 20 seconds. This PR does not change anything related to vsock, so let's move forward and merge it.

PR #980, opened 2018-12-05
virtcontainers: set private propagation in rootfs

devimc:
When overlay is used as storage driver, kata runtime creates a new bind mount
point to the merged directory, that way this directory can be shared with the
VM through 9p. By default the mount propagation is shared, that means mount
events are propagated, but umount events not, to deal with this problem and to
avoid left mount points in the host once container finishes, the mount
propagation of bind mounts should be set to private.

fixes #794

Signed-off-by: Julio Montes <julio.montes@intel.com>

PR #1139, opened 2019-01-17
clean up container dir

bergwolf:
We need to clean up `/run/kata-containers/shared/sandboxes/sbid/` and `/run/vc/vm/sbid` properly.


virtcontainers/qemu.go
```diff
@@ -653,9 +654,24 @@ func (q *qemu) stopSandbox() error {
...
-	err = os.RemoveAll(filepath.Join(RunVMStoragePath, q.id))
+	// cleanup vm path
+	dir := filepath.Join(RunVMStoragePath, q.id)
+
+	// If it's a symlink, remove both dir and the target.
+	// This can happen when vm template links a sandbox to a vm.
+	link, err := filepath.EvalSymlinks(dir)
 	if err != nil {
-		q.Logger().WithError(err).Error("Fail to clean up vm directory")
+		q.Logger().WithError(err).Warnf("failed to resolve vm path %s", dir)
```
grahamwhaley: I'm presuming this is a non-fatal failure on purpose, so we continue to try and clean up even if we think we cannot find the dir? Maybe we should add an explicit comment for that.

virtcontainers/qemu.go
```diff
@@ -653,9 +654,24 @@ func (q *qemu) stopSandbox() error {
...
+	}
+	q.Logger().WithField("link", link).WithField("dir", dir).Infof("cleanup vm path")
+
+	if err = os.RemoveAll(dir); err != nil {
```
grahamwhaley: nit, niggle - the `; err != nil` idiom here is not used above - maybe as we only have the one return val?, but it might be nice to keep the flows the same?

devimc:
heads up! with this patch files in `/dev` are removed

```
Stderr: 
Running command '/usr/bin/ls [ls /dev]'
could no start command: open /dev/null: no such file or directory
command failed error 'exec: not started'
Running command '/usr/bin/docker [docker ps -a -f name=QKsYKbEcpoeFKUoDIuTe0p5XBi9PSg --format {{.Status}}]'
could no start command: open /dev/null: no such file or directory
command failed error 'exec: not started'
```

grahamwhaley:
**DNM** applied :-) thx @devimc

devimc:
@bergwolf I think you'll need this pr https://github.com/kata-containers/runtime/pull/980

devimc:
```
Stderr: docker: Error response from daemon: OCI runtime create failed: open /dev/pts/0: no such file or directory: unknown.
```

bergwolf:
@devimc Yes, you are right! We need the container rootfs to be private to avoid `docker cp` propagate mounts to it. I've added `Depends-on` tag on #980. Thanks!

/test

gnawux:
LGTM

PR #1151, opened 2019-01-18
runtime-v2: Make sure Shutdown() only shuts the server down

### LINKED ISSUE #1150: runtime-v2: Make sure Shutdown() only shuts the server down
We should not stop and delete the sandbox from `Shutdown()` since the caller cannot check the error on this function. It's going to make the server exit, which will close the communication with the RPC server.

sboeuf:
Because the runtime v2 runs as a RPC server, the caller will at some
point use the Shutdown() API to shut down the server. Because this
will cause the server to exit, the caller cannot expect any valid
answer when calling this. That's why we cannot afford stopping and
deleting the sandbox from this function.

Instead, we move sandbox.Stop() and sandbox.Delete() to a more
appropriate API, the Delete() one.

Fixes #1150

### DIFF
--- a/containerd-shim-v2/service.go
+++ b/containerd-shim-v2/service.go
@@ -386,6 +386,22 @@ func (s *service) Delete(ctx context.Context, r *taskAPI.DeleteRequest) (*taskAP
 			return nil, err
 		}
 
+		// Take care of the use case where it is a sandbox.
+		// Right after the container representing the sandbox has
+		// been deleted, let's make sure we stop and delete the
+		// sandbox.
+		if c.cType.IsSandbox() {
+			if err = s.sandbox.Stop(); err != nil {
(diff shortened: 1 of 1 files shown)


lifupan:
LGTM

sboeuf:
@egernst @amshinde @bergwolf PTAL

sboeuf:
`jenkins-ci-fedora-vsocks` fails after 20 seconds, and this is nothing related to this PR. Let's merge.

grahamwhaley:
Yep, seems vsocks build is bust - @chavafg , feels like a new CI issue. Might this be an Azure related thing?
```
PR/go/src/github.com/kata-containers/tests
fatal: Couldn't find remote ref pull/1151/head
```

First candidate for the 'CI status' page/issue maybe?

Issue #1161, opened 2019-01-23
makefile: install does not work with DESTDIR


jcvenegas:
# Description of problem

make install DESTDIR=/tmp/t 
fails due to symlink to configuration.toml does not point to 
the rigth path.

PR #1162, opened 2019-01-23
makefile: honor DESDIR on install

jcvenegas:
- Do symlink to a relative path to hypervisor config.
- Create symlink on DESTDIR

Fixes: #1161


Makefile
```diff
@@ -546,7 +546,7 @@ install-bin-libexec: $(BINLIBEXECLIST)
 
 install-configs: $(CONFIGS)
 	$(QUIET_INST)$(foreach f,$(CONFIGS),$(call INSTALL_CONFIG,$f,$(dir $(CONFIG_PATH))))
-	$(QUIET_INST)ln -sf $(DEFAULT_HYPERVISOR_CONFIG_PATH) $(CONFIG_PATH)
+	$(QUIET_INST)ln -sf $(DEFAULT_HYPERVISOR_CONFIG) $(DESTDIR)/$(CONFIG_PATH)
```
jodh-intel: I'm not sure this is correct. Previously, if I ran `make install DESTDIR=/tmp/james` it would do:

```sh
ln -sf /usr/share/defaults/kata-containers/configuration-qemu.toml /usr/share/defaults/kata-containers/configuration.toml
```

... but with this change, it does something like:

```sh
ln -sf configuration-qemu.toml /tmp/james//usr/share/defaults/kata-containers/configuration.toml
```

However, what we want is:

```sh
ln -sf /tmp/james//usr/share/defaults/kata-containers/configuration-qemu.toml /tmp/james//usr/share/defaults/kata-containers/configuration.toml
```

jcvenegas: @jodh-intel if we make the symlink to a absulute path in the case the the content of /tmp/james is moved to root (`/`) wont be a valid reference anymore. But if we do a relative link it will point the config we want all the time.

jodh-intel: But won't that only work if you create that relative link if you're `cd`'d to the directory where `configuration-qemu.toml` lives?

```sh
(cd $(dir $(DESTDIR)/$(CONFIG_PATH)) && ln -sf $(DEFAULT_HYPERVISOR_CONFIG) $(DESTDIR)/$(CONFIG_PATH))
```

That's starting to look like lisp ;)

devimc:
thanks @jcvenegas , I need this patch to fix snap build, any plan to include it in the release 1.5.0? or do you want to release a new version?

jcvenegas:
@devimc  sadly this was found at post release, we will need to create a new release for this. I add it to obs packages temporally

Makefile
bryteise: Why do we want absolute paths for symlinks? Shouldn't the location for the symlink file be consistent so the relative path is well defined?

devimc:
@jodh-intel @jcvenegas 

I found another interesting issue

```
$ make PREFIX=/snap
./kata-runtime --kata-show-default-config-paths
/etc/kata-containers/configuration.toml
/usr/share/defaults/kata-containers/configuration.toml
```
why `/usr` for the configuration file? it should be `/snap`, isn't it?

jodh-intel:
Ignoring arm ci which has a jenkins issue fwics.

We have full approvals and since this change is critical and is not architecture specific, merging...

jodh-intel:
@devimc - captured in new issue https://github.com/kata-containers/runtime/issues/1185.

Issue #1163, opened 2019-01-23
checking status destroys vm factory initialization


liwei:
# Description of problem

After vm factory initialization, when check it's status with `kata-runtime factory status`, the initialized vm factory was destroyed.

# Expected result

vm factory stays initialized after status checking.

# Actual result

vm factory destroyed after status checking.

```bash
$ sudo kata-runtime factory init
vm factory initialized
$ sudo kata-runtime factory status
vm factory is on
$ sudo kata-runtime factory status
vm factory is off
```

PR #1164, opened 2019-01-23
factory: do not destroy vm factory when checking status

liwei:
Fixes: #1163 

Signed-off-by: Wei Li <liwei@anbutu.com>


jodh-intel:
Yeah, nice catch!!

lgtm

PR #1168, opened 2019-01-23
Makefile: Provide default hypervisor CMD settings

bryteise:
As arch detection isn't being done and KNOWN_HYPERVISORS was only
being populated via environment variable settings (and is a runtime
rather than build time dependency anyway) just provide a set of
defaults that can be overridden as needed at build time.

This will also cause any hypervisor configuration to always be
installed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,21 +23,19 @@ ifeq ($(SKIP_GO_VERSION_CHECK),)
     include golang.mk
 endif
 
-ifneq ($(GOPATH),)
-    GOARCH=$(shell go env GOARCH)
-    ifeq ($(ARCH),)
-        ARCH = $(GOARCH)
-    endif
-
-    ARCH_DIR = arch
-    ARCH_FILE_SUFFIX = -options.mk
-    ARCH_FILE = $(ARCH_DIR)/$(ARCH)$(ARCH_FILE_SUFFIX)
-    ARCH_FILES = $(wildcard arch/*$(ARCH_FILE_SUFFIX))
-    ALL_ARCHES = $(patsubst $(ARCH_DIR)/%$(ARCH_FILE_SUFFIX),%,$(ARCH_FILES))
-
-    # Load architecture-dependent settings
-    include $(ARCH_FILE)
+GOARCH=$(shell go env GOARCH)
+ifeq ($(ARCH),)
+    ARCH = $(GOARCH)
 endif
+
+ARCH_DIR = arch
+ARCH_FILE_SUFFIX = -options.mk
+ARCH_FILE = $(ARCH_DIR)/$(ARCH)$(ARCH_FILE_SUFFIX)
+ARCH_FILES = $(wildcard arch/*$(ARCH_FILE_SUFFIX))
+ALL_ARCHES = $(patsubst $(ARCH_DIR)/%$(ARCH_FILE_SUFFIX),%,$(ARCH_FILES))
+
+# Load architecture-dependent settings
+include $(ARCH_FILE)
 
 PROJECT_TYPE = kata
 PROJECT_NAME = Kata Containers


bryteise:
@grahamwhaley Thanks for the pointer, I had just tossed it up and was looking to see what stuck. I'll open an issue as that seems like the only thing I'm failing on with the patch-format now after fixing the signoff.

Makefile
```diff
@@ -84,6 +84,10 @@ FCBINDIR      := $(PREFIXDEPS)/bin
 SYSCONFDIR    := /etc
 LOCALSTATEDIR := /var
 
+# default CMD names if not
+QEMUCMD       ?= kata-qemu-lite-system-x86_64
```
jodh-intel: What's this? We don't use a kata prefix as this is not a Kata-developed component.

jodh-intel:
Ah - I think I might see what you were trying to do now; the build currently requires `GOPATH`. To resolve that, something like the following would work:

```diff
diff --git a/Makefile b/Makefile
index 1c11f74..86ce694 100644
--- a/Makefile
+++ b/Makefile
@@ -23,22 +23,20 @@ ifeq ($(SKIP_GO_VERSION_CHECK),)
     include golang.mk
 endif
 
-ifneq ($(GOPATH),)
-    GOARCH=$(shell go env GOARCH)
-    ifeq ($(ARCH),)
-        ARCH = $(GOARCH)
-    endif
-
-    ARCH_DIR = arch
-    ARCH_FILE_SUFFIX = -options.mk
-    ARCH_FILE = $(ARCH_DIR)/$(ARCH)$(ARCH_FILE_SUFFIX)
-    ARCH_FILES = $(wildcard arch/*$(ARCH_FILE_SUFFIX))
-    ALL_ARCHES = $(patsubst $(ARCH_DIR)/%$(ARCH_FILE_SUFFIX),%,$(ARCH_FILES))
-
-    # Load architecture-dependent settings
-    include $(ARCH_FILE)
+GOARCH=$(shell go env GOARCH)
+ifeq ($(ARCH),)
+    ARCH = $(GOARCH)
 endif
 
+ARCH_DIR = arch
+ARCH_FILE_SUFFIX = -options.mk
+ARCH_FILE = $(ARCH_DIR)/$(ARCH)$(ARCH_FILE_SUFFIX)
+ARCH_FILES = $(wildcard arch/*$(ARCH_FILE_SUFFIX))
+ALL_ARCHES = $(patsubst $(ARCH_DIR)/%$(ARCH_FILE_SUFFIX),%,$(ARCH_FILES))
+
+# Load architecture-dependent settings
+include $(ARCH_FILE)
+
 PROJECT_TYPE = kata
 PROJECT_NAME = Kata Containers
 PROJECT_TAG = kata-containers
```

bryteise:
@jodh-intel So the issue raised is when the GOPATH is set, the autodetection using the arch scripts isn't run is that something that could be modified then?

bryteise:
@jodh-intel Okay, I'll give that a try.

jcvenegas:
@jodh-intel I tested your patch, it as I was expecting, what I dont really get is that the GOPATH is actually set en my environment

Makefile
```diff
@@ -23,22 +23,20 @@ ifeq ($(SKIP_GO_VERSION_CHECK),)
...
-    ARCH_DIR = arch
-    ARCH_FILE_SUFFIX = -options.mk
-    ARCH_FILE = $(ARCH_DIR)/$(ARCH)$(ARCH_FILE_SUFFIX)
-    ARCH_FILES = $(wildcard arch/*$(ARCH_FILE_SUFFIX))
-    ALL_ARCHES = $(patsubst $(ARCH_DIR)/%$(ARCH_FILE_SUFFIX),%,$(ARCH_FILES))
-
-    # Load architecture-dependent settings
-    include $(ARCH_FILE)
+GOARCH=$(shell go env GOARCH)
+ifeq ($(ARCH),)
```
jodh-intel: It might be worth adding a comment here explaining this is to allow the user to set `ARCH` before we overwrite it.

# More info

A quick look makes it appear as though this is wrong since `ARCH` isn't set yet. However, since `ARCH` is added to the `USER_VARS` variable, that check is to allow the user to define the architecture they want to build with (cross build). If not specified, we default to the host architecture (using go's naming rather than that of `arch(1)`):

```sh
$ make ARCH=foo ...
```

teawater:
> Ah - I think I might see what you were trying to do now; the build currently requires `GOPATH`. To resolve that, something like the following would work:
> 
> ```diff
> diff --git a/Makefile b/Makefile
> index 1c11f74..86ce694 100644
> --- a/Makefile
> +++ b/Makefile
> @@ -23,22 +23,20 @@ ifeq ($(SKIP_GO_VERSION_CHECK),)
>      include golang.mk
>  endif
>  
> -ifneq ($(GOPATH),)
> -    GOARCH=$(shell go env GOARCH)
> -    ifeq ($(ARCH),)
> -        ARCH = $(GOARCH)
> -    endif
> -
> -    ARCH_DIR = arch
> -    ARCH_FILE_SUFFIX = -options.mk
> -    ARCH_FILE = $(ARCH_DIR)/$(ARCH)$(ARCH_FILE_SUFFIX)
> -    ARCH_FILES = $(wildcard arch/*$(ARCH_FILE_SUFFIX))
> -    ALL_ARCHES = $(patsubst $(ARCH_DIR)/%$(ARCH_FILE_SUFFIX),%,$(ARCH_FILES))
> -
> -    # Load architecture-dependent settings
> -    include $(ARCH_FILE)
> +GOARCH=$(shell go env GOARCH)
> +ifeq ($(ARCH),)
> +    ARCH = $(GOARCH)
>  endif
>  
> +ARCH_DIR = arch
> +ARCH_FILE_SUFFIX = -options.mk
> +ARCH_FILE = $(ARCH_DIR)/$(ARCH)$(ARCH_FILE_SUFFIX)
> +ARCH_FILES = $(wildcard arch/*$(ARCH_FILE_SUFFIX))
> +ALL_ARCHES = $(patsubst $(ARCH_DIR)/%$(ARCH_FILE_SUFFIX),%,$(ARCH_FILES))
> +
> +# Load architecture-dependent settings
> +include $(ARCH_FILE)
> +
>  PROJECT_TYPE = kata
>  PROJECT_NAME = Kata Containers
>  PROJECT_TAG = kata-containers
> ```

@jodh-intel does makefile call "go" without GOPATH?  If so, maybe I can update "GOPATH not set mode" to non-go mode because:
```
$ sudo go
sudo: go: command not found
```

jodh-intel:
I think you probably want `sudo -E go` - that's what we do in our CI. I personally set `GOPATH` to make my life easier ;)

Issue #1082, opened 2019-01-02
[RFC] Support multiple isolation mechanisms using Kata


mcastelino:
# Description of problem
Kata recently added support for the Firecracker VMM. What that means is that we can now support different hypervisors in Kata.

However it is more than just that. Firecracker is a VMM, but in reality it is a different enough that it can be considered a different isolation mechanism even though the underlying hardware framework is VT-x.
For example the resource profile/model of a Firecracker POD will be significantly different. This is in addition to any security differences (i.e. sandbox capabilities) and limitations.
 

We have multiple ways of exposing these isolation mechanisms to the end user of kubernetes

1. Single runtime with different annotations within the POD (which becomes more kata specific). 
    This may not be ideal just like non standard annotations that came before. 

2. Expose it directly as different runtime classes.
    This means that there may have to be two different kata binaries with different configurations, or the same kata binary that chooses a different configuration based on argv[0] (like busybox).
So `kata-runtime-firecracker` would chooser `configuration-firecracker.toml` as kata already has support for 
```
GLOBAL OPTIONS:
   --kata-config value               Kata Containers config file path
```

Note: Both 1 and 2 are very kata specific.

3. Expose at the runtime class level by allowing the runtime class spec to have more properties. This would allow us to use the same runtime binary but expose it at the top level as multiple runtime classes each with different properties and limitations.

```
type RuntimeClassSpec struct {
    // RuntimeHandler specifies the underlying runtime the CRI calls to handle pod and/or container
    // creation. The possible values are specific to a given configuration & CRI implementation.
    // The empty string is equivalent to the default behavior.
    // +optional
    RuntimeHandler    string

    // Properties to be passed to the runtime on each invocation
    RuntimeProperties map[string]string
}
```

Note: 3 is requires changes to both kubernetes and crio/containerd.

4. Enhance crio and containerd to support specifying configuration along with the runtime binary. This is the equivalent of 2 but a cleaner and natural abstraction for other runtimes to reuse. This also allows us to use runtime class the way it is defined and implemented in kubernetes today.

Hence the crio.conf would look like

```
[crio.runtime.runtimes.kata]
runtime_path = "/usr/bin/kata-runtime"

[crio.runtime.runtimes.katafirecracker]
runtime_path = "/usr/bin/kata-runtime"
runtime_params = "--kata-config /etc/kata-runtime/firecracker.toml"
```

And the RuntimeHandler would need to be enhanced along the lines of

[...]

PR #1173, opened 2019-01-25
shimv2: use the runtime config file passed from containerd/cri

lifupan:
containerd/cri's different runtime handlers can pass different
config files to shimv2 by a generic runtime options, by this kata
can launch the pods using different VMM for different runtime handlers.

Fixes:#1082


gnawux:
env var and the protobuf field, which should have higher privilege?

lifupan:
> env var and the protobuf field, which should have higher privilege?

protobuf > env var > /etc dir > /usr/share/defualt dif

gnawux:
I have another question with protobuf. if work with old version of containerd, will shimv2 panic due to the lack of a field?

lifupan:
> I have another question with protobuf. if work with old version of containerd, will shimv2 panic due to the lack of a field?

No, it will check the pointer first as below:
		if r.Options != nil {

For the old version containerd, here the r.Options will be nil.

lifupan:
@grahamwhaley Is there any thing wrong with CI, the tests have been launched a day ago, but none of them has been scheduled?

gnawux:
/retest
try again

grahamwhaley:
hi @lifupan - yep, there was a problem yesterday when a disk filled up and things got stalled/backed up in the queu. I think some jobs then got dropped..... looks like the above action has kicked it into action again ...

Issue #1174, opened 2019-01-25
runtime: Kata errors out if only initrd/rootfs image is installed


nitkon:
# Description of problem

runtime: Kata errors out if only initrd/rootfs image is installed

# Expected result

Having any one of the two, initrd/image rootfs is fine.

# Actual result

Errors out saying that one of them is not installed. 

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #1175, opened 2019-01-25
runtime: Do not error if only initrd/rootfs image installed

nitkon:
If only initrd or rootfs image is installed,
allow to start Kata Containers without erroring
out.

Fixes:  #1174

### DIFF
--- a/pkg/katautils/config.go
+++ b/pkg/katautils/config.go
@@ -166,7 +166,7 @@ func (h hypervisor) initrd() (string, error) {
 	p := h.Initrd
 
 	if p == "" {
-		return "", nil
+		return "", errors.New("initrd is not set")
 	}
 
 	return ResolvePath(p)
@@ -176,7 +176,7 @@ func (h hypervisor) image() (string, error) {
 	p := h.Image
 
 	if p == "" {
-		return "", nil
+		return "", errors.New("image is not set")
 	}
 
 	return ResolvePath(p)
@@ -340,20 +340,16 @@ func (h hypervisor) guestHookPath() string {
 }
 
 func (h hypervisor) getInitrdAndImage() (initrd string, image string, err error) {
-	if initrd, err = h.initrd(); err != nil {
-		return
-	}
+	initrd, errInitrd := h.initrd()
 
-	if image, err = h.image(); err != nil {
-		return
-	}
+	image, errImage := h.image()
 
 	if image != "" && initrd != "" {
 		return "", "", errors.New("having both an image and an initrd defined in the configuration file is not supported")
 	}
 
-	if image == "" && initrd == "" {
-		return "", "", errors.New("either image or initrd must be defined in the configuration file")
+	if errInitrd != nil && errImage != nil {
+		return "", "", fmt.Errorf("Either initrd or image must be set to a valid path (initrd: %v) (image: %v)", errInitrd, errImage)
 	}
 
 	return
--- a/pkg/katautils/config_test.go
+++ b/pkg/katautils/config_test.go
@@ -1046,14 +1046,14 @@ func TestHypervisorDefaultsInitrd(t *testing.T) {
 	defaultInitrdPath = testInitrdPath
 	h := hypervisor{}
 	p, err := h.initrd()
-	assert.NoError(err)
+	assert.Error(err)
 	assert.Equal(p, "", "default Image path wrong")
 
 	// test path resolution
 	defaultInitrdPath = testInitrdLinkPath
 	h = hypervisor{}
 	p, err = h.initrd()
-	assert.NoError(err)
+	assert.Error(err)
 	assert.Equal(p, "")
 }
 
@@ -1083,14 +1083,14 @@ func TestHypervisorDefaultsImage(t *testing.T) {
 	defaultImagePath = testImagePath
 	h := hypervisor{}
 	p, err := h.image()
-	assert.NoError(err)
+	assert.Error(err)
 	assert.Equal(p, "", "default Image path wrong")
 
 	// test path resolution
 	defaultImagePath = testImageLinkPath
 	h = hypervisor{}
 	p, err = h.image()
-	assert.NoError(err)
+	assert.Error(err)
 	assert.Equal(p, "")
 }
 


pkg/katautils/config.go
```diff
@@ -356,6 +352,12 @@ func (h hypervisor) getInitrdAndImage() (initrd string, image string, err error)
 		return "", "", errors.New("either image or initrd must be defined in the configuration file")
 	}
 
+	if errInitrd != nil {
```
grahamwhaley: This doesn't feel right to me, according to your commit message. This looks like if either fails then we return an error, and I think the caller will check that and fail?
Should we not return an error only if both image and initrd error out? I might be missing some subtlety though.

nitkon: @grahamwhaley : How about ...

```
if (image != "" && errImage != nil)  {
		return initrd, image, errImage
} else if ( initrd != "" && errInitrd != nil) {
             	return initrd, image, errInitrd
}
```

nitkon: @grahamwhaley : We should fail if

1. Both are defined.
2. Both are not defined.
3. One which is defined, that file does not exist.

nitkon:
I see the metrics CI failing. cc @grahamwhaley @chavafg 
```
18:05:14 time="2019-01-25T12:35:14Z" level=warning msg="Check for [boot-times] failed [Failed]"
18:05:14 time="2019-01-25T12:35:14Z" level=warning msg=" with [[*F* boot-times 93.0% 110.1% 107.0% 14.0% 108.4% 112.4% 3.7% 1.0% 20]]"
18:05:14 time="2019-01-25T12:35:14Z" level=warning msg="Failed Maxval (156664.500000 < 162729.470000) for [memory-footprint]"
18:05:14 time="2019-01-25T12:35:14Z" level=warning msg="Check for [memory-footprint] failed [Failed]"
18:05:14 time="2019-01-25T12:35:14Z" level=warning msg=" with [[*F* memory-footprint 85.0% 119.5% 115.0% 30.0% 119.5% 119.5% 0.0% 0.0% 1]]"
18:05:14 time="2019-01-25T12:35:14Z" level=warning msg="Overall we failed"
18:05:14 
18:05:14 Report Summary:
18:05:14 +-----+----------------------+-------+--------+--------+-------+--------+--------+------+------+-----+
18:05:14 | P/F |         NAME         |  FLR  |  MEAN  |  CEIL  |  GAP  |  MIN   |  MAX   | RNG  | COV  | ITS |
18:05:14 +-----+----------------------+-------+--------+--------+-------+--------+--------+------+------+-----+
18:05:14 | *F* | boot-times           | 93.0% | 110.1% | 107.0% | 14.0% | 108.4% | 112.4% | 3.7% | 1.0% |  20 |
18:05:14 | *F* | memory-footprint     | 85.0% | 119.5% | 115.0% | 30.0% | 119.5% | 119.5% | 0.0% | 0.0% |   1 |
18:05:14 | P   | memory-footprint-ksm | 95.0% | 99.8%  | 105.0% | 10.0% | 99.8%  | 99.8%  | 0.0% | 0.0% |   1 |
18:05:14 +-----+----------------------+-------+--------+--------+-------+--------+--------+------+------+-----+
18:05:14 Fails: 2, Passes 1
18:05:14 Failed
18:05:14 checkmetrics FAILED (1)
18:05:14 restoring KSM settings
```

grahamwhaley:
Hi @nitkon - thanks for reporting!
Looking at the job build trend graph: http://jenkins.katacontainers.io/job/kata-metrics-runtime-ubuntu-16-04-PR/buildTimeTrend
It seems that the metrics build has been fairly constantly failing for the last couple of days.
Generally that either means something has gotten 'sticky' on the build machine (but we reboot it once for each build now (which is why you see alternate blue/red status on the timeline), which tends to clear out such stickyness....

The other reason then that things start to get nacked by the metrics is that we've merged something that has changed our figures. Let me have a look back through the history and see if anything obvious pops out.

My guess is this is not related to your PR.

pkg/katautils/utils.go
```diff
@@ -65,7 +65,7 @@ func ResolvePath(path string) (string, error) {
 	if err != nil {
 		if os.IsNotExist(err) {
 			// Make the error clearer than the default
-			return "", fmt.Errorf("file %v does not exist", absolute)
+			return absolute, fmt.Errorf("file %v does not exist", absolute)
```
jodh-intel: This seems a bit unusual, but since an error *is* being returned is safe. Do we really need this though as the absolute path is already being returned in the error itself?

nitkon: Yes we need this so as to check that path was indeed specified in the configuration.toml file(uncommented). Btw do you mean we should extract absolute path  from the error returned?

pkg/katautils/config.go
```diff
@@ -356,6 +352,12 @@ func (h hypervisor) getInitrdAndImage() (initrd string, image string, err error)
 		return "", "", errors.New("either image or initrd must be defined in the configuration file")
 	}
 
+	if initrd != "" && errInitrd != nil {
```
bergwolf: We still err out when initrd is not installed?

nitkon: If initrd is specified in the toml file but that file does not exist, then we error out.

nitkon: Original Code: If I just install initrd and then run Kata, it errors out stating that path/to/rootfs.image does not exist. 

In this patch: I  check just one of the two,  initrd OR image path are specified(uncommented) in the configuration.toml and then check the one that was specified, that file actually exist.

nitkon:
ARM CI is failing because of ` Error: Failed to synchronize cache for repo 'updates'` @Pennyzct

pkg/katautils/config.go
bergwolf: `ResolvePath()` returns either a valid path with nil error, or empty path with an error. So we never get to the point here that `initrd` is not empty while `errInitrd` is non-nil, nor the line below that `image` is not empty while `errImage` is set.

So to make the code work, I think you only need to make sure `errInitrd` and `errImage` are not both set.

nitkon: Thanks @bergwolf

nitkon:
Ping for review. cc @grahamwhaley @jodh-intel 
@Pennyzct ARM CI is failing due to non-PR related issues.

ydjainopensource:
@GabyCT @chavafg ping

nitkon:
Good to merge? @grahamwhaley @jodh-intel

grahamwhaley:
has two acks.
Failed CIs are known issues (ARM, fc)
pullapprove is dead, long live pullapprove
the zuul WIP/SoB have not reported - but, I've manually checked they are present etc.
So, force merging..

Issue #1177, opened 2019-01-25
kata-runtime kata-env error out when there is no VERSION_ID, eg under debian buster/sid


yyyeerbo:
# Description of problem

kata-runtime kata-env error out when there is no VERSION_ID, eg under debian buster/sid

# Expected result

those information should be provided as possible as we can instead of error out

# Actual result

error out..

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #1178, opened 2019-01-25
kata-env: kata-env error out when there is no VERSION_ID.

yyyeerbo:
For example, under debian buster/sid. Those information should be
provide with best effort instead of error out.

Fixes: #1177

Signed-off-by: Yang Bo <bo@hyper.sh>


cli/utils.go
```diff
@@ -71,7 +71,7 @@ func getDistroDetails() (name, version string, err error) {
 			}
 		}
 
-		if name != "" && version != "" {
+		if name != "" || version != "" {
```
grahamwhaley: I agree, if we found one of them we should probably return it.
I suspect it might be good if we set the unfound item to something like `<not set>` - and maybe do that for both if neither is found, and not error out at all. @jodh-intel - wdyt?

yyyeerbo: we return what we know and what we don't know(set to <not set> or <unknown> looks better, I think. :)

amshinde:
lgtm, with returning the not found item as `unknown` or `not-set`

yyyeerbo:
Update to return "<<unknown>>" when name/version is not defined.

jodh-intel:
Thanks @yyyeerbo. I'm hoping we can overcome this issue (https://github.com/kata-containers/documentation/pull/358#pullrequestreview-196982129), but I agree this PR is a good idea regardless.

lgtm

yyyeerbo:
fix bug and testcase..

yyyeerbo:
fix testcases

jodh-intel:
Restarted failing CIs (metrics timed out).

devimc:
lgtm

ARM and metrics CIs are not stable, merging...

Issue #1180, opened 2019-01-25
Decouple the network API from the sandbox one


sameo:
# Description of problem

In order to fix #1059, we want to create a `hypervisor` package. Some of the hypervisor implementations (`qemu`) depend on the `network` and `endpoint` interfaces. We can not have a virtcontainers -> hypervisor -> network, endpoint -> virtcontainers cyclic dependency.
So before creating the `hypervisor` package, we need to decouple the network API from the virtcontainers one.

Issue #1179, opened 2019-01-25
The virtcontainers network interface is not needed


sameo:
# Description of problem

The following network interface:

```Go
type network interface {
	// run runs a callback function in a specified network namespace.
	run(networkNSPath string, cb func() error) error

	// add adds all needed interfaces inside the network namespace.
	add(sandbox *Sandbox, hotplug bool) error

	// remove unbridges and deletes TAP interfaces. It also removes virtual network
	// interfaces and deletes the network namespace.
	remove(sandbox *Sandbox, hotunplug bool) error
}
```

is only implemented by the default network implementation and the noop one. The noop one is only there for testing purposes and at the same time there is no known need for another network model implementation.

# Expected result

Remove the network interface and only use what's currently the default network interface implementation.

sameo:
cc @mcastelino @sboeuf @egernst @bergwolf

sboeuf:
@sameo I'm not against the removal of the Go interface since, as you mentioned it, we're only using one single implementation. But I'm concerned about the unit tests, because the noop implementation is pretty convenient to test a bunch of code that would be complex to test if we had to mimic the real network behavior. WDYT?

mcastelino:
> @sameo I'm not against the removal of the Go interface since, as you mentioned it, we're only using one single implementation. But I'm concerned about the unit tests, because the noop implementation is pretty convenient to test a bunch of code that would be complex to test if we had to mimic the real network behavior. WDYT?

@sboeuf could you not test it with the equivalent of net==none? With the noop you were actually not exercising any real code paths right. It was just there to satisfy the interface?

sboeuf:
@mcastelino 
>@sboeuf could you not test it with the equivalent of net==none?

That's a good idea, and I didn't think about this as it's been introduced fairly recently. This might be enough for some cases.

>With the noop you were actually not exercising any real code paths right. It was just there to satisfy the interface?

Yes, exactly.

egernst:
Great!  I think we should remove, no reason to have an interface in case, and we can fix the tests.

sameo:
> What will this mean with respect to semantic versioning?

This is purely an internal interface, there is no reason to bump version for this kind of change.

PR #1181, opened 2019-01-25
Network interface removal and decoupling

sameo:
This PR does 2 things:

- Remove the network interface (See #1179)
- Decouple the network API from the virtcontainers one (See #1180)


sameo:
> @sameo I'm not against the removal of the Go interface since, as you mentioned it, we're only using one single implementation. But I'm concerned about the unit tests, because the noop implementation is pretty convenient to test a bunch of code that would be complex to test if we had to mimic the real network behavior. WDYT?

FWIW all the unit tests are passing with the default network implementation. When first thinking about doing this I thought they will all fail but they don't. I get CI errors here but there not UT related.

virtcontainers/network.go
```diff
@@ -1464,17 +1417,91 @@ func createEndpoint(netInfo NetworkInfo, idx int, model NetInterworkingModel) (E
...
-// Container network plugins are used to setup virtual network
-// between VM netns and the host network physical interface.
-type network interface {
-	// run runs a callback function in a specified network namespace.
-	run(networkNSPath string, cb func() error) error
+// Network is the virtcontainer network structure
+type Network struct {
+}
+
+func (n *Network) trace(ctx context.Context, name string) (opentracing.Span, context.Context) {
```
mcastelino: @sameo This is code moved mostly unmodified from virtcontainers/default_network.go?

Where there any changes needed?

virtcontainers/sandbox.go
```diff
@@ -206,7 +205,7 @@ type Sandbox struct {
 	hypervisor hypervisor
 	agent      agent
 	storage    resourceStorage
-	network    network
+	network    Network
```
mcastelino: @sameo why did this need to be made public?

sameo:
@sboeuf @mcastelino https://github.com/kata-containers/runtime/pull/1181/commits/b39cb1d13a283322a489f9a746eeb8f99e173eda#diff-bca2c2b5b04e0e74533b492fd23a3cc8R784 is what makes several unit tests pass with this PR.

virtcontainers/sandbox.go
sameo: I'm preparing for moving the network code into its own package.

virtcontainers/network.go
sameo: No functional changes, no.

amshinde:
@sameo Change looks good overall, its just the noop implementation has been useful in mocking our network behaviour in unit tests.

sameo:
> @sameo Change looks good overall, its just the noop implementation has been useful in mocking our network behaviour in unit tests.

noop is useful for sure, but:

* It does not exercise much of the code path
* We need to maintain an artificial abstraction for it
* All unit tests pass without it

amshinde:
@sameo If thats the case
lgtm

bergwolf:
LGTM! It makes sense to simplify the code and `CNM` is not really CNM specific anyway.

sboeuf:
@chavafg which of those CI results can be ignored?

GabyCT:
@sboeuf here it is the page where it says the `CI known failures` http://jenkins.katacontainers.io/view/CI%20Status/, what I see that is missing is the `metrics` job. @grahamwhaley could you please add that with the proper link to the issue. Thanks

sboeuf:
thanks @GabyCT 
I've just restarted the fedora and metrics builds, let see how this goes.

PR #1187, opened 2019-01-28
cli: set config options before showing config paths

### LINKED ISSUE #1185: Allow build to set default config file path once again
As shown in https://github.com/kata-containers/runtime/issues/1144#issuecomment-455243197, PR #870 inadvertently broke setting the default runtime config file path:

```sh
$ make CONFIG_PATH=/foo/bar/configuration.toml
$ ./kata-runtime --kata-show-default-config-paths
/etc/kata-containers/configuration.toml
/usr/share/defaults/kata-containers/configuration.toml
```

That output is wrong as we'd expect to see `/foo/bar/configuration.toml`. Here's why:

```sh
$ grep -r "var defaultRuntimeConfiguration = "|grep -v CONFIG_PATH 
cli/config-generated.go:var defaultRuntimeConfiguration = "/foo/bar/configuration.toml"
pkg/katautils/config-settings.go:var defaultRuntimeConfiguration = "/usr/share/defaults/kata-containers/configuration.toml"
```

File `cli/config-generated.go`, which contains the correct value is not being honoured.

devimc:
Config paths are set correctly but they must be set before handleShowConfig

fixes #1185

cc @jodh-intel 

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -250,6 +250,8 @@ func beforeSubcommands(c *cli.Context) error {
 	var runtimeConfig oci.RuntimeConfig
 	var err error
 
+	katautils.SetConfigOptions(name, defaultRuntimeConfiguration, defaultSysConfRuntimeConfiguration)
+
 	handleShowConfig(c)
 
 	if userWantsUsage(c) || (c.NArg() == 1 && (c.Args()[0] == checkCmd)) {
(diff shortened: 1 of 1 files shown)


jodh-intel:
Ugh - firecracker CI failed with:

```
Check vsock is supported
modprobe: FATAL: Module vhost_vsock not found in directory /lib/modules/4.4.0-128-generic
ERROR: vsock is not supported on your host system
```

/cc @mnaser, @GabyCT.

GabyCT:
@jodh-intel , I made the changes at the `CI` to use another resource.

jodh-intel:
Thanks @GabyCT. Can we be sure this issue won't happen again? It sounds a bit like the problem we had where jobs were being scheduled on systems without VT-x support.

jodh-intel:
The only 2 failing CI's are known issues, so merging...

PR #1188, opened 2019-01-28
[WIP] [DNM]: virtcontainers: Jailer: Add jailer support for firecracker

mcastelino:
Firecracker provides a jailer to constrain the VMM. Use this
jailer to launch the firecracker VMM instead of launching it
directly from the kata-runtime.

The jailer will ensure that the firecracker VMM will run
in its own network and mount namespace. All assets required
by the VMM have to be present within these namespaces.

This also means that more sandbox context needs to be made
available to the hypervisor/jailer.

### DIFF
--- a/pkg/katautils/config-settings.go
+++ b/pkg/katautils/config-settings.go
@@ -9,6 +9,7 @@
 package katautils
 
 var defaultHypervisorPath = "/usr/bin/qemu-lite-system-x86_64"
+var defaultJailerPath = "/usr/bin/jailer"
 var defaultImagePath = "/usr/share/kata-containers/kata-containers.img"
 var defaultKernelPath = "/usr/share/kata-containers/vmlinuz.container"
 var defaultInitrdPath = "/usr/share/kata-containers/kata-containers-initrd.img"
--- a/virtcontainers/fc.go
+++ b/virtcontainers/fc.go
@@ -8,6 +8,7 @@ package virtcontainers
 import (
 	"context"
 	"fmt"
+	"os"
 	"os/exec"
 	"path/filepath"
 
@@ -45,7 +46,9 @@ const (
 const (
 	//fcTimeout is the maximum amount of time in seconds to wait for the VMM to respond
 	fcTimeout            = 10
-	fireSocket           = "firecracker.sock"
+	fcSocket             = "api.socket"
+	fcKernel             = "vmlinux"
+	fcRootfs             = "rootfs"
 	fcStopSandboxTimeout = 15
 	// This indicates the number of block devices that can be attached to the
 	// firecracker guest VM.
@@ -93,9 +96,15 @@ type firecracker struct {
 	state firecrackerState
 	info  FirecrackerInfo
 
-	firecrackerd *exec.Cmd           //Tracks the firecracker process itself
-	fcClient     *client.Firecracker //Tracks the current active connection
-	socketPath   string
+	firecrackerd  *exec.Cmd           //Tracks the firecracker process itself
+	connection    *client.Firecracker //Tracks the current active connection
+	chrootBaseDir string              //chroot base for the jailer. All VM assets need to be under this
+	socketPath    string
+	kernelPath    string
+	rootfsPath    string
+	netNSPath     string
+	uid           string //UID and GID to be used for the VMM
+	gid           string
 
 	storage        resourceStorage
 	config         HypervisorConfig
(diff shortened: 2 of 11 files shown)


jodh-intel:
@devimc - reading @mcastelino's summary above made me think of https://github.com/kata-containers/runtime/pull/847 - is there any overlap with your PR and FC's jailer?

devimc:
@jodh-intel yes, and some of the problems that @mcastelino will face are:

- How to update the mnt ns with new devices/rootfs (in the case of 2 or more containers in the same VM)
- crio tests

pkg/katautils/config-settings.go
```diff
@@ -9,6 +9,7 @@
 package katautils
 
 var defaultHypervisorPath = "/usr/bin/qemu-lite-system-x86_64"
+var defaultJailerPath = "/usr/bin/jailer"
```
jodh-intel: Wow, that's not a great name for a seemingly FC-specific command ;(

mcastelino: Agree. Right now it is specific to firecracker. But...

@jodh-intel @devimc @egernst @sboeuf we have a more fundamental problem. In the case of QEMU, we implement functionality that in the case of firecracker is delegated to the jailer. 

However the jailing functionality implemented for QEMU places the runtime itself in the jail vs just QEMU. Some of this is because of limitation in go and not necessarily the way we want it function. This will cause problems for us.

I am trying to see if we can create a cleaner interface that will work for both QEMU and firecracker.

You could envisage creating something similar to jailer created for QEMU so that we work around the go limitations in a cleaner manner.

Currently I am leaning towards disabling the self jailing code as a part of the hypervisor configuration. (i.e if a jailer is present, do not jail the runtime itself). The runtime can create the namespaces, but will not run in it all the time.

sboeuf: If we can reach a point where we have a jailer for QEMU, and let's go even further by saying we would have a jailer capable of isolating any process as a simple wrapper exec'ing the given process (reminds me about containers...), we would need to modify the `hypervisor` Go interface defined in virtcontainers so that it would support passing a few extra information about those expected isolation layers.

mcastelino:
This is gated by jailer issue https://github.com/firecracker-microvm/firecracker/issues/911

virtcontainers/fc.go
```diff
@@ -138,10 +147,41 @@ func (fc *firecracker) createSandbox(ctx context.Context, id string, hypervisorC
...
-	fc.socketPath = filepath.Join(runStoragePath, fc.id, fireSocket)
+	fc.state.set(notReady)
+
+	// unix domain socket names have a hard limit
+	// #define UNIX_PATH_MAX   108
+	// Keep it short and live within the jailer expected paths
+	// <chroot_base>/<exec_file_name>/<id>/
+	// Also jailer based on the id implicitly sets up cgroups under
+	// <cgroups_base>/<exec_file_name>/<id>/
+	hypervisorName := filepath.Base(hypervisorConfig.HypervisorPath)
```
jodh-intel: Please can you use [`BuildSocketPath`](https://github.com/kata-containers/runtime/blob/master/virtcontainers/utils/utils.go#L197) here.

virtcontainers/fc.go
```diff
@@ -226,17 +266,27 @@ func (fc *firecracker) fcInit(timeout int) error {
...
 	defer span.Finish()
 
-	args := []string{"--api-sock", fc.socketPath}
+	args := []string{"--id", fc.id,
+		"--node", "0",
+		"--exec-file", fc.config.HypervisorPath,
+		"--uid", "0", //hack: we need to setup user and group for kata
+		"--gid", "0",
+		"--chroot-base-dir", fc.chrootBaseDir,
+		//"--netns", fc.netNSPath, //the path seems to be unpopulated
```
jodh-intel: Commented out code.

virtcontainers/fc.go
```diff
@@ -226,17 +266,27 @@ func (fc *firecracker) fcInit(timeout int) error {
 	span, _ := fc.trace("fcInit")
 	defer span.Finish()
 
-	args := []string{"--api-sock", fc.socketPath}
+	args := []string{"--id", fc.id,
+		"--node", "0",
+		"--exec-file", fc.config.HypervisorPath,
+		"--uid", "0", //hack: we need to setup user and group for kata
```
jodh-intel: Hack?

virtcontainers/fc.go
```diff
@@ -266,16 +316,22 @@ func (fc *firecracker) fcSetBootSource(path, params string) error {
 	fc.Logger().WithFields(logrus.Fields{"kernel-path": path,
 		"kernel-params": params}).Debug("fcSetBootSource")
 
+	//Hardlink the file to place it within the jailed root dir
+	//TODO: Setup with uid/gid of the firecracker user
```
jodh-intel: Could you convert all the TODOs to github issues and reference the URLs.

jodh-intel:
Hi @mcastelino - this PR is in need of an update to resolve conflicts.

raravena80:
Hi @mcastelino any updates on this? Thanks!

raravena80:
@mcastelino nudge

mcastelino:
Closing this and creating a new PR based on the latest codebase.

PR #1189, opened 2019-01-28
virtcontainers: reimplement sandbox cgroup

### LINKED ISSUE #1125: [RFC] [Proposal] support cgroups in the host
# Description of the problem

Unlike other runtimes, kata starts at least 2 processes per POD, kata-shim and a hypervisor (and kata-proxy when vsocks is disable). From host perspective, containers run in a guest (virtual machine) whose resources can be monitored by monitoring the hypervisor process and its threads. kata-shims are used to interact with the containers, sending signals, reading/writing data, etc. A POD can have any number of containers that run in the same guest. One of the major gaps that can be found in kata containers currently is that cgroups are applied in the guest, but not in the host or not in the right cgroup path (see https://github.com/kata-containers/runtime/issues/1021) causing problem in some implementations for resource governance, like service fabric https://github.com/Microsoft/service-fabric/.

# Proposal

Move kata-shim (pause workload) and hypervisor (and kata-proxy if vsocks is disable) to sandbox cgroup, and apply a cgroup constraint to the sandbox that is defined by the sum of all its containers' cgroups. Container's cgroups are still created and applied but just to the kata-shim that interact with the container. 2 examples:
  
A pod with 2 containers,  container A with 1 cpu and container B without cpu constraints (1 vcpu is hotplugged).
  * in the host: no cpu constraints for container B and sandbox, 1 cpu for container A.
  * in the guest: no cpu constraints for container B, 1 cpu for container A.

A pod with 2 containers, container A with 1 cpu and container B with 2 cpus (3 vcpus are hotplugged).

proposal-cgroups

**cgroup directory hierarchy**

```
/sys/fs/cgroup/cpu,cpuacct
└── kubepods/burstable/pod644e04f3
      ├── crio-89e60b84  # Sandbox cgroupsPath=/kubepods/burstable/pod644e04f3/crio-89e60b84
      ├── crio-2a95e0f8  # Container A cgroupsPath=/kubepods/burstable/pod644e04f3/crio-2a95e0f8
      └── crio-d037bce7  # Container B cgroupsPath=/kubepods/burstable/pod644e04f3/crio-d037bce7
```

cc @WeiZhang555 @mocolic

devimc:
All containers run in different cgroups even the sandbox, with this new
implementation the sandbox cpu cgroup wil be equal to the sum of all its
containers and the hypervisor process will be placed there impacting to the
containers running in the sandbox (VM). The default number of vcpus is
used when the sandbox has no constraints. For example, if default_vcpus
is 2, then quota will be 200000 and period 100000.

**c-ray test**
http://www.futuretech.blinkenlights.nl/c-ray.html

```
+=============================================+
|         | 6 threads 6cpus | 1 thread 1 cpu  |
+=============================================+
| current |   40 seconds    |   122 seconds   |
+==============================================
|   new   |   37 seconds    |   124 seconds   |
+==============================================
```

current = current cgroups implementation
new = new cgroups implementation

**workload**

```yaml
apiVersion: v1
kind: Pod
metadata:
name: c-ray
annotations:
spec:
restartPolicy: Never
containers:
- name: c-ray-1
image: docker.io/devimc/c-ray:latest
imagePullPolicy: IfNotPresent
args: ["-t", "6", "-s", "1600x1200", "-r", "8", "-i",
      "/c-ray-1.1/sphfract", "-o", "/tmp/output.ppm"]
resources:
  limits:
    cpu: 6
- name: c-ray-2
image: docker.io/devimc/c-ray:latest
imagePullPolicy: IfNotPresent
args: ["-t", "1", "-s", "1600x1200", "-r", "8", "-i",
      "/c-ray-1.1/sphfract", "-o", "/tmp/output.ppm"]
resources:
  limits:
    cpu: 1
```

fixes #1125

### DIFF
--- a/virtcontainers/cgroups.go
+++ b/virtcontainers/cgroups.go
@@ -1,4 +1,5 @@
 // Copyright (c) 2018 Huawei Corporation
+// Copyright (c) 2019 Intel Corporation
 //
 // SPDX-License-Identifier: Apache-2.0
 //
@@ -6,95 +7,218 @@
 package virtcontainers
 
 import (
-	"encoding/json"
+	"bufio"
 	"fmt"
+	"math"
+	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/containerd/cgroups"
 	"github.com/kata-containers/runtime/virtcontainers/pkg/annotations"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 )
 
-const (
-	vcpuGroupName       = "vcpu"
-	defaultCgroupParent = "/kata"
-)
-
-type sandboxCgroups struct {
-	commonParent cgroups.Cgroup
-	sandboxSub   cgroups.Cgroup
-	vcpuSub      cgroups.Cgroup
+type cgroupPather interface {
+	cgroups.Subsystem
+	Path(path string) string
 }
 
-func (s *Sandbox) newCgroups() error {
-	// New will still succeed when cgroup exists
-	// create common parent for all kata-containers
-	// e.g. /sys/fs/cgroup/cpu/vc
-	parent, err := cgroups.New(cgroups.V1,
-		cgroups.StaticPath(defaultCgroupParent), &specs.LinuxResources{})
-	if err != nil {
-		return fmt.Errorf("failed to create cgroup for %q", defaultCgroupParent)
-	}
+// unconstrained cgroups are placed here.
+// for example /sys/fs/cgroup/memory/kata/$CGPATH
+// where path is defined by the containers manager
+const cgroupKataPath = "/kata/"
 
-	// create sub-cgroup for each sandbox
-	// e.g. /sys/fs/cgroup/cpu/vc/<sandbox>
-	sandboxSub, err := parent.New(s.id, &specs.LinuxResources{})
-	if err != nil {
-		return fmt.Errorf("failed to create cgroup for %s/%s", defaultCgroupParent, s.id)
-	}
+var cgroupsLoadFunc = cgroups.Load
+var cgroupsNewFunc = cgroups.New
 
-	// create sub-cgroup for vcpu threads
-	vcpuSub, err := sandboxSub.New(vcpuGroupName, &specs.LinuxResources{})
+// V1Constraints returns the cgroups that are compatible with th VC architecture
+// and hypervisor, constraints can be applied to these cgroups.
+func V1Constraints() ([]cgroups.Subsystem, error) {
+	root, err := cgroupV1MountPoint()
 	if err != nil {
-		return fmt.Errorf("failed to create cgroup for %s/%s/%s", defaultCgroupParent, s.id, vcpuGroupName)
+		return nil, err
 	}
-
-	s.cgroup = &sandboxCgroups{
-		commonParent: parent,
-		sandboxSub:   sandboxSub,
-		vcpuSub:      vcpuSub,
+	subsystems := []cgroups.Subsystem{
+		cgroups.NewCputset(root),
+		cgroups.NewCpu(root),
+		cgroups.NewCpuacct(root),
 	}
-	return nil
+	return cgroupsSubsystems(subsystems)
 }
 
-func (s *Sandbox) destroyCgroups() error {
-	if s.cgroup == nil {
-		s.Logger().Warningf("cgroup is not initialized, no need to destroy")
+// V1NoConstraints returns the cgroups that are *not* compatible with th VC
+// architecture and hypervisor, constraints MUST NOT be applied to these cgroups.
+func V1NoConstraints() ([]cgroups.Subsystem, error) {
+	root, err := cgroupV1MountPoint()
+	if err != nil {
+		return nil, err
+	}
+	subsystems := []cgroups.Subsystem{
+		// Some constainers managers, like k8s, take the control of cgroups.
+		// k8s: the memory cgroup for the dns containers is small to place
+		// a hypervisor there.
+		cgroups.NewMemory(root),
+	}
+	return cgroupsSubsystems(subsystems)
+}
+
+func cgroupsSubsystems(subsystems []cgroups.Subsystem) ([]cgroups.Subsystem, error) {
+	var enabled []cgroups.Subsystem
+	for _, s := range cgroupPathers(subsystems) {
+		// check and remove the default groups that do not exist
+		if _, err := os.Lstat(s.Path("/")); err == nil {
+			enabled = append(enabled, s)
+		}
+	}
+	return enabled, nil
+}
+
+func cgroupPathers(subystems []cgroups.Subsystem) []cgroupPather {
+	var out []cgroupPather
+	for _, s := range subystems {
+		if p, ok := s.(cgroupPather); ok {
+			out = append(out, p)
+		}
+	}
+	return out
+}
+
+// v1MountPoint returns the mount point where the cgroup
+// mountpoints are mounted in a single hiearchy
+func cgroupV1MountPoint() (string, error) {
+	f, err := os.Open("/proc/self/mountinfo")
+	if err != nil {
+		return "", err
+	}
+	defer f.Close()
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		if err := scanner.Err(); err != nil {
+			return "", err
+		}
+		var (
+			text   = scanner.Text()
+			fields = strings.Split(text, " ")
+			// safe as mountinfo encodes mountpoints with spaces as \040.
+			index               = strings.Index(text, " - ")
+			postSeparatorFields = strings.Fields(text[index+3:])
+			numPostFields       = len(postSeparatorFields)
+		)
+		// this is an error as we can't detect if the mount is for "cgroup"
+		if numPostFields == 0 {
+			return "", fmt.Errorf("Found no fields post '-' in %q", text)
+		}
+		if postSeparatorFields[0] == "cgroup" {
+			// check that the mount is properly formated.
+			if numPostFields < 3 {
+				return "", fmt.Errorf("Error found less than 3 fields post '-' in %q", text)
+			}
+			return filepath.Dir(fields[4]), nil
+		}
+	}
+	return "", cgroups.ErrMountPointNotExist
+}
+
+func cgroupNoConstraintsPath(path string) string {
+	return filepath.Join(cgroupKataPath, path)
+}
+
+// return the parent cgroup for the given path
+func parentCgroup(path string) (cgroups.Cgroup, error) {
+	// append '/' just in case CgroupsPath doesn't start with it
+	parent := filepath.Dir("/" + path)
+
+	parentCgroup, err := cgroupsLoadFunc(cgroups.V1,
+		cgroups.StaticPath(parent))
+	if err != nil {
+		return nil, fmt.Errorf("Could not load parent cgroup %v: %v", parent, err)
+	}
+
+	return parentCgroup, nil
+}
+
+func (s *Sandbox) updateCgroups() error {
+	if s.state.CgroupPath == "" {
+		s.Logger().Warn("sandbox's cgroup won't be updated: cgroup path is empty")
 		return nil
 	}
 
-	// first move all processes in subgroup to parent in case live process blocks
-	// deletion of cgroup
-	if err := s.cgroup.sandboxSub.MoveTo(s.cgroup.commonParent); err != nil {
-		return fmt.Errorf("failed to clear cgroup processes")
+	cgroup, err := cgroupsLoadFunc(V1Constraints, cgroups.StaticPath(s.state.CgroupPath))
+	if err != nil {
+		return fmt.Errorf("Could not load cgroup %v: %v", s.state.CgroupPath, err)
 	}
 
-	return s.cgroup.sandboxSub.Delete()
-}
-
-func (s *Sandbox) setupCgroups() error {
-	if s.cgroup == nil {
-		return fmt.Errorf("failed to setup uninitialized cgroup for sandbox")
+	if err := s.constrainHypervisor(cgroup); err != nil {
+		return err
 	}
 
-	resource, err := s.mergeSpecResource()
+	if len(s.containers) <= 1 {
+		// nothing to update
+		return nil
+	}
+
+	resources, err := s.resources()
 	if err != nil {
 		return err
 	}
 
-	if err := s.applyCPUCgroup(resource); err != nil {
-		return err
+	if err := cgroup.Update(&resources); err != nil {
+		return fmt.Errorf("Could not update cgroup %v: %v", s.state.CgroupPath, err)
 	}
+
 	return nil
 }
 
-func (s *Sandbox) applyCPUCgroup(rc *specs.LinuxResources) error {
-	if s.cgroup == nil {
-		return fmt.Errorf("failed to setup uninitialized cgroup for sandbox")
+func (s *Sandbox) deleteCgroups() error {
+	s.Logger().Debug("Deleting sandbox cgroup")
+
+	path := cgroupNoConstraintsPath(s.state.CgroupPath)
+	s.Logger().WithField("path", path).Debug("Deleting no constraints cgroup")
+	noConstraintsCgroup, err := cgroupsLoadFunc(V1NoConstraints, cgroups.StaticPath(path))
+	if err == cgroups.ErrCgroupDeleted {
+		// cgroup already deleted
+		return nil
 	}
 
-	// apply cpu constraint to vcpu cgroup
-	if err := s.cgroup.vcpuSub.Update(rc); err != nil {
-		return err
+	if err != nil {
+		return fmt.Errorf("Could not load cgroup without constraints %v: %v", path, err)
+	}
+
+	// move running process here, that way cgroup can be removed
+	parent, err := parentCgroup(path)
+	if err != nil {
+		// parent cgroup doesn't exist, that means there are no process running
+		// and the no constraints cgroup was removed.
+		s.Logger().WithError(err).Warn("Parent cgroup doesn't exist")
+		return nil
+	}
+
+	if err := noConstraintsCgroup.MoveTo(parent); err != nil {
+		// Don't fail, cgroup can be deleted
+		s.Logger().WithError(err).Warn("Could not move process from no constraints to parent cgroup")
+	}
+
+	return noConstraintsCgroup.Delete()
+}
+
+func (s *Sandbox) constrainHypervisor(cgroup cgroups.Cgroup) error {
+	pid := s.hypervisor.pid()
+	if pid <= 0 {
+		return fmt.Errorf("Invalid hypervisor PID: %d", pid)
+	}
+
+	// Move hypervisor into cgroups without constraints,
+	// those cgroups are not yet supported.
+	resources := &specs.LinuxResources{}
+	path := cgroupNoConstraintsPath(s.state.CgroupPath)
+	noConstraintsCgroup, err := cgroupsNewFunc(V1NoConstraints, cgroups.StaticPath(path), resources)
+	if err != nil {
+		return fmt.Errorf("Could not create cgroup %v: %v", path, err)
+	}
+
+	if err := noConstraintsCgroup.Add(cgroups.Process{Pid: pid}); err != nil {
+		return fmt.Errorf("Could not add hypervisor PID %d to cgroup %v: %v", pid, path, err)
 	}
 
 	// when new container joins, new CPU could be hotplugged, so we
@@ -103,33 +227,21 @@ func (s *Sandbox) applyCPUCgroup(rc *specs.LinuxResources) error {
 	if err != nil {
 		return fmt.Errorf("failed to get thread ids from hypervisor: %v", err)
 	}
-	if tids == nil {
+	if tids == nil || len(tids.vcpus) == 0 {
 		// If there's no tid returned from the hypervisor, this is not
 		// a bug. It simply means there is nothing to constrain, hence
 		// let's return without any error from here.
 		return nil
 	}
 
-	// use Add() to add vcpu thread to s.cgroup, it will write thread id to
-	// `cgroup.procs` which will move all threads in qemu process to this cgroup
-	// immediately as default behaviour.
-	if len(tids.vcpus) > 0 {
-		if err := s.cgroup.sandboxSub.Add(cgroups.Process{
-			Pid: tids.vcpus[0],
-		}); err != nil {
-			return err
-		}
-	}
-
+	// We are about to move just the vcpus (threads) into cgroups with constraints.
+	// Move whole hypervisor process whould be easier but the IO/network performance
+	// whould be impacted.
 	for _, i := range tids.vcpus {
-		if i <= 0 {
-			continue
-		}
-
 		// In contrast, AddTask will write thread id to `tasks`
 		// After this, vcpu threads are in "vcpu" sub-cgroup, other threads in
 		// qemu will be left in parent cgroup untouched.
-		if err := s.cgroup.vcpuSub.AddTask(cgroups.Process{
+		if err := cgroup.AddTask(cgroups.Process{
 			Pid: i,
 		}); err != nil {
 			return err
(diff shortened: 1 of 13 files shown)


devimc:
@WeiZhang555 please take a look

devimc:
@jcvenegas please take a look

virtcontainers/container.go
```diff
@@ -762,7 +764,13 @@ func (c *Container) create() (err error) {
 	}
 	c.process = *process
 
-	// If this is a sandbox container, store the pid for sandbox
+	cgroupsPath, err := c.newCgroups()
+	if err != nil {
+		return
+	}
+	c.state.CgroupPath = cgroupsPath
```
WeiZhang555: Why don't put this line into `c.newCgroups()` ?

virtcontainers/cgroups.go
```diff
@@ -1,197 +1,251 @@
...
+	if err := noConstraintsCgroup.Add(cgroups.Process{Pid: pid}); err != nil {
+		return fmt.Errorf("Could not add hypervisor PID %d to cgroup %v: %v", pid, path, err)
 	}
 
-	for _, i := range tids.vcpus {
-		if i <= 0 {
-			continue
-		}
+	// Add qemu into cgroup
+	return cgroup.Add(cgroups.Process{Pid: pid})
```
WeiZhang555: `pid` is added both to `noConstraintsCgroup` and `cgroup`, in the end only the last `Add` can take effect.
So I think `noConstraintsCgroup` isn't necessary and can be removed.

By the way, I didn't see `noConstraintsCgroup` created on my disk, not sure which part I missed.

virtcontainers/cgroups.go
```diff
@@ -1,197 +1,251 @@
...
-			return err
-		}
+	if err := noConstraintsCgroup.Add(cgroups.Process{Pid: pid}); err != nil {
+		return fmt.Errorf("Could not add hypervisor PID %d to cgroup %v: %v", pid, path, err)
 	}
 
-	for _, i := range tids.vcpus {
-		if i <= 0 {
-			continue
-		}
```
WeiZhang555: Adding only the `vcpus` threads is an optimization, which can benefit to small chunks IO performance according to our previous test.

It's OK to remove this if it makes the handling logic too complicated, I'm just wishing to share the performance tuning  result to community.

virtcontainers/cgroups.go
```diff
@@ -1,197 +1,251 @@
...
+			out = append(out, p)
+		}
 	}
+	return out
+}
 
-	// first move all processes in subgroup to parent in case live process blocks
-	// deletion of cgroup
-	if err := s.cgroup.sandboxSub.MoveTo(s.cgroup.commonParent); err != nil {
-		return fmt.Errorf("failed to clear cgroup processes")
```
WeiZhang555: In my previous test, a left-over process (zoombie or removed too slowly) can block the cgroup's deletion which will cause cgroup dir left on disk after sandbox destroyed. 

Have you tested this case? I don't see similar code in `deleteCgroups`

virtcontainers/cgroups.go
```diff
@@ -1,197 +1,251 @@
...
+			}
+
+			if c.state.Resources.CPU.Quota != nil {
+				quota += *c.state.Resources.CPU.Quota
+			}
+
+			if c.state.Resources.CPU.Period != nil {
+				period = uint64(math.Max(float64(*c.state.Resources.CPU.Period), float64(period)))
+			}
+
```
WeiZhang555: Missing `RealtimePeriod` and `RealtimeRuntime` ?

virtcontainers/types/sandbox.go
```diff
@@ -44,6 +46,16 @@ type State struct {
 
 	// GuestMemoryBlockSizeMB is the size of memory block of guestos
 	GuestMemoryBlockSizeMB uint32 `json:"guestMemoryBlockSize"`
+
+	// CgroupPath is the cgroup hierarchy where sandbox's processes
+	// including the hypervisor are placed.
+	CgroupPath string `json:"cgroupPath,omitempty"`
```
WeiZhang555: This can bring potential broken compatibility.

Suppose sandboxA is created by old kata-runtime and runs for some time, after we upgrade kata-runtime, and when using the new kata-runtime to delete the old container, it will search `state.json` file from disk and end up to find nothing, The cgroup dir can fail to delete and garbage left.

That's why `https://github.com/kata-containers/runtime/pull/883` is important.

The influence is not that big caused by this modification, so I can accept the modification so far.

virtcontainers/cgroups.go
devimc: > So I think noConstraintsCgroup isn't necessary and can be removed.

no, it's needed, since memory cgroup provided by k8s is too small (~50 mb) to run a VM, currently we don't see any failure because we are not honouring the `CgroupsPath` and our memory cgroup doesn't have any constraint

> By the way, I didn't see noConstraintsCgroup created on my disk, not sure which part I missed.

With this patch, only memory cgroup is part of `noConstraintsCgroup` and it's created under `kata` directory, for example `/sys/fs/cgroup/memory/kata/kubepods/burstable/poddbbadeac-2326-11e9-9b35-525400123456/crio-c44e31931c5b67260c977bfeb49d0fd6b2ebe0ff618e201115bf71d4eba9a29b`

virtcontainers/container.go
```diff
@@ -1246,3 +1262,107 @@ func (c *Container) detachDevices() error {
...
+	}
+
+	var spec specs.Spec
+	if err := json.Unmarshal([]byte(config), &spec); err != nil {
+		return "", err
+	}
+
+	// https://github.com/kata-containers/runtime/issues/168
+	resources := specs.LinuxResources{
+		CPU: cpuResources(spec.Linux.Resources.CPU),
```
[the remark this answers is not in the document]
devimc: good catch!, thanks

virtcontainers/container.go
devimc: sounds good

virtcontainers/cgroups.go
devimc: sure I can check that, what tests did you run?

virtcontainers/cgroups.go
devimc: I ran a simple test, and seems like cgroups are removed properly

virtcontainers/cgroups.go
devimc: fixed

devimc:
@WeiZhang555 changes applied, thanks

virtcontainers/container.go
```diff
@@ -1246,3 +1260,88 @@ func (c *Container) detachDevices() error {
 	}
 	return nil
 }
+
+// creates a new cgroup and return the cgroups path
+func (c *Container) newCgroups() error {
+	ann := c.GetAnnotations()
+
+	config, ok := ann[annotations.ConfigJSONKey]
```
jcvenegas: We use in many places the config json, at some point better to move it from annotations, virtcontainers is not agnostic at all.

virtcontainers/cgroups.go
```diff
@@ -1,197 +1,293 @@
...
+// for example /sys/fs/cgroup/memory/kata/$CGPATH
+// where path is defined by the containers manager
+const cgroupKataPath = "/kata/"
 
-	// create sub-cgroup for each sandbox
-	// e.g. /sys/fs/cgroup/cpu/vc/<sandbox>
-	sandboxSub, err := parent.New(s.id, &specs.LinuxResources{})
+// V1Constraints returns the cgroups that are compatible with th VC architecture
+// and hypervisor, constraints can be applied to these cgroups.
+func V1Constraints() ([]cgroups.Subsystem, error) {
```
jcvenegas: I dont fully get the naming of the cunftions  it is because of cgroups v1? I prefer remove that prefix if does not add any information.

virtcontainers/cgroups.go
```diff
@@ -1,197 +1,293 @@
...
+	for scanner.Scan() {
+		if err := scanner.Err(); err != nil {
+			return "", err
+		}
+		var (
+			text   = scanner.Text()
+			fields = strings.Split(text, " ")
+			// safe as mountinfo encodes mountpoints with spaces as \040.
+			index               = strings.Index(text, " - ")
+			postSeparatorFields = strings.Fields(text[index+3:])
```
jcvenegas: Magic number

virtcontainers/cgroups.go
```diff
@@ -1,197 +1,293 @@
...
+			numPostFields       = len(postSeparatorFields)
+		)
+		// this is an error as we can't detect if the mount is for "cgroup"
+		if numPostFields == 0 {
+			return "", fmt.Errorf("Found no fields post '-' in %q", text)
+		}
+		if postSeparatorFields[0] == "cgroup" {
```
jcvenegas: Index access without check len

virtcontainers/cgroups.go
```diff
@@ -1,197 +1,293 @@
...
+			text   = scanner.Text()
+			fields = strings.Split(text, " ")
+			// safe as mountinfo encodes mountpoints with spaces as \040.
+			index               = strings.Index(text, " - ")
+			postSeparatorFields = strings.Fields(text[index+3:])
+			numPostFields       = len(postSeparatorFields)
+		)
+		// this is an error as we can't detect if the mount is for "cgroup"
+		if numPostFields == 0 {
+			return "", fmt.Errorf("Found no fields post '-' in %q", text)
```
jcvenegas: Not really useful error (at least for me)

virtcontainers/cgroups.go
```diff
@@ -1,197 +1,293 @@
...
+		}
+		if postSeparatorFields[0] == "cgroup" {
+			// check that the mount is properly formated.
+			if numPostFields < 3 {
```
jcvenegas: Magic number

virtcontainers/cgroups.go
```diff
@@ -1,197 +1,293 @@
...
+				return "", fmt.Errorf("Error found less than 3 fields post '-' in %q", text)
+			}
+			return filepath.Dir(fields[4]), nil
```
jcvenegas: Index access with out check length

virtcontainers/cgroups.go
```diff
@@ -1,197 +1,293 @@
...
+		}
+	}
+	return "", cgroups.ErrMountPointNotExist
```
jcvenegas: I wonder if cgroups api does not provide a function like this.
From this file what is string is supose to return ?
```
cat /proc/self/mountinfo  | grep cgroup
27 20 0:24 / /sys/fs/cgroup ro,nosuid,nodev,noexec shared:4 - tmpfs tmpfs ro,seclabel,mode=755
28 27 0:25 / /sys/fs/cgroup/unified rw,nosuid,nodev,noexec,relatime shared:5 - cgroup2 cgroup2 rw,seclabel,nsdelegate
29 27 0:26 / /sys/fs/cgroup/systemd rw,nosuid,nodev,noexec,relatime shared:6 - cgroup cgroup rw,seclabel,xattr,name=systemd
33 27 0:30 / /sys/fs/cgroup/cpu,cpuacct rw,nosuid,nodev,noexec,relatime shared:7 - cgroup cgroup rw,seclabel,cpu,cpuacct
34 27 0:31 / /sys/fs/cgroup/freezer rw,nosuid,nodev,noexec,relatime shared:8 - cgroup cgroup rw,seclabel,freezer
35 27 0:32 / /sys/fs/cgroup/blkio rw,nosuid,nodev,noexec,relatime shared:9 - cgroup cgroup rw,seclabel,blkio
36 27 0:33 / /sys/fs/cgroup/cpuset rw,nosuid,nodev,noexec,relatime shared:10 - cgroup cgroup rw,seclabel,cpuset
37 27 0:34 / /sys/fs/cgroup/devices rw,nosuid,nodev,noexec,relatime shared:11 - cgroup cgroup rw,seclabel,devices
38 27 0:35 / /sys/fs/cgroup/perf_event rw,nosuid,nodev,noexec,relatime shared:12 - cgroup cgroup rw,seclabel,perf_event
39 27 0:36 / /sys/fs/cgroup/hugetlb rw,nosuid,nodev,noexec,relatime shared:13 - cgroup cgroup rw,seclabel,hugetlb
40 27 0:37 / /sys/fs/cgroup/memory rw,nosuid,nodev,noexec,relatime shared:14 - cgroup cgroup rw,seclabel,memory
41 27 0:38 / /sys/fs/cgroup/net_cls,net_prio rw,nosuid,nodev,noexec,relatime shared:15 - cgroup cgroup rw,seclabel,net_cls,net_prio
42 27 0:39 / /sys/fs/cgroup/pids rw,nosuid,nodev,noexec,relatime shared:16 - cgroup cgroup rw,seclabel,pids

```

virtcontainers/cgroups.go
devimc: This implementation is based on containerd cgroups https://github.com/kata-containers/runtime/blob/master/vendor/github.com/containerd/cgroups/v1.go#L28

virtcontainers/cgroups.go
devimc: copy&paste of a containerd private function https://github.com/kata-containers/runtime/blob/master/vendor/github.com/containerd/cgroups/v1.go#L49

virtcontainers/cgroups.go
devimc: https://github.com/kata-containers/runtime/blob/master/vendor/github.com/containerd/cgroups/v1.go#L74

virtcontainers/cgroups.go
devimc: https://github.com/kata-containers/runtime/blob/master/vendor/github.com/containerd/cgroups/v1.go#L77

virtcontainers/cgroups.go
devimc: it does, but it's private https://github.com/kata-containers/runtime/blob/master/vendor/github.com/containerd/cgroups/v1.go#L49

devimc:
ping @raravena80

raravena80:
@devimc did you mean to ping @WeiZhang555? 😄

devimc:
@raravena80 nop, this needs review :smile:

devimc:
@raravena80 thanks

devimc:
Shimv2 tests are failing with this change

```
STEP: stop PodSandbox
E0206 21:45:40.234455   28101 remote_runtime.go:115] StopPodSandbox "762c340ac84f4b554cfbee6b8fd058d9b7dfbd832dd30b77acf6e8b2431a4d54" from runtime service failed: rpc error: code = Unknown desc = failed to stop container "91cd574ea84883dc442e90fb04f1a0af1880437212bf53720825c087f2f3054f": failed to kill container "91cd574ea84883dc442e90fb04f1a0af1880437212bf53720825c087f2f3054f": Could not find the container "91cd574ea84883dc442e90fb04f1a0af1880437212bf53720825c087f2f3054f" from the sandbox "762c340ac84f4b554cfbee6b8fd058d9b7dfbd832dd30b77acf6e8b2431a4d54" containers list: unknown
STEP: delete PodSandbox
E0206 21:45:40.236155   28101 remote_runtime.go:229] StopContainer "91cd574ea84883dc442e90fb04f1a0af1880437212bf53720825c087f2f3054f" from runtime service failed: rpc error: code = Unknown desc = failed to kill container "91cd574ea84883dc442e90fb04f1a0af1880437212bf53720825c087f2f3054f": Could not find the container "91cd574ea84883dc442e90fb04f1a0af1880437212bf53720825c087f2f3054f" from the sandbox "762c340ac84f4b554cfbee6b8fd058d9b7dfbd832dd30b77acf6e8b2431a4d54" containers list: unknown
Feb  6 21:45:40.236: INFO: Unexpected error occurred: rpc error: code = Unknown desc = failed to kill container "91cd574ea84883dc442e90fb04f1a0af1880437212bf53720825c087f2f3054f": Could not find the container "91cd574ea84883dc442e90fb04f1a0af1880437212bf53720825c087f2f3054f" from the sandbox "762c340ac84f4b554cfbee6b8fd058d9b7dfbd832dd30b77acf6e8b2431a4d54" containers list: unknown
E0206 21:45:40.236587   28101 remote_runtime.go:132] RemovePodSandbox "762c340ac84f4b554cfbee6b8fd058d9b7dfbd832dd30b77acf6e8b2431a4d54" from runtime service failed: rpc error: code = Unknown desc = sandbox container "762c340ac84f4b554cfbee6b8fd058d9b7dfbd832dd30b77acf6e8b2431a4d54" is not fully stopped
[AfterEach] [k8s.io] Container
  /tmp/cri-install-crictl.sDoz/src/github.com/kubernetes-sigs/cri-tools/pkg/framework/framework.go:51


• Failure [67.015 seconds]
[k8s.io] Container
/tmp/cri-install-crictl.sDoz/src/github.com/kubernetes-sigs/cri-tools/pkg/framework/framework.go:72
  runtime should support basic operations on container
  /tmp/cri-install-crictl.sDoz/src/github.com/kubernetes-sigs/cri-tools/pkg/validate/container.go:68
    runtime should support stopping container [Conformance] [It]
    /tmp/cri-install-crictl.sDoz/src/github.com/kubernetes-sigs/cri-tools/pkg/validate/container.go:100

    Feb  6 21:45:40.229: stop container "91cd574ea84883dc442e90fb04f1a0af1880437212bf53720825c087f2f3054f" timeout.
```

I'm trying to debug it but I don't see logs in the journal.  @lifupan any thoughts?

devimc:
@lifupan seem like if the hypervisor is moved into a cgroup then tests will fail :thinking:

virtcontainers/container.go
WeiZhang555: Agree. We can definitely save the config in `Container` . This way we can save many disk read and json Unmarshaling.

virtcontainers/cgroups.go
```diff
@@ -1,197 +1,310 @@
...
+// V1Constraints returns the cgroups that are compatible with th VC architecture
+// and hypervisor, constraints can be applied to these cgroups.
+func V1Constraints() ([]cgroups.Subsystem, error) {
+	root, err := cgroupV1MountPoint()
 	if err != nil {
-		return fmt.Errorf("failed to create cgroup for %s/%s", defaultCgroupParent, s.id)
+		return nil, err
+	}
+	subsystems := []cgroups.Subsystem{
+		cgroups.NewCputset(root),
```
WeiZhang555: Cputset or cpuset?

virtcontainers/cgroups.go
WeiZhang555: I get your point. This sounds reasonable.

virtcontainers/cgroups.go
WeiZhang555: You can test with some network benchmark tools such as netperf, or disk IO benchmark tools such as FIO or iozone.  Try to send/receive with small chunks(e.g. 4k package), for one VM with 1 cpu, you can compare W/ or W/O "cpu cgroup of 1 cpu constraint", I believe you can see how the performance is different.

virtcontainers/cgroups.go
devimc: Cputset is correct https://github.com/kata-containers/runtime/blob/master/vendor/github.com/containerd/cgroups/cpuset.go#L29

virtcontainers/sandbox.go
```diff
@@ -1462,6 +1469,20 @@ func (s *Sandbox) setSandboxPid(pid int) error {
 	return s.storage.storeSandboxResource(s.id, stateFileType, s.state)
 }
 
+// setCgroupPath sets the cgroup for the sandbox, sandbox's processes including
+// the hypervisor are placed there.
+func (s *Sandbox) setCgroupPath(c *Container) error {
+	if s.state.Pid != c.process.Pid {
```
lifupan: For shimv2, there is no kata shim process for a container, and the container's process pid is set to "-1", the sandbox's state Pid is also set "-1", thus here we cannot assert whether a container is sandbox container by test "s.state.Pid != c.process.Pid".  it's better to use  the line https://github.com/kata-containers/runtime/blob/1c27ab79d9ce76202aade8b2124a003363419fc3/virtcontainers/container.go#L764 to assert.  Or you can move calling the setCgroupPath() to there.

virtcontainers/container.go
```diff
@@ -791,6 +797,10 @@ func (c *Container) delete() error {
 		return err
 	}
 
+	if err := c.deleteCgroups(); err != nil {
```
lifupan: For a sandbox container, the hypervisor process has been set in its cgroup, before deleteCgroups, you should remove the tasks from it.

lifupan:
> @lifupan seem like if the hypervisor is moved into a cgroup then tests will fail 🤔

Hi @devimc  See my inline comments,  I think my comments can address the errors occurred on shimv2.

virtcontainers/cgroups.go
WeiZhang555: I wonder this is a typo of `cgroups` package, I can't understand what this "t" means...

devimc:
@lifupan thank you! , it works

virtcontainers/cgroups.go
devimc: yes, cgroups are removed, I filed an issue to test it once this pr land

devimc: issue https://github.com/kata-containers/tests/issues/1177

virtcontainers/cgroups.go
devimc: you're right, performance is not good moving the whole hypervisor, I have updated the pull request, now the performance is good, thanks

cgroup1

cgroup2


current - current implementation
new - new implementation
no-cgroup - container without constraints

devimc:
@jcvenegas @WeiZhang555 @lifupan  changes applied, ready for review, thank you!

devimc:
ping @kata-containers/runtime

devimc:
@WeiZhang555 @lifupan @jcvenegas please take a look

virtcontainers/container.go
```diff
@@ -1243,3 +1251,104 @@ func (c *Container) detachDevices() error {
...
+	cgroup, err := cgroupsNewFunc(cgroups.V1,
+		cgroups.StaticPath(spec.Linux.CgroupsPath), &resources)
+	if err != nil {
+		return fmt.Errorf("Could not create cgroup for %v: %v", spec.Linux.CgroupsPath, err)
+	}
+
+	// Add shim into cgroup
+	if c.process.Pid > 0 {
+		if err := cgroup.Add(cgroups.Process{Pid: c.process.Pid}); err != nil {
+			return fmt.Errorf("Could not add PID %d to cgroup %v: %v", c.process.Pid, spec.Linux.CgroupsPath, err)
```
jcvenegas: What about deleted cgroup if the operation was not successful ? In a defer maybe?

devimc: I'd prefer to leave it there and delete it once `delete` command is used, otherwise we'll face some issues with tools that monitor the cgroups

jcvenegas: In that case, I wonder should not delay `c.state.CgroupPath = spec.Linux.CgroupsPath` to the end ? If return intermediately we dont save the path in the state.

devimc: you're right, fixing, thanks

jcvenegas:
lgtm

sboeuf:
Oh no... I wanted to review this, and I was expecting more approval for this PR since that is a pretty important/big change...

virtcontainers/container.go
```diff
@@ -928,6 +928,13 @@ func (c *Container) stop() error {
 	// get failed if the process hasn't exited.
 	c.sandbox.agent.waitProcess(c, c.id)
 
+	// container was killed by force, container MUST change its state
+	// as soon as possible just in case one of below operations fail leaving
+	// the containers in a bad state.
+	if err := c.setContainerState(types.StateStopped); err != nil {
```
sboeuf: How is it related to the PR? I'm not saying this is not fixing something, but it's been hidden into the PR that is supposed to fix cgroups, not the lifecycle of a container. This is the type of changes that can have a big impact and that is good to get reviewed standalone through its own PR.

virtcontainers/kata_agent.go
```diff
@@ -1569,6 +1569,13 @@ func (k *kataAgent) sendReq(request interface{}) (interface{}, error) {
 	span.SetTag("request", request)
 	defer span.Finish()
 
+	if k.state.ProxyPid > 0 {
```
sboeuf: Same thing here, I don't think this is related to cgroups, and this should not be part the same PR.

virtcontainers/qemu.go
```diff
@@ -1567,3 +1571,23 @@ func (q *qemu) cleanup() error {
 
 	return nil
 }
+
+func (q *qemu) pidFile() string {
+	return filepath.Join(store.RunVMStoragePath, q.id, "pid")
+}
+
+func (q *qemu) pid() int {
+	data, err := ioutil.ReadFile(q.pidFile())
```
sboeuf: The PID has been stored through its own file? Don't you think we could improve this by having the PID part of the `qemu.state` structure? This would be saved as part of the `state.json` file associated with the VM, and it would avoid yet another file to be created.

Issue #1204, opened 2019-02-01
Implement the Shimv2 async task model for kata


lifupan:
# Description of problem

The Runtime v2 supports an async event model. In order for the an upstream
caller (such as Docker) to get these events in the correct order a Runtime
v2 shim MUST implement the following events:

#### Tasks

| Topic | Compliance | Description |
| ----- | ---------- | ----------- |
| `runtime.TaskCreateEventTopic`       | MUST                                                                          | When a task is successfully created |
| `runtime.TaskStartEventTopic`        | MUST (follow `TaskCreateEventTopic`)                                          | When a task is successfully started |
| `runtime.TaskExitEventTopic`         | MUST (follow `TaskStartEventTopic`)                                           | When a task exits expected or unexpected |
| `runtime.TaskDeleteEventTopic`       | MUST (follow `TaskExitEventTopic` or `TaskCreateEventTopic` if never started) | When a task is removed from a shim |
| `runtime.TaskPausedEventTopic`       | SHOULD                                                                        | When a task is successfully paused |
| `runtime.TaskResumedEventTopic`      | SHOULD (follow `TaskPausedEventTopic`)                                        | When a task is successfully resumed |
| `runtime.TaskCheckpointedEventTopic` | SHOULD                                                                        | When a task is checkpointed |
| `runtime.TaskOOMEventTopic`          | SHOULD                                                                        | If the shim collects Out of Memory events |

#### Execs

| Topic | Compliance | Description |
| ----- | ---------- | ----------- |
| `runtime.TaskExecAddedEventTopic`   | MUST (follow `TaskCreateEventTopic` )     | When an exec is successfully added |
| `runtime.TaskExecStartedEventTopic` | MUST (follow `TaskExecStartedEventTopic`) | When an exec is successfully started |

For much more info, please see:
https://github.com/containerd/containerd/blob/master/runtime/v2/README.md#events

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

sboeuf:
Hi @lifupan, from an API perspective, it's not clear why this is needed. For instance, let's take the container creation example. When we ask for a container to be created, send the gRPC request `CreateTaskRequest`. This will return only when the creation of the container is done, so no need for any `TaskEvent` IMO.

[...]

PR #1205, opened 2019-02-01
shimv2: Send async task events to containerd/cri

lifupan:
The Runtime v2 supports an async event model. In order for the an upstream
caller (such as Docker) to get these events in the correct order a Runtime
v2 shim MUST implement some events.

For much more info, please see:
https://github.com/containerd/containerd/blob/master/runtime/v2/README.md#events

Fixes:#1204

### DIFF
--- a/containerd-shim-v2/service.go
+++ b/containerd-shim-v2/service.go
@@ -91,7 +91,8 @@ type exit struct {
 
 // service is the shim implementation of a remote shim over GRPC
 type service struct {
-	sync.Mutex
+	mu          sync.Mutex
+	eventSendMu sync.Mutex
 
 	// pid Since this shimv2 cannot get the container processes pid from VM,
 	// thus for the returned values needed pid, just return this shim's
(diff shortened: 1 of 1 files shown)


containerd-shim-v2/service.go
```diff
@@ -328,44 +344,73 @@ func (s *service) Create(ctx context.Context, r *taskAPI.CreateTaskRequest) (_ *
...
+	defer s.mu.Unlock()
 
 	c, err := s.getContainer(r.ID)
 	if err != nil {
 		return nil, err
 	}
 
+	// hold the send lock so that the start events are sent before any exit events in the error case
+	s.eventSendMu.Lock()
+
```
Ace-Tang: If I not miss something, can we use `defer s.eventSendMu.Unlock()` here ?

lifupan: Yes, of course. 
Thanks for the reminder.

containerd-shim-v2/service.go
```diff
@@ -760,20 +823,22 @@ func (s *service) processExits() {
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
 
 	id := e.execid
 	if id == "" {
 		id = e.id
 	}
-	s.events <- &eventstypes.TaskExit{
+
+	s.sendL(&eventstypes.TaskExit{
```
Ace-Tang: Another confused me is there is two function for send event, `send`(// for unit test, it will not initialize s.events) and `sendL`,  `sendL` is only used here, so that means `Start` and `checkProcesses` need lock, but other function do not need, a little confused.

lifupan: Yes, here the lock is mainly used for preventing the case of start events are sent before any exit events in the error case.
For example, shimv2 start a container, but for some reasons the start failed and the container process exit immediately, without this lock, we cannot guarantee which event send first, the create event or the exit event? So this lock is used to make sure the create event is send first.

teawater:
LGTM

Issue #1215, opened 2019-02-06
Remove .pullapprove.yaml


grahamwhaley:
We are moving off [pullapprove](https://pullapprove.com/pulls/).
https://github.com/kata-containers/ci/issues/98

Thus, remove the `.pullapprove.yaml` in this repo.

PR #1216, opened 2019-02-07
pullapprove: remove it

grahamwhaley:
We are moving off pullapprove. Remove its config file.

Fixes: #1215

Signed-off-by: Graham Whaley <graham.whaley@intel.com>


jodh-intel:
We don't need CI results to pass here - we're simply removing a config file. Merging...

jodh-intel:
Webhook removed.

Issue #1219, opened 2019-02-07
delete: force: Do not fail on non exiting container


jcvenegas:
# Description of problem

When a container does not exist, runc does not fail.  Lets mimic this
behavior, sometimes kuberentes will try to force delete containers that
could not be created and gets confused if delete --force fails.
    

# Expected result

```
kata-runtime delete --force THIS_CONTAINER_NOT_EXIST 
echo $?
1
```

# Actual result

```
kata-runtime delete --force THIS_CONTAINER_NOT_EXIST 
echo $?
0
```

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #1220, opened 2019-02-07
delete: force: Do not fail on non exiting container

jcvenegas:
When a container does not exist, runc does not fail.  Lets mimic this
behavior, sometimes kuberentes will try to force delete containers that
could not be created and gets confused if delete --force fails.

Fixes: #1219

### DIFF
--- a/cli/delete.go
+++ b/cli/delete.go
@@ -71,6 +71,10 @@ func delete(ctx context.Context, containerID string, force bool) error {
 	// Checks the MUST and MUST NOT from OCI runtime specification
 	status, sandboxID, err := getExistingContainerInfo(ctx, containerID)
 	if err != nil {
+		if force {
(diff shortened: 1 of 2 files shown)


cli/delete.go
```diff
@@ -71,7 +71,12 @@ func delete(ctx context.Context, containerID string, force bool) error {
 	// Checks the MUST and MUST NOT from OCI runtime specification
 	status, sandboxID, err := getExistingContainerInfo(ctx, containerID)
 	if err != nil {
-		return err
+		if force {
+			kataLog.Warnf("Failed to get container, force wont fail: %s", err)
```
jodh-intel: I'd say `force will not fail` to avoid an apostrophe in `won't`.

jodh-intel:
I think we need to *use the force* to update the unit-tests for this change to exercise `force=false` and `force=true`.

bergwolf:
LGTM

PR #1232, opened 2019-02-12
versions: use clear linux latest rootfs.

jcvenegas:
Update Clear Linux rootfs to latest.


grahamwhaley:
The one thing we now need to be aware of is that the version of the OS in the rootfs will change, potentially without us obviously knowing or realising. It is not infeasible that one day something will change in the rootfs that might affect the agent say - so, let's just keep that in mind if/when we are chasing down 'something broke' type bugs.

Issue #1234, opened 2019-02-13
virtcontainers: Improve debug messages when deleting files


nitkon:
# Description of problem

Context: https://github.com/kata-containers/tests/issues/1157

Adding debug messages which state which files are being deleted could be helpful in analyzing situations like leaky pod issues.

PR #1235, opened 2019-02-13
virtcontainers: Improve debug messages when deleting files

nitkon:
Adding debug messages which state which files
are being deleted could be helpful in analyzing
situations like leaky pod issues.

Fixes: #1234

### DIFF
--- a/virtcontainers/store/filesystem_backend.go
+++ b/virtcontainers/store/filesystem_backend.go
@@ -152,11 +152,13 @@ func (f *filesystem) initialize() error {
 	// The root directory, a lock file and a raw files directory.
 
 	// Root directory
+	f.logger().WithField("path", f.path).Debugf("Creating root directory")
 	if err := os.MkdirAll(f.path, DirMode); err != nil {
 		return err
 	}
 
 	// Raw directory
+	f.logger().WithField("path", f.rawPath).Debugf("Creating raw directory")
 	if err := os.MkdirAll(f.rawPath, DirMode); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


virtcontainers/store/filesystem_backend.go
```diff
@@ -186,6 +186,7 @@ func (f *filesystem) new(ctx context.Context, path string, host string) error {
 }
 
 func (f *filesystem) delete() error {
+	f.logger().Debugf("Deleting files %s", f.path)
```
devimc: I think something like this is better

`f.logger().WithField("path", f.path).Debugf("Deleting files)`

nitkon:
@grahamwhaley : Updated my PR. _and the ack is gone ;-)_

virtcontainers/store/filesystem_backend.go
```diff
@@ -152,11 +152,13 @@ func (f *filesystem) initialize() error {
 	// The root directory, a lock file and a raw files directory.
 
 	// Root directory
+	f.logger().Debugf("Creating root directory if it already does not exist %s", f.path)
```
grahamwhaley: thx. I think we should take on board @devimc suggestion of using `WithField`.
I'd also rephrase these to drop the 'if it already....' bit, as unless we do the check (stat) and print a status, then I think it will just be confusing noise. So, just `Creating root directory` will be enough (along with the WithField(path)) I think.

nitkon:
Updated my PR @grahamwhaley 
/retest

virtcontainers/store/filesystem_backend.go
nitkon: Thanks @devimc !

nitkon:
Good to merge.

Issue #1292, opened 2019-02-27
Allocated vCPUs doesn't match expectations from podspec


kennymacleod:
# Description of problem

Under Kata 1.5 and Kube 1.12, the number of VCPUs allocated to the pod's VM doesn't seem to correspond to the CPU resource restrictions in the pod spec. - but I only see 5.

# Expected result

For example, if I have a pod spec with 3 containers that have cpu limit of 8, 4 and none respectively, I'd expect 13 vCPUs in the pod

# Actual result

When I exec into one of the containers in the pod, and run `cat /proc/cpuinfo`, I see 5 vCPUs.

Here's my pod spec:
```
apiVersion: v1
kind: Pod
metadata:
  name: log-explorer
  annotations: 
    io.kubernetes.cri.untrusted-workload: "true"
spec:
  containers:
  - image: busybox:latest
    name: container1
    args:
    - sleep
    - "1000000"
    resources:
      limits:
        cpu: "8"  
  - image: busybox:latest
    name: container2
    args:
    - sleep
    - "1000000"
    resources:
      limits:
        cpu: "4"  
  - image: busybox:latest
    name: container3
    args:
    - sleep
    - "1000000"
  nodeSelector:
    customer: kata
  tolerations:
  - effect: NoSchedule
    key: customer
    operator: Equal
    value: kata    
```

egernst:
I'll work to reproduce this - thanks for filing @kennymacleod

egernst:
I see the issue.  Unfortunately, when we do the summation of container requests, we do a running total for both the period and quota.  This doesn't make sense - for a given period, we should just do a summation of the quota.  

This can be seen at https://github.com/kata-containers/runtime/blob/master/virtcontainers/sandbox.go#L1628

If I simplify your yaml to just include `container1` and `containerd2`, on the iterations we'd see:
```
|               | period | shares | running total period | running total shars | calculation |
|---------------|--------|--------|----------------------|---------------------|-------------|
| 1st iteration | 100000 | 800000 | 100000               | 800000              | 8vCPU       |
| 2nd iteration | 100000 | 400000 | 200000               | 120000              | 6vCPU       |
|               |        |        |                      |                     |             |
```

PR #1294, opened 2019-02-28
virtcontainers: fix vCPU calculation errors

egernst:
We were grabbing a running total of quota and period for each container
and then calculating the number of resulting vCPUs. Summing period
doesn't make sense.  To simplify, let's just calculate mCPU per
container, keep a running total of mCPUs requested, and then translate
to sandbox vCPUs after.

Fixes: #1292

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -1608,7 +1608,6 @@ func (s *Sandbox) AddDevice(info config.DeviceInfo) (api.Device, error) {
 func (s *Sandbox) updateResources() error {
 	// the hypervisor.MemorySize is the amount of memory reserved for
 	// the VM and contaniners without memory limit
-
 	sumResources := specs.LinuxResources{
 		Memory: &specs.LinuxMemory{
 			Limit: new(int64),
@@ -1619,19 +1618,22 @@ func (s *Sandbox) updateResources() error {
 		},
 	}
 
+	var mCPU uint32
+
+	// Calculate running total of memory and mCPUs requested
 	for _, c := range s.config.Containers {
 		if m := c.Resources.Memory; m != nil && m.Limit != nil {
 			*sumResources.Memory.Limit += *m.Limit
 		}
 		if cpu := c.Resources.CPU; cpu != nil {
 			if cpu.Period != nil && cpu.Quota != nil {
-				*sumResources.CPU.Period += *cpu.Period
-				*sumResources.CPU.Quota += *cpu.Quota
+				mCPU += utils.CalculateMilliCPUs(*cpu.Quota, *cpu.Period)
 			}
+
 		}
 	}
 
-	sandboxVCPUs := uint32(utils.ConstraintsToVCPUs(*sumResources.CPU.Quota, *sumResources.CPU.Period))
+	sandboxVCPUs := utils.CalculateVCpusFromMilliCpus(mCPU)
 	sandboxVCPUs += s.hypervisor.hypervisorConfig().NumVCPUs
 
 	sandboxMemoryByte := int64(s.hypervisor.hypervisorConfig().MemorySize) << utils.MibToBytesShift
--- a/virtcontainers/utils/utils.go
+++ b/virtcontainers/utils/utils.go
@@ -105,6 +105,25 @@ func WriteToFile(path string, data []byte) error {
 	return nil
 }
 
+//CalculateMilliCPUs converts CPU quota and period to milli-CPUs
+func CalculateMilliCPUs(quota int64, period uint64) uint32 {
+
+	// If quota is -1, it means the CPU resource request is
+	// unconstrained.  In that case, we don't currently assign
+	// additional CPUs.
+	if quota >= 0 && period != 0 {
+		return uint32((uint64(quota) * 1000) / period)
+	}
+
+	return 0
+}
+
+//CalculateVCpusFromMilliCpus converts from mCPU to CPU, taking the ceiling
+// value when necessary
+func CalculateVCpusFromMilliCpus(mCPU uint32) uint32 {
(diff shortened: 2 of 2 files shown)


egernst:
This is an important fix which is applicable for stable-1.5, but not stable-1.4.

virtcontainers/utils/utils.go
```diff
@@ -105,6 +105,15 @@ func WriteToFile(path string, data []byte) error {
 	return nil
 }
 
+//CalculatemCPUs converts CPU quota and period to milli-CPUs
+func CalculatemCPUs(quota int64, period uint64) uint {
+	if quota != 0 && period != 0 {
+		return uint((uint64(quota) * 100) / period)
```
WeiZhang555: As an int64, `quota` can be -1 to indicate unconstrainted resources. And it's better to also check int overflow as an utils package.

WeiZhang555: Other parts LGTM

virtcontainers/sandbox.go
```diff
@@ -1619,19 +1622,23 @@ func (s *Sandbox) updateResources() error {
...
-				*sumResources.CPU.Quota += *cpu.Quota
+				mCPU += uint32(utils.CalculatemCPUs(*cpu.Quota, *cpu.Period))
 			}
+
 		}
 	}
 
-	sandboxVCPUs := uint32(utils.ConstraintsToVCPUs(*sumResources.CPU.Quota, *sumResources.CPU.Period))
+	//For a sandbox, we need to take the ceiling when converting from mCPU to CPU
+	sandboxVCPUs := (mCPU + 99) / 100
```
bergwolf: Since the math is based on 100, could you please add a helper function for it in utils.go just like `utils.CalculatemCPUs` that is also basing on 100 (milli-CPUs).

bergwolf:
Nice catch @egernst !

virtcontainers/sandbox.go
```diff
@@ -1253,6 +1253,8 @@ func (s *Sandbox) createContainers() error {
 	span, _ := s.trace("createContainers")
 	defer span.Finish()
 
+	s.Logger().Info("createContainersr")
```
jodh-intel: Typo: "Containers". Do we really need this and the other calls though?

virtcontainers/utils/utils.go
jodh-intel: Could you rename this to something like `CalculateMilliCPUs()` - the "`m`" in `CalculatemCPUs()` isn't immediately obvious so we may as well expand it for clarity.

grahamwhaley:
I know this is an important fix, and we should land asap....
are we also going to improve the tests so we don't regress again in the future? Sure, we don't have to block the merge without the tests - I think as long as we get a promise to update the tests asap as well...

grahamwhaley:
Ah, I see test request on #1296 already - thx.

virtcontainers/sandbox.go
```diff
@@ -1608,6 +1610,7 @@ func (s *Sandbox) AddDevice(info config.DeviceInfo) (api.Device, error) {
 func (s *Sandbox) updateResources() error {
 	// the hypervisor.MemorySize is the amount of memory reserved for
 	// the VM and contaniners without memory limit
+	s.Logger().Info("updateResources")
```
jcvenegas: Good to have in the log, helps to get a better view of the flow

virtcontainers/sandbox.go
```diff
@@ -1619,19 +1622,23 @@ func (s *Sandbox) updateResources() error {
...
+	// Calculate running total of memory and mCPUs requested
 	for _, c := range s.config.Containers {
 		if m := c.Resources.Memory; m != nil && m.Limit != nil {
 			*sumResources.Memory.Limit += *m.Limit
 		}
 		if cpu := c.Resources.CPU; cpu != nil {
 			if cpu.Period != nil && cpu.Quota != nil {
-				*sumResources.CPU.Period += *cpu.Period
-				*sumResources.CPU.Quota += *cpu.Quota
+				mCPU += uint32(utils.CalculatemCPUs(*cpu.Quota, *cpu.Period))
```
jcvenegas: Please sync data types with `CalculatemCPUs` instad of do casting

virtcontainers/sandbox.go
egernst: A function for going from milli to vCPU?  Sure.

virtcontainers/utils/utils.go
egernst: ACK - renamed, and handling the -1 case.

virtcontainers/sandbox.go
egernst: done

virtcontainers/utils/utils.go
```diff
@@ -105,6 +105,25 @@ func WriteToFile(path string, data []byte) error {
 	return nil
 }
 
+//CalculatemCPUs converts CPU quota and period to milli-CPUs
```
sboeuf: The comment should start with the function name here.

egernst:
@chavafg - what's the delta in testing on fedora vs ubuntu?

I see a CPU test failure that i'm going to look at now, but i'm wondering how it wasn't exercused with the other jobs?

chavafg:
We currently have the same resources for both - 4 vCPUs.
I think that `[Fail] Update CPU constraints Update CPU set [It] cpuset should be equal to 2 
` is the one that we have seen lately and that @devimc was trying to fix.
See: https://github.com/kata-containers/tests/pull/1238

egernst:
OK, I don't think we should gate based on this then, correct?

chavafg:
Correct, I have restarted the job anyway.

Issue #1308, opened 2019-03-01
ppc64le: qemu major/minor version not set when running unit test


nitkon:
Set test qemu major/minor version when
running unit test TestQemuPPC64leMemoryTopology

Issue #363, opened 2018-06-01
ppc64le: Restrict maxmem to avoid  HTAB allocation failure


nitkon:
# Description of problem

PowerPC guests need a data structure called HTAB (hash table) that stores the virtual to physical page mappings for the guest. HTAB for guest is allocated by the host in host contiguous memory area (CMA) which is a limited resource (by default 5% of host RAM is CMA region). All guests running in the host get their HTAB allocated from this CMA region. HTAB size depends on the maxmem size and specifying huge values of maxmem for guest could result in failures like below:

qemu-system-ppc64 … -m 4G,slots=32,maxmem=1T
qemu-system-ppc64: Failed to allocate HTAB of requested size, try with smaller maxmem
Aborted

In such cases lowering the maxmem is recommended.

# Expected result
```
docker run -it --runtime kata-runtime busybox /bin/sh
/ # 
```
# Actual result
```
docker run -it --runtime kata-runtime busybox /bin/sh
docker: Error response from daemon: OCI runtime create failed: Unexpected error in spapr_alloc_htab() at /build/qemu-ozKJoj/qemu-2.5+dfsg/hw/ppc/spapr.c:1030:
qemu-system-ppc64: Failed to allocate HTAB of requested size, try with smaller maxmem: unknown.
```
---

mpe:
This is supposed to have been fixed in upstream kernel/qemu.

The qemu support was merged in qemu 2.10, and the kernel support in 4.11.

Is there some reason that's not working for you?

jodh-intel:
Hi @nitkon - please could you confirm which version of qemu you're using?

You could just paste in the output of `sudo kata-collect-data.sh` (build from the runtime repo from https://github.com/kata-containers/runtime/blob/master/data/kata-collect-data.sh.in) as that will provide all details on qemu.

jodh-intel:
Re-opening until we understand this issue more fully...

nitkon:
@jodh-intel : When I hit this issue, I was using the following version of qemu.
```QEMU emulator version 2.5.0 (Debian 1:2.5+dfsg-5ubuntu10.25), Copyright (c) 2003-2008 Fabrice Bellard```

jodh-intel:
Hi @nitkon - thanks, that aligns with the info provided by @mpe. Out of interest, can you run newer versions of qemu on ppc64el with Kata?

We record the latest versions in the packaging versions database:

[...]

PR #1309, opened 2019-03-01
virtcontainers: Set test qemu version for unit test

nitkon:
Set test qemu major/minor version when
running unit test TestQemuPPC64leMemoryTopology

Fixes: #1308

### DIFF
--- a/virtcontainers/qemu_ppc64le_test.go
+++ b/virtcontainers/qemu_ppc64le_test.go
@@ -7,11 +7,16 @@ package virtcontainers
 
 import (
 	"fmt"
+	"os/exec"
+	"regexp"
+	"strconv"
 	"testing"
 
 	govmmQemu "github.com/intel/govmm/qemu"
 	"github.com/stretchr/testify/assert"
 )
+
+var qemuVersionArgs = "--version"
 
 func newTestQemu(machineType string) qemuArch {
 	config := HypervisorConfig{
@@ -34,6 +39,31 @@ func TestQemuPPC64leCPUModel(t *testing.T) {
 	assert.Equal(expectedOut, model)
 }
 
+func getQemuVersion() (qemuMajorVersion int, qemuMinorVersion int) {
+
+	cmd := exec.Command(defaultQemuPath, qemuVersionArgs)
+	additionalEnv := "LANG=C"
+	cmd.Env = append(cmd.Env, additionalEnv)
+	out, err := cmd.Output()
+	if err != nil {
+		err = fmt.Errorf("Could not execute command %s %s", defaultQemuPath, qemuVersionArgs)
+		fmt.Println(err.Error())
+	}
+
+	re := regexp.MustCompile("[0-9]+")
+	qVer := re.FindAllString(string(out), -1)
+
+	qMajor, err := strconv.Atoi(qVer[0])
+	qMinor, err1 := strconv.Atoi(qVer[1])
+
+	if err != nil || err1 != nil {
+		err = fmt.Errorf("Could not convert string to int")
+		fmt.Println(err.Error())
+	}
+
+	return qMajor, qMinor
+}
+
 func TestQemuPPC64leMemoryTopology(t *testing.T) {
 	assert := assert.New(t)
 	ppc64le := newTestQemu(QemuPseries)
@@ -42,12 +72,19 @@ func TestQemuPPC64leMemoryTopology(t *testing.T) {
 	hostMem := uint64(1024)
 	mem := uint64(120)
 	slots := uint8(10)
+
+	qemuMajorVersion, qemuMinorVersion = getQemuVersion()
+	m := ppc64le.memoryTopology(mem, hostMem, slots)
+
+	if qemuMajorVersion <= 2 && qemuMinorVersion < 10 {
+		hostMem = uint64(defaultMemMaxPPC64le)
+	}
+
 	expectedMemory := govmmQemu.Memory{
 		Size:   fmt.Sprintf("%dM", mem),
 		Slots:  slots,
 		MaxMem: fmt.Sprintf("%dM", hostMem+uint64(memoryOffset)),
 	}
 
-	m := ppc64le.memoryTopology(mem, hostMem, slots)
 	assert.Equal(expectedMemory, m)
 }


jodh-intel:
Hi @nitkon - please could you provide some details as to why this is necessary?

nitkon:
@jodh-intel : When we run the unit test, `go test -run TestQemuPPC64leMemoryTopology` the `qemuMajorVersion, qemuMinorVersion` do not get set (since `func (q *qemu) waitSandbox(timeout int) error` does not get called?)
So we set it explicitly when running unit test.

grahamwhaley:
@nitkon I think you're going to have to explain this more to us, with some details. I don't see this being done for any of the other architectures for instance, so what is different for ppc64le?

I don't like the coding of 2.11 in the test either, as that will just rot pretty soon over time I suspect. If we really really have to pull in a version, then maybe we can extract it from the versions.yaml (I don't know if we do that for any other tests already for instance for other vars - maybe it isn't trivial).

/me mutters that showing lots of detail in the original report, including the failure logs and why this is very arch specific probably would have avoided us going around the loop ;-)

jodh-intel:
@nitkon - any update?

jodh-intel:
Ping @nitkon.

raravena80:
@nitkon any updates? Thx!

raravena80:
@nitkon nudge

jodh-intel:
@nitkon - is this PR still required? I'm still curious as to why this change is required for ppc but not other architectures?

jodh-intel:
Re-ping @nitkon. Is this still required or can we close?

nitkon:
Apologies for the late response. For the record, I still see the issue on ppc64le... 
```
=== RUN   TestQemuPPC64leMemoryTopology
--- FAIL: TestQemuPPC64leMemoryTopology (0.00s)
    assertions.go:239: ^M                          ^M   Error Trace:    qemu_ppc64le_test.go:52
        ^M      Error:          Not equal:
        ^M                      expected: qemu.Memory{Size:"120M", Slots:0xa, MaxMem:"2048M", Path:""}
        ^M                      actual: qemu.Memory{Size:"120M", Slots:0xa, MaxMem:"33280M", Path:""}
        ^M
        ^M                      Diff:
        ^M                      --- Expected
        ^M                      +++ Actual
        ^M                      @@ -3,3 +3,3 @@
        ^M                        Slots: (uint8) 10,
        ^M                      - MaxMem: (string) (len=5) "2048M",
        ^M                      + MaxMem: (string) (len=6) "33280M",
        ^M                        Path: (string) ""
```

nitkon:
@jodh-intel @grahamwhaley: On ppc64le, max memory `maxmem` possible depends on the qemu version. See this [issue](https://github.com/kata-containers/runtime/issues/363). Since `qemuMajorVersion` & `qemuMinorVersion` for the unit test `TestQemuPPC64leMemoryTopology` does not get set, hence it always defaults to [this](https://github.com/kata-containers/runtime/blob/86d51f59d427c914132bf0147733646c58e7a8fd/virtcontainers/qemu_ppc64le.go#L128). Hence the test fails as expected and actual maxmem never match. That's why this PR to set a qemu version and run this unit test against it qemu version 2.11 & above.

jodh-intel:
Could the test check the version of qemu and then alter it's behaviour based on that? That approach will be resilient to newer versions of qemu I think.

tal at the test constraints tests as they calculate the distro version and the kernel version, compute new versions based on the actual running versions and then test with them:

- https://github.com/kata-containers/runtime/blob/master/pkg/katatestutils/constraints_test.go#L1192

nitkon:
@jodh-intel : ptal

jodh-intel:
Hi @nitkon - I'm afraid I'm confused now :) I thought this PR was to set a qemu version for testing? But that's not what this PR is doing....?!?

nitkon:
@jodh-intel: Earlier qemu version did not get set and hence defaulted to 0,0 always. Now we are querying the installed qemu version and the test case behaves accordingly.

jodh-intel:
Stepping back a moment, the code here is being modified to suit the tests and I don't think we should be doing that.

Instead, why don't you change `qemuPPC64le.memoryTopology()` so that it does something like this:

```diff
diff --git a/virtcontainers/qemu_ppc64le.go b/virtcontainers/qemu_ppc64le.go
index 6ef4015..29d0a1e 100644
--- a/virtcontainers/qemu_ppc64le.go
+++ b/virtcontainers/qemu_ppc64le.go
@@ -119,8 +119,10 @@ func (q *qemuPPC64le) cpuModel() string {
        return cpuModel
 }
 
-func (q *qemuPPC64le) memoryTopology(memoryMb, hostMemoryMb uint64, slots uint8) govmmQemu.Memory {
+// Defined as a variable to allow tests to modify the behaviour
+var qemuPPC64leMemoryTopology = qemuPPC64leMemoryTopologyImplementation
 
+func qemuPPC64leMemoryTopologyImplementation(q *qemuPPC64le, memoryMb, hostMemoryMb uint64, slots uint8, memoryOffset uint32) govmmQemu.Memory {
        if qemuMajorVersion >= 2 && qemuMinorVersion >= 10 {
                q.Logger().Debug("Aligning maxmem to multiples of 256MB. Assumption: Kernel Version >= 4.11")
                hostMemoryMb -= (hostMemoryMb % 256)
@@ -129,7 +131,11 @@ func (q *qemuPPC64le) memoryTopology(memoryMb, hostMemoryMb uint64, slots uint8)
                hostMemoryMb = defaultMemMaxPPC64le
        }
 
-       return genericMemoryTopology(memoryMb, hostMemoryMb, slots, q.memoryOffset)
+       return genericMemoryTopology(memoryMb, hostMemoryMb, slots, memoryOffset)
+}
+
+func (q *qemuPPC64le) memoryTopology(memoryMb, hostMemoryMb uint64, slots uint8) govmmQemu.Memory {
+       return qemuPPC64leMemoryTopology(q, memoryMb, hostMemoryMb, slots, q.memoryOffset)
 }
 
 func (q *qemuPPC64le) appendImage(devices []govmmQemu.Device, path string) ([]govmmQemu.Device, error) {
```

That way, your test can simply redefine the `qemuPPC64leMemoryTopology` variable to do something different: call `getQemuVersion()`, but only in the test code.

nitkon:
@jodh-intel: Updated my PR. Ptal at this approach. 
/retest

jodh-intel:
Ping @kata-containers/runtime - can we get some more eyes on this folks?

@nitkon - I'm afraid I really don't like the idea of launching an external program to check its version using regex's. Yes, some of that is done in `kata-env`, but that's a utility that is not -- by design -- in the fastpath.

I'll take another look tomorrow with "fresh eyes" but I still think there must be a way to do what you want entirely in test code without potentially impacting container start time purely to make a test pass.

nitkon:
Thnx @jodh-intel. Moved the code `out of the fastpath`. 
/retest

nitkon:
/retest 
Fixed: travis-ci failure `virtcontainers/qemu_ppc64le_test.go:50:16: cannot use err (variable of type error) as assert.TestingT value in argument to assert.Error: missing method Errorf (typecheck)`

nitkon:
Hi @grahamwhaley , ptal at this PR. Once this gets landed, can we get Power CI attached to this repo?  :-) CI checks like `make test` & `make check` run clean on my local ppc64le machine

grahamwhaley:
@chavafg @GabyCT - can we clone the Power8 proxy job over to the runtime repo as well pls?, and let's see how it flies...

chavafg:
I've cloned the job, should start testing PRs from now on

Issue #1320, opened 2019-03-05
aarch64 fail in No GO command or GOPATH not set mode


teawater:
Got aarch64 fail in No GO command or GOPATH not set mode:
Makefile:38: arch/aarch64-options.mk: No such file or directory
Makefile:237: *** "ERROR: No hypervisors known for architecture aarch64 (looked for: firecracker qemu)".  Stop.

PR #1321, opened 2019-03-05
Makefile: Fix aarch64 fail in No GO command or GOPATH not set mode

teawater:
Got aarch64 fail in No GO command or GOPATH not set mode:
Makefile:38: arch/aarch64-options.mk: No such file or directory
Makefile:237: *** "ERROR: No hypervisors known for architecture aarch64 (looked for: firecracker qemu)".  Stop.

The cause is the machine hardware name of aarch64 is aarch64 but
runtime's arch make file's name is arm64-options.mk.

This commit convert aarch64 to arm64 to fix the issue.

Fixes: #1320

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,9 @@ else
     ifeq ($(ARCH),x86_64)
         ARCH = amd64
     endif
+    ifeq ($(ARCH),aarch64)
+        ARCH = arm64
+    endif
 endif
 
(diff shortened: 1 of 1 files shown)


grahamwhaley:
vsocks and metrics CIs are known flaky right now. ARM and other CIs are happy - merging!

Issue #1318, opened 2019-03-04
Can't connect to qmp/console's unix socket if created by shimv2


zhsj:
# Description of problem

I want to debug the VM, and I follow the developer guide, to start a shell in the VM.
Now with container created by docker(via `docker run  --runtime kata-runtime`), I've successfully entered the VM, with command `socat "stdin,raw,echo=0,escape=0x11" "unix-connect:${console}"`

But for the container created by shimv2 (via `ctr run --runtime io.containerd.run.kata.v2`), the unix socket doesn't response, neither console.sock nor qmp.sock.

This looks like that other process is using these socket files. (I just observe that these unix socket files can only response to one client.) But I use `fuser` to search, there's only qemu process using the socket file.

I'm using the default kata configuration(debug is false), default containerd configuration.

kata version is 1.5.0 and containerd is 1.2.4.

# Expected result

The qmp/console unix socket files can response.

# Actual result

The socket files of container created by shimv2 can't response.

amshinde:
@sboeuf Any idea about this beheaviour?

sboeuf:
@zhsj can you check if the QEMU process that has been started by `containerd-shim-kata-v2` has a `console.sock` socket being created?
I'm not entirely sure in case of `containerd-shim-kata-v2`, if you're not using `vsock`, the runtime might be the one connecting to the `console.sock`, even if `enable_debug=false` (that would be a bug obviously). Or the other possibility would be that in this case the `console.sock` is not created (check by looking at qemu process `ps -ef | grep qemu`).

egernst:
@lifupan @gnawux ptal?

zhsj:
> I'm not entirely sure in case of containerd-shim-kata-v2, if you're not using vsock, the runtime might be the one connecting to the console.sock, even if enable_debug=false (that would be a bug obviously).

So I tried the vsock. Although I failed to run (not sure if #1319 is related), I can connect to the qmp and console socket.
> Or the other possibility would be that in this case the console.sock is not created (check by looking at qemu process ps -ef | grep qemu).

It's created, I can see it's opened by qemu process.
```
$ sudo lsof -n -p 9765|grep sock
qemu-syst 9765 root   11u     sock                0,9      0t0 20218250 protocol: UNIX
qemu-syst 9765 root   12u     sock                0,9      0t0 20218252 protocol: UNIX
qemu-syst 9765 root   13u     sock                0,9      0t0 20218254 protocol: UNIX
[...]

PR #1322, opened 2019-03-05
virtcontainer: watch the qemu's console when proxy's debug enabled

lifupan:
kata builtin proxy has always watched the qemu's console
whether proxy's debug is set or not, this is not aligned
with kata cli. This patch will change it and watch the
qemu's console only when proxy's debug is set in kata config.

Fixes: #1318


sboeuf:
Thanks @lifupan for fixing this :)

Issue #1325, opened 2019-03-05
Containerd block snapshotter support


mcastelino:
# Kata does not support block based containerd snapshotters

Kata does not support block based containerd snapshotters. This is due to an incorrect assumption in Kata that there will be a directory called "rootfs" within the block device that holds the snapshot.

However with containerd block based snapshotters this no longer true. 

e..g In the past the built in devicemapper would report the rootfs at
```
/var/lib/docker/devicemapper/mnt/999042f4b5dea95ed2b91e58ca5dbd865fe064f8a17cca5415489930f1a34df5/rootfs
```
The backing device is at
```
/var/lib/docker/devicemapper/mnt/999042f4b5dea95ed2b91e58ca5dbd865fe064f8a17cca5415489930f1a34df5/
```
***and the rootfs is a directory under the backing device.***

The Kata logic always assumed that this was the case.

However with the current container snapshotter implementation the rootfs is reported as follows
```
/dev/mapper/vgthin-47 on /run/containerd/io.containerd.runtime.v2.task/default/hello-kata-lvm/rootfs type xfs
```

***the rootfs directory is the backing device.***

i.e. the volume/block device is mounted directly at `rootfs` which breaks Kata logic

# Kata Logic

Kata today has multiple paths and options in handling the rootfs

1. Hypervisor does not support block device hotplug

   Kata bind mounts the rootfs location to the shared directory at location `c.id/rootfs`
   a. For overlay based snapshotters
   b. For block based snapshotters

2. Hypervisor supports block device hotplug
   a. overlay based snapshotters
      Kata bind mounts the rootfs location to the shared directory at location `c.id/rootfs`
   b. block based graph drivers (e.g devicemapper) where the rootfs is a directory within the device
   c. block based snapshotters  (e.g firecracker devicemapper, kata lvm) where the rootfs is the device

3. Hypervisor supports block device hotplug but user chooses to not use block hotplug
   Kata bind mounts the rootfs location to the shared directory at location `c.id/rootfs`
   a. For overlay based snapshotters
   b. For block based snapshotters

We need to handle all of these cases   

# Fix

```
diff --git a/virtcontainers/container.go b/virtcontainers/container.go
index 5408d85..f11fe99 100644
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -276,6 +276,7 @@ type Container struct {
        runPath       string
        configPath    string
        containerPath string
+       rootfsSuffix  string
 
        state types.State
 
[...]

PR #1326, opened 2019-03-05
Fix rootfs mount assumptions

ganeshmaharaj:
This patch fixes the issue where vrious version of snapshotters,
overlay, block based graphdriver, containerd-shim-v2 overlay, block
based snapshotters mount & create rootfs differently and kata should be
able to handle them all.

The current version of the code always assumes that a folder named
'rootfs' exists within the mount device and that is the path the
container should start at. This patch checks the existing mount point
and if it is the same as the rootFs passed to the container, we no
longer add a suffix to the container's rootfs path.

Fixes: #1325

Co-Authored-by: Manohar Castelino <manohar.r.castelino@intel.com>

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -276,6 +276,7 @@ type Container struct {
 	runPath       string
 	configPath    string
 	containerPath string
+	rootfsSuffix  string
 
 	state types.State
 
@@ -640,6 +641,7 @@ func newContainer(sandbox *Sandbox, contConfig ContainerConfig) (*Container, err
 		runPath:       store.ContainerRuntimeRootPath(sandbox.id, contConfig.ID),
 		configPath:    store.ContainerConfigurationRootPath(sandbox.id, contConfig.ID),
 		containerPath: filepath.Join(sandbox.id, contConfig.ID),
+		rootfsSuffix:  "rootfs",
 		state:         types.State{},
 		process:       Process{},
 		mounts:        contConfig.Mounts,
@@ -1129,6 +1131,10 @@ func (c *Container) hotplugDrive() error {
 
 	if !isDM {
 		return nil
+	}
+
+	if dev.mountPoint == c.rootFs {
+		c.rootfsSuffix = ""
 	}
 
 	// If device mapper device, then fetch the full path of the device
--- a/virtcontainers/container_test.go
+++ b/virtcontainers/container_test.go
@@ -275,6 +275,57 @@ func TestContainerAddDriveDir(t *testing.T) {
 	}
 }
 
+func TestContainerRootfsPath(t *testing.T) {
+
+	testRawFile, loopDev, fakeRootfs, err := testSetupFakeRootfs(t)
+	defer cleanupFakeRootfsSetup(testRawFile, loopDev, fakeRootfs)
+	assert.Nil(t, err)
+
+	truecheckstoragedriver := checkStorageDriver
+	checkStorageDriver = func(major, minor int) (bool, error) {
+		return true, nil
+	}
+	defer func() {
+		checkStorageDriver = truecheckstoragedriver
+	}()
+
+	sandbox := &Sandbox{
+		ctx:        context.Background(),
+		id:         "rootfstestsandbox",
+		agent:      &noopAgent{},
+		hypervisor: &mockHypervisor{},
+		config: &SandboxConfig{
+			HypervisorConfig: HypervisorConfig{
+				DisableBlockDeviceUse: false,
+			},
+		},
+	}
+	vcstore, err := store.NewVCSandboxStore(sandbox.ctx, sandbox.id)
+	sandbox.store = vcstore
+	assert.Nil(t, err)
+	container := Container{
+		id:           "rootfstestcontainerid",
+		sandbox:      sandbox,
+		rootFs:       fakeRootfs,
+		rootfsSuffix: "rootfs",
+	}
+	cvcstore, err := store.NewVCContainerStore(context.Background(),
+		sandbox.id,
+		container.id)
+	assert.Nil(t, err)
+	container.store = cvcstore
+
+	container.hotplugDrive()
+	assert.Empty(t, container.rootfsSuffix)
+
+	// Reset the value to test the other case
+	container.rootFs = fakeRootfs + "/rootfs"
+	container.rootfsSuffix = "rootfs"
+
+	container.hotplugDrive()
+	assert.Equal(t, container.rootfsSuffix, "rootfs")
+}
+
 func TestCheckSandboxRunningEmptyCmdFailure(t *testing.T) {
 	c := &Container{}
 	err := c.checkSandboxRunning("")
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -986,7 +986,7 @@ func (k *kataAgent) createContainer(sandbox *Sandbox, c *Container) (p *Process,
 
 	// This is the guest absolute root path for that container.
 	rootPathParent := filepath.Join(kataGuestSharedDir, c.id)
-	rootPath := filepath.Join(rootPathParent, rootfsDir)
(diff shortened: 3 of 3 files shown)


egernst:
nit on typo in commit mg.

I get that it was broken before - can you tell me how and what you did to fix in the commit msg?  

Nice work!

virtcontainers/container.go
```diff
@@ -640,6 +641,7 @@ func newContainer(sandbox *Sandbox, contConfig ContainerConfig) (*Container, err
 		runPath:       store.ContainerRuntimeRootPath(sandbox.id, contConfig.ID),
 		configPath:    store.ContainerConfigurationRootPath(sandbox.id, contConfig.ID),
 		containerPath: filepath.Join(sandbox.id, contConfig.ID),
+		rootfsSuffix:  "rootfs",
```
amshinde: @mcastelino @ganeshmaharaj This works for us since the storage drivers provide the root file system under "rootfs" path. We talked about making this generic, so maybe we can deduce the suffix at runtime by looking at the difference between the rootfs mountpoint and the rootfs path provided in the OCI spec. 
I am fine if you want to address this in a separate PR.

ganeshmaharaj: @amshinde if it can be a seperate PR, that would be good. Will also give us time to look through the code and clean-up anything else we might have missed.

virtcontainers/container.go
```diff
@@ -1131,6 +1133,10 @@ func (c *Container) hotplugDrive() error {
 		return nil
 	}
 
+	if dev.mountPoint == c.rootFs {
+		c.rootfsSuffix = ""
```
jodh-intel: Would it be possible to contrive a unit test for the two possible values here?

virtcontainers/mount.go
```diff
@@ -89,7 +89,9 @@ var errMountPointNotFound = errors.New("Mount point not found")
 //		manor : minor(/dev/sda1)
 //		mountPoint: /a/b/c
 //	}
-func getDeviceForPath(path string) (device, error) {
+var getDeviceForPath = findDeviceForPath
```
mcastelino: @ganeshmaharaj cute but ouch.

@jodh-intel I see the value in adding this unit test. But instead of doing it this way, it may be better to just enhance our system level tests to include this path. Using devicemapper vs snapshotter.

virtcontainers/container.go
mcastelino: I saw the unit test that got added. But I feel we would be better served with a system level test here. :(

ganeshmaharaj: the function `hotplugDrive` does a lot of system/OS level calls to identify  the characteristic of the drive that is being handed for rootfs. To changes i made to code to unit test this, does not sit well with me, but i was unable to find another cleaner way to do this test without doing a full blown device creation and mocking a few more things. Could you check the code and let me know your thoughts. This case i think is one that might benefit with more functional tests against unit tests.

virtcontainers/mount.go
amshinde: agree with @mcastelino ..we are going down the hole of adding too many indirections.
@ganeshmaharaj Is is possible to test this by actually creating an img and mounting it on a loop device?
Something like https://github.com/kata-containers/runtime/blob/6431f1f28883f7f60548e2651d68a6dd6aa7dc34/virtcontainers/container_test.go#L139 ?

ganeshmaharaj: @amshinde I did think about it and I thought it was an overkill (creating a drive to just test this). But if you think it's worth it, I can hack this function to allow a mount point be passed and then handle the rootfs scenario.

ganeshmaharaj:
the CI job failed with the below text.
```
Stderr: 
Running command '/usr/bin/docker [docker pull alpine:3.7]'
[docker pull alpine:3.7]
Timeout: 600 seconds
Exit Code: 1
Stdout: 3.7: Pulling from library/alpine

Stderr: no matching manifest for unknown in the manifest list entries

--- FAIL: TestIntegration (6.16s)
    main_test.go:60: failed to pull docker image: alpine:3.7
FAIL

Ginkgo ran 1 suite in 13.319596948s
Test Suite Failed
Makefile:63: recipe for target 'docker' failed
make: *** [docker] Error 1
Performing Post build task...
```
http://jenkins.katacontainers.io/job/kata-containers-runtime-ARM-18.04-PR/233/consoleText

I do not think my change would have caused this.

sboeuf:
@ganeshmaharaj no I don't think so. @Pennyzct PTAL

Pennyzct:
Hi~@sboeuf 
`Stderr: no matching manifest for unknown in the manifest list entries`
This error looks odd to me. I have checked the image `alpine:3.7`, it's a multi-arch image and have `arm64/v8` entry. `docker pull alpine:3.7` locally is also working well. Let's see if it's repeatable.;)
/test

virtcontainers/mount.go
ganeshmaharaj: @mcastelino @amshinde fixed the patch to not hack any more files. Let me know how this one looks.

grahamwhaley:
@ganeshmaharaj - which CI ;-) - a few have failed, and I think for different reasons. I think we have had a reasonably sized CI instability overnight (I suspect mostly due to either/both an image change and/or some network issues).
I'm going to hold off re-triggering them here until we have diagnosed and fixed.

virtcontainers/container_test.go
```diff
@@ -275,6 +275,57 @@ func TestContainerAddDriveDir(t *testing.T) {
 	}
 }
 
+func TestContainerRootfsPath(t *testing.T) {
+
+	testRawFile, loopDev, fakeRootfs, err := testSetupFakeRootfs(t)
+	defer cleanupFakeRootfsSetup(testRawFile, loopDev, fakeRootfs)
+	assert.Nil(t, err)
+
+	truecheckstoragedriver := checkStorageDriver
```
mcastelino: @ganeshmaharaj do you still need this hack?

ganeshmaharaj: @mcastelino yes. That particular functions check if the device is a devicemapper based device and return early if it is not. We need it to assume that it is a devicemapper for it to do the shorting of rootfs.

virtcontainers/mount.go
amshinde: lgtm @ganeshmaharaj

virtcontainers/container_test.go
mcastelino: So it was already a variable

https://github.com/kata-containers/runtime/blob/45fe8700b8e984fece021aac3e3bb9ddbe3c7aa3/virtcontainers/mount.go#L208

ganeshmaharaj:
@amshinde @grahamwhaley @chavafg alright to kick the CI now and see what happens?

sboeuf:
Here we go @ganeshmaharaj ! Let's see what the CI gods decide muhaha :smile:

ganeshmaharaj:
@sboeuf seems like the CI gods really do not like me this week. 😀

grahamwhaley:
@ganeshmaharaj - I don't know, I've seen much worse ;-)

18.04-initrd failed with a network test:

```
Summarizing 1 Failure:

[Fail] run host networking Run with host networking [It] should error out 
/tmp/jenkins/workspace/kata-containers-runtime-ubuntu-18-04-PR-initrd/go/src/github.com/kata-containers/tests/integration/docker/run_test.go:259

Ran 219 of 237 Specs in 749.092 seconds
FAIL! -- 218 Passed | 1 Failed | 0 Pending | 18 Skipped 
```

and vsocks failed with what feels like an infra level timeout (aka, not you...)
```
Install test dependencies
Install kata containers dependencies
Build timed out (after 5 minutes). Marking the build as aborted.
```

Let me nudge both of those for rebuilds....

jodh-intel:
vsocks CI timed out so restarted. initrd CI fell over on a network test. Might be transitory so restarted to see if it is ;)

Issue #1333, opened 2019-03-07
 Factory: Fix fake return value issue on creating template


jongwu:
# Description of problem

Now, function NewFactory will return nil even create template
does't complete. As for this, it will tell user that factory
has been initialized no matter whether the template is created
or not. 
# Expected result

# Actual result

---

PR #1334, opened 2019-03-07
Factory: Fix fake return value issue on creating template

jongwu:
Now, function NewFactory will return nil even create template
does't complete. As for this, it will tell user that factory
has been initialized no matter whether the template is created
or not. This patch correct it by adding another return value
of error in NewFactory.

Fixes: #1333

Signed-off-by: Jianyong Wu  <jianyong.wu@arm.com>
@jodh-intel @grahamwhaley @Pennyzct

### DIFF
--- a/virtcontainers/factory/factory.go
+++ b/virtcontainers/factory/factory.go
@@ -67,7 +67,10 @@ func NewFactory(ctx context.Context, config Config, fetchOnly bool) (vc.Factory,
 				return nil, err
 			}
 		} else {
-			b = template.New(ctx, config.VMConfig)
+			b, err = template.New(ctx, config.VMConfig)
+			if err != nil {
+				return nil, err
+			}
 		}
 	} else if config.VMCache && config.Cache == 0 {
 		b, err = grpccache.New(ctx, config.VMCacheEndpoint)
--- a/virtcontainers/factory/template/template.go
+++ b/virtcontainers/factory/template/template.go
@@ -15,7 +15,6 @@ import (
 
 	vc "github.com/kata-containers/runtime/virtcontainers"
 	"github.com/kata-containers/runtime/virtcontainers/factory/base"
-	"github.com/kata-containers/runtime/virtcontainers/factory/direct"
 	"github.com/kata-containers/runtime/virtcontainers/store"
 )
 
@@ -42,14 +41,13 @@ func Fetch(config vc.VMConfig) (base.FactoryBase, error) {
 }
 
 // New creates a new VM template factory.
-func New(ctx context.Context, config vc.VMConfig) base.FactoryBase {
+func New(ctx context.Context, config vc.VMConfig) (base.FactoryBase, error) {
 	statePath := store.RunVMStoragePath + "/template"
 	t := &template{statePath, config}
 
 	err := t.prepareTemplateFiles()
 	if err != nil {
-		// fallback to direct factory if template is not supported.
-		return direct.New(ctx, config)
+		return nil, err
 	}
 	defer func() {
 		if err != nil {
@@ -59,11 +57,10 @@ func New(ctx context.Context, config vc.VMConfig) base.FactoryBase {
 
 	err = t.createTemplateVM(ctx)
 	if err != nil {
-		// fallback to direct factory if template is not supported.
-		return direct.New(ctx, config)
+		return nil, err
 	}
 
-	return t
+	return t, nil
 }
 
 // Config returns template factory's configuration.
@@ -116,7 +113,9 @@ func (t *template) createTemplateVM(ctx context.Context) error {
 	config.HypervisorConfig.MemoryPath = t.statePath + "/memory"
 	config.HypervisorConfig.DevicesStatePath = t.statePath + "/state"
 	// template vm uses builtin proxy
-	config.ProxyType = templateProxyType
+	if config.ProxyType != "noopProxy" {
+		config.ProxyType = templateProxyType
+	}
 
 	vm, err := vc.NewVM(ctx, config)
 	if err != nil {
--- a/cli/factory.go
+++ b/cli/factory.go
@@ -185,6 +185,7 @@ var initFactoryCommand = cli.Command{
 					HypervisorConfig: runtimeConfig.HypervisorConfig,
 					AgentType:        runtimeConfig.AgentType,
 					AgentConfig:      runtimeConfig.AgentConfig,
+					ProxyType:        runtimeConfig.ProxyType,
 				},
 			}
 			kataLog.WithField("factory", factoryConfig).Info("create vm factory")
--- a/cli/factory_test.go
+++ b/cli/factory_test.go
@@ -17,6 +17,8 @@ import (
 	vc "github.com/kata-containers/runtime/virtcontainers"
 )
 
+const testDisabledAsNonRoot = "Test disabled as requires root privileges"
+
 func TestFactoryCLIFunctionNoRuntimeConfig(t *testing.T) {
 	assert := assert.New(t)
 
@@ -63,9 +65,14 @@ func TestFactoryCLIFunctionInit(t *testing.T) {
 	assert.Nil(err)
 
 	// With template
+	if os.Geteuid() != 0 {
+		t.Skip(testDisabledAsNonRoot)
+	}
+
 	runtimeConfig.FactoryConfig.Template = true
 	runtimeConfig.HypervisorType = vc.MockHypervisor
 	runtimeConfig.AgentType = vc.NoopAgentType
+	runtimeConfig.ProxyType = vc.NoopProxyType
 	ctx.App.Metadata["runtimeConfig"] = runtimeConfig
 	fn, ok = initFactoryCommand.Action.(func(context *cli.Context) error)
 	assert.True(ok)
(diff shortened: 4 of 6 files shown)


virtcontainers/factory/factory.go
```diff
@@ -63,7 +63,7 @@ func NewFactory(ctx context.Context, config Config, fetchOnly bool) (vc.Factory,
 				return nil, err
 			}
 		} else {
-			b = template.New(ctx, config.VMConfig)
+			b, err = template.New(ctx, config.VMConfig)
```
liwei: Seems the error value not used at all, IMHO, we need more thoughts on error processing of factory initialization, maybe @bergwolf could give more comments.

bergwolf: Yeah, please return `err` here.

@liwei Are you referring to the fact that we currently fall back to direct vm when vm templating is unsupported?

liwei: yep, I have no strong favorite on which(fallback or error out) is better, my concern is that silently fallback may hide some information to users and make them think factory works as they expected.

bergwolf: hmm, I agree. Do you want to make a PR for it?

bergwolf: NVM, it seems such a change should be folded into this PR instead.

virtcontainers/factory/template/template.go
```diff
@@ -42,14 +42,14 @@ func Fetch(config vc.VMConfig) (base.FactoryBase, error) {
...
-func New(ctx context.Context, config vc.VMConfig) base.FactoryBase {
+func New(ctx context.Context, config vc.VMConfig) (base.FactoryBase, error) {
 	statePath := store.RunVMStoragePath + "/template"
 	t := &template{statePath, config}
 
 	err := t.prepareTemplateFiles()
 	if err != nil {
 		// fallback to direct factory if template is not supported.
-		return direct.New(ctx, config)
+		return direct.New(ctx, config), err
```
bergwolf: Please just return `nil, err` here.

virtcontainers/factory/template/template.go
```diff
@@ -60,10 +60,10 @@ func New(ctx context.Context, config vc.VMConfig) base.FactoryBase {
 	err = t.createTemplateVM(ctx)
 	if err != nil {
 		// fallback to direct factory if template is not supported.
-		return direct.New(ctx, config)
+		return direct.New(ctx, config), err
```
bergwolf: Same here.

virtcontainers/factory/template/template.go
```diff
@@ -60,10 +60,10 @@ func New(ctx context.Context, config vc.VMConfig) base.FactoryBase {
...
 	}
 
-	return t
+	return t, err
```
bergwolf: `return t, nil` is more explicit.

jongwu: @bergwolf thanks, I will fix it.

jodh-intel:
CI's are unhappy:

```
--- FAIL: TestFactoryCLIFunctionInit (0.00s)
    <autogenerated>:1: 
                          
	Error Trace:	factory_test.go:73
        
	Error:      	Expected nil, but got: &os.PathError{Op:"mkdir", Path:"/run/vc", Err:0xd}
```

amshinde:
@jongwu CI is failing , can you fix the tests?

jongwu:
> @jongwu CI is failing , can you fix the tests?

@amshinde maybe this error is existed originally and revealed by my commit. let me have a try to fix it.

amshinde:
@jongwu The error is at this line https://github.com/kata-containers/runtime/blob/47bde46ceb52d1475e7d13a223e22fe33b1bf625/virtcontainers/factory/factory_test.go#L73

You may need to adjust that test.

jongwu:
> @jongwu The error is at this line
> 
> [runtime/virtcontainers/factory/factory_test.go](https://github.com/kata-containers/runtime/blob/47bde46ceb52d1475e7d13a223e22fe33b1bf625/virtcontainers/factory/factory_test.go#L73)
> 
> Line 73 in [47bde46](/kata-containers/runtime/commit/47bde46ceb52d1475e7d13a223e22fe33b1bf625)
> 
>  assert.Nil(err) 
> You may need to adjust that test.

@amshinde thanks, em, this error maybe at cli/factory_test.go. It caused by the wrong proxytype used in test. The proxytype is forced as katabuiltinproxy when creating template even in testing, which should have been noopproxy.

jongwu:
@amshinde @jodh-intel ci fail at the same test case, but I will not get that failure in my machine.

jodh-intel:
@jongwu - the latest failure is:

```
--- FAIL: TestFactoryCLIFunctionInit (0.00s)
    <autogenerated>:1: 
                          
	Error Trace:	factory_test.go:74
        
	Error:      	Expected nil, but got: &os.PathError{Op:"mkdir", Path:"/run/vc", Err:0xd}
```

That problem looks to me like the factory is attempting to create `/run/vc`. But for the tests that isn't / should not be valid - the test should be calling `ioutil.TempDir("", "")` and setting `Path` to that value.

jongwu:
> @jongwu - the latest failure is:
> 
> ```
> --- FAIL: TestFactoryCLIFunctionInit (0.00s)
>     <autogenerated>:1: 
>                           
> 	Error Trace:	factory_test.go:74
>         
> 	Error:      	Expected nil, but got: &os.PathError{Op:"mkdir", Path:"/run/vc", Err:0xd}
> ```
> That problem looks to me like the factory is attempting to create `/run/vc`. But for the tests that isn't / should not be valid - the test should be calling `ioutil.TempDir("", "")` and setting `Path` to that value.

@jodh-intel thanks a lot, it really help me. 
This error caused by lack of permission to create directory under /run. I can pass tmp dir to fix it. But tmpfs must be mounted at somewhere when create template which only can be operated by root. so factory test could't pass by no-root user.  just skip it for no-root user.
Refer to function prepareTemplateFiles in /runtime/virtcontainers/factory/template/template.go to find more.

jodh-intel:
Hi @jongwu - ok, in which case please could you add this to the top of your test:

```go
if os.Geteuid() != 0 {
    t.Skip(testDisabledNeedRoot)             
}
```

You can then create a temporary `tmpfs` mount when running as the `root` user. Note that we run the tests twice - once as non-`root` and once as `root` (to handle the`testDisabledNeedNonRoot` and `testDisabledNeedRoot` scenarios respectively).

jongwu:
> Hi @jongwu - ok, in which case please could you add this to the top of your test:
> 
> ```go
> if os.Geteuid() != 0 {
>     t.Skip()             
> }
> ```
> 
> You can then create a temporary `tmpfs` mount when running as the `root` user. Note that we run the tests twice - once as non-`root` and once as `root` (to handle the`testDisabledNeedNonRoot` and `testDisabledNeedRoot` scenarios respectively).

@jodh-intel this test case contains no template part and with template part, only with template need root
privilege, so I put it here. should we skip all of them?

jongwu:
@jodh-intel also, dir for creating template is set in template.New as "statePath := store.RunVMStoragePath + "/template"" , maybe it have not taken  consideration for test. So it's not easy to pass tmpdir from test func.

jodh-intel:
@jongwu - `store.RunVMStoragePath` is a *variable*, so you can do the following in your test:

```golang
stavedStoragePath := store.RunVMStoragePath
defer func() {
    store.RunVMStoragePath = savedStoragePath
}()

store.RunVMStoragePath = "/some/where/else"

// run test with modified store.RunVMStoragePath  here...
```

However, I don't know the factory code well so maybe @bergwolf can comment on the overall test strategy.

bergwolf:
@jongwu IIRC it was intentional to assert success at line [73](https://github.com/kata-containers/runtime/blob/dd2acd26eb63c0eaf861bb21cebe28fed387f5fd/cli/factory_test.go#L73) as it checks for the fallback strategy from template factory to direct factory. Since the PR removes such fallback, I think you can either just remove that part of test, or assert failure there.

jongwu:
@bergwolf that's easy way to go. another problem is how to do with TestTemplateFactory in virtcontainers/factory/tmplate/template_test.go, as template.New will return factory instance as nil, the following test depending on this factory will panic.

jongwu:
@bergwolf thanks, skip those tests which need root privilege related to create factory.

Pennyzct:
/retest
Arm CI timed out. Let’s see if it’s repeatable.:)

Pennyzct:
thanks to [#osbuilder/247](https://github.com/kata-containers/osbuilder/pull/247), ARM CI should be working well now. let's try it. ;) 
/retest

jongwu:
@jodh-intel 3 failing and 11 successful checks, what's that meaning?

caoruidong:
You can click the 'Details' link to see what's wrong. I re-trigger them all.

jongwu:
vm factory destroyed
++ mount
++ grep template
++ wc -l
+ res=1
+ '[' 1 -eq 0 ']'
+ die 'template factory is not cleaned up'
+ msg='template factory is not cleaned up'
+ echo 'ERROR: template factory is not cleaned up'
ERROR: template factory is not cleaned up
+ exit 1
Makefile:123: recipe for target 'vm-factory' failed

@jodh-intel @bergwolf I'm not sure if this failure caused by my commit. any idea?

Issue #1336, opened 2019-03-07
Factory: enable template feature for arm64


jongwu:
# Description of problem

now, template feature can't work on arm64. when create template using bypass-shared-memory cap, new sandbox incoming from the template will fail on arm64. but it will work if bypass-shared-memory cap disabled. I'm not sure about the root cause. I assume that the bypass-shared-memory may not work well in some scenarios. then I offer a switch in configuration-qemu.toml to give the user a choice of opening or closing bypass-shared-memory cap.
@jodh-intel @grahamwhaley @Pennyzct @bergwolf 
# Expected result

# Actual result

---

PR #1337, opened 2019-03-07
Factory: enable template feature for arm64

jongwu:
Template: enable template for arm64

    Now, template feature can't work on arm64. When using bypass-shared-memory
    cap to saving sandbox as a template, new sandbox cloning from the template
    will fail. From initial investigation, it maybe a qemu issue. Further
    research should be done.
    To enable template on arm64, this patch adds a switch to offer the capability
    to open or close the bypass-shared-memory cap.
    While closing bypass-shared-memory cap, saving vm will occupy more space
    and consume more time. So increase 300M for mount size and bump the time
    waiting for migration to 10 seconds.

    Fixes: #1336

    Signed-off-by: Jianyong Wu  <jianyong.wu@arm.com>

@jodh-intel @grahamwhaley  @bergwolf  @Pennyzct

### DIFF
--- a/virtcontainers/factory/template/template.go
+++ b/virtcontainers/factory/template/template.go
@@ -90,7 +90,7 @@ func (t *template) prepareTemplateFiles() error {
 		return err
 	}
 	flags := uintptr(syscall.MS_NOSUID | syscall.MS_NODEV)
-	opts := fmt.Sprintf("size=%dM", t.config.HypervisorConfig.MemorySize+8)
+	opts := fmt.Sprintf("size=%dM", t.config.HypervisorConfig.MemorySize+templateDeviceStateSize)
 	if err = syscall.Mount("tmpfs", t.statePath, "tmpfs", flags, opts); err != nil {
 		t.close()
 		return err
--- a/virtcontainers/factory/template/template_amd64.go
+++ b/virtcontainers/factory/template/template_amd64.go
@@ -1,0 +1,14 @@
+// Copyright (c) 2019 HyperHQ Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// template implements base vm factory with vm templating.
+
+package template
+
+// templateDeviceStateSize denotes device state size when
+// mount tmpfs.
+// when bypass-shared-memory is not support like arm64,
+// creating template will occupy more space. That's why we
+// put it here.
+const templateDeviceStateSize = 8
--- a/virtcontainers/factory/template/template_s390x.go
+++ b/virtcontainers/factory/template/template_s390x.go
@@ -1,0 +1,14 @@
+// Copyright (c) 2019 HyperHQ Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// template implements base vm factory with vm templating.
+
+package template
+
+// templateDeviceStateSize denotes device state size when
+// mount tmpfs.
+// when bypass-shared-memory is not support like arm64,
+// creating template will occupy more space. That's why we
+// put it here.
+const templateDeviceStateSize = 8
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -89,10 +89,8 @@ const (
 	consoleSocket = "console.sock"
 	qmpSocket     = "qmp.sock"
 
-	qmpCapErrMsg                      = "Failed to negoatiate QMP capabilities"
-	qmpCapMigrationBypassSharedMemory = "bypass-shared-memory"
-	qmpExecCatCmd                     = "exec:cat"
-	qmpMigrationWaitTimeout           = 5 * time.Second
+	qmpCapErrMsg  = "Failed to negoatiate QMP capabilities"
+	qmpExecCatCmd = "exec:cat"
 
 	scsiControllerID  = "scsi0"
 	rngID             = "rng0"
@@ -1320,12 +1318,7 @@ func (q *qemu) saveSandbox() error {
 	// BootToBeTemplate sets the VM to be a template that other VMs can clone from. We would want to
 	// bypass shared memory when saving the VM to a local file through migration exec.
 	if q.config.BootToBeTemplate {
-		err = q.qmpMonitorCh.qmp.ExecSetMigrationCaps(q.qmpMonitorCh.ctx, []map[string]interface{}{
-			{
-				"capability": qmpCapMigrationBypassSharedMemory,
-				"state":      true,
-			},
-		})
+		err := q.arch.setBypassSharedMemoryMigrationCaps(q.qmpMonitorCh.ctx, q.qmpMonitorCh.qmp)
 		if err != nil {
 			q.Logger().WithError(err).Error("set migration bypass shared memory")
 			return err
--- a/virtcontainers/qemu_arm64.go
+++ b/virtcontainers/qemu_arm64.go
@@ -6,10 +6,12 @@
 package virtcontainers
 
 import (
+	"context"
 	"io/ioutil"
 	"os"
 	"runtime"
 	"strings"
+	"time"
 
 	govmmQemu "github.com/intel/govmm/qemu"
 	"github.com/kata-containers/runtime/virtcontainers/types"
@@ -24,6 +26,10 @@ type qemuArm64 struct {
 const defaultQemuPath = "/usr/bin/qemu-system-aarch64"
 
 const defaultQemuMachineType = QemuVirt
+
+const qmpMigrationWaitTimeout = 10 * time.Second
+
+const qmpCapMigrationBypassSharedMemory = "bypass-shared-memory"
 
 var defaultQemuMachineOptions = "usb=off,accel=kvm,nvdimm,gic-version=" + getGuestGICVersion()
 
@@ -188,4 +194,9 @@ func (q *qemuArm64) appendImage(devices []govmmQemu.Device, path string) ([]govm
 	devices = append(devices, object)
 
 	return devices, nil
+}
+
+func (q *qemuArm64) setBypassSharedMemoryMigrationCaps(_ context.Context, _ *govmmQemu.QMP) error {
+	// bypass-shared-memory not support in arm64 for now
+	return nil
 }
--- a/virtcontainers/factory/template/template_arm64.go
+++ b/virtcontainers/factory/template/template_arm64.go
@@ -1,0 +1,14 @@
+// Copyright (c) 2019 HyperHQ Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// template implements base vm factory with vm templating.
+
+package template
+
+// templateDeviceStateSize denotes device state size when
+// mount tmpfs.
+// when bypass-shared-memory is not support like arm64,
+// creating template will occupy more space. That's why we
+// put it here.
+const templateDeviceStateSize = 300
--- a/virtcontainers/factory/template/template_ppc64le.go
+++ b/virtcontainers/factory/template/template_ppc64le.go
@@ -1,0 +1,14 @@
+// Copyright (c) 2019 HyperHQ Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// template implements base vm factory with vm templating.
+
+package template
+
+// templateDeviceStateSize denotes device state size when
+// mount tmpfs.
+// when bypass-shared-memory is not support like arm64,
+// creating template will occupy more space. That's why we
+// put it here.
+const templateDeviceStateSize = 8
--- a/virtcontainers/qemu_amd64.go
+++ b/virtcontainers/qemu_amd64.go
@@ -7,6 +7,7 @@ package virtcontainers
 
 import (
 	"os"
+	"time"
 
 	"github.com/kata-containers/runtime/virtcontainers/types"
 
@@ -23,6 +24,10 @@ const defaultQemuPath = "/usr/bin/qemu-system-x86_64"
 const defaultQemuMachineType = QemuPC
 
 const defaultQemuMachineOptions = "accel=kvm,kernel_irqchip,nvdimm"
+
+const qmpCapMigrationBypassSharedMemory = "bypass-shared-memory"
+
+const qmpMigrationWaitTimeout = 5 * time.Second
 
 var qemuPaths = map[string]string{
 	QemuPCLite: "/usr/bin/qemu-lite-system-x86_64",
--- a/virtcontainers/qemu_arch_base.go
+++ b/virtcontainers/qemu_arch_base.go
@@ -6,6 +6,7 @@
 package virtcontainers
 
 import (
+	"context"
 	"encoding/hex"
 	"fmt"
 	"os"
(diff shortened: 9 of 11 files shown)


virtcontainers/factory/template/template.go
```diff
@@ -93,7 +94,7 @@ func (t *template) prepareTemplateFiles() error {
 		return err
 	}
 	flags := uintptr(syscall.MS_NOSUID | syscall.MS_NODEV)
-	opts := fmt.Sprintf("size=%dM", t.config.HypervisorConfig.MemorySize+8)
+	opts := fmt.Sprintf("size=%dM", t.config.HypervisorConfig.MemorySize+templateDeviceStateSize)
```
jodh-intel: This change is specifically for arm64 so although it won't stop the feature working for other architectures, it clearly will have an impact. In summary, why not create a `virtcontainers/factory/template/template_arm64.go` with the modified value specifically for arm64?

virtcontainers/qemu.go
```diff
@@ -90,7 +90,7 @@ const (
 	qmpCapErrMsg                      = "Failed to negoatiate QMP capabilities"
 	qmpCapMigrationBypassSharedMemory = "bypass-shared-memory"
 	qmpExecCatCmd                     = "exec:cat"
-	qmpMigrationWaitTimeout           = 5 * time.Second
+	qmpMigrationWaitTimeout           = 10 * time.Second
```
jodh-intel: Same comment as above - why not make this value arm64-specific?

jongwu: @jodh-intel @bergwolf thanks, I agree with that. I will put all the value to arch base file. offering usingbypassSharedMemory cap is not necessary for other arch, so I will offer a const value as a  switch to decide if  ExecSetMigrationCaps should be executed.

virtcontainers/factory/template/template_amd64.go
```diff
@@ -0,0 +1,14 @@
...
+// template implements base vm factory with vm templating.
+
+package template
+
+// templateDeviceStateSize denotes device state size when
+// mount tmpfs.
+// when bypass-shared-memory is not support like arm64,
+// creating template will occupy more space. That's why we
+// put it here.
+const templateDeviceStateSize = 8
```
jodh-intel: Nit: I'd personally probably create a `genericTemplateDeviceStateSize` variable and then for all the non-arm64 files:

```go
const templateDeviceStateSize = genericTemplateDeviceStateSize
```

That's the approach we've taken in `cli/kata-check.go` (`genericCPUFlagsTag`, *etc*).

jongwu: @jodh-intel that's really better.

jodh-intel:
Ping @kata-containers/runtime - review required over here y'all! ;)

amshinde:
@jongwu Agree with @bergwolf here. It would make sense to have an architecture specific implementation for
`ExecSetMigrationCaps`. You would need to open a PR in this repo https://github.com/intel/govmm for that.

jongwu:
@amshinde thanks, I agree with @bergwolf , so I just issue an identical workaround for it, maybe an easier way. also I can fellow @bergwolf , but I assume that it is not necessary  to open a PR in https://github.com/intel/govmm, for we can fix it here.

jodh-intel:
@jongwu - if you need to modify govmm files (below `vendor/`), then yes, you do need to raise a PR on the govmm project. But if you can resolve the issue without modifying govmm files, you can just use this PR - see https://github.com/kata-containers/community/blob/master/VENDORING.md and https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#re-vendor-prs.

Let us know when this PR is ready to re-review.

amshinde:
@jongwu I will mark this as wip till the PR to `govmm` is submitted  and merged. Let us know when this PR is ready.

jodh-intel:
Hi @markdryan - is ^ something you might be able to help @jongwu with?

markdryan:
@jodh-intel Sure, but first could some clearly spell out what is required, either in this issue or in a new govmm issue?

Do you just want a version of ExecSetMigrationCaps that does nothing on arm?  What about ExecSetMigrateArguments?  What is the underlying reason for this?  Are these QMP methods not supported on ARM versions of QEMU?

jongwu:
@jodh-intel @markdryan thanks, we need an arm64 version of ExecSetMigrationCaps which just return nil. nothing should be done for ExecSetMigrateArguments. in additional, some copys of the normal implementation of ExecSetMigrationCaps  for other archs like amd64, ppc, should be added.
we need that PR because bypass-shared-memory caps does not support on arm64 in kata-runtime for now. so we need disable ExecSetMigrationCaps for arm64.

markdryan:
@jongwu Thanks for the clarification.  However, I'm not sure disabling ExecSetMigrationCaps on ARM is the correct thing to do.  If I understand correctly, the function itself should work on ARM but one of the capabilities kata is passing to this function is not supported on ARM.  Is that the case?

If so we can't really disable ExecSetMigrationCaps in govmm.  Another project may want to use GoVMM on ARM to set a migration capability that was supported.

I note that there does seem to be a QMP function called query-migrate-capabilities.  We could add a method for that which kata could call and then only pass "bypass-shared-memory" if it is supported.  Would that work?

jongwu:
@markdryan yeah, that's really the case. maybe disabling ExecSetMigrationCaps for arm64 doesn't make sense. @amshinde @bergwolf any comments?
howerver bypass-shared-memory caps is supported by qemu on arm64 now. running vm will go into failure if it cloning from a template vm which using bypass-shared-memory in migration.  That's the point. so query-migrate-capabilities would not help.

markdryan:
@jongwu I see.  That is unfortunate.   I don't know the kata code base but looking at where ExecSetMigrationCaps is called, could we not achieve what we want by ensuring BootToBeTemplate is never set to true on ARM or do you want the code in [setupTemplate](https://github.com/kata-containers/runtime/blob/9f96da2014f0894687e2935e92a20e3815ca7bc1/virtcontainers/qemu.go#L405) to run, even though ultimately, the caps should not be set.

markdryan:
The other thing we could do would be to provide an ARM version of ExecSetMigrationCaps in govmm that returns a specific error if " bypass-shared-memory" is passed as a parameter.  Then kata could could detect this error and ignore it.  The thing that worries me about this solution is that, just from looking at the code, it seems to me that the BootToBeTemplate functionality would still be half enabled as the code in setupTemplate would still run.

jongwu:
@markdryan don't worry. template can be enabled for arm64 just by disable bypass-shared-memroy cap when create the template. if saving vm as a template without using bypass-shared-memory, running vm cloning from it will start successfully. so if we can implement a version of ExecSetMigrationCaps returning nil when detect bypass-shared-memory, it will do what we want.

markdryan:
> so if we can implement a version of ExecSetMigrationCaps returning nil when detect bypass-shared-memory, it will do what we want.

We could do this but it would be a bit misleading.  Someone call could call ExecSetMigrationCaps with "bypass-shared-memory" have the function return nil and reasonably expect that the capability had been set, when in fact nothing had happened.  If we're going to do this we need to return an error to correctly signal that the parameter was not set.  This specific error could be trapped in kata and ignored.

BTW is there any bug open in qemu related to "bypass-shared-memory" and ARM we could reference when we make the change?

jongwu:
@markdryan it sounds great that the error can trap into kata and be ignored. but how can this error trapped into kata  be ignored?
there is no issue related bypass-shared-memory proposed to qemu community for now. but you can reference it freely.

markdryan:
Something like

```
	err = q.qmpMonitorCh.qmp.ExecSetMigrationCaps(q.qmpMonitorCh.ctx, []map[string]interface{}{
			{				{
				"capability": qmpCapMigrationBypassSharedMemory,					"capability": qmpCapMigrationBypassSharedMemory,
				"state":      true,					"state":      true,
			},				},
		})			})
		if err != nil && err != qemu.Unsupported {	
```

or something like that.

jongwu:
@markdryan yeah, I see. however I'm not sure about if this  additional check can be accept in kata.

markdryan:
@amshinde, @bergwolf, @jodh-intel Would this solution work for you, i.e., returning a specific error on ARM if "bypass-shared-memory" is passed to ExecSetMigrationCaps?  We can't disable the function entirely on ARM and we can't do nothing and return nil either, so this is the only option I can see for a govmm based fix.

bergwolf:
@markdryan yeah, it makes sense to return an error from govmm. However, I was actually thinking of adding a new function in the `qemuArch` interface and then qemu.go has something like:

```
       if q.config.BootToBeTemplate {
                err = q.arch.setBypassSharedMemoryMigrationCaps()
                if err != nil {
                        q.Logger().WithError(err).Error("set migration bypass shared memory")
                        return err
                }
        }
```

Then arm64 can just return nil doing nothing in `setBypassSharedMemoryMigrationCaps` until this is actually fixed in qemu. Other arch can use the default qemu_arch_base version implementing the original qmp action.

wdyt?

markdryan:
@bergwolf Okay, that sounds good to me.  If I understand correctly this wouldn't actually require any modifications to govmm at all.  The knowledge about bypassing migration caps not working on ARM would be coded into kata rather in govmm.  Is that correct?  I didn't realise that there was already an architectural abstraction in kata, but seeing as there is, I agree that this seems like a good place to make the fix.

bergwolf:
@markdryan yes, you are right. It doesn't need any govmm modification.

jongwu:
@amshinde @bergwolf @jodh-intel follow workaround from @bergwolf . Add setBypassSharedMemoryMigrationCaps to enable bypass-shared-memory cap for no-arm arch and return nil for arm64 only.

jongwu:
@amshinde any comments?

jongwu:
Error Trace:	linters_test.go:34
    			            				linters_test.go:138
    			            				linters_test.go:54
    			Error:      	Received unexpected error:
    			            	
    			            	unmatched errors
    			            	2019/03/23 06:52:27 unknown architecture wasm
@jodh-intel @amshinde I'm confused with these lots of similar odd failures.  what should I do?

teawater:
The travis-ci cannot pass until https://github.com/kata-containers/runtime/pull/1372 get merged.

jongwu:
@teawater thanks, does it means that this commit won't merge before #1372?

teawater:
> @teawater thanks, does it means that this commit won't merge before #1372?

Looks new policy is that.  We can do test first.

raravena80:
@teawater @jongwu ping. Any updates? Thx!

teawater:
> @teawater @jongwu ping. Any updates? Thx!

#1372 is still not merged.  And looks most of test failed.

ganeshmaharaj:
The issue is with k8s tests and there is a workaround in place that fuentess is working on. I can rekick #1372 as soon as that workaround lands.

teawater:
continuous-integration/travis-ci/pr still try to use go 1.10 but  #1372 changed it to go1.11 to handle the issue.
Is there any way to let travis-ci change to use 1.11 except re-push the code?

virtcontainers/factory/template/template_amd64.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2018 HyperHQ Inc.
```
devimc: 2019

virtcontainers/factory/template/template_s390x.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2018 HyperHQ Inc.
```
devimc: wrong year in all new files

virtcontainers/qemu_arm64.go
```diff
@@ -189,3 +193,8 @@ func (q *qemuArm64) appendImage(devices []govmmQemu.Device, path string) ([]govm
 
 	return devices, nil
 }
+
+func (q *qemuArm64) setBypassSharedMemoryMigrationCaps(_ context.Context, _ *govmmQemu.QMP) error {
+	// bypass-shared-memory not support in arm64 for now
```
devimc: cc @Pennyzct

virtcontainers/factory/template/template_s390x.go
jongwu: @devimc thanks, I will fix all of them.

virtcontainers/qemu_arm64.go
Pennyzct: thanks for the heads-up, ;), yes, sadly, that's the current status.😥

jongwu:
@jodh-intel  @bergwolf why travis-ci can't pass?

grahamwhaley:
Hi @jongwu - it has hit a strange compilation (link) error - I think we have seen this on another PR, and I'm not sure it is because of your changes. @jodh-intel @sboeuf @chavafg - is this something we have tried to diagnose yet?:

```
# github.com/kata-containers/runtime/cli/containerd-shim-kata-v2
/home/travis/.gimme/versions/go1.11.7.linux.amd64/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/bin/ld: /tmp/go-link-053240943/000010.o: unrecognized relocation (0x2a) in section `.text'
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
```

jongwu:
@grahamwhaley thx.

devimc:
ARM CI is not happy

```
level=error msg="Create new sandbox failed" error="QEMU path (/tmp/vc-tmp-578885730/hypervisor) does not exist" sandbox=7f49d00d-1995-4156-8c79-5f5ab24ce138 sandboxid=7f49d00d-1995-4156-8c79-5f5ab24ce138 source=virtcontainers subsystem=sandbox
```

@Pennyzct can we merge this either way?

Pennyzct:
Hi~ @devimc I will /retest to see if it's repeatable. ;)

devimc:
@Pennyzct same error

Pennyzct:
@devimc thanks. I will check if it's related. ;)

jongwu:
@devimc @Pennyzct help retest.

jcvenegas:
ARM gut stuck at build triggered on github, need a new retests.
@GabyCT could you trigger a rebuild just for arm?

Also in Fedora got this error: Looks like is a non stable test I've seen with a few times 
@devimc @chavafg @grahamwhaley 
```
• Failure [21.767 seconds]
state
/tmp/jenkins/workspace/kata-containers-runtime-fedora-PR/go/src/github.com/kata-containers/tests/functional/state_test.go:26
  container
  /tmp/jenkins/workspace/kata-containers-runtime-fedora-PR/go/src/github.com/kata-containers/tests/vendor/github.com/onsi/ginkgo/extensions/table/table.go:92
    with workload [true], timeWait 5 [It]
    /tmp/jenkins/workspace/kata-containers-runtime-fedora-PR/go/src/github.com/kata-containers/tests/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:46

    Expected
        <int>: 1
    to equal
        <int>: 0

    /tmp/jenkins/workspace/kata-containers-runtime-fedora-PR/go/src/github.com/kata-containers/tests/functional/state_test.go:45


Summarizing 1 Failure:

[Fail] state container [It] with workload [true], timeWait 5 
/tmp/jenkins/workspace/kata-containers-runtime-fedora-PR/go/src/github.com/kata-containers/tests/functional/state_test.go:45

Ran 8 of 8 Specs in 24.580 seconds
```

grahamwhaley:
That second fail @jcvenegas looks like the one I'm seeing over on https://github.com/kata-containers/packaging/pull/314 I think...

GabyCT:
@jcvenegas it looks like it is already running

jongwu:
@devimc what's next?

devimc:
@jongwu thanks

PR #1340, opened 2019-03-07
pkg: reduce memory footprint

### LINKED ISSUE #1339: reduce memory footprint
reduce memory footprint by disabling systemd udevd and journald

devimc:
Reduce memory footprint ~7% by disabling some systemd services like
systemd-journald and systemd-udevd, those services are just consuming memory
and are not needed. For example kata-agent logs the errors through the proxy.

screenshot from 2019-03-07 09-25-08

fixes #1339

### DIFF
--- a/pkg/katautils/create.go
+++ b/pkg/katautils/create.go
@@ -37,11 +37,72 @@ var systemdKernelParam = []vc.Param{
 	},
 }
 
-func getKernelParams(needSystemd bool) []vc.Param {
+// kernel params to improve memory footprint
+var noTraceKernelParam = []vc.Param{
+	// No logs: agent has its own logging system
+	{
+		Key:   "systemd.mask",
+		Value: "systemd-journald.service",
+	},
+	{
+		Key:   "systemd.mask",
+		Value: "systemd-journald.socket",
+	},
+	{
+		Key:   "systemd.mask",
+		Value: "systemd-journal-flush.service",
+	},
+	// No udev events: agent implements udev events
+	{
+		Key:   "systemd.mask",
+		Value: "systemd-udevd.service",
+	},
+	{
+		Key:   "systemd.mask",
+		Value: "systemd-udevd.socket",
+	},
+	{
+		Key:   "systemd.mask",
+		Value: "systemd-udev-trigger.service",
+	},
+	// No timesync: kata is able to setup the time and this service consume network
+	{
+		Key:   "systemd.mask",
+		Value: "systemd-timesyncd.service",
+	},
+	// No update audit logs
+	{
+		Key:   "systemd.mask",
+		Value: "systemd-update-utmp.service",
+	},
+	// No temporal files
+	{
+		Key:   "systemd.mask",
+		Value: "systemd-tmpfiles-setup.service",
+	},
+	{
+		Key:   "systemd.mask",
+		Value: "systemd-tmpfiles-cleanup.service",
+	},
+	{
+		Key:   "systemd.mask",
+		Value: "systemd-tmpfiles-cleanup.timer",
+	},
+	// No mounts
+	{
+		Key:   "systemd.mask",
+		Value: "tmp.mount",
+	},
+}
+
+func getKernelParams(needSystemd, trace bool) []vc.Param {
 	p := []vc.Param{}
 
 	if needSystemd {
 		p = append(p, systemdKernelParam...)
+		if !trace {
+			p = append(p, noTraceKernelParam...)
+		}
 	}
 
 	return p
--- a/pkg/katautils/config.go
+++ b/pkg/katautils/config.go
@@ -663,7 +663,7 @@ func updateRuntimeConfigShim(configPath string, tomlConf tomlConfig, config *oci
 // SetKernelParams adds the user-specified kernel parameters (from the
 // configuration file) to the defaults so that the former take priority.
 func SetKernelParams(runtimeConfig *oci.RuntimeConfig) error {
-	defaultKernelParams := GetKernelParamsFunc(needSystemd(runtimeConfig.HypervisorConfig))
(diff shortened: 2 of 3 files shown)


pkg/katautils/create.go
```diff
@@ -35,6 +35,22 @@ var systemdKernelParam = []vc.Param{
 		Key:   "systemd.mask",
 		Value: "systemd-networkd.socket",
 	},
+	{
+		Key:   "systemd.mask",
+		Value: "systemd-journald.service",
```
jodh-intel: I think this will break the agent tracing PR - https://github.com/kata-containers/agent/pull/415, specifically the `kata-journald-host-redirect.service` servivce. That isn't technically part of tracing, but allows guest journal messages to propagate out of the VM when vsock is enabled.

devimc: I guess we can do something like this

```go
if configuration.enable_tracing == false {
    // Disable systemd udev and journald
}
```

devimc:
metrics CI is not happy, I think that's fine

```
+-----+----------------------+-------+-------+--------+-------+-------+--------+-------+------+-----+
| P/F |         NAME         |  FLR  | MEAN  |  CEIL  |  GAP  |  MIN  |  MAX   |  RNG  | COV  | ITS |
+-----+----------------------+-------+-------+--------+-------+-------+--------+-------+------+-----+
| P   | boot-times           | 95.0% | 96.3% | 105.0% | 10.0% | 89.3% | 103.7% | 16.1% | 4.2% |  20 |
| *F* | memory-footprint     | 95.0% | 91.8% | 105.0% | 10.0% | 91.8% | 91.8%  | 0.0%  | 0.0% |   1 |
| *F* | memory-footprint-ksm | 95.0% | 87.2% | 105.0% | 10.0% | 87.2% | 87.2%  | 0.0%  | 0.0% |   1 |
+-----+----------------------+-------+-------+--------+-------+-------+--------+-------+------+-----+
```

grahamwhaley:
In fact, I'd call it 'expected'. If you did not fail the metrics CI on the footprint tests, then I would have questioned if the patch really worked ;-)
so, for the footprint tests, it is showing a more like 8% and 12.8% saving (different machine sizes may make the difference). So, right ballpark with your measures I think.

devimc:
@jodh-intel please take a look

devimc:
Pull request updated, now boot time and memory footprint are better then before :smile: 

```
14:44:52 +-----+----------------------+-------+-------+--------+-------+-------+-------+-------+------+-----+
14:44:52 | P/F |         NAME         |  FLR  | MEAN  |  CEIL  |  GAP  |  MIN  |  MAX  |  RNG  | COV  | ITS |
14:44:52 +-----+----------------------+-------+-------+--------+-------+-------+-------+-------+------+-----+
14:44:52 | *F* | boot-times           | 95.0% | 89.6% | 105.0% | 10.0% | 82.9% | 99.1% | 19.4% | 4.0% |  20 |
14:44:53 | *F* | memory-footprint     | 95.0% | 88.6% | 105.0% | 10.0% | 88.6% | 88.6% | 0.0%  | 0.0% |   1 |
14:44:53 | *F* | memory-footprint-ksm | 95.0% | 83.2% | 105.0% | 10.0% | 83.2% | 83.2% | 0.0%  | 0.0% |   1 |
14:44:53 +-----+----------------------+-------+-------+--------+-------+-------+-------+-------+------+-----+
```

devimc:
@amshinde change applied, thanks

jodh-intel:
```
22:34:05 Report Summary:
22:34:05 +-----+----------------------+-------+-------+--------+-------+-------+-------+-------+------+-----+
22:34:05 | P/F |         NAME         |  FLR  | MEAN  |  CEIL  |  GAP  |  MIN  |  MAX  |  RNG  | COV  | ITS |
22:34:05 +-----+----------------------+-------+-------+--------+-------+-------+-------+-------+------+-----+
22:34:05 | *F* | boot-times           | 95.0% | 89.7% | 105.0% | 10.0% | 84.1% | 96.0% | 14.2% | 3.7% |  20 |
22:34:05 | *F* | memory-footprint     | 95.0% | 88.4% | 105.0% | 10.0% | 88.4% | 88.4% | 0.0%  | 0.0% |   1 |
22:34:05 | *F* | memory-footprint-ksm | 95.0% | 83.2% | 105.0% | 10.0% | 83.2% | 83.2% | 0.0%  | 0.0% |   1 |
22:34:05 +-----+----------------------+-------+-------+--------+-------+-------+-------+-------+------+-----+
22:34:05 Fails: 3, Passes 0
```

Still failing ;(

grahamwhaley:
It's expected to fail @jodh-intel I think - this PR reduces the memory footprint etc., which means it falls below the limits set by the metrics CI. Yeah, the metrics CI does not 'automatically level', and even if it did, it would still flag that this PR has changed the metrics measures....
So, for this PR, we get to not-block on the metrics CI - *but*, once this PR lands then we have to go re-adjust the metrics CI expectations or all other PRs will start to fail.

Speaking of which then....
 - I'd like somebody else to do that this time, so we have an [anti-bus](https://en.wikipedia.org/wiki/Bus_factor) strategy for if I am unavailable (which will happen at some point...)
 - currently the base level metrics are stored on the CI slave machines themselves. I think we should make it easier to tweak them, by adding them into one of the repos and having the slaves pull their specific config files down for each new build (the config files are machine specific).... let me at least open an Issue for that
- longer term it would be lovely if the CI could either reference recent historical data to dynamically derive a baseline (but we have no historic store yet), or do two runs and check for a difference between them (but that would rely on runs being *very* reproducible, which they are not as reproducible as I'd like, so the implication here is that we'd probably have to widen the 'pass' bounds checks).

devimc:
can we merge this?

jodh-intel:
@grahamwhaley's comment says yes to me...

grahamwhaley:
Excellent. Now, expect the metrics CI to start failing for PRs until we have gathered a few runs to be able to reset its expectations....
/cc @chavafg @sboeuf

Issue #1296, opened 2019-02-28
unit-tests: add tests for resource requests for pods


egernst:
We had an issue where vCPUs weren't being accounted correctly, [1].  I noticed we don't have unit tests -- let's make sure we add some to cover the potential scenarios!

[1] - https://github.com/kata-containers/runtime/issues/1292

grahamwhaley:
I've assigned to @devimc , who I think has the most history and knowledge of that bit of code etc.

PR #1346, opened 2019-03-08
virtcontainers: move resource calculation to its own function

jcvenegas:
Make cpu and memory calculation in a different function this help to reduce
the function complexity and easy  unit test.

Fixes: #1296

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -7,6 +7,7 @@ package virtcontainers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net"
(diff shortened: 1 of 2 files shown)


virtcontainers/sandbox.go
```diff
@@ -1666,3 +1652,27 @@ func (s *Sandbox) updateResources() error {
 	}
 	return nil
 }
+
+func calculateSandboxMemory(containers []ContainerConfig) int64 {
```
jcvenegas: self review: probably can convert to a sandbox method and do not receive containers as function parameter. 
@devimc @jodh-intel any preference here?

devimc: or you can move them to utils

egernst:
@gabibeyer PTAL?

virtcontainers/sandbox.go
jodh-intel: Yeah, I'd either make it a sandbox method, or just pass it the sandbox and put it in utils as @devimc suggests.

/cc @sboeuf, @bergwolf.

sboeuf: I would keep that in `sandbox.go` as a sandbox method. This makes more sense IMO as I don't see this function being reused by anything else since it is very sandbox specific. The code we put in utils should be very generic so that it could be reused by any piece of code.
And I think `calculateSandboxCPUs()` is actually the perfect example where it is a sandbox method, that relies onto some very generic `utils.CalculateMilliCPUs()` function calls.

jcvenegas: thank you all, moved as sandbox method.

PR #1374, opened 2019-03-15
volumes: Handle k8s empty-dirs of "default" medium type

### LINKED ISSUE #1341: EmptyDir in k8s handled incorrectly
# EmptyDir in k8s handled incorrectly

See https://github.com/kata-containers/runtime/issues/61#issuecomment-440645443

> EmptyDir defined in k8s supposed having three types of medium, default, tmpfs, hugepage. And the default medium should be node disk instead of memdisk, should it be better we consider mount option before pass ephemeral volumes into guest and back to use 9pfs when it was backed by default medium

https://kubernetes.io/docs/concepts/storage/volumes/#emptydir

By default, emptyDir volumes are stored on whatever medium is backing the node - that might be disk or SSD or network storage, depending on your environment. However, you can set the emptyDir.medium field to "Memory" to tell Kubernetes to mount a tmpfs (RAM-backed filesystem) for you instead. While tmpfs is very fast, be aware that unlike disks, tmpfs is cleared on node reboot and any files you write will count against your Container’s memory limit.

https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#enforcing-node-allocatable

@harche @amshinde @linxiulei

# Expected result

When medium is not specified or default we should use 9p/virtio-fs

# Actual result

We use tmpfs within the VM resulting in incorrect behaviour, scheduling and memory accounting'

amshinde:
We were considering all empty-dir k8s volumes as backed by tmpfs.
However they can be backed by a host directory as well.
Pass those as 9p volumes, while tmpfs volumes are handled as before,
namely creating a tmpfs directory inside the guest.
The only way to detect "Memory" empty-dirs is to actually check if the
volume is mounted as a tmpfs mount, since any information of k8s
"medium" is lost at the OCI layer.

Fixes #1341

### DIFF
--- a/pkg/katautils/utils.go
+++ b/pkg/katautils/utils.go
@@ -14,6 +14,8 @@ import (
 	"path/filepath"
 	"strings"
 	"syscall"
+
+	vc "github.com/kata-containers/runtime/virtcontainers"
 )
 
 const (
@@ -43,7 +45,9 @@ func IsEphemeralStorage(path string) bool {
 	if len(splitSourceSlice) > 1 {
 		storageType := splitSourceSlice[len(splitSourceSlice)-2]
 		if storageType == k8sEmptyDir {
-			return true
+			if _, fsType, _ := vc.GetDevicePathAndFsType(path); fsType == "tmpfs" {
+				return true
+			}
 		}
 	}
 	return false
--- a/pkg/katautils/create_test.go
+++ b/pkg/katautils/create_test.go
@@ -16,6 +16,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"testing"
 
 	vc "github.com/kata-containers/runtime/virtcontainers"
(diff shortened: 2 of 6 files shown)


jodh-intel:
Hi @amshinde - The CI has caught a static analysis issue:

```
virtcontainers/mount.go:167:1:warning: exported function GetDevicePathAndFsType should have comment or be unexported (golint)

```

amshinde:
@jodh-intel thanks, I have fixed the issue now.

mcastelino:
@amshinde so a tmpfs backed empty dir will have a tmpfs based empty dir on host and one more in the VM. But the emptyDir on host is not really used. So should not chew up memory.

pkg/katautils/utils.go
```diff
@@ -43,7 +45,9 @@ func IsEphemeralStorage(path string) bool {
 	if len(splitSourceSlice) > 1 {
 		storageType := splitSourceSlice[len(splitSourceSlice)-2]
 		if storageType == k8sEmptyDir {
-			return true
+			if _, fsType, _ := vc.GetDevicePathAndFsType(path); fsType == "tmpfs" {
```
mcastelino: @amshinde under what other conditions is IsEphemeralStorage called. Only for volumes?

amshinde: @mcastelino Thats right.

amshinde:
@mcastelino Yes, since we dont use the host tmpfs, it should not consume any memory on the host side.

amshinde:
@sboeuf `empty-dir` volumes can be of medium type "Memory" or "default".
In the first case, k8s creates a tmpfs backed volume. It makes sense for us to allocate memory inside the guest instead and mount it in the container namespace.
In case of "default" type, k8s creates a directory on the host and that directory is passed as a volume.
It makes sense to pass the host-directory to the guest using 9p in that case.I hope its clearer.

Issue #1347, opened 2019-03-08
Unable to use PHC time source in qemu guest.


dylanzr:
# Description of problem

Similar to #1279, we need to be able to keep guest time synchronized. Using ptp_kvm for a virtual PHC seems like an appropriate solution.

# Expected result

According to the kernel configs in the packaging repository, it appears that `CONFIG_PTP_1588_CLOCK_KVM` is enabled. I'd expect a `/dev/ptp0` interface to exist in the qemu guest.

# Actual result

The char special file does not appear to be created. In a debug console, it looks like the kernel has support for ptp_kvm, but I'm unable to see or create the char device.
```
bash-4.4# dmesg -T | grep -i ptp
[Thu Mar  7 22:49:24 2019] PTP clock support registered
bash-4.4# grep ptp /proc/devices
250 ptp
bash-4.4# ls -l /dev/ptp0
ls: cannot access '/dev/ptp0': No such file or directory
bash-4.4# mknod /dev/ptp0 c 250 0
mknod: /dev/ptp0: Read-only file system
```

---

This issue is present in our kubernetes environments, but the below data was generated from a test environment using docker. I suspect the actual implementation matters little for this issue.

<details><summary>click here for data</summary>
<p>
# Meta details

Running `kata-collect-data.sh` version `1.5.1 (commit b1ef5212114ca866fd831f1620a374e02aad61e1-dirty)` at `2019-03-08.11:01:14.224200419-0600`.

---

Runtime is `/root/go/src/github.com/kata-containers/runtime/kata-runtime`.

# `kata-env`

Output of "`/root/go/src/github.com/kata-containers/runtime/kata-runtime kata-env`":
```toml
[Meta]
  Version = "1.0.20"

[Runtime]
  Debug = true
  Trace = false
  DisableGuestSeccomp = false
  DisableNewNetNs = false
  Path = "/root/go/src/github.com/kata-containers/runtime/kata-runtime"
  [Runtime.Version]
    Semver = "1.5.1"
    Commit = "b1ef5212114ca866fd831f1620a374e02aad61e1-dirty"
    OCI = "1.0.1-dev"
  [Runtime.Config]
    Path = "/etc/kata-containers/configuration.toml"

[Hypervisor]
  MachineType = "pc"
  Version = "QEMU emulator version 3.1.50\nCopyright (c) 2003-2018 Fabrice Bellard and the QEMU Project developers"
  Path = "/opt/virtiofs/bin/qemu-system-x86_64"
[...]

PR #1387, opened 2019-03-19
Allow data collection to be hidden

### LINKED ISSUE #1386: Allow data collection script output to be hidden
The output of the [data collection script](https://github.com/kata-containers/runtime/blob/master/data/kata-collect-data.sh.in) is designed to be pasted directly into a github issue.

That results in a *lot* of output so using a cute trick which @dylanzr introduced us to, allow the output of the script to be hidden / unhidden by clicking on a "twistie". See the example at the top of https://github.com/kata-containers/runtime/issues/1347.

jodh-intel:
Use a clever HTML trick to allow the output of the data collection script to be hidden / unhidden in the github.com interface.

See the example at the top of https://github.com/kata-containers/runtime/issues/1347.

Fixes #1386.

### DIFF
--- a/data/kata-collect-data.sh.in
+++ b/data/kata-collect-data.sh.in
@@ -672,8 +672,27 @@ read_osbuilder_file()
 	cat "$file"
 }
 
+show_header()
+{
+	cat <<EOT
+<details>
+<summary>Show <tt>$script_name</tt> details</summary>
+<p>
+EOT
+}
+
+show_footer()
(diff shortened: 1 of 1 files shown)


jodh-intel:
https://github.github.com/gfm/#raw-html ;)

jodh-intel:
This script change cannot be causing (only a few) CI's to fail in this way. Merging...

PR #1390, opened 2019-03-19
virtcontainers: improve security and mount the rootfs as read-only fs

### LINKED ISSUE #1389: rootfs: rootfs is mounted with rw access
Currently the rootfs is mounted with rw access, let's improve security and mount it as a read-only filesystem, benefits:

* Security: binaries can't be modified
* Experimental: snap will support rootfs (finally :tada: !!)

devimc:
Mounting the rootfs as read-only fs the binaries can't be modified.

fixes #1389

### DIFF
--- a/virtcontainers/qemu_amd64.go
+++ b/virtcontainers/qemu_amd64.go
@@ -32,7 +32,7 @@ var qemuPaths = map[string]string{
 
 var kernelRootParams = []Param{
 	{"root", "/dev/pmem0p1"},
-	{"rootflags", "dax,data=ordered,errors=remount-ro rw"},
+	{"rootflags", "dax,data=ordered,errors=remount-ro ro"},
 	{"rootfstype", "ext4"},
 }
 
--- a/pkg/katautils/create.go
+++ b/pkg/katautils/create.go
@@ -93,6 +93,11 @@ var noTraceKernelParam = []vc.Param{
 		Key:   "systemd.mask",
 		Value: "tmp.mount",
 	},
+	// No random seed
+	{
+		Key:   "systemd.mask",
+		Value: "systemd-random-seed.service",
+	},
 }
 
 func getKernelParams(needSystemd, trace bool) []vc.Param {


amshinde:
@devimc Will this not prevent changes to /sys as well? I am asking because we work with the guest in a way that requires changes to sysfs such as hotplugging devices.

virtcontainers/qemu_amd64.go
```diff
@@ -32,7 +32,7 @@ var qemuPaths = map[string]string{
 
 var kernelRootParams = []Param{
 	{"root", "/dev/pmem0p1"},
-	{"rootflags", "dax,data=ordered,errors=remount-ro rw"},
+	{"rootflags", "dax,data=ordered,errors=remount-ro ro"},
```
sboeuf: I thought our rootfs was already read-only? Was I the only one thinking this?

devimc:
@amshinde 
> Will this not prevent changes to /sys as well?

no, `/sys` is volatile and mounted with rw

lifupan:
Does the other arch need this change, such as arm64?

virtcontainers/qemu_amd64.go
grahamwhaley: I was thinking 'why did we not do this already, forever - maybe there is a reason...' :-)
@jodh-intel - I don't suppose you remember from wayback do you?
I suspect @gnawux @bergwolf might have some thoughts as well.

jodh-intel: I'm frankly surprised this wasn't raised on one of the https://github.com/clearcontainers repos but cannot find any mention of it.

After a bit of digital archaeology, I found a very old comment from a test tool I wrote a few years ago that stated we needed to mount the rootfs `rw` with DAX/NVDIMM to avoid a kernel panic like this:

```
Kernel panic - not syncing: Requested init /usr/lib/systemd/systemd failed (error -14)
```

I believe we were seeing that behaviour with a `linux-container-4.2.5-44` kernel but cannot be 100% sure. 

However, since this PR does indeed work, something changed :smile:. It would be good to get input from the DAX experts on this PR.

jodh-intel: We even mounted `rw` in the predecessor to `clearcontainers`, `cc-oci-runtime`:

- https://github.com/intel/cc-oci-runtime/blob/master/data/kernel-cmdline

devimc:
@lifupan 
> Does the other arch need this change, such as arm64?

good question

 cc @Pennyzct

devimc:
@jodh-intel  CI is happy, so I think this change is fine

virtcontainers/qemu_amd64.go
grahamwhaley: @djbw - any thoughts from the DAX side wrt mounting ro instead of rw

devimc:
@grahamwhaley  do we have a smaller memory footprint with this change?

```
19:16:51 +-----+----------------------+-------+--------+--------+-------+--------+--------+-------+------+-----+
19:16:51 | P/F |         NAME         |  FLR  |  MEAN  |  CEIL  |  GAP  |  MIN   |  MAX   |  RNG  | COV  | ITS |
19:16:51 +-----+----------------------+-------+--------+--------+-------+--------+--------+-------+------+-----+
19:16:51 | P   | boot-times           | 95.0% | 99.1%  | 105.0% | 10.0% | 93.2%  | 107.6% | 15.5% | 4.4% |  20 |
19:16:51 | P   | memory-footprint     | 95.0% | 102.1% | 105.0% | 10.0% | 102.1% | 102.1% | 0.0%  | 0.0% |   1 |
19:16:51 | *F* | memory-footprint-ksm | 95.0% | 124.6% | 105.0% | 10.0% | 124.6% | 124.6% | 0.0%  | 0.0% |   1 |
19:16:51 +-----+----------------------+-------+--------+--------+-------+--------+--------+-------+------+-----+
```

grahamwhaley:
@devimc - the results are showing *bigger* KSM footprint (the mean is above the ceiling), but about the same for non-KSM. I wonder if qemu is mapping it differently (not mergeable) if `ro`. That would almost feel backwards though?
Given metrics is a bit up/down right now, probably best doing a local report tool test to see how it looks on your system I think.

virtcontainers/qemu_amd64.go
devimc: works for me with DAX enabled

```
[    0.282411] EXT4-fs (pmem0p1): DAX enabled. Warning: EXPERIMENTAL, use at your own risk
[    0.282687] EXT4-fs (pmem0p1): mounted filesystem with ordered data mode. Opts: dax,data=ordered,errors=remount-ro
```

sboeuf: Yes pmem metadata simply need to be read, so no issues from this perspective.

devimc:
@amshinde @grahamwhaley @jodh-intel @sboeuf please take a look

grahamwhaley:
Well, nobody is objecting, and everybody is promising this won't break stuff ... so, I'll merge... 🤞

Issue #1379, opened 2019-03-18
Make get fail without curl


teawater:
teawater@teawater-virtual-machine:~/gopath/src/github.com/kata-containers/runtime$ make
INFO: yq was not found, installing it
.ci/install-yq.sh: 行 56: curl: 未找到命令
.ci/install-yq.sh: 行 62: curl: 未找到命令
chmod: 无法访问'/home/teawater/gopath//bin/yq': 没有那个文件或目录
ERROR: Cannot not get /home/teawater/gopath//bin/yq executable
make: /home/teawater/gopath//bin/yq：命令未找到
golang.mk:33: *** "ERROR: cannot determine minimum golang version"。 停止。

PR #1398, opened 2019-03-21
ci: check curl before use it

teawater:
install-yq.sh use curl but not check if curl is available or not.
Add code to check curl before use it.

Fixes: #1379

### DIFF
--- a/.ci/install-yq.sh
+++ b/.ci/install-yq.sh
@@ -51,6 +51,11 @@ function install_yq() {
 
 	mkdir -p "${GOPATH}/bin"
 
+	# Check curl
+	if ! command -v "curl" >/dev/null; then
+		die "Please install curl"
+	fi
+
 	# Workaround to get latest release from github (to not use github token).
 	# Get the redirection to latest release on github.
 	yq_latest_url=$(curl -Ls -o /dev/null -w %{url_effective} "https://${yq_pkg}/releases/latest")


teawater:
jenkins-metrics-ubuntu-16-04 failed:
Report Summary:
+-----+----------------------+-------+--------+--------+-------+--------+--------+-------+------+-----+
| P/F |         NAME         |  FLR  |  MEAN  |  CEIL  |  GAP  |  MIN   |  MAX   |  RNG  | COV  | ITS |
```
+-----+----------------------+-------+--------+--------+-------+--------+--------+-------+------+-----+
| P   | boot-times           | 95.0% | 102.8% | 105.0% | 10.0% | 95.9%  | 110.7% | 15.5% | 4.5% |  20 |
| P   | memory-footprint     | 95.0% | 102.2% | 105.0% | 10.0% | 102.2% | 102.2% | 0.0%  | 0.0% |   1 |
| *F* | memory-footprint-ksm | 95.0% | 122.4% | 105.0% | 10.0% | 122.4% | 122.4% | 0.0%  | 0.0% |   1 |
+-----+----------------------+-------+--------+--------+-------+--------+--------+-------+------+-----+
Fails: 1, Passes 2
Failed
checkmetrics FAILED (1)
```
This pr doesn't change the default behavior of kata-runtime.
So I think this fail does not relate to this fail.

.ci/install-yq.sh
```diff
@@ -51,6 +51,11 @@ function install_yq() {
 
 	mkdir -p "${GOPATH}/bin"
 
+	# Check curl
+	if ! command -v "curl" >/dev/null; then
+		die "Please install curl"
```
gnawux: It may confuse users because he/she don't want to install curl originally. Is the following one better:

> Failed to install yq: no curl installed

PR #1403, opened 2019-03-21
Add more build targets

jodh-intel:
Add `install-runtime` and `install-netmon` targets. This allows the `install` target to be simplified and also allows the runtime to be built without having to build the `containerd-shim-v2` binary which is slow to build:

```
$ make runtime && sudo -E PATH=$PATH make install-runtime
```


jodh-intel:
@grahamwhaley - yes :) And this change gives me an almost 100% build speed improvement when I build just the runtime.

jodh-intel:
Err, 50% even ;)

jodh-intel:
All CI's passed except for 18.04:

```
not ok 36 ctr with non-root user has no effective capabilities
# (in test file ctr.bats, line 1243)
# 0
```

Restarting...

Issue #1415, opened 2019-03-25
Remove duplicate fields "Resources" from `State`


WeiZhang555:
In new sandbox cgroups implementation, `Resources` was added to `State` structure which will be also saved to disk.
The field is already saved once in `ContainerConfig` and also saved to disk already, so we don't need a duplicated fields here.

PR #1416, opened 2019-03-25
cgroups: remove duplicate fields from state

WeiZhang555:
Fixes: #1415

Container resources have been saved to ContainerConfig so there's no
need to save it again in state.json.

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -1295,7 +1295,7 @@ func (c *Container) newCgroups() error {
 		return fmt.Errorf("Could not create cgroup for %v: %v", c.state.CgroupPath, err)
 	}
 
-	c.state.Resources = resources
+	c.config.Resources = resources
 
 	// Add shim into cgroup
 	if c.process.Pid > 0 {
@@ -1359,8 +1359,8 @@ func (c *Container) updateCgroups(resources specs.LinuxResources) error {
 	}
 
 	// store new resources
-	c.state.Resources = r
-	if err := c.store.Store(store.State, c.state); err != nil {
+	c.config.Resources = r
+	if err := c.storeContainer(); err != nil {
 		return err
 	}
 
--- a/virtcontainers/types/sandbox.go
+++ b/virtcontainers/types/sandbox.go
@@ -8,8 +8,6 @@ package types
 import (
 	"fmt"
 	"strings"
-
-	specs "github.com/opencontainers/runtime-spec/specs-go"
 )
 
 // StateString is a string representing a sandbox state.
@@ -50,12 +48,6 @@ type State struct {
 	// CgroupPath is the cgroup hierarchy where sandbox's processes
 	// including the hypervisor are placed.
 	CgroupPath string `json:"cgroupPath,omitempty"`
-
-	// Resources contains the resources assigned to the container.
-	// When a container is created resources specified in the config json
-	// are used, those resources change when a container is updated but
-	// the config json is not updated.
-	Resources specs.LinuxResources `json:"resources,omitempty"`
 }
 
 // Valid checks that the sandbox state is valid.
--- a/virtcontainers/cgroups.go
+++ b/virtcontainers/cgroups.go
@@ -281,36 +281,36 @@ func (s *Sandbox) cpuResources() *specs.LinuxCPU {
 			continue
 		}
 
-		if c.state.Resources.CPU == nil {
+		if c.config.Resources.CPU == nil {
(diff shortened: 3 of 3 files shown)


WeiZhang555:
I'm trying to simplify our persist data, `Resources` isn't an necessary field in `State`, maybe `CgroupPath` is not either. Still investigating.

cc @devimc

devimc:
@WeiZhang555 thanks, this change looks good

virtcontainers/types/sandbox.go
```diff
@@ -50,12 +48,6 @@ type State struct {
 	// CgroupPath is the cgroup hierarchy where sandbox's processes
 	// including the hypervisor are placed.
 	CgroupPath string `json:"cgroupPath,omitempty"`
-
-	// Resources contains the resources assigned to the container.
-	// When a container is created resources specified in the config json
-	// are used, those resources change when a container is updated but
-	// the config json is not updated.
-	Resources specs.LinuxResources `json:"resources,omitempty"`
```
sboeuf: @WeiZhang555 I don't think you should remove this. Look at the comment here, it states that at first the container resources are going to be identical to the container config, but later if those container resources are updated, the container config will not be updated, and that's why this resource structure is important. It will hold the new values.

WeiZhang555: Let me test it. I guess you're right. How about update it back to config?

WeiZhang555: Commit is updated. @sboeuf 

The updated resource will be write back to config in https://github.com/kata-containers/runtime/pull/1416/files#diff-8a56f5a1fa5971cc43e650ffa24f1e2bR1362 .

The problem here is, we store container resources both in state file and config file, it's not necessary. I am trying to remove duplicated fields and only keep necessary fields.

If we want to original unchanged resources, we can still find it in Annotations[configJSONKey], right? So from every point, there's one unnecessary copy here.

virtcontainers/container.go
```diff
@@ -1359,8 +1359,8 @@ func (c *Container) updateCgroups(resources specs.LinuxResources) error {
 	}
 
 	// store new resources
-	c.state.Resources = r
-	if err := c.store.Store(store.State, c.state); err != nil {
+	c.config.Resources = r
```
devimc: is valid to do this? can we modify the config.json ?

virtcontainers/types/sandbox.go
sboeuf: I think the bottom question is, do we expect applied container resources to be different from the expected container resources? For instance, could we decide to give a container more memory because we're rounding things up, but when you do the computation when a new container is created, it's actually better to sum up from the config than the state?
I'm trying to understand if we have added value from keeping config information.

virtcontainers/container.go
WeiZhang555: Why not? The config.json is not the original OCI spec from other components(K8S or docker), it's another copy maintained by kata-runtime. I don't see a reason that we can't modify it.

virtcontainers/types/sandbox.go
WeiZhang555: @sboeuf I think you're talking about another issue. Current implementation keeps 3 copy of resources:
1. `Annotations[configJSONKey]`: keeps original unchanged OCI spec with resources.
2. `config.json`: one copy of OCI spec resources without change.
3. `state.json`: resources container really get.

My idea is that we can merge 2 and 3. because we already have a copy of unchanged resources, we don't need a duplicated copy.

If the PR is removing `spec.Resources` from `config.json`, it may be easier to understand. But I prefer container resources stay in config because it sounds more reasonable.

WeiZhang555: And actually we have fourth copy of `Resources`, it's from `bundle/config.json`

sboeuf: That's sounds fair to me. As long as we always have a way to know about the initial resources of the container and to know about the current resources, this is fine.
And `config` sounds like a good candidate since `state` is something that is supposed to hold values that have been generated by our runtime. If we consider the container resources fully dictated by the configuration, I'm happy with having those information sitting in the stored container config.

WeiZhang555: Thanks @sboeuf :)

WeiZhang555:
@bergwolf I agree. It makes more sense to seperate container state to sandbox state

gabibeyer:
+1 on not having unnecessary duplication of data 🙌🏻

PR #1418, opened 2019-03-25
mounts: Add check for system volumes

### LINKED ISSUE #1417: System mounts: Use 9p in case of docker volume and k8s host empty-dir.
System mounts are handled as a special case in Kata. If there exists a bind mount for a system directory, we skip passing the host directory to the guest via 9p. This is done as it is not useful to share the host system directories with the guest. We instead mounted the guest system directory in the container rootfs.
This behaviour is useful as long as the source directory for the system directory is not a docker volume or a k8s host empty-dir.

amshinde:
We handle system directories differently, if its a bind mount
    we mount the guest system directory to the container mount and
    skip the 9p share mount.
    However, we should not do this for docker volumes which are directories
    created by Docker.
    
    This introduces a Docker specific check, but that is the only
    information available to us at the OCI layer.
    Fixes #1417 
 
### DIFF
--- a/virtcontainers/mount.go
+++ b/virtcontainers/mount.go
@@ -342,3 +342,53 @@ func IsDockerVolume(path string) bool {
 	return false
 }
 
+const (
+	// K8sEmptyDir is the k8s specific path for `empty-dir` volumes
+	K8sEmptyDir = "kubernetes.io~empty-dir"
+)
+
+// IsEphemeralStorage returns true if the given path
+// to the storage belongs to kubernetes ephemeral storage
+//
+// This method depends on a specific path used by k8s
+// to detect if it's of type ephemeral. As of now,
+// this is a very k8s specific solution that works
+// but in future there should be a better way for this
+// method to determine if the path is for ephemeral
+// volume type
+func IsEphemeralStorage(path string) bool {
+	if !isEmptyDir(path) {
+		return false
+	}
+
+	if _, fsType, _ := GetDevicePathAndFsType(path); fsType == "tmpfs" {
+		return true
+	}
+
+	return false
+}
+
+// Isk8sHostEmptyDir returns true if the given path
+// to the storage belongs to kubernetes empty-dir of medium "default"
+// i.e volumes that are directories on the host.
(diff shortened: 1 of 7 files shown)


egernst:
@lifupan @bergwolf PTAL?

devimc:
fixes https://github.com/kata-containers/runtime/issues/1472 ?

virtcontainers/mount.go
```diff
@@ -326,3 +326,69 @@ func bindUnmountAllRootfs(ctx context.Context, sharedDir string, sandbox *Sandbo
 		}
 	}
 }
+
+const (
+	dockerVolumePrefix = "/var/lib/docker/volumes"
```
jcvenegas: It would be great to check where is defined this is docker code just to add as a comment reference in case it changes in the future.

virtcontainers/mount.go
```diff
@@ -326,3 +326,69 @@ func bindUnmountAllRootfs(ctx context.Context, sharedDir string, sandbox *Sandbo
...
+	if _, fsType, _ := GetDevicePathAndFsType(path); fsType != "tmpfs" {
+		return true
+	}
+	return false
+}
+
+func isEmptyDir(path string) bool {
+	splitSourceSlice := strings.Split(path, "/")
+	if len(splitSourceSlice) > 1 {
+		storageType := splitSourceSlice[len(splitSourceSlice)-2]
```
jcvenegas: magic number

virtcontainers/mount.go
jcvenegas: `var/lib/docker` is configurable at docker startup I may check only from volumes

virtcontainers/mount_test.go
```diff
@@ -282,3 +289,46 @@ func TestIsDeviceMapper(t *testing.T) {
 		t.Fatal()
 	}
 }
+
+func TestIsDockerVolume(t *testing.T) {
+	path := "/var/lib/docker/volumes/00da1347c7cf4f15db35f/_data"
```
jcvenegas: dockerVolumePrefix

PR #1372, opened 2019-03-15
lint: Change go linter from gometalinter to golangci-lint

ganeshmaharaj:
gometalinter is being deprecated and will be archived April '19. Their suggestion is to switch to using golangci-lint.

Signed-off-by: Ganesh Maharaj Mahalingam <ganesh.mahalingam@intel.com>

Issue #1338, opened 2019-03-07
Update golang versions


jodh-intel:
The [versions database](https://github.com/kata-containers/runtime/blob/master/versions.yaml) currently specifies:

```sh
$ yq read versions.yaml languages.golang.version
1.10.4
$ yq read versions.yaml languages.golang.meta.newest-version
1.11.1
```

FWICS...

- The snap build is using golang 1.10.4:

  https://github.com/kata-containers/packaging/blob/master/snap/snapcraft.yaml#L16

- The packaging is using golang 1.11.1:

  https://github.com/kata-containers/packaging/blob/master/obs-packaging/gen_versions_txt.sh#L49

- The CI system is using golang 1.11.1:

  https://github.com/kata-containers/tests/blob/master/.ci/install_go.sh#L79

However, from https://golang.org/dl:

- The latest stable golang is now: 1.12.
- The oldest stable golang version is now: 1.11.5

# TL;DR

We should consider updating golang versions to:

```yaml
languages.golang.version = 1.11.5
languages.golang.meta.newest-version = 1.12
```

Related to discussions about `go mod` (#1331).

/cc @chavafg.

Issue #1413, opened 2019-03-22
[RFC] Unify all kata components under a single repository


ganeshmaharaj:
# Description of problem
I will start off stating that I am writing this all down as someone who is new to the project and with my limited understanding. Please help me educate myself on things that i might be missing/mis-stating in this RFC.

The main components of kata are split across a few repositories. `runtime`, `shim`, `agent`, `proxy`. There are a couple of additional repositories that help test all the components together and a few more to aid with what is the final product. from talking to a few folks, the reasons for creating seperate repositories for each of those components were two-fold.

1. Allow folks to write an alternate `shim` or `proxy` which can be plug & play with all the other existing components.
2. Make sure that we are able to keep the components truly decoupled and use standard interfaces to communicate with each other without being heavily tied to each other.

I have come to understand that kata is now fairly stable in how the components are going to be working with each other and interfaces are also fairly well defined. With that as premise, __I was wondering how the community would feel about merging the code bases of the key kata components into a single repository?__

### Pros:
> * Less Paperwork (in the sense, one does not have to touch multiple repositories, open multiple issues for their fixes and modify yet another repository to add tests to their code and tell Jenkins to make sure it picks up all the changes while testing).
> * Better developer workflow where users can have a single repository, make changes, be able to run both functional and unit tests with some ease (It is not really that hard today and has been very well automated but this will reduce the additional checks one might do to make sure how well they test their code).
> * Release will be simplified on how many repositories need tagging, pinning, packaging, etc.
> * Folks can still write an alternate `agent` or `proxy` and have it replace the existing code.

### Cons:
> * We are opening the door where changes might break the strict decoupling of components without realizing.
> * Anything else you can think of?

I believe if we add add checks/processes that can verify we maintain the design ethos we have today for Kata even after moving all the sources into a single repository.

If we choose to move to a single repository, some things to consider.
1. What components/repositories will be merged into the new repository and what will still remain as a seperate repository?
   * I am thinking `proxy`, `shim`, `agent`, `runtime`, `tests`, `documentation` will all be merged into the new repository and all the others will remain as they are today.
2. how are we going to maintain each components history?
   * We can archive the current repositories and tell folks to refer to them in case they wish to view previous commits for each of those components.
3. Any changes to the CI systems to handle this?

sboeuf:
Thanks for raising this point @ganeshmaharaj.
[...]

Issue #42, opened 2018-03-01
Update cc-runtime to 3.0.20


jodh-intel:
Update `cc-runtime` to release 3.0.20.

PR #1420, opened 2019-03-26
Migrate to go modules (deprecate 'vendor')

### LINKED ISSUE #1331: Migration from 'go dep' to 'go mod'
Coming from [#1152](https://github.com/kata-containers/runtime/pull/1152#issuecomment-470052791):

> [@umarcor]
> @Pennyzct, @jodh-intel just because of curiosity, why are you using `dep` instead of `go mod`?

> [@grahamwhaley]
> @umarcor - at this point, it may be mostly due to historical reasons - the kata, and forerunner, code bases have been around a few years now... I think `go mod` came in with golang 1.11 ?, and we still have our baseline golang supported version as 1.10.4 :-)
> 
> Now, if you'd like to propose a move to `go mod`, and submit a PR ;-) But, first probably best if it was discussed. /cc @sboeuf

> [@jodh-intel]
Yes - at the time we introduced it, `dep` was the best tool available.

> [@sboeuf]
> @umarcor @grahamwhaley @jodh-intel I recently discussed this with @mcastelino  and I think it is time for us to move to go modules. We need to open  issues on every Kata repositories for this. And then let's start submit  PRs. @umarcor is it something you're interested in helping with?

Let's continue here, to avoid getting @Pennyzct's PR dirtier.

# Description of problem

This is neither a problem nor a requirement, because `dep` is not deprecated, yet. Certainly, it is still actively maintained (golang/dep#2052). Nonetheless, as `go mod` will eventually cease to be experimental, it is worth considering the implications.

# Expected result

- Allow developers to use either `dep` or `go mod`, at least until the baseline supported version in kata is above 1.11. Consider removing `dep` support afterwards.
- Remove `vendor` from the repository.

# Actual result

- Only `dep` is supported.
- The `vendor` repository is checked in the repo and it must be explicitly updated. See [#1152](https://github.com/kata-containers/runtime/pull/1152), [community#88](https://github.com/kata-containers/community/issues/88) and [community#89](https://github.com/kata-containers/community/issues/89).

---

So, regarding

> [@grahamwhaley]
> Now, if you'd like to propose a move to `go mod`, and submit a PR ;-) But, first probably best if it was discussed.

> [@sboeuf]
> @umarcor is it something you're interested in helping with?

I said I was asking because of curiosity since I don't really understand how `go mod` is expected to work. I see that it always works in clean environments (such as CI jobs), but when used in a machine with other golang packages, it freaks out easily. Precisely, the behaviour that I have seen is the following:

- If a dependency does not exist locally, the version specified in `go.mod` is retrieved.
- If the dependency exists locally, it may work or it may complain because the version does not match.
  - This is sometimes desirable, because the local dependency is in a valid dirty state, i.e. some patches have been applied.
  - In other contexts, the checked out or modified local state corresponds to any other project, and it is not compatible, so it breaks the build.

Moreover, `go mod` allows to generate a `vendor` directory, but it is not honoured. I.e., I would expect it to search for dependencies in common paths (`/go`, `~/go`, `GOPATH`) and save additional clones in the `vendor` only when there is some conflict. Then, when `go build` or `go run` is executed, it would use the versions in `vendor` and, if not present, it would search in the common paths.

What is happening instead is that you can copy absolutely all the dependencies to `vendor` and then tell go tools to use it exclusively. This is quite nonesense, because it effectively treats each repository as a completely independent project and the same dependencies are duplicated unnecessarily. At this point, I don't know if there is any benefit over using git submodules, provided that all the go dependencies are available as git repos.

Therefore, an 'easy' workaround is to just use a `golang:alpine` container and build the binaries with `CGO_ENABLED=0 go build -a`. This allows to execute each build in a clean environment, so there is no conflict with other projects in the host. The main disadvantage of this approach is that all the dependencies are downloaded each time a container is started, unless `vendor` is used exclusively.

Ideally, checking `vendor` into the repo should not be required neither with `dep` nor with `go mod`. The `vendor` directory should be a local temporal resource that can be 'generated' unequivocally with `go.mod`, `Gopkg.lock` and/or `Gopkg.toml`.

Last, the behaviour of `go mod` is different if the repository is inside or outside of the `GOTPATH`.

---

Anyway, the 'clean' approach to migrate from `dep` to `go mod` should be as 'simple' as:

``` bash
docker run --rm -it golang:alpine sh -c "$(cat <<-EOF
# Get 'regular' dependencies to build the runtime
apk update
apk add git
go get -u github.com/golang/dep/cmd/dep
go get -u github.com/kata-containers/runtime
cd \$GOPATH/src/github.com/kata-containers/runtime
dep ensure

# Remove 'dep' related sources
git rm -rf Gopkg.lock Gopkg.toml vendor/

# Move the repo outside of GOPATH
cd ..
mv runtime ~/
cd ~/runtime

# Initialize go modules
go mod init
git add go.mod

# Profit
git config --local user.name "dep2mod"
git config --local user.email "dep2mod@script.example"
git status
git commit -m 'migrate from dep to go mod'
git push
EOF
)"
```

> NOTE: optionally, replace `git push` with `sh` to execute further commands before pushing the result.

> NOTE: on MSYS2 (Win10), prepend the first line with `winpty`.

---

Summarizing, @grahamwhaley, @sboeuf, I am interested on the discusion, on hearing your thoughs, and on seen how a project like this handles this specific issue (provided that dependency management in golang is a flaw in the language/ecosystem). Nevertheless, I can sure open PRs applying the script above to the repos you deem appropriate.

umarcor:
This is an attempt to migrate from `dep` and a `vendor` directory to go modules, and hopefully close #1331. Ref #1338.

The first commit does remove `vendor` and `Gopkg.*`; `go.mod` and `go.sum` are added instead. `.travis.yml` is updated accordingly.

The build is successful, but static-checks fail because the testing infrastructure assumes that go packages installed through `go get` are located in `$GOPATH/src`: https://travis-ci.com/umarcor/kata-runtime/builds/105771341. That's not true anymore. With go modules, packages are located in `$GOPATH/pkg/mod`.

In the second commit, static-checks are disabled, just to ensure that other steps are successful: https://travis-ci.com/umarcor/kata-runtime/builds/105771550

The third commit is an attempt at fixing the issue with `kata-runtime/tests`. The repository is downloaded through `curl -fsSL https://codeload.github.com/kata-containers/tests/tar.gz/master | tar xzf - -C ../tests --strip-components=1` instead of `go get`. So, `export tests_repo_dir="$(pwd)/../tests"`. There is no problem with downloading and calling the scripts in it, but it fails because scripts in `kata-runtime/tests` do not check their own location: https://travis-ci.com/umarcor/kata-runtime/builds/105772078 (it is hardcoded https://github.com/kata-containers/tests/blob/master/.ci/lib.sh#L30-L31).

I could make a dirty hack by overwritting the file after downloading the tarball. But then I got another error related to the same issue: golangci-lint is installed through `go get` (which is quite strange I must say). Therefore, scripts assume that it is available in `GOPATH/src...`. That's not true now. I removed it from https://github.com/kata-containers/tests/blob/master/versions.yaml#L52-L62 and installed it in the `before_install` field of `.travis.yml`. At some point it worked, but linting errors were produced: https://travis-ci.com/umarcor/kata-runtime/jobs/187675554. I tried to rebase on top of #1372. Rebasing produced no conflicts, but linting issues were not solved. Furthermore, in #1372 `golangci-lint` is defined in a different `versions.yaml` the specified version does not match the one `kata-containers/tests`: https://github.com/kata-containers/runtime/blob/f4428761cbd0418c8024f0356897f10417ab38bd/versions.yaml.

As a result, I think that it would be desirable that someone with better knowledge about `kata-containers/test` could think about the issue. How much effort is required to avoid using hardcoded locations in the tests? Why are packages such as golangci-lint installed through `go get` instead of `go install`?

Also, related to `.travis.yml`, I wonder why `install` and `before_script` are used instead of `script` and `after_success`.


raravena80:
@umarcor nudge. any updates?

umarcor:
Hi @raravena80! Thanks for coming back to this. As commented in the first message:

> As a result, I think that it would be desirable that someone with better knowledge about `kata-containers/test` could think about the issue. How much effort is required to avoid using hardcoded locations in the tests? Why are packages such as `golangci-lint` installed through `go get` instead of `go install`?

Overall, there is no issue at all with replacing `dep` with 'go modules', as far as dependency management is concerned. However, the testing infrastructure in the project as a whole relies on some hardcoded paths which are broken with this change. I believe that this is an indicator of the fragility of the testing environment. Thus, those hardcoded paths should be fixed. However, since that is out of the scope of this PR, I'm waiting for some thoughts, comments or help about it.

umarcor:
I rebased on top of master. This is the result with static-checks disabled: https://travis-ci.com/umarcor/kata-runtime/builds/108843843

jodh-intel:
@umarcor - since there seem to be various pieces we need to change to switch to `go mod`, I've created a tracking issue:

- https://github.com/kata-containers/kata-containers/issues/42

I've tried to distill your fairly detailed explanations, but please update the issue above if I've missed anything. If you would like to help out with some of these preparatory tasks, please do! :smile:

raravena80:
@umarcor any updates? Thx.

umarcor:
Hi @raravena80. Not really. Some clarificative comments where posted in kata-containers/kata-containers#42, but there is no further progress AFAIK.

grahamwhaley:
@jodh-intel - did you say you were contemplating looking at this as well, or did I dream that? :-)

jodh-intel:
That would have been a dream :smile:

I like the idea but no spare cycles to do this atm. Any volunteers?

caoruidong:
I think we can do this when golang 1.13 is out

raravena80:
Waiting for golang 1.13...

grahamwhaley:
@ganeshmaharaj @amshinde - were you working on this on new PRs now? If so, can this one be closed do we think?

caoruidong:
ping @umarcor Can we close it ?

umarcor:
@caoruidong, #1911 is a kind of duplicate of this. Hence, you can either close that or this. However, note that there is a relevant difference. Here, I proposed to remove `vendor`, since it should not be required with go modules. However, in #1911 the vendor directory is updated, not removed.

I am not familiar with using repositories outside of GOPATH, with go modules enabled and with a vendor subdir. How does this work when there is a version of a package in the system, another one in the cache of the go modules and a last one in the vendor dir? Is the expected approach for each kata-containers  repo to vendor everything (probably duplicating lots of common dependencies), instead of potentially having advantage of a built-in mechanism?

caoruidong:
Leaving vendor in repo will benefits in downloading process. User can directly run `make`. I think both ways are OK. The community now are thinking about merging several repos into two or three.

umarcor:
> 
> 
> Leaving vendor in repo will benefits in downloading process. User can directly run `make`. I think both ways are OK. The community now are thinking about merging several repos into two or three.

Thant's equivalent to `go mod download` (which can be added to the `make` target), isn't it?

caoruidong:
Yes, you're right. Maybe it can avoid some network issue. Different version of packages should follow a priority rule during building. Some k8s related projects hold vendor directory too. So it may not be a big problem. We can discuss further in https://github.com/kata-containers/runtime/pull/1911.

umarcor:
@caoruidong, after having a better look at #1911, I can provide some further insight:

The title of this PR is `Replace 'dep' (vendor) with go modules`, but what I actually propose is to migrate to go modules (in a proper idiomatic way). The title of #1911 is `wip: Move to Go modules`, but @amshinde is only replacing the mechanism to update `vendor`. Therefore, we should probably swtich the titles.

Since @amshinde adds `export GO111MODULE=on` to `make vendor`, and because `make vendor` is NOT executed in `.travis.yml`, none of the CI tests is using go modules, in practice. I am afraid that it will fail in any environment where `GO111MODULE=on`, from the beginning. Please, see [this](https://travis-ci.com/umarcor/kata-runtime/jobs/231319730) job, which is a [single commit](https://github.com/umarcor/kata-runtime/compare/9f74e1fa52df%5E...1b211ca56d1e) ahead of #1911. By the same token, here I set `GO111MODULE` globally, so every task in the CI job uses go modules. 

As a result, all (if not all) the issues with the unidiomatic way of handling the tests_repo, `setup.sh`, `lib.sh`, etc. still apply; no matter if #1911 is merged only, this one, or both of them. Please, read https://github.com/kata-containers/kata-containers/issues/42#issuecomment-488680416 and https://github.com/kata-containers/runtime/pull/1420#issue-264351427 carefully.

Summarizing, it is ok if you want to merge #1911 first, as it will work for now. However, I will keep this PR open, I will update it and I will update the title. I believe this is the way to go in the mid term. No because of vendoring, but because the current repo structure of kata-containers as a project is broken from a 'go modules workspace' perspective.

caoruidong:
This is just a weekly ping, not a force merge or close. I think repo structure may change some day. Your PR may be the final resolution.  You can keep this PR open anyway.

umarcor:
> I think repo structure may change some day.

I think this is the relevant issue here. I don't mind if it is this PR or any other. But I believe that kata-containers/kata-containers#42 deserves some love from the community of maintainers and contributors of kata-containers.

caoruidong:
FYI, it's https://github.com/kata-containers/runtime/issues/1413

umarcor:
> FYI, it's #1413

I'm sorry, but I don't understand the relation. #1413 is about unifying all the repositories in a monorepo. I [commented](https://github.com/kata-containers/runtime/issues/1413#issuecomment-476341755) about there.

What I mean with 'the structure of the project' here is not to unify, but to think about it. For example: why are tests and other non-build dependencies installed through `go get`, which is precisely meant to handle build dependencies? It might be ok to use `go install`, should the tests repo be a go tool. Otherwise, `curl`, `git`, `wget`, etc. should be used. These are standard tools to retrieve assets; and tests are assets. See https://github.com/kata-containers/runtime/pull/1420/commits/761534e1673397f436aea72971e5950f6e730308 for a taste of what I mean.

caoruidong:
Got it. It is another choice. Personally I'm OK with it. Let's hear from the community.

raravena80:
@umarcor any updates on this PR? Thx!

umarcor:
Hi @raravena80! Unfortunately... no. This is issue likely to be stalled until anyone tries to reorganise the repo/testing layout: https://github.com/kata-containers/tests. Since that's a (not pleasant) chore, and not a new feature or an improvement, I assume that nothing will be done until things break really badly and the whole project is blocked.

raravena80:
@umarcor any updates?

Your weekly Kata herder 😄

umarcor:
@raravena80, see https://github.com/kata-containers/runtime/pull/1911#issuecomment-566237979

raravena80:
Btw, one of the downsides of go modules is that in some cases the build might take longer due to the dependencies being pulled across the network.

umarcor:
Agree. A mechanism to explicitly define a local cache folder (a la `vendor` but without checking it into git) would be really useful.

grahamwhaley:
as this is effectively replaced by #1911 , can we close this one @umarcor @raravena80

umarcor:
@grahamwhaley, as discussed in the comments above, and as explained in https://github.com/kata-containers/runtime/pull/1911#issuecomment-528181290 , #1911 is likely to be a (probably broken) subset of this PR. I'm ok with closing this PR, as it didn't get enough attention in almost a year. Moreover, I'm ok with @amshinde taking over this PR. Nonetheless, #1911 does NOT solve the problem that has prevented this PR from going forward.

grahamwhaley:
Ah, thanks for the clarification @umarcor - there were a lot of comments to read through, and I'm not a `go mod` expert ;-).
Let's leave this here for a little while longer then - now us gatekeepers should see this note and not nudge you again for a while!

raravena80:
@umarcor this is old, should we close this and move the work to 2.0?

umarcor:
@raravena80, as explained, the blocking factor for this PR is unrelated to the runtime. Hence, I think that closing this PR and opening a new one in [kata-containers/kata-containers](https://github.com/kata-containers/kata-containers) is just a cosmetic change. The problem is with [https://github.com/kata-containers/tests](kata-containers/tests), which did not change as far as I can tell:

Precisely:

> https://github.com/kata-containers/tests#prepare-an-environment
>
> Clone the kata-container/tests repository:
> 
> $ go get -d github.com/kata-containers/tests

`go get` is NOT a command to "clone a repository". `go get` is for ADDING DEPENDENCIES and/or INSTALLING TOOLS. Assets should (need to) be installed using  git, wget, curl... or ANY OTHER mechanism.

IMHO, unless this organization stops using `go get` to retrieve assets, it is NOT possible to use go modules.

bergwolf:
We have already moved to `go mod` in kata 2.0, see https://github.com/kata-containers/kata-containers/blob/2.0-dev/src/runtime/go.mod.

IMO we can close this unless it is strongly desired to use `go mod` in 1.x as well.

umarcor:
@bergwolf is kata-containers/tests used in kata-containers/kata-containers? Or are all tests in the new repo written in golang?

umarcor:
@bergwolf, @raravena80, @grahamwhaley I did a quick test to prove that the problem still exists in v2.0. It's exactly the same problem that was explained in this thread, in #1331, in #42 and in #1911. Very precisely, go modules are NOT enabled in Travis by default, so neither #1911 nor 2.0-dev in kata-containers/kata-containers are using go modules. Those branches/repos have a `go.mod` file which is ignored. You can easily check that in any of the CI runs of the repo: https://travis-ci.org/github/kata-containers/kata-containers/jobs/709153391#L190

I just added `GO111MODULE=on` (see https://github.com/umarcor/kata-containers/commit/a659b5423edd212ec7667101005f2f8dad894540) and everything broke (https://travis-ci.com/github/umarcor/kata-containers/jobs/363292420#L252) because of kata-containers/tests. This is because, again, `go get` is NOT a procedure to retrieve assets, and hardcoding the default checkout path of pre-gomodules everywhere is a BROKEN procedure.

@bergwolf, see https://github.com/kata-containers/runtime/pull/1911#issuecomment-597547275

amshinde:
This has been fixed in our 2.0 branch instead. Will not fix in 1.x for now.

umarcor:
@amshinde, as I said 19 months ago, and I've been repeating each time some maintainer tried to closed this as "fixed", this will NOT be fixed until ALL improper usages of `go get` are removed. At this point, I'm starting to feel the maintenance to be quite disrespectful towards me trying to just explain a problem that exists because of an inadequate design (structural) decision. So, once again:

**Enabling `GO111MODULE` breaks 2.0. Travis CI is passing just because go modules are NOT used in practice.**

See https://github.com/umarcor/kata-containers/commit/1cc0749b03dbb7da6dcdb93725228acd8b00262c and https://travis-ci.com/github/umarcor/kata-containers/jobs/392554689.

Maintainers, please read https://github.com/kata-containers/runtime/pull/1420#issuecomment-661846793 CAREFULLY. I'd be so glad to reword any explanation, should the problem be not clear enough.

amshinde:
@umarcor I understand that this is still an outstanding issue and needs to be resolved with not only moving to go modules but also addressing all erroneous invocations of `go get`, replacing them with a wget or curl instead. 
I simply closed this as I am trying to close any stale PRs that are not actively being worked on. With 2.0, we are consolidating all repos into `kata-containers/kata-containers` repo including runtime, tests, packaging among others. While we do want to backport important bugs and features, I think resolving all the issues you mentioned including changing the package management tooling would be worth doing in our 2.0 repo. 
I closed this PR since I was under the impression that you are not actively working on this PR and considering we have an issue open in 2.0 repo to track remaining changes that need to be done.

umarcor:
@amshinde, thanks for clarifying. I think it is ok to have this PR closed. The title is outdated and the content of branch `umarcor:go-mod` is not going to be merged. However, this was also a placeholder for acknowledging that a problem with the testing infrastructure exists. Closing it without explicitly keeping track of that requirement is what I considered unfair.

Should the plan be to fix the tests repo after everything is merged in a monolithic approach, that's absolutely acceptable. Please, just be careful for it not to be diluted.

Anyway, I opened kata-containers/kata-containers#823 with an as short and specific description as I could write.

Issue #1422, opened 2019-03-26
Test case `TestEnterContainer` failed


WeiZhang555:
Test case failed but we didn't find it in Jenkins jobs, see related issue: https://github.com/kata-containers/tests/issues/1355

PR #1423, opened 2019-03-26
tests: do cleanUp() always in the end

WeiZhang555:
Fixes: #1422

Detect failing test case:

```
....
=== RUN   TestEnterContainerFailingContNotStarted
--- PASS: TestEnterContainerFailingContNotStarted (0.01s)
=== RUN   TestEnterContainer
--- FAIL: TestEnterContainer (0.00s)
 Error Trace: sandbox_test.go:1154
 Error:      	Expected value not to be nil.
 Messages:   	Entering non-running container should fail
 Error Trace: sandbox_test.go:1157
 Error:      	Expected nil, but got: &errors.errorString{s:"Can not
move from running to running"}
 Messages:   	Failed to start sandbox: Can not move from running to
running
FAIL
```

`TestEnterContainerFailingContNotStarted` calls `cleanUp` at function
begging but it doesn't clean its garbage after it ends.
`TestEnterContainer` only call `cleanUp` in the end but it doesn't do
cleanUp in the begging, that gives first test case a chance to impact
latter one.

This commit modifies all the test cases, let them all do the cleanUp()
in the end.

The policy here is: "everyone needs to take their garbage away when they
leave" :)

### DIFF
--- a/virtcontainers/api_test.go
+++ b/virtcontainers/api_test.go
@@ -199,7 +199,7 @@ func newTestSandboxConfigKataAgent() SandboxConfig {
 }
 
 func TestCreateSandboxNoopAgentSuccessful(t *testing.T) {
-	cleanUp()
+	defer cleanUp()
 
 	config := newTestSandboxConfigNoop()
 
@@ -232,7 +232,7 @@ func TestCreateSandboxHyperstartAgentSuccessful(t *testing.T) {
 		t.Skip(testDisabledAsNonRoot)
 	}
 
-	cleanUp()
+	defer cleanUp()
 
 	config := newTestSandboxConfigHyperstartAgent()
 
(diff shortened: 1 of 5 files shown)


WeiZhang555:
The policy here is: "everyone needs to take their garbage away when they leave" :)

So if we treat the unit test as a "clean room", some test case enter the room, then do something, it should take its garbage away(`defer cleanUp()`) from the room when it leaves. 
If the room isn't clean when one test case enters, we should blame the last one, and the new test case can do one extra `cleanUp()` in the begging, I'm not against this :-) .
That is something like:
```
func Testxxxx(){
cleanUp()
defer cleanUp()
....
}
```

WeiZhang555:
Another issue is Jenkins doesn't run the failing test case at all, see https://github.com/kata-containers/tests/issues/1355

Issue #1424, opened 2019-03-26
kata-env does not validate proxy path


bergwolf:
When proxy path is configured but the destination file does not exists, `kata-runtime kata-env` still thinks it is there and prints proxy version as `unknown`.

OTOH, shim path is validated and an error will be shown if the destination file does not exist.

Let's unify the behavior.

Reported-by @laijs

PR #1425, opened 2019-03-26
config: validate proxy path

bergwolf:
Like shim, we should validate the proxy path if it is provided.

Fixes: #1424


WeiZhang555:
I believe failing test cases are unrelated.

WeiZhang555:
LGTM

Issue #1426, opened 2019-03-26
qemu leak if container failed to start 


Ace-Tang:
# Description of problem

https://github.com/kata-containers/runtime/blob/c6804c8e2bc9e2d290970ff39d982ffffe291146/virtcontainers/api.go#L100-L109

kata do cleanup after startVm(), but in startVM, function may return error but vm already start  , kata-agent failed to start server and return error, 
```
failed to create task for container(3a13403c6e5b13c6e2db5274b5e0b9882f37d82f5b61b1a8da84e08e7236f962): OCI runtime create failed: create error: unknown"
```

The error found in test I mention in https://github.com/kata-containers/runtime/issues/1368, maybe hundreds qemu run on the host make host load high, and kata-agent can not work.

The qemu version is `QEMU emulator version 2.11.0`, I think it should be a bug if use other qemu version

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

egernst:
@Ace-Tang do you have a simple example for reproducing this bug?

Ace-Tang:
@egernst ， run many container parallel may reproduce, but it also simple to reproduce if you created an error in `startVM()`, the qemu will leak

grahamwhaley:
@egernst - afaik, the only parallel launch test code we have is in the [fast footprint density test](https://github.com/kata-containers/tests/blob/master/metrics/density/fast_footprint.sh#L484) - maybe you can use that as a test case?

PR #1427, opened 2019-03-26
qemu: fix qemu leak when failed to start container

Ace-Tang:
add vm cleanup before startVM()

Fixes: #1426

### DIFF
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -108,7 +108,7 @@ func createSandboxFromConfig(ctx context.Context, sandboxConfig SandboxConfig, f
 		}
 	}()
 
-	if err := s.getAndStoreGuestDetails(); err != nil {
+	if err = s.getAndStoreGuestDetails(); err != nil {
 		return nil, err
 	}
 
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -908,7 +908,7 @@ func (s *Sandbox) ListRoutes() ([]*vcTypes.Route, error) {
 }
 
 // startVM starts the VM.
-func (s *Sandbox) startVM() error {
+func (s *Sandbox) startVM() (err error) {
 	span, ctx := s.trace("startVM")
 	defer span.Finish()
 
@@ -938,6 +938,12 @@ func (s *Sandbox) startVM() error {
 	}); err != nil {
 		return err
 	}
+
+	defer func() {
+		if err != nil {
+			s.hypervisor.stopSandbox()
+		}
+	}()
 
 	// In case of vm factory, network interfaces are hotplugged
 	// after vm is started.


devimc:
thanks @Ace-Tang for raising this

Ace-Tang:
I add one more fix 
```
diff --git a/virtcontainers/api.go b/virtcontainers/api.go
index bb3841d..f7859e6 100644
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -108,7 +108,7 @@ func createSandboxFromConfig(ctx context.Context, sandboxConfig SandboxConfig, f
                return nil, err
        }
 
-       if err := s.getAndStoreGuestDetails(); err != nil {
+       if err = s.getAndStoreGuestDetails(); err != nil {
                return nil, err
        }
```
avoid defer not catch err

Ace-Tang:
make sense, any idea to check vm is running? still do the cleanup first, but check if vm is started?

sboeuf:
@Ace-Tang IMO the way to fix it properly will be to do the proper rollback from the function `startVM()`. You need to augment the rolling back that is done from `startVM()` itself, without calling `stopVM()`, but only the functions which need to be deferred, depending where it failed in `startVM()`.

Ace-Tang:
@sboeuf :+1: , stopVM() is do more than shutdown the vm, some don't need in this bug, the problem can be divided to,
1. vm not start -> nothing need to do 
2. vm start, agent not start,  -> just stop vm
3. vm start, agent start, and fail after startVM function, this already cleanup in code
Then the second should be fix , how about call `stopSandbox()` inside `startVM` do the cleanup

virtcontainers/api.go
```diff
@@ -96,19 +96,19 @@ func createSandboxFromConfig(ctx context.Context, sandboxConfig SandboxConfig, f
...
-
 	// rollback to stop VM if error occurs
 	defer func() {
 		if err != nil {
 			s.stopVM()
 		}
 	}()
 
-	if err := s.getAndStoreGuestDetails(); err != nil {
+	// Start the VM
```
lifupan: I don't think it's a good idea to this move. Since the s.startVM() will do two things: launch the hypervisor process and start the sandbox in hypervisor, on the contrary, the s.stopVM() will do the stop sandbox in the hypervisor and then stop the hypervisor process. 

Suppose  it failed on launching the hypervisor process, then the defer function s.stopVM() will be redundant since there is no need to do the cleanup. On the other hand, if it failed on starting the sandbox in hypervisor, then the cleanup function of stop the sandbox in hypervisor will be failed since the sandbox failed to be started, and the s.stopVM() will return directly without trying to stop the hypervisor process. This isn't what we want.

Ace-Tang: yes, it is what I said in the comment
```
Then the second should be fix , how about call stopSandbox() inside startVM do the cleanup
```

virtcontainers/sandbox.go
```diff
@@ -908,12 +908,18 @@ func (s *Sandbox) ListRoutes() ([]*vcTypes.Route, error) {
...
 // startVM starts the VM.
-func (s *Sandbox) startVM() error {
+func (s *Sandbox) startVM() (err error) {
 	span, ctx := s.trace("startVM")
 	defer span.Finish()
 
 	s.Logger().Info("Starting VM")
 
+	defer func() {
+		if err != nil && s.hypervisor.pid() > 0 {
```
lifupan: I think it's better to register this cleanup function after startVM successfully, thus you shouldn't have to check the hypervisor's pid.

Ace-Tang: I am not sure if it will happen that `s.hypervisor.startSandbox` return error but vm start, since it complicated when related to qemu, so I choose to locate cleanup before start

Ace-Tang:
@sboeuf  would you help to check this fix ? it cause some qemu leak when run lots of parallel test

virtcontainers/sandbox.go
```diff
@@ -939,6 +939,12 @@ func (s *Sandbox) startVM() error {
 		return err
 	}
 
+	defer func() {
+		if err != nil {
+			s.hypervisor.stopSandbox()
+		}
+	}()
```
sboeuf: Is this new `defer` function fixing the issue you're seeing? The patch looks good here, and if it's fixing the problem, then let's merge it.

Ace-Tang: > run many container parallel may reproduce, but it also simple to reproduce if you created an error in startVM(), the qemu will leak

@sboeuf ， like I said in the https://github.com/kata-containers/runtime/issues/1426, if you donnot mind, you can create an error in the startVM function, and should see the fail

sboeuf: You didn't answer my question here. Is the current version of this PR fixing the issue that you are seeing when running many containers in parallel?
If it is, then please let me know and I'll merge the patch.
If it's not, then extend this patch until you get it fixed.

Ace-Tang: Yes, it does fix the problem here, and I provide a way to reproduce the problem.

sboeuf: Great, then let's merge it.

wenlxie:
I met the same issue, and find the same root cause.  I am using stable-1.6.
Qemu process leaked a lot in some scenarios. Thanks for the fix.

PR #1436, opened 2019-03-28
hypervisor: return cpu->threadID mapping

bergwolf:
We need such information to be able to apply container-level cpu cgroup constraints on the host side. This is in preparation for k8s static CPU manager support as being tracked in #878.


virtcontainers/cgroups.go
```diff
@@ -227,7 +227,7 @@ func (s *Sandbox) constrainHypervisor(cgroup cgroups.Cgroup) error {
 	if err != nil {
 		return fmt.Errorf("failed to get thread ids from hypervisor: %v", err)
 	}
-	if tids == nil || len(tids.vcpus) == 0 {
```
sboeuf: You should change the comment below.

bergwolf: hmmm, I think `len(tids.vcpus) == 0` is a failure case. Did you add it just for the FC case?

bergwolf: fixed.

virtcontainers/fc.go
```diff
@@ -687,16 +683,53 @@ func (fc *firecracker) resizeVCPUs(reqVCPUs uint32) (currentVCPUs uint32, newVCP
...
+			fields := strings.Fields(line)
+			if len(fields) != 5 {
+				return vcpuInfo, errors.Errorf("Invalid fc thread info: %v", line)
+			}
+			tid, err := strconv.ParseInt(fields[1], 10, 32)
+			if err != nil {
+				return vcpuInfo, errors.Wrapf(err, "Invalid fc thread info: %v", line)
+			}
+			cpus := strings.SplitAfter(line, "fc_vcpu")
+			cpuID, err := strconv.ParseInt(cpus[1], 10, 32)
```
jcvenegas: required check size before index?

virtcontainers/fc.go
```diff
@@ -687,16 +683,53 @@ func (fc *firecracker) resizeVCPUs(reqVCPUs uint32) (currentVCPUs uint32, newVCP
...
+	if err != nil {
+		return vcpuInfo, errors.Wrapf(err, "fail to get fc thread info: %s", output)
+	}
+
+	// Sample out:
+	//   PID  SPID TTY          TIME CMD
+	// 14328 14328 ?        00:00:00 firecracker
+	// 14328 14330 ?        00:00:00 fc_vmm
+	// 14328 14337 ?        00:00:05 fc_vcpu0
+	lines := strings.Split(string(output), "\n")
```
devimc: I wonder if won't be easier to read the content of `/proc/[pid]/task/`  :thinking:

bergwolf: hmm, I've found `github.com/shirou/gopsutil/process` and will make use of it. Just so that we don't need to fork with `exec.Command`.

virtcontainers/fc.go
bergwolf: Good point!

virtcontainers/fc.go
bergwolf: Turns out `github.com/shirou/gopsutil/process` is also calling `exec.command` so I don't think we should use it.

PR #1437, opened 2019-03-28
Make factory can control VMCache server

teawater:
* factory destroy stop VMCache server
* factory status show status of VMCache server


devimc:
travis is not happy, please rebase your pr

Issue #1438, opened 2019-03-28
Pass correct mount type to agent for ephemeral volumes


YongjiXie:
Now we set volume's mount type to "ephemeral" to indicate ephemeral volumes. And this type will be passed to agent later. It's incorrect. Instead, "bind" should be the correct mount type to be passed.

mcastelino:
@YongjiXie @amshinde a question. We do want ephemeral `memory` volumes to be created within the VM. Not on the host. Will this continue to be the case after this change.

YongjiXie:
> 
> 
> @YongjiXie @amshinde a question. We do want ephemeral `memory` volumes to be created within the VM. Not on the host. Will this continue to be the case after this change.

Sure. We have passed storages like: storages:<driver:\"ephemeral\" source:\"tmpfs\" fstype:\"tmpfs\" mount_point:\"/run/kata-containers/sandbox/ephemeral/xxx\" > to agent. This will make agent mount tmpfs to the mount point. Then IIUC, we need a bind mount between container path and this mountpoint.

Thanks,
Yongji

amshinde:
@YongjiXie What you described is how it is currently working in Kata.

PR #1439, opened 2019-03-28
agent: pass correct mount type to agent for ephemeral volumes

YongjiXie:
The "ephemeral" is just used to indicate ephemeral volumes in
runtime. We should not pass it to agent. Instead, "bind" should be
the correct mount type to be passed.

Fixes: #1438

Signed-off-by: Xie Yongji <xieyongji@baidu.com>

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -1119,6 +1119,8 @@ func (k *kataAgent) handleEphemeralStorage(mounts []specs.Mount) []*grpc.Storage
 		if mnt.Type == kataEphemeralDevType {
 			// Set the mount source path to a path that resides inside the VM
 			mounts[idx].Source = filepath.Join(ephemeralPath, filepath.Base(mnt.Source))
+			// Set the mount type to "bind"
+			mounts[idx].Type = "bind"
 
 			// Create a storage struct so that kata agent is able to create
 			// tmpfs backed volume inside the VM


virtcontainers/kata_agent.go
```diff
@@ -1119,6 +1119,8 @@ func (k *kataAgent) handleEphemeralStorage(mounts []specs.Mount) []*grpc.Storage
 		if mnt.Type == kataEphemeralDevType {
 			// Set the mount source path to a path that resides inside the VM
 			mounts[idx].Source = filepath.Join(ephemeralPath, filepath.Base(mnt.Source))
+			// Set the mount type to "bind"
+			mounts[idx].Type = "bind"
```
bergwolf: I guess the original code works only because k8s would set the mount type to `bind` as well to tell runc to bindmount the host tmpfs mountpoint to container's ephemeral volume. But it does feel safer to explicitly set it. Thanks!

amshinde: @bergwolf yes, that is the case today, which is why it works. k8s already sets this to "bind".
@YongjiXie This change is not absolutely required, I guess its good to have it set explicitly for clarity.

YongjiXie: @bergwolf @amshinde Actually what I did in this patch is changing the mount type from “ephemeral” to "bind". Without this patch, we will pass something like: Mounts:<destination:\"/mnt\" source:\"/run/kata-containers/sandbox/ephemeral/cache\" type:\"ephemeral\" options:\"rbind\" options:\"rprivate\" options:\"rw\" > to kata agent.

And I guess the original code works only because we have "rbind" option. And libcontainer will still do mounting although it doesn't know the mount type in kata agent.

Thanks,
Yongji

bergwolf: @YongjiXie I see. The problem is that instead of exposing a proper sandbox level storage structure, we modified oci spec to indicate an `ephemeral` mount. Thanks for the explanation!

jcvenegas:
@GabyCT  does the `suse job` errors are fixed now ?
@amshinde  I'd like your ack here to merge this PR

GabyCT:
@jcvenegas no, the opensuse issue is still there https://github.com/kata-containers/ci/issues/136

amshinde:
@jcvenegas @GabyCT Are the opensuse and nemy CIs supposed to pass?
If not, please merge.

GabyCT:
@amshinde opensuse is not but nemu yes

amshinde:
Ok. I am getting a 404 on the nemu one.Restarting all
/test

PR #1467, opened 2019-04-03
Stable 1.5 backports

egernst:
relevant backports since 1.6:
096fa04 qemu: fix qemu leak when failed to start container
fb64a3e doc: update architecture.md link
config: validate proxy path
virtcontainers: makefile fix .ci path

PR #1477, opened 2019-04-03
Stable 1.5 linter fixes

jcvenegas:
linter fixes for 1.5


jcvenegas:
@egernst  ops fails for missing header ,seems that is better put on top https://github.com/kata-containers/runtime/pull/1467

egernst:
We shouldn't add lint fixes to stable.  Should only be bug fixes/security fixes.  Let's close this one so we make sure we don't eat more CI cycles :)

PR #1485, opened 2019-04-05
storage: create k8s emptyDir inside VM

### LINKED ISSUE #1472: Kubernetes EmptyDir should be not be using 9p
Kubernetes EmptyDir performances are very slow (9p), while there is no real need to use to use 9pfs for the default medium. The EmptyDir volumes are only intended to share data between containers within a pod, and not with the host.

This was a recent related change in https://github.com/kata-containers/runtime/issues/1341 where tmpfs is not handled correctly. 

For the default medium type (disk), `kata-agent` should probably create a directory in the VM, and mount it to the containers ?

cc @mcastelino @amshinde ?

awprice:
**Some notes about this PR:**
 - I've based it off https://github.com/kata-containers/runtime/pull/1418, as it had logic around host k8s emptyDir completed for me.
 - There is a corresponding PR in the kata-agent repo which actually creates the directories: https://github.com/kata-containers/agent/pull/521

---

This introduces a new storage type: local. Local storage type will
tell the kata-agent to create an empty directory in the sandbox
directory within the VM.

K8s host emptyDirs will then use the local storage type and mount it
inside each container. By doing this, we utilise the storage medium
that the sandbox uses. In most cases this will be 9p.

If the VM is using device mapper for container storage, the containers
will benefit from the better performance of device mapper for
host emptyDir.

Fixes #1472

Signed-off-by: Alex Price <aprice@atlassian.com>

### DIFF
--- a/pkg/katautils/create.go
+++ b/pkg/katautils/create.go
@@ -162,7 +162,10 @@ func HandleFactory(ctx context.Context, vci vc.VC, runtimeConfig *oci.RuntimeCon
 func SetEphemeralStorageType(ociSpec oci.CompatOCISpec) oci.CompatOCISpec {
 	for idx, mnt := range ociSpec.Mounts {
 		if vc.IsEphemeralStorage(mnt.Source) {
-			ociSpec.Mounts[idx].Type = "ephemeral"
+			ociSpec.Mounts[idx].Type = vc.KataEphemeralDevType
+		}
+		if vc.Isk8sHostEmptyDir(mnt.Source) {
+			ociSpec.Mounts[idx].Type = vc.KataLocalDevType
 		}
 	}
 	return ociSpec
--- a/pkg/katautils/create.go
+++ b/pkg/katautils/create.go
@@ -161,7 +161,7 @@ func HandleFactory(ctx context.Context, vci vc.VC, runtimeConfig *oci.RuntimeCon
 // of the same pod the already existing volume is reused.
 func SetEphemeralStorageType(ociSpec oci.CompatOCISpec) oci.CompatOCISpec {
 	for idx, mnt := range ociSpec.Mounts {
-		if IsEphemeralStorage(mnt.Source) {
+		if vc.IsEphemeralStorage(mnt.Source) {
 			ociSpec.Mounts[idx].Type = "ephemeral"
 		}
 	}
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -43,6 +43,15 @@ import (
 	grpcStatus "google.golang.org/grpc/status"
 )
 
+const (
+	// KataEphemeralDevType creates a tmpfs backed volume for sharing files between containers.
+	KataEphemeralDevType = "ephemeral"
+
+	// KataLocalDevType creates a local directory inside the VM for sharing files between
+	// containers.
+	KataLocalDevType = "local"
+)
+
 var (
 	checkRequestTimeout   = 30 * time.Second
 	defaultKataSocketName = "kata.sock"
@@ -59,17 +68,17 @@ var (
 	vsockSocketScheme     = "vsock"
 	// port numbers below 1024 are called privileged ports. Only a process with
 	// CAP_NET_BIND_SERVICE capability may bind to these port numbers.
-	vSockPort            = 1024
-	kata9pDevType        = "9p"
-	kataMmioBlkDevType   = "mmioblk"
-	kataBlkDevType       = "blk"
-	kataSCSIDevType      = "scsi"
-	kataNvdimmDevType    = "nvdimm"
-	sharedDir9pOptions   = []string{"trans=virtio,version=9p2000.L,cache=mmap", "nodev"}
-	shmDir               = "shm"
-	kataEphemeralDevType = "ephemeral"
-	ephemeralPath        = filepath.Join(kataGuestSandboxDir, kataEphemeralDevType)
-	grpcMaxDataSize      = int64(1024 * 1024)
+	vSockPort          = 1024
+	kata9pDevType      = "9p"
+	kataMmioBlkDevType = "mmioblk"
+	kataBlkDevType     = "blk"
+	kataSCSIDevType    = "scsi"
+	kataNvdimmDevType  = "nvdimm"
+	sharedDir9pOptions = []string{"trans=virtio,version=9p2000.L,cache=mmap", "nodev"}
+	shmDir             = "shm"
+	ephemeralPath      = filepath.Join(kataGuestSandboxDir, KataEphemeralDevType)
+	grpcMaxDataSize    = int64(1024 * 1024)
+	localDirOptions    = []string{"mode=0777"}
 )
 
 // KataAgentConfig is a structure storing information needed
@@ -672,7 +681,7 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
 		shmSizeOption := fmt.Sprintf("size=%d", sandbox.shmSize)
 
 		shmStorage := &grpc.Storage{
-			Driver:     kataEphemeralDevType,
+			Driver:     KataEphemeralDevType,
 			MountPoint: path,
 			Source:     "shm",
 			Fstype:     "tmpfs",
@@ -1038,6 +1047,9 @@ func (k *kataAgent) createContainer(sandbox *Sandbox, c *Container) (p *Process,
 	epheStorages := k.handleEphemeralStorage(ociSpec.Mounts)
 	ctrStorages = append(ctrStorages, epheStorages...)
 
+	localStorages := k.handleLocalStorage(ociSpec.Mounts, sandbox.id)
+	ctrStorages = append(ctrStorages, localStorages...)
+
 	// We replace all OCI mount sources that match our container mount
 	// with the right source path (The guest one).
 	if err = k.replaceOCIMountSource(ociSpec, newMounts); err != nil {
@@ -1116,14 +1128,14 @@ func (k *kataAgent) createContainer(sandbox *Sandbox, c *Container) (p *Process,
 func (k *kataAgent) handleEphemeralStorage(mounts []specs.Mount) []*grpc.Storage {
 	var epheStorages []*grpc.Storage
 	for idx, mnt := range mounts {
-		if mnt.Type == kataEphemeralDevType {
+		if mnt.Type == KataEphemeralDevType {
 			// Set the mount source path to a path that resides inside the VM
 			mounts[idx].Source = filepath.Join(ephemeralPath, filepath.Base(mnt.Source))
 
 			// Create a storage struct so that kata agent is able to create
 			// tmpfs backed volume inside the VM
 			epheStorage := &grpc.Storage{
-				Driver:     kataEphemeralDevType,
+				Driver:     KataEphemeralDevType,
 				Source:     "tmpfs",
 				Fstype:     "tmpfs",
 				MountPoint: mounts[idx].Source,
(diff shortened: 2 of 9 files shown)


egernst:
Thanks for the PR @awprice.  Travis is picking up some issues if you can take a look, while do the actual review.

egernst:
@bergwolf can you PTAL?

virtcontainers/mount_test.go
```diff
@@ -311,7 +311,7 @@ func TestIsEphemeralStorage(t *testing.T) {
 	}
 	defer os.RemoveAll(dir)
 
-	sampleEphePath := filepath.Join(dir, k8sEmptyDir, "tmp-volume")
+	sampleEphePath := filepath.Join(dir, K8sEmptyDir, "tmp-volume")
```
egernst: The type is k8sEmptyDir, not K8sEmptyDir, right?

awprice: It's this const - https://github.com/kata-containers/runtime/blob/9628c2493bd2e6760b1ca75eb8185b28d079e0e1/virtcontainers/mount.go#L347

Which is also used here - https://github.com/kata-containers/runtime/blob/9628c2493bd2e6760b1ca75eb8185b28d079e0e1/virtcontainers/mount.go#L389

virtcontainers/kata_agent.go
```diff
@@ -1134,6 +1138,30 @@ func (k *kataAgent) handleEphemeralStorage(mounts []specs.Mount) []*grpc.Storage
...
 
+// handleLocalStorage handles local storage within the VM
+// by creating a directory in the VM from the source of the mount point
+func (k *kataAgent) handleLocalStorage(mounts []specs.Mount, sandboxID string) []*grpc.Storage {
+	var localStorages []*grpc.Storage
+	for idx, mnt := range mounts {
+		if mnt.Type == kataLocalDevType {
+			// Set the mount source path to a the desired directory point in the VM.
+			// In this case it is located in the sandbox directory.
+			mounts[idx].Source = filepath.Join(kataGuestSharedDir, sandboxID, kataLocalDevType, filepath.Base(mnt.Source))
```
amshinde: What if two empty-dirs have the same Base? Could you make sure that this path is unique.

amshinde:
While this PR would help us to get away from 9p, this PR does introduce creating directories in the rootfs of the sandbox, which happens to be the pause container today. So, this PR adds some explicit dependency for the pause container to be present.

cc @bergwolf

virtcontainers/kata_agent.go
awprice: It isn't possible for two emptyDirs to have the same base, see example pod spec:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  containers:
    - name: test
      image: ubuntu
      args:
        - sleep
        - "3600"
  volumes:
    - name: test123
      emptyDir:
        medium: ""
    - name: test123
      emptyDir:
        medium: ""
```
```
$ k create -f test.yaml
The Pod "test" is invalid: spec.volumes[1].name: Duplicate value: "test123"
```

Should I still make it unique?

awprice:
> While this PR would help us to get away from 9p, this PR does introduce creating directories in the rootfs of the sandbox, which happens to be the pause container today. So, this PR adds some explicit dependency for the pause container to be present.
> 
> cc @bergwolf

Yep, this change definitely does add a dependency on the pause container.

pkg/katautils/create.go
```diff
@@ -161,9 +161,12 @@ func HandleFactory(ctx context.Context, vci vc.VC, runtimeConfig *oci.RuntimeCon
 // of the same pod the already existing volume is reused.
 func SetEphemeralStorageType(ociSpec oci.CompatOCISpec) oci.CompatOCISpec {
 	for idx, mnt := range ociSpec.Mounts {
-		if IsEphemeralStorage(mnt.Source) {
+		if vc.IsEphemeralStorage(mnt.Source) {
 			ociSpec.Mounts[idx].Type = "ephemeral"
 		}
+		if vc.Isk8sHostEmptyDir(mnt.Source) {
+			ociSpec.Mounts[idx].Type = "local"
```
lifupan: is it better to replace "local" with vc.kataLocalDevType ?

mcastelino:
Using the pause container for holding on to the ephemeral volumes makes a lot of sense. It keeps the individual container behaviour unchanged. This also allows the volume to grow to the limit afforded by the graph driver.

amshinde:
Looks like the tests are failing as we are explicitly checking for 9p mount for host directory based empty-dir volumes in our k8s integration tests:
https://github.com/kata-containers/tests/blob/master/integration/kubernetes/k8s-empty-dirs.bats#L26

@awprice Can you a raise a PR in the tests repo to fix that? 
@chavafg This PR would then need to carry a `depends-on` tag for the PR in the tests repo for the CI to pass?

chavafg:
yes, when you open the PR in the tests repo please add a `Depends-on: github.com/kata-containers/runtime#nnn`  on your commit message where `nnn` is your PR number here. You can also add the tag here using the PR on the tests repo.

pkg/katautils/create.go
awprice: Yea it makes sense to do that, unfortunately `vc.kataLocalDevType` & `vc.kataEphemeralDevType` are private, should I make them public?

awprice:
@amshinde & @chavafg 

I've raised https://github.com/kata-containers/tests/pull/1433 in the tests repo with `depends-on` in the commit message.

I'm not sure the change will fix the issue with the tests failing however, I think they might be reliant on the agent changes - https://github.com/kata-containers/agent/pull/521

We can see in the tests that the test pod for the emptyDir tests doesn't transition into the ready condition - 
```
not ok 1 Empty dir volumes
# (in test file k8s-empty-dirs.bats, line 22)
#   `kubectl wait --for=condition=Ready pod "$pod_name"' failed
# INFO: k8s configured to use trusted and untrusted annotations
# pod/sharevol-kata created
# error: timed out waiting for the condition
# pod "sharevol-kata" deleted
```

This is most likely because the agent doesn't know how to provision "local" storage types and doesn't create the pod properly.

amshinde:
@awprice I almost forgot about the agent PR! The agent PR will need to be merged first and then this one.

pkg/katautils/create.go
lifupan: Yes,  please.

virtcontainers/kata_agent.go
bergwolf: `kataGuestSharedDir` is actually tmpfs in the guest. Is it desirable to use memory for k8s empty dir? The other options are to use guest rootfs or container rootfs.

awprice: It is **NOT** desirable to use memory for k8s emptyDir (unless it is of `medium: Memory`) and in this case it is **NOT** on a tmpfs. The full path for the emptyDir directory will be: `/run/kata-containers/shared/containers/<sandbox id>/local/<emptyDir name>`, which is on whatever storage the sandbox container (i.e. the pause container is using).

See comment by @mcastelino:
> Using the pause container for holding on to the ephemeral volumes makes a lot of sense. It keeps the individual container behaviour unchanged. This also allows the volume to grow to the limit afforded by the graph driver.

bergwolf: @awprice OK, I see. You are relying on the fact that the sandbox id is the same as the first container id. We don't guarantee that but it is also the current fact. Please add some comments above the lines to highlight it so that we don't forget to fix it in case we need to change it.

awprice:
Thanks for the feedback @bergwolf and @lifupan! I've addressed your comments and updated the PR.

jcvenegas:
```
1..1
not ok 1 Empty dir volumes
# (in test file k8s-empty-dirs.bats, line 22)
#   `kubectl wait --for=condition=Ready pod "$pod_name"' failed
# INFO: k8s configured to use trusted and untrusted annotations
# pod/sharevol-kata created
# error: timed out waiting for the condition
# pod "sharevol-kata" deleted
```

Nice failing as expected, need agent PR merged first

amshinde:
@jcvenegas Let's include my commits in this PR. I am going to close the other one.

awprice:
Looks like the tests are going to fail until https://github.com/kata-containers/agent/pull/521 is merged so we should prioritise merging that PR first.

chavafg:
I think next step is to merge the tests PR and then this, right?

awprice:
@chavafg Is it possible to re-run the tests for this PR? I'd like to see if they are passing now that the agent PR is merged.

ganeshmaharaj:
Ok. That worked. Let's see how this goes @awprice

awprice:
@ganeshmaharaj :( they seem to have all failed, but with a java exception? They didn't even manage to get to the emptyDir tests.

amshinde:
@chavafg @GabyCT Seeing some very unusual java exceptions here, these look like an  issue with the Azure instance itself.
```
Running command '/usr/bin/docker [docker run --cidfile /tmp/cid798623508/d3KddNfu1XXIn7VEoFV1c9jWWMomP3 --runtime kata-runtime --name d3KddNfu1XXIn7VEoFV1c9jWWMomP3 --rm --cap-drop linux_immutable centos sh -c capsh --print]'
FATAL: command execution failed
java.io.EOFException
	at java.io.ObjectInputStream$PeekInputStream.readFully(ObjectInputStream.java:2681)
	at java.io.ObjectInputStream$BlockDataInputStream.readShort(ObjectInputStream.java:3156)
	at java.io.ObjectInputStream.readStreamHeader(ObjectInputStream.java:862)
	at java.io.ObjectInputStream.<init>(ObjectInputStream.java:358)
	at hudson.remoting.ObjectInputStreamEx.<init>(ObjectInputStreamEx.java:49)
	at hudson.remoting.Command.readFrom(Command.java:140)
	at hudson.remoting.Command.readFrom(Command.java:126)
	at hudson.remoting.AbstractSynchronousByteArrayCommandTransport.read(AbstractSynchronousByteArrayCommandTransport.java:36)
	at hudson.remoting.SynchronousCommandTransport$ReaderThread.run(SynchronousCommandTransport.java:63)
Caused: java.io.IOException: Unexpected termination of the channel
	at hudson.remoting.SynchronousCommandTransport$ReaderThread.run(SynchronousCommandTransport.java:77)
Caused: java.io.IOException: Backing channel 'fedora28-azurec446b0' is disconnected.
	at hudson.remoting.RemoteInvocationHandler.channelOrFail(RemoteInvocationHandler.java:214)
	at hudson.remoting.RemoteInvocationHandler.invoke(RemoteInvocationHandler.java:283)
```
We need to investigate these.

Meanwhile I am going to try running the CI one more time.
/test

grahamwhaley:
The Java exceptions are an artifact after the point of failure - they are just (normally) the jenkins stack complaining that the test aborted. Normally we need to look at the lines just above them and:
 - see what stage of the scripts failed
 - if we find a point in the script where it can fail but not give us enough clues, then we should PR an improvement to the scripts ;-)

but, I think what maybe happened here is the jenkins server had to go down for some maintenance - and all CIs have now passed.
I am seeing a big green merge button! - but, I'd feel more comfortable if somebody who has been tracking this PR series did the actual merge ;-)

virtcontainers/mount.go
```diff
@@ -326,3 +326,69 @@ func bindUnmountAllRootfs(ctx context.Context, sharedDir string, sandbox *Sandbo
...
+	if _, fsType, _ := GetDevicePathAndFsType(path); fsType != "tmpfs" {
+		return true
+	}
+	return false
+}
+
+func isEmptyDir(path string) bool {
+	splitSourceSlice := strings.Split(path, "/")
+	if len(splitSourceSlice) > 1 {
+		storageType := splitSourceSlice[len(splitSourceSlice)-2]
```
sachinmsft: this check would not work if container is mounting sub path in the emptyDir volume. In that case subpath would be append to the host path to be mounted and len(splitSourceSlice)-2 will not have kubernetes.io~empty-dir string.

kfox1111:
when 9p based, this seems to still leave it as 9p... why isn't it using storage in the vm itself in this case?

awprice:
> when 9p based, this seems to still leave it as 9p... why isn't it using storage in the vm itself in this case?

This PR changes the storage location for emptyDir to be in the rootfs of the sandbox, i.e. the pause container. In your case the rootfs of the sandbox is 9p. If you are using devicemapper, then the rootfs of the sandbox will benefit from the speed of devicemapper.

The storage of the VM itself is actually a ram disk, see comment by @mcastelino on https://github.com/kata-containers/runtime/issues/1472 - 

> The VM rootfs today is a NVDIMM or initrd. So the VM rootfs is not backed by any host side writable storage.

kfox1111:
That doesn't do what https://github.com/kata-containers/runtime/issues/1472 requested at all then. :(

What about making a qcow2 volume for the emptydirs?

Issue #1452, opened 2019-04-01
Store related vm template leak


teawater:
```
root@teawater-virtual-machine:/home/teawater/gopath/src/github.com/kata-containers/tests# ls -al /var/lib/vc/sbs
total 16
drwxr-x--- 2 root root 12288 Apr  1 14:35 .
drwxr-x--- 3 root root  4096 Mar 18 16:14 ..
root@teawater-virtual-machine:/home/teawater/gopath/src/github.com/kata-containers/tests# kata-runtime factory init
vm factory initialized
root@teawater-virtual-machine:/home/teawater/gopath/src/github.com/kata-containers/tests# ls -al /var/lib/vc/sbs
total 16
drwxr-x--- 2 root root 12288 Apr  1 14:37 .
drwxr-x--- 3 root root  4096 Mar 18 16:14 ..
root@teawater-virtual-machine:/home/teawater/gopath/src/github.com/kata-containers/tests# docker run --runtime=kata --rm -it tea0water/ubuntu
root@3948422be05e:/# exit
root@teawater-virtual-machine:/home/teawater/gopath/src/github.com/kata-containers/tests# ls -al /var/lib/vc/sbs
total 20
drwxr-x--- 3 root root 12288 Apr  1 14:37 .
drwxr-x--- 3 root root  4096 Mar 18 16:14 ..
drwxr-x--- 3 root root  4096 Apr  1 14:37 1426e6b5-7f1d-4f1e-aa93-e84ed8d761bf
root@teawater-virtual-machine:/home/teawater/gopath/src/github.com/kata-containers/tests# docker run --runtime=kata --rm -it tea0water/ubuntu
root@63bc4aacfcfa:/# exit
root@teawater-virtual-machine:/home/teawater/gopath/src/github.com/kata-containers/tests# ls -al /var/lib/vc/sbs
total 24
drwxr-x--- 4 root root 12288 Apr  1 14:38 .
drwxr-x--- 3 root root  4096 Mar 18 16:14 ..
drwxr-x--- 3 root root  4096 Apr  1 14:37 1426e6b5-7f1d-4f1e-aa93-e84ed8d761bf
drwxr-x--- 3 root root  4096 Apr  1 14:38 bf5886f0-276f-445a-8faa-a0a7fa8317c5
root@teawater-virtual-machine:/home/teawater/gopath/src/github.com/kata-containers/tests#
```
[...]

PR #1491, opened 2019-04-08
qemu: Remove the storage directories if qemu get from the factory

teawater:
Store related in directory /var/lib/vc/sbs and /run/vc/sbs if
vm template is enabled.
The cause is NewVM and NewVMFromGrpc will create vcStore with
VM's ID and set it as store of hypervisor if the factory is enabled.

This commit record the VM's ID to HypervisorConfig.VMid and remove
directories in qemu.cleanupVM to handle the issue.

Fixes: #1452

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -690,6 +690,17 @@ func (q *qemu) cleanupVM() error {
 		}
 	}
 
+	if q.config.VMid != "" {
+		dir = store.SandboxConfigurationRootPath(q.config.VMid)
+		if err := os.RemoveAll(dir); err != nil {
+			q.Logger().WithError(err).WithField("path", dir).Warnf("failed to remove vm path")
+		}
+		dir = store.SandboxRuntimeRootPath(q.config.VMid)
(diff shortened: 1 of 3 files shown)


virtcontainers/qemu.go
```diff
@@ -690,6 +690,17 @@ func (q *qemu) cleanupVM() error {
...
 	}
 
+	if q.config.VMid != "" {
+		dir = store.SandboxConfigurationRootPath(q.config.VMid)
+		if err := os.RemoveAll(dir); err != nil {
+			q.Logger().WithError(err).Warnf("failed to remove vm path %s", dir)
+		}
+		dir = store.SandboxRuntimeRootPath(q.config.VMid)
+		if err := os.RemoveAll(dir); err != nil {
+			q.Logger().WithError(err).Warnf("failed to remove vm path %s", dir)
```
devimc: please use `WithField` before calling `Warn`

teawater: > please use `WithField` before calling `Warn`

Updated.

Issue #1493, opened 2019-04-08
shimv2 testcase always failed after libcontainer updated to latest


lifupan:
# Description of problem

For much info, please https://github.com/kata-containers/agent/issues/502

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #1494, opened 2019-04-08
shimv2: fix the issue of stop container failed

lifupan:
There is no need to send another SIGKILL signal following
SIGTERM signal, otherwise, sending the SIGKILL signal would
get failed since the container process would has been terminated
by the previous SIGTERM signal.

Fixes:#1493


devimc:
dnm - waiting for https://github.com/kata-containers/agent/pull/526

Issue #1495, opened 2019-04-08
katautils: shimv2 will not output the proxy's log when enabled proxy.debug


lifupan:
# Description of problem

When used builtIn proxy, the  proxy.debug option is missed, thus there is no chance to
output the proxy's log even when proxy.debug enabled in kata configuration.

(replace this text with the list of steps you followed)

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #1496, opened 2019-04-08
katautils: fix the issue of missing proxy debug config

lifupan:
When used builtin proxy, it's better to config the proxy
debug based on setting in kata configuration.

Fixes:#1495

### DIFF
--- a/pkg/katautils/config.go
+++ b/pkg/katautils/config.go
@@ -598,6 +598,9 @@ func updateRuntimeConfigHypervisor(configPath string, tomlConf tomlConfig, confi
 func updateRuntimeConfigProxy(configPath string, tomlConf tomlConfig, config *oci.RuntimeConfig, builtIn bool) error {
 	if builtIn {
 		config.ProxyType = vc.KataBuiltInProxyType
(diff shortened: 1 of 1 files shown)


pkg/katautils/config.go
```diff
@@ -596,12 +596,14 @@ func updateRuntimeConfigHypervisor(configPath string, tomlConf tomlConfig, confi
...
-	if builtIn {
-		config.ProxyType = vc.KataBuiltInProxyType
-		return nil
-	}
-
 	for k, proxy := range tomlConf.Proxy {
+		if builtIn {
+			config.ProxyType = vc.KataBuiltInProxyType
+			config.ProxyConfig = vc.ProxyConfig{
+				Debug: proxy.debug(),
```
devimc: I'm not sure if it's correct to take the configuration of the first proxy that appears in the configuration file, since: 

1. This behaviour it's not documented. 
2. Proxy built-in is part of the runtime.

I'd prefer to use the runtime information instead.

```go
	Debug: config.Debug,
```

what do you think?

lifupan: Yeah, that's a good idea. 
Had updated!

PR #1502, opened 2019-04-08
network: Make tcfilter model as default

### LINKED ISSUE #1501: Change default networking model to `tcfilter`
tcfilter requires no changes to the interface provided by the network plugin and supports a larger set of plugins.

amshinde:
tcfilter requires no changes to the interface provided by the network
plugin and supports a larger set of plugins.

Fixes #1501

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -210,7 +210,7 @@ ifneq (,$(QEMUCMD))
 
     # qemu-specific options (all should be suffixed by "_QEMU")
     DEFBLOCKSTORAGEDRIVER_QEMU := virtio-scsi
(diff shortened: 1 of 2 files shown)


amshinde:
cc @egernst 
We probably need to evaluate our CI matrix, after this change. I think we had a CI job specifically for `tcfilter`, we would need to change that to `macvtap` in order tests runs for `macvtap`(or not?)

amshinde:
Test failing due to a bug in the way netmon processes routes manifested only in tcfilter mode as routes are actually handled in this mode. 
#1524 should fix this.

amshinde:
rebased changes..
/test

Issue #1503, opened 2019-04-08
Bump kernel to 5.x stable (5.0.7 for now)


ganeshmaharaj:
# Description of problem
Update kernel to 5.x stable releases.

ganeshmaharaj:
This issue is no longer valid as currently kata is not going to pursue moving to 5.x kernel. Will re-open when that becomes an active work item.

PR #1504, opened 2019-04-08
kernel: Move kernel to 5.0.y

ganeshmaharaj:
Fixes: #1503

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -161,8 +161,8 @@ assets:
     description: "Linux kernel optimised for virtual machines"
     url: "https://cdn.kernel.org/pub/linux/kernel/v4.x/"
     uscan-url: >-
-      https://mirrors.edge.kernel.org/pub/linux/kernel/v4.x/linux-(4\.19\..+)\.tar\.gz
-    version: "v4.19.28"
+      https://mirrors.edge.kernel.org/pub/linux/kernel/v5.x/linux-(5\.0\..+)\.tar\.gz
+    version: "v5.0.7"
 
 components:
   description: "Core system functionality"


jcvenegas:
You may need a config file in packaging repository too. 

Because config add kernel config live in different repository you will need to add `depends-on` in this commit.

https://github.com/kata-containers/tests#breaking-compatibility

egernst:
s/5\.x/5\.0\.y/ ?

ganeshmaharaj:
@jcvenegas will test all kata systems with the new kernel and current config to make sure it is alright to delve into the new configs. i will push a follow-on patch with the same soon.

bergwolf:
We carry a pretty large arm64 kernel patch in packaging. Is it still needed once we move to 5.0.y? @Weichen81

Weichen81:
Hi @pennyzct , @jongwu Could you please have a look at 5.0 kernel?

Weichen81:
I think the 52bit ipa feature has been included in the 5.x kernels. So I think it’s ok to remove the patch

bergwolf:
Thanks for confirmation @Weichen81 

@ganeshmaharaj please add a PR to packaging repo to drop [the arm64 kernel patch](https://github.com/kata-containers/packaging/blob/master/kernel/patches/0003-arm64-backport-Arm64-KVM-Dynamic-IPA-and-52bit-IPA-s.patch) otherwise this PR will break CI as the arm64 patch will fail to apply. -- sadly we apparently are not catching it in CI here.

bergwolf:
Adding `dnm` until the arm patch is dropped since otherwise it breaks CI.

ganeshmaharaj:
Currently waiting on https://github.com/kata-containers/tests/pull/1442 and https://github.com/kata-containers/packaging/pull/432 to land before proceeding with this change.

egernst:
depends on ?  https://github.com/kata-containers/packaging/pull/314

Pennyzct:
Hi~ @bergwolf @ganeshmaharaj As you may not notice, we also backport one patch for kernel v4.9.x for memory hotplug, which has been already included in v5.0.x, you could confirm from [here](https://github.com/torvalds/linux/commit/4ab215061554ae2a4b78744a5dd3b3c6639f16a7). Btw, not all parts, still left one little part regarding probe interface under review. But, we can wait for this PR merged and do the refinement. ;)
So,  @ganeshmaharaj let's also drop [arm64 memory hotplug](https://github.com/kata-containers/packaging/commit/c33b51911d821106726109b74875503d661aa276) patch to avoid break CI.

grahamwhaley:
ack @Pennyzct - I tried to build a 5.0.7 kernel yesterday and the arm memory patches failed to apply, so I (locally) deleted them :-) @ganeshmaharaj - I think we can thus nuke the arm patch subdir in this PR (can/must/need to).

ganeshmaharaj:
@Pennyzct @grahamwhaley Yes, i do have a commit with new config changes and the patch updates here. https://github.com/ganeshmaharaj/packaging/commit/b86b07c818ebb8f74cd28e993d3759e218162e0c I was going to re-work the PR right after we finish the kernel fragment PR to avoid too many clashes. If you wish to get that going now, i can get it ready.

versions.yaml
```diff
@@ -161,8 +161,8 @@ assets:
     description: "Linux kernel optimised for virtual machines"
     url: "https://cdn.kernel.org/pub/linux/kernel/v4.x/"
```
devimc: you should update this url, right?

grahamwhaley: yep: https://cdn.kernel.org/pub/linux/kernel/v5.x/
otherwise my (somewhere pending, maybe merged) 'version check scripts' will fail to find which versions are available, and probably complain loudly.

ganeshmaharaj: Yup. I am holding off any changes to this patch and CI churn until we land kernel fragments change. Once that is done, i can update this patch and also PR the kernel patches update in packaging repo.

jodh-intel:
Still blocked on https://github.com/kata-containers/packaging/pull/314 fwics.

ganeshmaharaj:
Right. the current plan is to hold off with 5.x move and work on getting virtio-fs & kernel fragments based off 4.x onto the tree. Following which we can work out a plan to move to 5.x

raravena80:
@ganeshmaharaj any updates? Still blocked by https://github.com/kata-containers/packaging/pull/314?

grahamwhaley:
Yeah, I think so, which is then blocked on the firecracker CI seeming less stable with the fragment kernel config.... we may have to bite the bullet on kernel version updates if we don't figure that out soon.

egernst:
@ganeshmaharaj I think this can be closed at this point?

Issue #1195, opened 2019-01-30
kata-env: if vsock cannot be enabled due to missing module, say so in the kata-env


grahamwhaley:
# Description of problem

If you enable vsock support with `use_vsock = true` in the runtime config file, *but*, you don't happen to have the vsock kernel modules loaded, the `kata-env` output is a bit confusing, as it implies it has not seen the `use_vsock`...

```bash
$ kata-runtime kata-env | fgrep -i vsocks
  UseVSock = false
  SupportVSocks = false
```

The hint is there, that the host does not support vsocks (for whatever reason) - the second line comes from the `[Host]` section - but, it is not obvious, as the `UseVSock` line comes out first in the kata-env, and 'feels' like it reflects the config file setting... I started off wondering if the runtime had managed to pick up my config file change for instance.

# Expected result

I think it would be nice if on the `UseVSock` line, maybe it could be appended with `kernel module not loaded?` - **if** `use_vsock` is set to true in the config file.

I think this is a touch tricky, as the kata-env code afaik pretty much just prints out the internal status structures, and thus this level of detail doesn't currently fit in with that. But, if we can make it a little more obvious that the runtime knows you asked for vsock, but could not load it... that'd be nice.

grahamwhaley:
Related #581

Issue #581, opened 2018-08-14
vsock support check requires manual kernel module loading


maximilianriemensberger:
# Description of problem

I configured kata-runtime to use vsocks on ubuntu 18.04 with kata-runtime 1.2.  But the detection of vsock support failed.  It seems that the code explicitly looks for /dev/vsock and /dev/vhost-vsock. Those are only present after loading the vsock and vhost_vsock modules.

```
Aug 14 07:33:39 hix kata-runtime[29715]: time="2018-08-14T07:33:39.953266567+02:00" level=warning msg="No vsock support, falling back to legacy serial port" arch=amd64 command=create name=kata-runtime pid=29715 source=runtime
```
After loading the module:

```
Aug 14 07:48:03 hix kata-runtime[30209]: time="2018-08-14T07:48:03.900473483+02:00" level=info msg="vsock supported" arch=amd64 command=create name=kata-runtime pid=30209 source=runtime
Aug 14 07:48:03 hix kata-runtime[30209]: time="2018-08-14T07:48:03.90062639+02:00" level=info msg="VSOCK supported, configure to not use proxy" arch=amd64 command=create name=kata-runtime pid=30209 source=runtime
```

kata-runtime  : 1.2.0
   commit   : 0bcb32f
   OCI specs: 1.0.1

# Expected result

kata runtime detects vsock support even if the kernel modules are not loaded.
The modules are autoloaded when used, probably by calling socket(AF_VSOCK, ...).

# Actual result

Vsock support was detected only after manually loading all relevant kernel modules, i.e., modprobe vhost_vsock.

jodh-intel:
Thanks for reporting @maximilianriemensberger.

/cc @devimc.

maximilianriemensberger:
I played around with `AF_VSOCK` sockets.  The straightforward call to `socket(AF_VSOCK, SOCK_STREAM, 0)` autoloads the vsock module but not the vhost_vsock module. Instead it various loads vmw_vmci modules (vmw_vsock_vmci_transport, vmw_vmci). I don't know but autoloading vhost_vsock might not be as easy as I thought.

devimc:
@maximilianriemensberger any update on this? I'll close this, please feel free to reopen it if you wish to add something

PR #1513, opened 2019-04-10
support-vsock: load vhost_vsock module if it isn't built-in

Pennyzct:
# Description of problem
In a few host machines, when vhost_vsock isn't built-in, but could be loadable, value `SupportVoscks` from `kata-env` still outputs false.
We add `modprobe -i vhost_vsock` in func `SupportsVsocks` to try to avoid above scenario.

# Expected result
```
[Host]
  Kernel = "5.0.7"
  Architecture = "arm64"
  VMContainerCapable = true
  SupportVSocks = true
  [Host.Distro]
    Name = "Ubuntu"
    Version = "18.04"
  [Host.CPU]
    Vendor = "3rd Party Limited"
    Model = "v8"
```

# Actual result
```
[Host]
  Kernel = "5.0.7"
  Architecture = "arm64"
  VMContainerCapable = true
  SupportVSocks = false
  [Host.Distro]
    Name = "Ubuntu"
    Version = "18.04"
  [Host.CPU]
    Vendor = "3rd Party Limited"
    Model = "v8"
```
Related: #1195 #581

### DIFF
--- a/virtcontainers/utils/utils.go
+++ b/virtcontainers/utils/utils.go
@@ -29,9 +29,6 @@ const MibToBytesShift = 20
 // See unix(7).
 const MaxSocketPathLen = 107
 
-// VSockDevicePath path to vsock device
-var VSockDevicePath = "/dev/vsock"
-
 // VHostVSockDevicePath path to vhost-vsock device
 var VHostVSockDevicePath = "/dev/vhost-vsock"
 
@@ -234,10 +231,6 @@ func BuildSocketPath(elements ...string) (string, error) {
 
 // SupportsVsocks returns true if vsocks are supported, otherwise false
 func SupportsVsocks() bool {
-	if _, err := os.Stat(VSockDevicePath); err != nil {
-		return false
-	}
-
 	if _, err := os.Stat(VHostVSockDevicePath); err != nil {
 		return false
 	}
--- a/cli/kata-check.go
+++ b/cli/kata-check.go
@@ -35,6 +35,9 @@ type kernelModule struct {
 
 	// maps parameter names to values
 	parameters map[string]string
+
+	// if it is definitely required
+	required bool
 }
 
 type vmContainerCapableDetails struct {
@@ -124,8 +127,11 @@ func haveKernelModule(module string) bool {
 	// Now, check if the module is unloaded, but available.
 	// And modprobe it if so.
 	cmd := exec.Command(modProbeCmd, module)
-	err := cmd.Run()
-	return err == nil
+	if output, err := cmd.CombinedOutput(); err != nil {
+		kataLog.WithField("module", module).WithError(err).Warnf("modprobe insert module failed: %s", string(output))
+		return false
+	}
+	return true
 }
 
 // checkCPU checks all required CPU attributes modules and returns a count of
@@ -200,7 +206,9 @@ func checkKernelModules(modules map[string]kernelModule, handler kernelParamHand
 
 		if !haveKernelModule(module) {
 			kataLog.WithFields(fields).Error("kernel property not found")
-			count++
+			if details.required {
+				count++
+			}
 			continue
 		}
 
--- a/cli/kata-check_amd64.go
+++ b/cli/kata-check_amd64.go
@@ -15,14 +15,15 @@ import (
 )
 
 const (
-	cpuFlagsTag        = genericCPUFlagsTag
-	archCPUVendorField = genericCPUVendorField
-	archCPUModelField  = genericCPUModelField
-	archGenuineIntel   = "GenuineIntel"
-	archAuthenticAMD   = "AuthenticAMD"
-	msgKernelVM        = "Kernel-based Virtual Machine"
-	msgKernelVirtio    = "Host kernel accelerator for virtio"
-	msgKernelVirtioNet = "Host kernel accelerator for virtio network"
+	cpuFlagsTag               = genericCPUFlagsTag
+	archCPUVendorField        = genericCPUVendorField
+	archCPUModelField         = genericCPUModelField
+	archGenuineIntel          = "GenuineIntel"
+	archAuthenticAMD          = "AuthenticAMD"
+	msgKernelVM               = "Kernel-based Virtual Machine"
+	msgKernelVirtio           = "Host kernel accelerator for virtio"
+	msgKernelVirtioNet        = "Host kernel accelerator for virtio network"
+	msgKernelVirtioVhostVsock = "Host Support for Linux VM Sockets"
 )
 
 // CPU types
@@ -75,17 +76,25 @@ func setCPUtype() error {
 		}
 		archRequiredKernelModules = map[string]kernelModule{
 			"kvm": {
-				desc: msgKernelVM,
+				desc:     msgKernelVM,
+				required: true,
 			},
 			"kvm_intel": {
 				desc:       "Intel KVM",
 				parameters: kvmIntelParams,
+				required:   true,
 			},
 			"vhost": {
-				desc: msgKernelVirtio,
+				desc:     msgKernelVirtio,
+				required: true,
 			},
 			"vhost_net": {
-				desc: msgKernelVirtioNet,
+				desc:     msgKernelVirtioNet,
+				required: true,
+			},
+			"vhost_vsock": {
+				desc:     msgKernelVirtioVhostVsock,
+				required: false,
 			},
 		}
 	} else if cpuType == cpuTypeAMD {
@@ -99,16 +108,24 @@ func setCPUtype() error {
 		}
 		archRequiredKernelModules = map[string]kernelModule{
 			"kvm": {
-				desc: msgKernelVM,
+				desc:     msgKernelVM,
+				required: true,
 			},
 			"kvm_amd": {
-				desc: "AMD KVM",
+				desc:     "AMD KVM",
+				required: true,
 			},
 			"vhost": {
-				desc: msgKernelVirtio,
+				desc:     msgKernelVirtio,
+				required: true,
 			},
 			"vhost_net": {
-				desc: msgKernelVirtioNet,
+				desc:     msgKernelVirtioNet,
+				required: true,
+			},
+			"vhost_vsock": {
+				desc:     msgKernelVirtioVhostVsock,
+				required: false,
 			},
 		}
 	}
--- a/cli/kata-check_amd64_test.go
+++ b/cli/kata-check_amd64_test.go
@@ -171,6 +171,7 @@ func TestCheckCheckKernelModulesNoNesting(t *testing.T) {
 				"nested":             "Y",
 				"unrestricted_guest": "Y",
 			},
+			required: true,
 		},
 	}
 
@@ -255,6 +256,7 @@ func TestCheckCheckKernelModulesNoUnrestrictedGuest(t *testing.T) {
 				"nested":             "Y",
 				"unrestricted_guest": "Y",
 			},
+			required: true,
 		},
 	}
 
--- a/cli/kata-check_arm64.go
+++ b/cli/kata-check_arm64.go
@@ -30,13 +30,20 @@ var archRequiredCPUAttribs = map[string]string{}
 // required module parameters.
 var archRequiredKernelModules = map[string]kernelModule{
 	"kvm": {
-		desc: "Kernel-based Virtual Machine",
+		desc:     "Kernel-based Virtual Machine",
+		required: true,
 	},
 	"vhost": {
-		desc: "Host kernel accelerator for virtio",
+		desc:     "Host kernel accelerator for virtio",
+		required: true,
 	},
 	"vhost_net": {
-		desc: "Host kernel accelerator for virtio network",
+		desc:     "Host kernel accelerator for virtio network",
+		required: true,
+	},
+	"vhost_vsock": {
+		desc:     "Host Support for Linux VM Sockets",
+		required: false,
 	},
 }
 
--- a/cli/kata-check_ppc64le.go
+++ b/cli/kata-check_ppc64le.go
@@ -42,10 +42,16 @@ var archRequiredCPUAttribs = map[string]string{}
 // required module parameters.
 var archRequiredKernelModules = map[string]kernelModule{
 	"kvm": {
-		desc: "Kernel-based Virtual Machine",
+		desc:     "Kernel-based Virtual Machine",
+		required: true,
 	},
 	"kvm_hv": {
-		desc: "Kernel-based Virtual Machine hardware virtualization",
+		desc:     "Kernel-based Virtual Machine hardware virtualization",
+		required: true,
+	},
+	"vhost_vsock": {
+		desc:     "Host Support for Linux VM Sockets",
+		required: false,
 	},
 }
 
--- a/cli/kata-check_s390x.go
+++ b/cli/kata-check_s390x.go
@@ -33,7 +33,12 @@ var archRequiredCPUAttribs = map[string]string{}
 // required module parameters.
 var archRequiredKernelModules = map[string]kernelModule{
 	"kvm": {
-		desc: "Kernel-based Virtual Machine",
+		desc:     "Kernel-based Virtual Machine",
+		required: true,
+	},
+	"vhost_vsock": {
+		desc:     "Host Support for Linux VM Sockets",
+		required: false,
 	},
 }
 
--- a/cli/kata-check_test.go
+++ b/cli/kata-check_test.go
@@ -502,6 +502,7 @@ func TestCheckCheckKernelModules(t *testing.T) {
 		"foo": {
 			desc:       "desc",
 			parameters: map[string]string{},
+			required:   true,
 		},
 		"bar": {
 			desc: "desc",
@@ -511,6 +512,7 @@ func TestCheckCheckKernelModules(t *testing.T) {
 				"param3": "a",
 				"param4": ".",
 			},
+			required: true,
 		},
 	}
 
@@ -570,6 +572,7 @@ func TestCheckCheckKernelModulesUnreadableFile(t *testing.T) {
 			parameters: map[string]string{
 				"param1": "wibble",
 			},
+			required: true,
 		},
 	}
 
@@ -617,6 +620,7 @@ func TestCheckCheckKernelModulesInvalidFileContents(t *testing.T) {
 			parameters: map[string]string{
 				"param1": "wibble",
 			},
+			required: true,
 		},
 	}
 
@@ -713,6 +717,7 @@ func TestCheckKernelParamHandler(t *testing.T) {
 		"foo": {
 			desc:       "desc",
 			parameters: map[string]string{},
+			required:   true,
 		},
 		"bar": {
 			desc: "desc",
@@ -720,6 +725,7 @@ func TestCheckKernelParamHandler(t *testing.T) {
 				"param1": "hello",
 				"param2": "world",
 			},
+			required: true,
 		},
 	}
 
@@ -731,6 +737,7 @@ func TestCheckKernelParamHandler(t *testing.T) {
 			parameters: map[string]string{
 				"param1": "moo",
 			},
+			required: true,
 		},
 	}
 
@@ -740,6 +747,7 @@ func TestCheckKernelParamHandler(t *testing.T) {
 			parameters: map[string]string{
 				"param1": "bar",
 			},
+			required: true,
 		},
 	}
 
--- a/pkg/katautils/config_test.go
+++ b/pkg/katautils/config_test.go
@@ -719,13 +719,10 @@ func TestMinimalRuntimeConfigWithVsock(t *testing.T) {
 	[agent.kata]
 `
 	orgVHostVSockDevicePath := utils.VHostVSockDevicePath
-	orgVSockDevicePath := utils.VSockDevicePath
 	defer func() {
 		utils.VHostVSockDevicePath = orgVHostVSockDevicePath
-		utils.VSockDevicePath = orgVSockDevicePath
 	}()
 	utils.VHostVSockDevicePath = "/dev/null"
-	utils.VSockDevicePath = "/dev/null"
 
 	configPath := path.Join(dir, "runtime.toml")
 	err = createConfig(configPath, runtimeMinimalConfig)
@@ -765,13 +762,10 @@ func TestNewQemuHypervisorConfig(t *testing.T) {
 	disableBlock := true
 	enableIOThreads := true
 	hotplugVFIOOnRootBus := true
-	orgVSockDevicePath := utils.VSockDevicePath
 	orgVHostVSockDevicePath := utils.VHostVSockDevicePath
 	defer func() {
-		utils.VSockDevicePath = orgVSockDevicePath
 		utils.VHostVSockDevicePath = orgVHostVSockDevicePath
 	}()
-	utils.VSockDevicePath = "/dev/abc/xyz"
 	utils.VHostVSockDevicePath = "/dev/abc/xyz"
 
 	hypervisor := hypervisor{
@@ -808,7 +802,6 @@ func TestNewQemuHypervisorConfig(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	utils.VSockDevicePath = "/dev/null"
 	utils.VHostVSockDevicePath = "/dev/null"
 
 	// all paths exist now
--- a/virtcontainers/utils/utils_test.go
+++ b/virtcontainers/utils/utils_test.go
@@ -298,23 +298,12 @@ func TestBuildSocketPath(t *testing.T) {
 func TestSupportsVsocks(t *testing.T) {
 	assert := assert.New(t)
 
-	orgVSockDevicePath := VSockDevicePath
 	orgVHostVSockDevicePath := VHostVSockDevicePath
 	defer func() {
-		VSockDevicePath = orgVSockDevicePath
 		VHostVSockDevicePath = orgVHostVSockDevicePath
 	}()
 
-	VSockDevicePath = "/abc/xyz/123"
 	VHostVSockDevicePath = "/abc/xyz/123"
-	assert.False(SupportsVsocks())
-
-	vSockDeviceFile, err := ioutil.TempFile("", "vsock")
-	assert.NoError(err)
-	defer os.Remove(vSockDeviceFile.Name())
-	defer vSockDeviceFile.Close()
-	VSockDevicePath = vSockDeviceFile.Name()
-
 	assert.False(SupportsVsocks())
 
 	vHostVSockFile, err := ioutil.TempFile("", "vhost-vsock")


Pennyzct:
Hi~ @grahamwhaley @devimc @chavafg @jodh-intel I don't know why a lot CI failed. I really tried to find the spot. Could anyone lend a hand here? a lot thanks!!!!! 😭😭😭

Pennyzct:
Hi~ @grahamwhaley @chavafg @devimc @jodh-intel sorry to bother. I know a lot CI failed. I really tried to find the failing spot. I literally have no clue. could anyone lend a hand here? a lot thanks!!!!!! ;)

grahamwhaley:
Hi @Pennyzct :-) Yeah, the error is not obvious ;-)
I think what has happened is the script has tried to run `kata-runtime kata-env`, and that has failed, and thus failed the build. Could you try to run that locally?

```
Logging kata-env information:
```

We should also get some opinion on if we should be auto-loading the `vsock` module, and `exec`ing a modprobe call. I think we probably originally decided to *not* do this when we first wrote the code - now we need to discuss/remember why...
/cc @sboeuf @egernst @gnawux @jodh-intel

sboeuf:
@grahamwhaley I don't recall the discussion about the decision, but the main drawback is that having this `modprobe` being performed by `kata-runtime` would affect the "hot path" of creating a container, which would affect performances.

Pennyzct:
Hi~ @grahamwhaley sooooooo thanks for the help!!!!
on AArch64, that's no problem, Here is the output from my local machine:
```
root@entos-thunderx2-desktop:~# kata-runtime kata-env
[Meta]
  Version = "1.0.21"

[Runtime]
  Debug = true
  Trace = false
  DisableGuestSeccomp = true
  DisableNewNetNs = false
  Path = "/usr/local/bin/kata-runtime"
  [Runtime.Version]
    Semver = "1.7.0-alpha0"
    Commit = "1684332d059f3698f4a1a7f6d18cf32c4f79a732"
    OCI = "1.0.1-dev"
  [Runtime.Config]
    Path = "/etc/kata-containers/configuration.toml"

[Hypervisor]
  MachineType = "virt"
  Version = "QEMU emulator version 3.0.92 (v3.1.0-rc2-dirty)\nCopyright (c) 2003-2018 Fabrice Bellard and the QEMU Project developers"
  Path = "/snap/kata-containers/140/usr/bin/qemu-system-aarch64"
  BlockDeviceDriver = "virtio-scsi"
  EntropySource = "/dev/urandom"
  Msize9p = 8192
  MemorySlots = 10
  Debug = true
  UseVSock = true

[Image]
  Path = ""

[Kernel]
  Path = "/snap/kata-containers/140/usr/share/kata-containers/vmlinuz-4.19.28.container"
  Parameters = "agent.log=debug initcall_debug"

[Initrd]
  Path = "/usr/share/kata-containers/kata-containers-alpine-3.7-osbuilder-edd7d9c-agent-74639b7.initrd"

[Proxy]
  Type = "noProxy"
  Version = ""
  Path = ""
  Debug = false

[Shim]
  Type = "kataShim"
  Version = "kata-shim version 1.6.1-39896627000d0305490e26440880a68f766ce45c"
  Path = "/snap/kata-containers/140/usr/libexec/kata-containers/kata-shim"
  Debug = true

[Agent]
  Type = "kata"

[Host]
  Kernel = "5.0.7"
  Architecture = "arm64"
  VMContainerCapable = true
  SupportVSocks = true
  [Host.Distro]
    Name = "Ubuntu"
    Version = "18.04"
  [Host.CPU]
    Vendor = "3rd Party Limited"
    Model = "v8"

[Netmon]
  Version = "kata-netmon version 1.6.1"
  Path = "/snap/kata-containers/current/usr/libexec/kata-containers/kata-netmon"
  Debug = true
  Enable = false

```
That the ARM CI got passed also confirmed the result. 
I will try to find one x86_64 machine to do the test. ;)

Pennyzct:
Hi~ @sboeuf sorry to bother, could you please explain what is "hot path"?? or just link me the similar issue?? a lot thanks. ;)

sboeuf:
@Pennyzct
>could you please explain what is "hot path"?? or just link me the similar issue??

By "hot path" I mean the code path that is used when calling into `kata-runtime create`. This is a critical path since we expect some performance from there, and by having the module loading part of this code flow, we might slow it down.

That being said, if the modprobe is not too slow and if others think that is not too critical, then we could consider having the `kata-runtime create` command checking if the module is loaded or not, and load it if that's not the case.

Pennyzct:
Hi~ @sboeuf thanks for the detailed explanation! ;)

grahamwhaley:
@Pennyzct - you should probably use the [report tool](https://github.com/kata-containers/tests/tree/master/metrics/report) to do a before/after comparison of boot times (`grabdata.sh -t` with the report tool), so we can see if there is any visible impact to the probe.

I have to agree, that if it has an impact (and I think it probably will, as it has to exec() another binary which goes and does some work), then we may not want to do the probe.

Pennyzct:
Hi~ @grahamwhaley 
Thanks for the `report tool`, everyday learn a new thing. ;) I will try and paste out the comparison lately.
previously, i was writing a simple test to catch the difference of boot time of `kata create`
here is the code, just calculate the execution time of the same `kata create` 100 times and see the mean.
```
 #!/bin/bash
count=0.000000
for((i=1;i<=100;i++))
do
startTime=`date +"%s.%N"`
kata-runtime create --bundle /tmp/bundle067106738 vsock
endTime=`date +"%s.%N"`
value=$(echo `awk -v x1="$(echo $endTime | cut -d '.' -f 1)" -v x2="$(echo $startTime | cut -d '.' -f 1)" -v y1="$[$(echo $endTime | cut -d '.' -f 2) / 1000]" -v y2="$[$(echo $startTime | cut -d '.' -f 2) /1000]" 'BEGIN{printf "%.6f",(x1-x2)+(y1-y2)/1000000}'`)
echo $value
count=$(echo `awk -v x1="$(echo $value)" -v x2="$(echo $count)" 'BEGIN{printf "%.6f",x1+x2}'`)
echo $count
kata-runtime delete vsock
done
mean=$(echo `awk -v x1="$(echo $count)" 'BEGIN{printf "%.6f\n",x1/100}'`)
```
Do you think it is also a feasible solution? ;)

grahamwhaley:
Hi @Pennyzct - looks feasible. Yes, running >=20 instances and taking the average is the right way, as we do get spikes due to other system activity etc.
That test is quite similar to the one the report tool makes use of: https://github.com/kata-containers/tests/blob/master/metrics/time/launch_times.sh
Be careful that you may be measuring the complete create/run/delete cycle - you may not want to measure that 'delete' bit ;-)

Yours:
 - uses the runtime directly, whereas the `launch_times` uses docker. In some ways that is good, as it just measures kata parts.
 - Only measures the complete cycle. the `launch_times` tries to measure some other times as well - time for the kernel to boot, time to get to the workload etc.

/me very happy to see somebody running metrics :-)

Pennyzct:
Hi~ @grahamwhaley @sboeuf I have changed my solution, from probing every time when creating a container to only the first time, if on the same host machine. ;) 
I will still do the performance test, using following script, I will `modprobe -r vhost_vsock` every time to make sure the same environment:
```
#!/bin/bash
count=0.000000
for((i=1;i<=20;i++))
do
startTime=`date +"%s.%N"`
kata-runtime create --bundle /tmp/bundle067106738 vsock
endTime=`date +"%s.%N"`
value=$(echo `awk -v x1="$(echo $endTime | cut -d '.' -f 1)" -v x2="$(echo $startTime | cut -d '.' -f 1)" -v y1="$[$(echo $endTime | cut -d '.' -f 2) / 1000]" -v y2="$[$(echo $startTime | cut -d '.' -f 2) /1000]" 'BEGIN{printf "%.6f",(x1-x2)+(y1-y2)/1000000}'`)
echo $value
count=$(echo `awk -v x1="$(echo $value)" -v x2="$(echo $count)" 'BEGIN{printf "%.6f",x1+x2}'`)
kata-runtime delete vsock
while :
do
        if [ "$(lsmod | awk '{ if($1 == "vhost_vsock") print $3;}')" == "0" ]; then
                modprobe -r vhost_vsock
                break
        fi
done
done
mean=$(echo `awk -v x1="$(echo $count)" 'BEGIN{printf "%.6f\n",x1/20}'`)
echo $mean
```
I'm afraid if I use [report tool](https://github.com/kata-containers/tests/tree/master/metrics/report) @grahamwhaley suggested, it couldn't provide the same environment before each run. ;)

Pennyzct:
Hi~ @grahamwhaley @sboeuf 
On AArch64, Each run has 20 samples. 
Before: 
```
2.924
2.942
2.948
2.942
2.941
2.944
2.932
2.931
2.944
2.941
2.939
2.951
2.947
2.942
2.936
2.952
2.942
2.937
2.939
2.932
mean: 2.940
```
After: (load my patch)
```
2.942
2.962
2.962
2.948
2.954
2.950
2.954
2.946
2.943
2.940
2.939
2.950
2.950
2.931
2.958
2.956
2.941
2.960
2.946
2.943
mean: 2.948
```
I will try to find one x86_64 machine to do the comparison. ;)

grahamwhaley:
@Pennyzct - you should be able to use the report tool to tell if the PR makes a different to the real use case - that is, the module would only get inserted once for the first container, but the check would happen for every container - so you would do something like:

```bash
# Have the last kata release installed
$ modprobe -r *vsock* (or whatever, to ensure we have a clean start - maybe even a reboot)
$ grabdata.sh -t
$ mkdir ../results/run1
$ mv ../results/*.json ../results/run1/
# Now check out, build and install your PR into your system
$ modprobe -r *vsock* (or whatever, to ensure we have a clean start - maybe even a reboot)
$ grabdata.sh -t
$ mkdir ../results/run2
$ mv ../results/*.json ../results/run2/
$ ./makereport.sh
# and view report in output dir
```

The idea is to compare the current version or HEAD performance, and the performance when the new code is installed. You do not want to remove the module between each container launch, as that is not how it would work in 'the real world' - the module would only get loaded once, the first time the probe is done. What we are really checking for is if the module/vsock *check* makes any difference to the performance.

grahamwhaley:
It looks to me that the `kata-runtime kata-env` is still failing, on the metrics machine at least?

```bash
10:20:02 /usr/local/bin/kata-runtime
10:20:02 Logging kata-env information:
10:20:02 Build step 'Execute shell' marked build as failure
```

Pennyzct:
Hi~ @grahamwhaley I found one x86_64 machine and patched my cold. 
No matter when I turned on `use_vsock` or not, `kata-runtime kata-env` works well. 
**when `use_vsock` is turned on:**
```
$ kata-runtime kata-env                
[Meta]
  Version = "1.0.21"

[Runtime]
  Debug = false
  Trace = false
  DisableGuestSeccomp = true
  DisableNewNetNs = false
  Path = "/usr/local/bin/kata-runtime"
  [Runtime.Version]
    Semver = "1.7.0-alpha0"
    Commit = "8d711073fe242ff01aee89e540eee1633705f92f"
    OCI = "1.0.1-dev"
  [Runtime.Config]
    Path = "/usr/share/defaults/kata-containers/configuration-qemu.toml"

[Hypervisor]
  MachineType = "pc"
  Version = "QEMU emulator version 2.11.0\nCopyright (c) 2003-2017 Fabrice Bellard and the QEMU Project developers"
  Path = "/usr/bin/qemu-lite-system-x86_64"
  BlockDeviceDriver = "virtio-scsi"
  EntropySource = "/dev/urandom"
  Msize9p = 8192
  MemorySlots = 10
  Debug = false
  UseVSock = true

[Image]
  Path = "/usr/share/kata-containers/kata-containers-clearlinux-28830-osbuilder-edd7d9c-agent-7720b93.img"

[Kernel]
  Path = "/usr/share/kata-containers/vmlinuz-4.19.28-33"
  Parameters = "init=/usr/lib/systemd/systemd systemd.unit=kata-containers.target systemd.mask=systemd-networkd.service systemd.mask=systemd-networkd.socket systemd.mask=systemd-journald.service systemd.mask=systemd-journald.socket systemd.mask=systemd-journal-flush.service systemd.mask=systemd-udevd.service systemd.mask=systemd-udevd.socket systemd.mask=systemd-udev-trigger.service systemd.mask=systemd-timesyncd.service systemd.mask=systemd-update-utmp.service systemd.mask=systemd-tmpfiles-setup.service systemd.mask=systemd-tmpfiles-cleanup.service systemd.mask=systemd-tmpfiles-cleanup.timer systemd.mask=tmp.mount systemd.mask=systemd-random-seed.service"

[Initrd]
  Path = ""

[Proxy]
  Type = "noProxy"
  Version = ""
  Path = ""
  Debug = false

[Shim]
  Type = "kataShim"
  Version = "kata-shim version 1.6.0-rc2-ea27044"
  Path = "/usr/libexec/kata-containers/kata-shim"
  Debug = false

[Agent]
  Type = "kata"

[Host]
  Kernel = "4.15.0-43-generic"
  Architecture = "amd64"
  VMContainerCapable = true
  SupportVSocks = true
  [Host.Distro]
    Name = "Ubuntu"
    Version = "18.04"
  [Host.CPU]
    Vendor = "GenuineIntel"
    Model = "Intel(R) Core(TM) i7-4790 CPU @ 3.60GHz"

[Netmon]
  Version = "kata-netmon version 1.7.0-alpha0"
  Path = "/usr/libexec/kata-containers/kata-netmon"
  Debug = false
  Enable = false
```
**when `use_vsock` is turned off:**
```
$ kata-runtime kata-env
[Meta]
  Version = "1.0.21"

[Runtime]
  Debug = false
  Trace = false
  DisableGuestSeccomp = true
  DisableNewNetNs = false
  Path = "/usr/local/bin/kata-runtime"
  [Runtime.Version]
    Semver = "1.7.0-alpha0"
    Commit = "8d711073fe242ff01aee89e540eee1633705f92f"
    OCI = "1.0.1-dev"
  [Runtime.Config]
    Path = "/usr/share/defaults/kata-containers/configuration-qemu.toml"

[Hypervisor]
  MachineType = "pc"
  Version = "QEMU emulator version 2.11.0\nCopyright (c) 2003-2017 Fabrice Bellard and the QEMU Project developers"
  Path = "/usr/bin/qemu-lite-system-x86_64"
  BlockDeviceDriver = "virtio-scsi"
  EntropySource = "/dev/urandom"
  Msize9p = 8192
  MemorySlots = 10
  Debug = false
  UseVSock = false

[Image]
  Path = "/usr/share/kata-containers/kata-containers-clearlinux-28830-osbuilder-edd7d9c-agent-7720b93.img"

[Kernel]
  Path = "/usr/share/kata-containers/vmlinuz-4.19.28-33"
  Parameters = "init=/usr/lib/systemd/systemd systemd.unit=kata-containers.target systemd.mask=systemd-networkd.service systemd.mask=systemd-networkd.socket systemd.mask=systemd-journald.service systemd.mask=systemd-journald.socket systemd.mask=systemd-journal-flush.service systemd.mask=systemd-udevd.service systemd.mask=systemd-udevd.socket systemd.mask=systemd-udev-trigger.service systemd.mask=systemd-timesyncd.service systemd.mask=systemd-update-utmp.service systemd.mask=systemd-tmpfiles-setup.service systemd.mask=systemd-tmpfiles-cleanup.service systemd.mask=systemd-tmpfiles-cleanup.timer systemd.mask=tmp.mount systemd.mask=systemd-random-seed.service"

[Initrd]
  Path = ""

[Proxy]
  Type = "kataProxy"
  Version = "kata-proxy version 1.6.0-rc2-d4aa0b2"
  Path = "/usr/libexec/kata-containers/kata-proxy"
  Debug = false

[Shim]
  Type = "kataShim"
  Version = "kata-shim version 1.6.0-rc2-ea27044"
  Path = "/usr/libexec/kata-containers/kata-shim"
  Debug = false

[Agent]
  Type = "kata"

[Host]
  Kernel = "4.15.0-43-generic"
  Architecture = "amd64"
  VMContainerCapable = true
  SupportVSocks = true
  [Host.Distro]
    Name = "Ubuntu"
    Version = "18.04"
  [Host.CPU]
    Vendor = "GenuineIntel"
    Model = "Intel(R) Core(TM) i7-4790 CPU @ 3.60GHz"

[Netmon]
  Version = "kata-netmon version 1.7.0-alpha0"
  Path = "/usr/libexec/kata-containers/kata-netmon"
  Debug = false
  Enable = false

```

Pennyzct:
Hi~ @grahamwhaley this is the performance report on AArch64. Since it is a pdf report, it isn't convenient to paste here. so I upload into my own repository [Pennyzct/metrics_report](https://github.com/Pennyzct/metrics_report/blob/master/metrics_report_aarch64.pdf), you could see from there.

Pennyzct:
Hi~ @grahamwhaley Hi~ finally knowing why a lot ci failed, getting error output from metrics CI:
`modprobe -i vhost_vsock failed: modprobe: FATAL: Module vhost_vsock not found in directory /lib/modules/4.4.0-134-generic
`

grahamwhaley:
Ah, good spot @Pennyzct - that might be as I think @jcvenegas added some extra debug in the fail case to the logs? Which machine (distro/arch) is that on? /cc @devimc

Pennyzct:
Besides the old kernel on metrics CI, vsock needing 4.8+, other failed because `modprobe -i vhost_vsock failed: modprobe: ERROR: could not insert 'vhost_vsock': Operation not permitted`. 
I think I should check if kernel version is more than 4.8 and also give this command root permission.

Pennyzct:
I'm thinking, does all failed CI run in VMware guest? 
Coping from [VSOCK.md](https://github.com/kata-containers/documentation/blob/b09a46909730600ede8a2c4cd34527ee6b1e7b55/design/VSocks.md), we need to
```
sudo systemctl stop vmware-tools
sudo modprobe -r vmw_vsock_vmci_transport
sudo modprobe -i vhost_vsock
```
if it's the reason, maybe we should check if it is `RunningOnVMM`, above workload is too much, we may just give a heads-up to users, let themselves do the work.

grahamwhaley:
Good question @Pennyzct . The other question that occurs to me is - is a failure to insert the module a fatal error? I think it probably should not be. So, in theory the `kata-runtime kata-env` in the CI scripts probably should have 'worked', even if it cannot insert the vsock module for whatever reason.

Pennyzct:
Hi~  @grahamwhaley my previous version is to omit the error, `if err != nil`, just use kataLog to convey the warning message. But a lot ci failed, `kata-runtime kate-env` outputted nothing.
And since I couldn't log into these failing CIs. So in this version, I tried to display the error.
and I also found the error code is `exit status 1`, is it a fatal error?

Pennyzct:
Hi~ @egernst modprobe works well rootless, testing it out on AArch64. In my latest version, when user use `kata-runtime kata-env`, if `SupportsVsocks` fails, I use `fmt.Printf("warning: check host vhost_vsock ability failed: %s", err)` to give use specific error info and let themselves decide what to do.

Pennyzct:
Hi~ @grahamwhaley 
metrics failed becuase dirty environment.
`fatal: destination path '/home/jenkins/workspace/workspace/kata-metrics-runtime-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests' already exists and is not an empty directory.
`
This error is weird to me, since we get `[ -d "${tests_repo_dir}" ] || git clone "https://${tests_repo}.git" "${tests_repo_dir}"`, if the directory exists, we shouldn't `git clone`.

other ci failed in the same spot, looks like a pattern to me. the creation of kubernetes master node failed.
```
NAME                     STATUS     ROLES    AGE   VERSION
ubuntu1804-azure618d50   NotReady   master   7s    v1.13.3
No resources found.
clusterrole.rbac.authorization.k8s.io/flannel created
clusterrolebinding.rbac.authorization.k8s.io/flannel created
serviceaccount/flannel created
configmap/kube-flannel-cfg created
daemonset.extensions/kube-flannel-ds-amd64 created
daemonset.extensions/kube-flannel-ds-arm64 created
daemonset.extensions/kube-flannel-ds-arm created
daemonset.extensions/kube-flannel-ds-ppc64le created
daemonset.extensions/kube-flannel-ds-s390x created
No resources found.
No resources found.
No resources found.
Makefile:91: recipe for target 'kubernetes' failed
make: *** [kubernetes] Error 1

```

grahamwhaley:
Hi @Pennyzct - metrics fail is indeed strange. There are a few reasons that should not happen (including CI should remove complete workspace between runs). I looked at the build slave history, and generally it is looking quite stable - so I have kicked off a rebuild of your job.

For the k8s fail - that is not one I know/recognise - @chavafg , a known k8s problem?, or sporadic we think?

grahamwhaley:
@egernst - I see you 👍 to @jodh-intel review of behavior - if you are OK with this PR now, can you ack it pls?

chavafg:
Seems like the k8s failures were produced by the issue we saw yesterday when pulling images... I'll restart the jobs

grahamwhaley:
re-span the metrics CI. Looks like it all ran, but the final checks hit some odd `yq` error. I'm going to guess sporadic and outside our control... let's try again...

Pennyzct:
Hi~ @grahamwhaley 
```
00:05:20 time="2019-04-18T16:05:20Z" level=warning msg="Failed to run [jq .\"boot-times\".Results | .[] | .\"to-workload\".Result /home/jenkins/workspace/workspace/kata-metrics-runtime-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests/metrics/results/boot-times.json][exit status 4]"
00:05:20 time="2019-04-18T16:05:20Z" level=warning msg="[/home/jenkins/workspace/workspace/kata-metrics-runtime-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests/metrics/results/boot-times.json][exit status 4]"
00:05:20 time="2019-04-18T16:05:20Z" level=warning msg="Failed to run [jq .\"memory-footprint\".Results | .[] | .average.Result /home/jenkins/workspace/workspace/kata-metrics-runtime-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests/metrics/results/memory-footprint.json][exit status 4]"
00:05:20 time="2019-04-18T16:05:20Z" level=warning msg="[/home/jenkins/workspace/workspace/kata-metrics-runtime-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests/metrics/results/memory-footprint.json][exit status 4]"
00:05:20 time="2019-04-18T16:05:20Z" level=warning msg="Failed to run [jq .\"memory-footprint-ksm\".Results | .[] | .average.Result /home/jenkins/workspace/workspace/kata-metrics-runtime-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests/metrics/results/memory-footprint-ksm.json][exit status 4]"
00:05:20 time="2019-04-18T16:05:20Z" level=warning msg="[/home/jenkins/workspace/workspace/kata-metrics-runtime-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests/metrics/results/memory-footprint-ksm.json][exit status 4]"
00:05:20 time="2019-04-18T16:05:20Z" level=warning msg="Overall we failed"
```
Looks like `jq` couldn't process properly. I will let you to decide whether to give it another run.;)

virtcontainers/utils/utils.go
```diff
@@ -232,17 +235,39 @@ func BuildSocketPath(elements ...string) (string, error) {
...
+	supportVsock, err := supportsVsocks()
+	if err != nil {
+		return false, err
 	}
 
-	if _, err := os.Stat(VHostVSockDevicePath); err != nil {
-		return false
+	// in case vhost_vsock module isn't built-in, but could be loaded
+	if !supportVsock {
+		if output, err := exec.Command("modprobe", "-i", VHostVsockModule).CombinedOutput(); err != nil {
```
devimc: IMO, we shouldn't trust in the envar `PATH`, the full path to `modprobe` should be used instead

Pennyzct: ok. I will use `/sbin/modprobe` to instead. ;)

Pennyzct: updated. plz review. ;)

Pennyzct:
Hi~ @grahamwhaley 
```
16:46:58 time="2019-04-23T08:46:58Z" level=warning msg="Failed to run [jq .\"boot-times\".Results | .[] | .\"to-workload\".Result /home/jenkins/workspace/workspace/kata-metrics-runtime-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests/metrics/results/boot-times.json][exit status 4]"
16:46:58 time="2019-04-23T08:46:58Z" level=warning msg="[/home/jenkins/workspace/workspace/kata-metrics-runtime-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests/metrics/results/boot-times.json][exit status 4]"
16:46:58 time="2019-04-23T08:46:58Z" level=warning msg="Failed to run [jq .\"memory-footprint\".Results | .[] | .average.Result /home/jenkins/workspace/workspace/kata-metrics-runtime-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests/metrics/results/memory-footprint.json][exit status 4]"
16:46:58 time="2019-04-23T08:46:58Z" level=warning msg="[/home/jenkins/workspace/workspace/kata-metrics-runtime-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests/metrics/results/memory-footprint.json][exit status 4]"
16:46:58 time="2019-04-23T08:46:58Z" level=warning msg="Failed to run [jq .\"memory-footprint-ksm\".Results | .[] | .average.Result /home/jenkins/workspace/workspace/kata-metrics-runtime-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests/metrics/results/memory-footprint-ksm.json][exit status 4]"
16:46:58 time="2019-04-23T08:46:58Z" level=warning msg="[/home/jenkins/workspace/workspace/kata-metrics-runtime-ubuntu-16-04-PR/go/src/github.com/kata-containers/tests/metrics/results/memory-footprint-ksm.json][exit status 4]"
16:46:58 time="2019-04-23T08:46:58Z" level=warning msg="Overall we failed"
```
same error three times. I'm afraid this not sporadic. Do you find this error anywhere besides here?

grahamwhaley:
Hi, Hi @Pennyzct . Aha, so, if you look in the .json files generated as part of the metrics tests (they are attached as artifacts to the Jenkins builds), then you find this in them:

```json
},
	"kata-env" :
	warning: check host vhost_vsock ability failed: modprobe -i vhost_vsock failed: modprobe: FATAL: Module vhost_vsock not found in directory /lib/modules/4.4.0-134-generic
{
```

which, is invalid json :-)

The big question then is I think, what should we do if the `modprobe` fails? I suspect we should silently fail, as some systems may not have `vsock`.
I'm open for some input from @egernst @jodh-intel here as well :-)

jodh-intel:
Agreed - `kata-env` should always show as much details as possible. Failing would stop that happening unless we are super-careful about how we handle failures (as we are in `kata-check`).

Speaking of which, tal at how `kata-env` displays `Host.VMContainerCapable` - it essentially calls `kata-check`. So we could modify `kata-env` to add a new `Host.VSockCapable` (or `Host.VHostCapable`, ...) option for example. If that shows as `false`, users could run `kata-check` to get the full details (and errors)?

jodh-intel:
Any update on this @Pennyzct?

Pennyzct:
Hi~ @jodh-intel I've been offline for attending Open Infrastructure Summit. I'll try to update asap.;)

Pennyzct:
Hi~ Sorry for the delayed update. ;) @grahamwhaley @egernst @devimc 
Thanks for the suggestion from @stefanha, quoting our discussion here, 
> This happens because /dev/vsock only exists after vhost_vsock.ko has been loaded. QEMU opens /dev/vhost-vsock and this causes vhost_vsock.ko to be loaded. Therefore SupportsVsocks() shouldn't call os.Stat(VSockDevicePath), just checking os.Stat(VHostVSockDevicePath) is enough.

> Hi~ @stefanha I'm Penny. ;). Correct me if I'm wrong, So as you say, even the vhost_vsock.ko hasn't been loaded at host firstly, we don't need to use`modprobe -i vhost_vsock` to load vhost_vsock.ko before using them, it will be automatically loaded when QEMU opens /dev/vhost-vosck.

> @Pennyzct Hi Penny! Yes, that's correct since Linux 4.13. Earlier Linux versions required manual loading of vhost_vsock.ko.

I updated the `SupportsVSocks` to only check `/dev/vhost_vsock` and I added the vhost_vsock.ko into acquired kernel modules in kata-check, just like @jodh-intel suggested, using `kata-check` to get the full details (and errors).

virtcontainers/utils/utils.go
```diff
@@ -234,9 +234,6 @@ func BuildSocketPath(elements ...string) (string, error) {
 
 // SupportsVsocks returns true if vsocks are supported, otherwise false
 func SupportsVsocks() bool {
-	if _, err := os.Stat(VSockDevicePath); err != nil {
```
jodh-intel: It looks like you can remove `VSockDevicePath` entirely now as it's only being used by the tests?

jodh-intel:
@chavafg, @mnaser - check out this sles 12 CI failure:

```
time="2019-05-10T08:40:21Z" level=error msg="ERROR: System is not capable of running Kata Containers" arch=amd64 name=kata-runtime pid=66647 source=runtime
```

jodh-intel:
@mnaser, @chavafg - and the same problem on Ubuntu 16.04:

```
09:25:16 time="2019-05-10T08:25:16Z" level=error msg="kernel property not found" arch=amd64 description="Host Support for Linux VM Sockets" name=vhost_vsock pid=15988 source=runtime type=module
09:25:16 time="2019-05-10T08:25:16Z" level=info msg="kernel property found" arch=amd64 description="Intel KVM" name=kvm_intel pid=15988 source=runtime type=module
09:25:16 time="2019-05-10T08:25:16Z" level=error msg="ERROR: System is not capable of running Kata Containers" arch=amd64 name=kata-runtime pid=15988 source=runtime
```

Pennyzct:
Hi~ @grahamwhaley @jodh-intel  For better understanding of the `modprobe -i module` failure, i tried to deliver more specific error info in this patch series.
And for `modprobe -i vhost_vsock`, i found that if the host is running on VMM, especially in a VMWare guest environment, the whole insertion procedure is a little bit complicated, so i also tried to give warning message to users to let them manually deal with this scenario.

virtcontainers/utils/utils.go
Pennyzct: yes, already updated. ;)

chavafg:
On sles 12, I see that the issue is:
```
time="2019-05-13T08:34:45Z" level=error msg="kernel property not found" arch=amd64 description="Host Support for Linux VM Sockets" name=vhost_vsock pid=64711 source=runtime type=module
```
maybe the sles kernel version does not support vsocks? 

I don't see the Ubuntu failure... Well, it failed but on the integration tests...

jodh-intel:
Looks like the sles CI has stalled but before it's been scheduled (so there is no restart link). All other CI's have passed. This code looks distro-agnostic to me so let's just merge to avoid having to re-crank every CI for that one issue...

Issue #1488, opened 2019-04-06
Kata metrics broken in Kubernetes v1.14.0 + Containerd 1.2.5 + kata-runtime shimv2 : 1.6.0


mcastelino:
# Description of problem

When running with latest Kubernetes, Containerd and Kata, Kata setup as shimv2

Kubernetes v1.14.0 + Containerd 1.2.5 + kata-runtime  : 1.6.0

```
[plugins]
  [plugins.cri]
   [plugins.cri.containerd]
     [plugins.cri.containerd.runtimes.kata]
        runtime_type = "io.containerd.kata.v2"
```

We get incorrect metrics reported (all metrics reported as 0). This means that scaling is also broken.

Early investigation shows that the metrics are being picked up purely from the container cgroups.
Moving QEMU vCPU threads into container cgroups shows metrics being reported properly.

This means that kubelet is not obtaining the metrics via cri, or that containerd is not obtaining the metrics from the kata-shim.

sboeuf:
@mcastelino A few comments there.

I only tried shim v2 and the ability to report metrics through CRI, with CRI-O, so I don't know what was the status for containers with 1.13.

Now, from your description of the issue here, it sounds like we need to identify if this is about the CRI call not being made or if something specific to containerd is broken.

FWIW, when I was testing with CRI-O, I did use the metrics-server to perform the autoscaling, and I was seeing the CRI calls going through CRI-O (I added extra logs for this in CRI-O).

Last thing, let's hear from @lifupan on this too, as he implemented kata-shimv2 basing his testing on containerd.

krsna1729:
> FWIW, when I was testing with CRI-O, I did use the metrics-server to perform the autoscaling, and I was seeing the CRI calls going through CRI-O (I added extra logs for this in CRI-O).

@sboeuf this is with patched kubelet correct? https://github.com/kubernetes/kubernetes/pull/73751

sboeuf:
@krsna1729 no I was changing the CRI socket name for this. Make sure it is not the default that is hardcoded in kubelet.

bergwolf:
ref: cAdvisor is brought in kubelet back by https://github.com/kubernetes/kubernetes/commit/84b0e82db4d485ac474e59a34e4eba878c755944

PR #1518, opened 2019-04-11
virtcontainers: prepend a kata specific string to host cgroups path

lifupan:
prepend a kata specific string to oci cgroup path to form a different cgroup path, thus cAdvisor couldn't find kata containers cgroup path on host to prevent it from grabbing the stats data.

Fixes:#1488

### DIFF
--- a/virtcontainers/api_test.go
+++ b/virtcontainers/api_test.go
@@ -863,6 +863,11 @@ func TestStatusSandboxSuccessfulStateReady(t *testing.T) {
 	cleanUp()
 
 	config := newTestSandboxConfigNoop()
+	cgroupPath, err := renameCgroupPath(utils.DefaultCgroupPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+
 	hypervisorConfig := HypervisorConfig{
 		KernelPath:        filepath.Join(testDir, testKernel),
 		ImagePath:         filepath.Join(testDir, testImage),
@@ -889,7 +894,7 @@ func TestStatusSandboxSuccessfulStateReady(t *testing.T) {
 				ID: containerID,
 				State: types.State{
 					State:      types.StateReady,
-					CgroupPath: utils.DefaultCgroupPath,
+					CgroupPath: cgroupPath,
 				},
 				PID:         0,
 				RootFs:      filepath.Join(testDir, testBundle),
@@ -922,6 +927,11 @@ func TestStatusSandboxSuccessfulStateRunning(t *testing.T) {
 	cleanUp()
 
 	config := newTestSandboxConfigNoop()
+	cgroupPath, err := renameCgroupPath(utils.DefaultCgroupPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+
 	hypervisorConfig := HypervisorConfig{
 		KernelPath:        filepath.Join(testDir, testKernel),
 		ImagePath:         filepath.Join(testDir, testImage),
@@ -948,7 +958,7 @@ func TestStatusSandboxSuccessfulStateRunning(t *testing.T) {
 				ID: containerID,
 				State: types.State{
 					State:      types.StateRunning,
-					CgroupPath: utils.DefaultCgroupPath,
+					CgroupPath: cgroupPath,
 				},
 				PID:         0,
 				RootFs:      filepath.Join(testDir, testBundle),
@@ -1734,7 +1744,12 @@ func TestStatusContainerStateReady(t *testing.T) {
 
 	// (homage to a great album! ;)
 	contID := "101"
+
 	config := newTestSandboxConfigNoop()
+	cgroupPath, err := renameCgroupPath(utils.DefaultCgroupPath)
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	ctx := context.Background()
 	p, err := CreateSandbox(ctx, config, nil)
@@ -1772,7 +1787,7 @@ func TestStatusContainerStateReady(t *testing.T) {
 		ID: contID,
 		State: types.State{
 			State:      types.StateReady,
-			CgroupPath: utils.DefaultCgroupPath,
+			CgroupPath: cgroupPath,
 		},
 		PID:         0,
 		RootFs:      filepath.Join(testDir, testBundle),
@@ -1800,7 +1815,12 @@ func TestStatusContainerStateRunning(t *testing.T) {
 
 	// (homage to a great album! ;)
 	contID := "101"
+
 	config := newTestSandboxConfigNoop()
+	cgroupPath, err := renameCgroupPath(utils.DefaultCgroupPath)
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	ctx := context.Background()
 	p, err := CreateSandbox(ctx, config, nil)
@@ -1848,7 +1868,7 @@ func TestStatusContainerStateRunning(t *testing.T) {
 		ID: contID,
 		State: types.State{
 			State:      types.StateRunning,
-			CgroupPath: utils.DefaultCgroupPath,
+			CgroupPath: cgroupPath,
 		},
 		PID:         0,
 		RootFs:      filepath.Join(testDir, testBundle),
--- a/virtcontainers/cgroups.go
+++ b/virtcontainers/cgroups.go
@@ -29,6 +29,12 @@ type cgroupPather interface {
 // where path is defined by the containers manager
 const cgroupKataPath = "/kata/"
 
+// prepend a kata specific string to oci cgroup path to
+// form a different cgroup path, thus cAdvisor couldn't
+// find kata containers cgroup path on host to prevent it
+// from grabbing the stats data.
+const cgroupKataPrefix = "kata"
+
 var cgroupsLoadFunc = cgroups.Load
 var cgroupsNewFunc = cgroups.New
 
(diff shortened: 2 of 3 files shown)


lifupan:
@egernst  If we accept the fix, we should change the test cases for cgroup test.

egernst:
I think this is the best option right now for fixing.   Let's get the tests updated!

Thanks @lifupan

devimc:
@mocolic  does this change works for you?  (I don't think so) :thinking:

egernst:
@devimc - can you clarify their assumptions on cgroups? How do they use our cgroups?

mocolic:
@devimc 
If this change doesn't honor the cgroup name we passed to kata, it will break the behavior, so won't work for us. Our resource governance implementation relies on honoring cgroup, as resource limits are set on cgroup level and expectation is that kata creates and activates a container inside this cgroup. 

 
However, recent tests have shown another problem with this check-in - https://github.com/kata-containers/runtime/pull/1189 
Please let me know if I should open a new issue for this and we can continue there.

I believe the problem is this code snippet in virtcontainers/cgroups.go - as this will govern containers even though they do not have constraints - which is not expected.

        // use a default constraint for sandboxes without cpu constraints
	if period == uint64(0) && quota == int64(0) {
		// set a quota and period equal to the default number of vcpus
		quota = int64(s.config.HypervisorConfig.NumVCPUs) * 100000
		period = 100000
	}

In our clusters default number of vcpus is -1 (entire machine) and as parent cgroup can’t have stricter limits than in child cgroup, container activation fails with this error:

"Could not update cgroup /fabric/SingleInstance_0_App1:myCpuServicePkg@3016e3dc-60b9-8746-a897-4f1bb73dd80a@0c665905-47af-8147-b616-8a0c099ca7a5/crio-5e4f661e9035b89d190a34bcec617a75ac0006b47b784dab207f02bda98736d1: write /sys/fs/cgroup/cpu/fabric/SingleInstance_0_App1:myCpuServicePkg@3016e3dc-60b9-8746-a897-4f1bb73dd80a@0c665905-47af-8147-b616-8a0c099ca7a5/crio-5e4f661e9035b89d190a34bcec617a75ac0006b47b784dab207f02bda98736d1/**cpu.cfs_quota_us: invalid argument**"  
Without the code snippet above, our resource governance would have worked out of the box.

devimc:
@mocolic 
> If this change doesn't honor the cgroup name we passed to kata, it will break the behavior, so won't work for us. Our resource governance implementation relies on honoring cgroup, as resource limits are set on cgroup level and expectation is that kata creates and activates a container inside this cgroup.

yes, with this change the cgroup path won't be honoured (again)

> However, recent tests have shown another problem with this check-in - #1189
Please let me know if I should open a new issue for this and we can continue there.

good finding, please file an issue

mocolic:
> @devimc 
> > However, recent tests have shown another problem with this check-in - #1189
> > Please let me know if I should open a new issue for this and we can continue there.
> 
> good finding, please file an issue

Thanks, opened https://github.com/kata-containers/runtime/issues/1521

egernst:
This forces a pretty tight coupling. As a data point; in Kubernetes, the node level part of the orchestrator/manager (kubelet) will create a cgroups under which it will place containers (in this specific instance, a pod cgroup). Kubelet is able to then go ahead and modify and restrict/enforce by modifying the cgroup that it created. 

In the other scenario (@mocolic), the upper layer relies on Kata creating container  cgroups which the upper layer can then go and modify. This makes a bit of a tight coupling, as can be seen in this scenario. 

I’m not sure what the right solution is yet, but the coupling isn’t ideal.

mocolic:
> In the other scenario (@mocolic), the upper layer relies on Kata creating container cgroups which the upper layer can then go and modify. This makes a bit of a tight coupling, as can be seen in this scenario.
> 
So, we create our cgroup and change limits on that level, we do not modify kata child cgroups. The only what we rely on is that container will be created in this cgroup. Also, all other container runtimes(docker/clear containers) work in this way.
On the other hand, with this solution, there is no way to create a container that is ungoverned, right? It is implicitly governed on the whole machine, even if the customer don't want that.

egernst:
can you provide a complete hierarchy example @mocolic?

It *sounds* like the name of the cgroup we create doesn't matter, so long as we place it in the correct place in the hierarchy, correct?

mocolic:
> can you provide a complete hierarchy example @mocolic?
> 
> It _sounds_ like the name of the cgroup we create doesn't matter, so long as we place it in the correct place in the hierarchy, correct?

Correct.
If we pass cgroup name to kata, for example _fabric_, the hierarchy should look like:
└── **fabric**
 ├── crio-89e60b84 # Sandbox cgroupsPath=/**fabric**/crio-89e60b84
 ├── crio-2a95e0f8 # Container A cgroupsPath=/**fabric**/crio-2a95e0f8
 └── crio-d037bce7 # Container B cgroupsPath=/**fabric**/crio-d037bce7

The hierarchy if parent cgroup is not provided, I expect something like this:
/sys/fs/cgroup/cpu,cpuacct
├── crio-89e60b84 # Sandbox cgroupsPath=/crio-89e60b84
├── crio-2a95e0f8 # Container A cgroupsPath=/crio-2a95e0f8
└── crio-d037bce7 # Container B cgroupsPath=/crio-d037bce7

One of the examples for the problem:
We set limits on _fabric_ cgroup (for example 1.3 cores which translates to cpu.cfs_quota_us = 130000). Let's say that machine has 2 cores (and as in our configuration default_vcpus=-1), the code snippet above will try to set limits of cpu.cfs_quota_us = 200000 on child cgroup fabric/crio-d037bce7. Linux doesn't allow this, so container activation fails.

devimc:
@mocolic 

> One of the examples for the problem:
We set limits on fabric cgroup (for example 1.3 cores which translates to cpu.cfs_quota_us = 130000). Let's say that machine has 2 cores (and as in our configuration default_vcpus=-1), the code snippet above will try to set limits of cpu.cfs_quota_us = 200000 on child cgroup fabric/crio-d037bce7. Linux doesn't allow this, so container activation fails.

then the constraint for the container should be -1 (no constraint) right?

mocolic:
> @mocolic
> 
> > One of the examples for the problem:
> > We set limits on fabric cgroup (for example 1.3 cores which translates to cpu.cfs_quota_us = 130000). Let's say that machine has 2 cores (and as in our configuration default_vcpus=-1), the code snippet above will try to set limits of cpu.cfs_quota_us = 200000 on child cgroup fabric/crio-d037bce7. Linux doesn't allow this, so container activation fails.
> 
> then the constraint for the container should be -1 (no constraint) right?

When I checked the configuration.toml file, it says this:
 #< 0                             --> will be set to the actual number of physical cores

If I understood correctly, this means that you would internally change this to the entire core machine (to 2), right? Also, when I created a container that requires whole machine (2 cores are set on fabric cgroup), the container activation was successful.

devimc:
@mocolic 

> If I understood correctly, this means that you would internally change this to the entire core machine (to 2), right? 

yes, currently it's now working because we try to set a constraint equal to the number of physical resources (in this case 2), but that won't be possible if the parent cgroup is smaller (for example 1.3), so we shouldn't apply a constraint

mcastelino:
@mocolic as Kata really does not (and cannot really) impose container level cgroups unless CPU Sets are used can you expand a little bit on what your container level cgroups is expected to have. We will place the kata-shim in this cgroup, but the actual workloads will be in what we will designate as the `pod` cgroup. So any stats or constraints you apply on the `container cgroup` will apply to the shim and not the actual workload. Do you place anything else in the container cgroup?

```
└── fabric
├── crio-89e60b84 # Sandbox cgroupsPath=/sys/fs/cgroup/cpu,cpuacct/fabric/crio-89e60b84
├── crio-2a95e0f8 # Container A cgroupsPath=/sys/fs/cgroup/cpu,cpuacct/fabric/crio-2a95e0f8
└── crio-d037bce7 # Container B cgroupsPath=/sys/fs/cgroup/cpu,cpuacct/fabric/crio-d037bce7
```

egernst:
@fupan approach looks good. Preprend would be better than append imo. Let’s fry the tests fixed.

lifupan:
> @fupan approach looks good. Preprend would be better than append imo. Let’s fry the tests fixed.

Got it, I'll change it to prepend and then change the testcases

lifupan:
The changes for tests is here: https://github.com/kata-containers/tests/pull/1462 
Let merge it first then trigger the tests for  this PR.

mcastelino:
@chavafg can we have a test case for this using something along the lines of a pod such as

```
apiVersion: v1
kind: Pod
metadata:
  name: test-metrics
spec:
  runtimeClassName: kata
  containers:
  - name: cont-2cpu-400m
    image: busybox
    resources:
      limits:
        cpu: 2
        memory: "400Mi"
    command: ["md5sum"]
    args: ["/dev/urandom"]
  - name: cont-3cpu-200m
    image: busybox
    resources:
      limits:
        cpu: 3
        memory: "200Mi"
    command: ["md5sum"]
    args: ["/dev/urandom"]   
```

And then check the actual results viakubectl get --raw "/apis/metrics.k8s.io/v1beta1/namespaces/default/pods/test-metrics"

which should yeild pretty high CPU utilization

along the lines of 

```
  "containers": [
    {
      "name": "cont-3cpu-200m",
      "usage": {
        "cpu": "398739040n",
        "memory": "744Ki"
      }
    },
    {
      "name": "cont-2cpu-400m",
      "usage": {
        "cpu": "200848120n",
        "memory": "852Ki"
      }
    }
```

PR #1528, opened 2019-04-12
shimv2 should return grpc error codes

bergwolf:
containerd relies on it to determine proper actions, e.g., fail the operation or wait for an exit event.[1]

ref: [1] https://github.com/containerd/cri/blob/master/pkg/server/container_stop.go#L141


devimc:
cc @lifupan

PR #1530, opened 2019-04-12
virtcontainers: inherit parent's CPU constraint

### LINKED ISSUE #1521: Resource governance doesn't work with kata 1.6.0
Recent tests have shown a problem with this check-in - #1189 where containers are not activated at all if parent cgroup with resource limits is passed to kata. When test was changed in a way that container is requesting the whole machine, this container activated (as parent and child cgroup will have the same limits).

In our clusters default number of vcpus is -1 (entire machine) and as parent cgroup can’t have stricter limits than the child cgroup, container activation fails with this error:

"Could not update cgroup /fabric/SingleInstance_0_App1:myCpuServicePkg@3016e3dc-60b9-8746-a897-4f1bb73dd80a@0c665905-47af-8147-b616-8a0c099ca7a5/crio-5e4f661e9035b89d190a34bcec617a75ac0006b47b784dab207f02bda98736d1: write /sys/fs/cgroup/cpu/fabric/SingleInstance_0_App1:myCpuServicePkg@3016e3dc-60b9-8746-a897-4f1bb73dd80a@0c665905-47af-8147-b616-8a0c099ca7a5/crio-5e4f661e9035b89d190a34bcec617a75ac0006b47b784dab207f02bda98736d1/cpu.cfs_quota_us: invalid argument"

I believe the problem is this code snippet in virtcontainers/cgroups.go - as this will govern containers even though they do not have constraints - which is not expected.

        // use a default constraint for sandboxes without cpu constraints
	if period == uint64(0) && quota == int64(0) {
		// set a quota and period equal to the default number of vcpus
		quota = int64(s.config.HypervisorConfig.NumVCPUs) * 100000
		period = 100000
	}

Without the code snippet above, our resource governance would have worked out of the box.

# Expected result

Container is activated and governed (restricted by limits on parent cgroup level that is passed to kata).

# Actual result

Container is not activated at all - the error received:

CreateContainer gRPC-error: 2:   container create failed: Could not update cgroup   /fabric/SingleInstance_0_App1:myCpuServicePkg@3016e3dc-60b9-8746-a897-4f1bb73dd80a@0c665905-47af-8147-b616-8a0c099ca7a5/crio-5e4f661e9035b89d190a34bcec617a75ac0006b47b784dab207f02bda98736d1:   write   /sys/fs/cgroup/cpu/fabric/SingleInstance_0_App1:myCpuServicePkg@3016e3dc-60b9-8746-a897-4f1bb73dd80a@0c665905-47af-8147-b616-8a0c099ca7a5/crio-5e4f661e9035b89d190a34bcec617a75ac0006b47b784dab207f02bda98736d1/cpu.cfs_quota_us:   invalid argument

devimc:
Don't set a default CPU constraint if period and quota are not specified, that
way the container will inherit the CPU constraint from its parent. Container
creation won't fail if the parent CPU constraint is smaller than the default
number of vCPUs.

fixes #1521

### DIFF
--- a/virtcontainers/cgroups.go
+++ b/virtcontainers/cgroups.go
@@ -266,6 +266,8 @@ func (s *Sandbox) resources() (specs.LinuxResources, error) {
 }
 
 func (s *Sandbox) cpuResources() *specs.LinuxCPU {
+	// Use default period and quota if they are not specified.
+	// Container will inherit the constraints from its parent.
 	quota := int64(0)
 	period := uint64(0)
 	shares := uint64(0)
@@ -321,13 +323,6 @@ func (s *Sandbox) cpuResources() *specs.LinuxCPU {
 	}
 
 	cpu.Cpus = strings.Trim(cpu.Cpus, " \n\t,")
-
-	// use a default constraint for sandboxes without cpu constraints
-	if period == uint64(0) && quota == int64(0) {
-		// set a quota and period equal to the default number of vcpus
-		quota = int64(s.config.HypervisorConfig.NumVCPUs) * 100000
-		period = 100000
-	}
 
 	return validCPUResources(cpu)
 }


jcvenegas:
@egernst @sboeuf  @bergwolf  please take a look ready to merge

devimc:
cc @mocolic

gkhanna79:
@teawater Can you please also backport this to 1.6 stable since it has also been back ported to 1.5?

CC @mocolic

gkhanna79:
@egernst Can you please help with this as well?

gkhanna79:
@egernst @teawater Can you please help with port of this fix to stable 1.6? Service Fabric resource governance with Kata is broken without this change.

PR #1532, opened 2019-04-12
stable 1.6 backport: netmon: Fix bug in how routes are converted

### LINKED ISSUE #1523: netmon: Retain dest CIDR when passing routes to the runtime
Netmon process incorrectly converts the dest today to an IP address dropping the subnet information. The agent expects a CIDR as the route dest. As a result, the netmon process receives an error and exits. This bug is manifested today only with `tcfilter` mode, as macvtap passes empty routes to the agent as we move the IP address from the veth to the tap within the VM.

amshinde:
The agent expects a IP CIDR for the route destination
rather than an IP address. netmon was incorrectly
converting route dest to an IP address and hence
exiting with an error.

We did not have an integration test for netmon with tcfilter mode.
macvtap mode did not uncover this, as with macvtap routes are
not really passed to the agent.
We delete the IP on the veth device, and netmon looks at the
routes after the IP is deleted with macvtap.

Fixes #1523

### DIFF
--- a/netmon/netmon.go
+++ b/netmon/netmon.go
@@ -309,7 +309,7 @@ func convertRoutes(netRoutes []netlink.Route) []vcTypes.Route {
 		dst := ""
 		if netRoute.Dst != nil {
 			if netRoute.Dst.IP.To4() != nil {
-				dst = netRoute.Dst.IP.String()
+				dst = netRoute.Dst.String()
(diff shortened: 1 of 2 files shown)


amshinde:
@egernst @jcvenegas Can we merge this?

Issue #1514, opened 2019-04-10
factory init can run many times if vm factory is enabled


teawater:
```
root@teawater-virtual-machine:~# kata-runtime factory init
vm factory initialized
root@teawater-virtual-machine:~# kata-runtime factory init
vm factory initialized
root@teawater-virtual-machine:~# kata-runtime factory init
vm factory initialized
root@teawater-virtual-machine:~# mount | grep template
tmpfs on /run/vc/vm/template type tmpfs (rw,nosuid,nodev,relatime,size=532480k)
tmpfs on /run/vc/vm/template type tmpfs (rw,nosuid,nodev,relatime,size=532480k)
tmpfs on /run/vc/vm/template type tmpfs (rw,nosuid,nodev,relatime,size=532480k)
```

teawater:
And need its issue for tests

PR #1547, opened 2019-04-17
factory: Check if vm templateing is on before vm templateing initialize

teawater:
Check if vm templateing is on before vm templateing initialize.

Fixes: #1514

### DIFF
--- a/virtcontainers/factory/template/template.go
+++ b/virtcontainers/factory/template/template.go
@@ -46,7 +46,12 @@ func New(ctx context.Context, config vc.VMConfig) (base.FactoryBase, error) {
 	statePath := store.RunVMStoragePath + "/template"
 	t := &template{statePath, config}
 
-	err := t.prepareTemplateFiles()
+	err := t.checkTemplateVM()
+	if err == nil {
+		return nil, fmt.Errorf("There is already a VM template in %s", statePath)
+	}
+
+	err = t.prepareTemplateFiles()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cli/factory.go
```diff
@@ -206,7 +206,12 @@ var initFactoryCommand = cli.Command{
 
 		if runtimeConfig.FactoryConfig.Template {
 			kataLog.WithField("factory", factoryConfig).Info("create vm factory")
-			_, err := vf.NewFactory(ctx, factoryConfig, false)
+			_, err := vf.NewFactory(ctx, factoryConfig, true)
+			if err == nil {
+				err = fmt.Errorf("vm factory is on")
```
jodh-intel: Nit: This can use `errors.New()`.

cli/factory.go
```diff
@@ -206,7 +206,12 @@ var initFactoryCommand = cli.Command{
...
+			} else {
+				_, err = vf.NewFactory(ctx, factoryConfig, false)
```
jodh-intel: Calling `NewFunction()` twice looks odd - can't we just rely on your new test rather than doing this?

- https://github.com/kata-containers/tests/pull/1481/files

amshinde: Agree, why not expose the init functionality for vm-factory  as a separate function and call that instead.

teawater: Pushed a new commit that move check to template.go:New.

Issue #1564, opened 2019-04-18
Fix markdown issues


jodh-intel:
Resolve issues with the markdown doc files in this repo.

PR #1565, opened 2019-04-18
Fix markdown

jodh-intel:
Fix various issues with the markdown files in this repo.

Fixes: #1564.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,11 +68,7 @@ $ kata-runtime kata-check
 > **Note:**
 >
 > If you run the previous command as the `root` user, further checks will be
-> performed (e.g. it will check if another incompatible hypervisor is running):
->
-> ```bash
-> $ sudo kata-runtime kata-check
-> ```
+> performed (e.g. it will check if another incompatible hypervisor is running).
 
 ## Download and install
 
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ $ kata-runtime --kata-config=/some/where/configuration.toml ...
 ```
 
 The runtime will log the full path to the configuration file it is using. See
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -68,11 +68,7 @@ $ kata-runtime kata-check
 > **Note:**
 >
 > If you run the previous command as the `root` user, further checks will be
-> performed (e.g. it will check if another incompatible hypervisor is running):
->
-> ```bash
```
grahamwhaley: Looked OK on my github page render?

README.md
```diff
@@ -120,7 +120,7 @@ $ kata-runtime --kata-config=/some/where/configuration.toml ...
 ```
 
 The runtime will log the full path to the configuration file it is using. See
```
grahamwhaley: grammar in commit - `must be allow`

jodh-intel: Fixed.

README.md
jodh-intel: See https://github.com/kata-containers/kata-containers/issues/40. We need a way to check the markdown that doesn't involve manually viewing on github and checking every link every time a doc changes. I appreciate that it might render fine in GitHub, but all the parsers I've found fail to handle this. Therefore, by making this change (a tiny sacrifice), we gain the ability to validate any future changes to this doc and all others.

grahamwhaley: I guess that depends if we are using 'plain markdown' or [github flavour markdown](https://github.github.com/gfm/) and which tools we intend to verify with?
pandoc has a github markdown type, but I don't know how accurate it is.
If we wanted to use the same tools as github itself, I think the code lives over at https://github.com/github/markup (it has been a little while since I had a look)

PR #1575, opened 2019-04-23
newstore:  removing deprecated files when use new store driver

### LINKED ISSUE #803: [RFC] baseline persist data for live-upgrade

WeiZhang555:
Fixes #803

It contains several commits:

1. Simplify new store API to make the code easier to understand and use.
2. newstore: remove file "devices.json" 

### DIFF
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -591,8 +591,8 @@ func statusContainer(sandbox *Sandbox, containerID string) (ContainerStatus, err
 
 				if !running {
 					virtLog.WithFields(logrus.Fields{
-						"state":       container.state.State,
-						"process pid": container.process.Pid}).
+						"state": container.state.State,
+						"pid":   container.process.Pid}).
 						Info("container isn't running")
 					if err := container.stop(); err != nil {
 						return ContainerStatus{}, err
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -590,6 +590,10 @@ func statusContainer(sandbox *Sandbox, containerID string) (ContainerStatus, err
 				}
 
 				if !running {
+					virtLog.WithFields(logrus.Fields{
+						"state":       container.state.State,
+						"process pid": container.process.Pid}).
+						Info("container isn't running")
 					if err := container.stop(); err != nil {
 						return ContainerStatus{}, err
 					}
--- a/virtcontainers/device/drivers/block.go
+++ b/virtcontainers/device/drivers/block.go
@@ -1,5 +1,5 @@
 // Copyright (c) 2017-2018 Intel Corporation
-// Copyright (c) 2018 Huawei Corporation
+// Copyright (c) 2018-2019 Huawei Corporation
 //
 // SPDX-License-Identifier: Apache-2.0
 //
@@ -147,9 +147,9 @@ func (device *BlockDevice) GetDeviceInfo() interface{} {
 	return device.BlockDrive
 }
 
-// Dump convert and return data in persist format
-func (device *BlockDevice) Dump() persistapi.DeviceState {
-	ds := device.GenericDevice.Dump()
+// Save converts Device to DeviceState
+func (device *BlockDevice) Save() persistapi.DeviceState {
+	ds := device.GenericDevice.Save()
 	ds.Type = string(device.DeviceType())
 
 	drive := device.BlockDrive
@@ -167,6 +167,28 @@ func (device *BlockDevice) Dump() persistapi.DeviceState {
 		}
 	}
 	return ds
+}
+
+// Load loads DeviceState and converts it to specific device
+func (device *BlockDevice) Load(ds persistapi.DeviceState) {
+	device.GenericDevice = &GenericDevice{}
+	device.GenericDevice.Load(ds)
+
+	bd := ds.BlockDrive
+	if bd == nil {
+		return
+	}
+	device.BlockDrive = &config.BlockDrive{
+		File:     bd.File,
+		Format:   bd.Format,
+		ID:       bd.ID,
+		Index:    bd.Index,
+		MmioAddr: bd.MmioAddr,
+		PCIAddr:  bd.PCIAddr,
+		SCSIAddr: bd.SCSIAddr,
+		NvdimmID: bd.NvdimmID,
+		VirtPath: bd.VirtPath,
+	}
 }
 
 // It should implement GetAttachCount() and DeviceID() as api.Device implementation
--- a/virtcontainers/device/manager/manager.go
+++ b/virtcontainers/device/manager/manager.go
@@ -16,6 +16,7 @@ import (
 	"github.com/kata-containers/runtime/virtcontainers/device/api"
 	"github.com/kata-containers/runtime/virtcontainers/device/config"
 	"github.com/kata-containers/runtime/virtcontainers/device/drivers"
+	persistapi "github.com/kata-containers/runtime/virtcontainers/persist/api"
 	"github.com/kata-containers/runtime/virtcontainers/utils"
 )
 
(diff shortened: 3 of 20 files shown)


WeiZhang555:
Second commit is only for test purpose which should be removed before merging.
After we enabled experimental Jenkins job, we don't need the test commit any more.

ping CI team @kata-containers/ci : 

https://github.com/kata-containers/runtime/pull/883#issuecomment-485056458

virtcontainers/api.go
```diff
@@ -590,6 +590,10 @@ func statusContainer(sandbox *Sandbox, containerID string) (ContainerStatus, err
 				}
 
 				if !running {
+					virtLog.WithFields(logrus.Fields{
+						"state": container.state.State,
+						"Pid":   container.process.Pid}).
```
jodh-intel: Nit: I'd make the field name lowercase: `"pid"`.

virtcontainers/device/drivers/block.go
```diff
@@ -169,5 +169,26 @@ func (device *BlockDevice) Dump() persistapi.DeviceState {
 	return ds
 }
 
+// Load loads DeviceState and converts it to specific device
+func (device *BlockDevice) Load(ds persistapi.DeviceState) {
+	device.GenericDevice = &GenericDevice{}
+	device.GenericDevice.Load(ds)
+
+	bd := ds.BlockDrive
+	if bd != nil {
```
jodh-intel: Nit: I tend to invert logic like this to reduce the indent and make the code as simple as possible:

```go

if bd == nil {
    return
}

device.BlockDrive = &config.BlockDrive{
      :
```

Same comment for some of the checks below...

virtcontainers/device/manager/manager.go
```diff
@@ -229,3 +230,39 @@ func (dm *deviceManager) IsDeviceAttached(id string) bool {
...
+			dm.devices[dev.DeviceID()] = dev
+		case config.VhostUserBlk:
+			dev := &drivers.VhostUserBlkDevice{}
+			dev.Load(ds)
+			dm.devices[dev.DeviceID()] = dev
+		case config.VhostUserNet:
+			dev := &drivers.VhostUserNetDevice{}
+			dev.Load(ds)
+			dm.devices[dev.DeviceID()] = dev
+		}
```
jodh-intel: How about adding a `default` case and logging a warning/error that the type cannot be handled?

netmon/netmon.go
```diff
@@ -567,26 +567,26 @@ func (n *netmon) handleRTMDelRoute(ev netlink.RouteUpdate) error {
 }
 
 func (n *netmon) handleLinkEvent(ev netlink.LinkUpdate) error {
-	n.logger().Debug("handleLinkEvent: netlink event received")
+	n.logger().Info("handleLinkEvent: netlink event received")
 
 	switch ev.Header.Type {
 	case unix.NLMSG_DONE:
-		n.logger().Debug("NLMSG_DONE")
+		n.logger().Info("NLMSG_DONE")
```
jodh-intel: How many of these events do we normally get? I'm wondering if `Info` might end up DoS'ing the logs so best to leave at `Debug` level?

virtcontainers/persist.go
```diff
@@ -61,7 +67,7 @@ func deviceToDeviceState(devices []api.Device) (dss []persistapi.DeviceState) {
 	return
```
gabibeyer: Although not really part of this PR, you could probably get rid of this return if you feel like it goes with some of the other clean up.

netmon/netmon.go
WeiZhang555: Yeah, this is only for debug purpose, I ran into some wired problems in CI and I don't have a good environment to reproduce, still struggling...

These changes will be removed in final version. Thanks :-)

virtcontainers/device/drivers/block.go
WeiZhang555: Oh, nice suggestion!

virtcontainers/persist.go
WeiZhang555: Right 👍

virtcontainers/api.go
WeiZhang555: done

WeiZhang555:
ping @chavafg  : Can you enable "experimental" in `cri-containerd` job so that we can test experimental features with K8s test cases? I think cri-containerd+Fedora can cover almost all cases for experimental features.

I can't find how to set the "KATA_EXPERIMENTAL_FEATURES" flag in our Jenkins Jobs, maybe that's because I don't have right permission.

WeiZhang555:
/test

@jodh-intel @gabibeyer  I fixed the CI now, so please take another look at this. Thanks :smile:

chavafg:
Hi @WeiZhang555, I've just enabled it in the cri-containerd job and also provided you with correct permissions on Jenkins.

WeiZhang555:
Thanks so much! @chavafg

WeiZhang555:
ping @kata-containers/runtime

virtcontainers/device/manager/manager.go
```diff
@@ -229,3 +230,41 @@ func (dm *deviceManager) IsDeviceAttached(id string) bool {
...
+
+	for _, ds := range devStates {
+		switch config.DeviceType(ds.Type) {
+		case config.DeviceGeneric:
+			dev := &drivers.GenericDevice{}
+			dev.Load(ds)
+			dm.devices[dev.DeviceID()] = dev
+		case config.DeviceBlock:
+			dev := &drivers.BlockDevice{}
+			dev.Load(ds)
```
jodh-intel: Nit: It would be great if we could avoid all the `dev.Load()` and `dm.devices[dev.DeviceID()]` duplication somehow  (an interface?) so that those calls could be made once, after the `switch`

virtcontainers/persist.go
```diff
@@ -90,59 +96,53 @@ func (s *Sandbox) dumpDevices(ss *persistapi.SandboxState, cs map[string]persist
...
+	return nil
 }
 
-func (s *Sandbox) getSbxAndCntStates() (*persistapi.SandboxState, map[string]persistapi.ContainerState, error) {
-	if err := s.newStore.FromDisk(s.id); err != nil {
-		return nil, nil, err
-	}
+func (s *Sandbox) loadState(ss persistapi.SandboxState) {
+	s.state.PersistVersion = ss.PersistVersion
+	s.state.GuestMemoryBlockSizeMB = ss.GuestMemoryBlockSizeMB
```
jodh-intel: The only element of `SandboxState` you are not setting here is `GuestMemoryHotplugProbe`. Is there any reason for that? Might be worth a comment explaining why if so.

virtcontainers/api.go
```diff
@@ -590,6 +590,10 @@ func statusContainer(sandbox *Sandbox, containerID string) (ContainerStatus, err
 				}
 
 				if !running {
+					virtLog.WithFields(logrus.Fields{
+						"state":       container.state.State,
+						"process pid": container.process.Pid}).
```
jodh-intel: Nit: By (admittedly undocumented) convention we try to ensure all log field names do not contain spaces, so could you change this to `process-id` (or even just `pid` as I think that's a well-understood term).

virtcontainers/persist.go
WeiZhang555: Nice catch! This is a new field recently added, I will add save/load logic here.

virtcontainers/device/manager/manager.go
WeiZhang555: Sounds good.

virtcontainers/api.go
WeiZhang555: I'll use the `process-pid` since we already have a `pid` field in `virtLog`, so if I use `pid` here that will confuse developers.

WeiZhang555:
/test

Thanks @jodh-intel ! Your comments are addressed.

virtcontainers/persist.go
jodh-intel: Thanks @WeiZhang555.

Thinking about this, I wonder if there is a way we could use reflection to ensure we always copy all elements in all structs? I think it would be possible as you could iterate through `ss` and then match the corresponding element in `s` based on name and type. It would complicate the code though.

Alternatively, maybe you could create a unit test that:

- Sets `ss` so that all elements have non-zero/non-`nil` values.
- Call `s.loadState()` with a new `Sandbox`.
- Iterate over all the fields in `s.state` and ensure they are all set, and set to the values in `ss`.

gabibeyer:
lgtm!

teawater:
jenkins-ci-centos-7-4-q-35 fail:
```
--- FAIL: TestHostNetworkingRequested (0.01s)
    assertions.go:239: 
                          
	Error Trace:	network_test.go:99
        
	Error:      	Should be true
```
This is a known issue https://github.com/kata-containers/runtime/issues/1604

WeiZhang555:
Thanks @teawater !

Issue #1598, opened 2019-04-28
runtime doesn't work OK with Firecracker v0.15.2


teawater:
https://github.com/kata-containers/runtime/pull/1560

PR #1265, opened 2019-02-21
versions: firecracker: update to latest

grahamwhaley:
Update our firecracker from v0.12.0 to v0.14.0

Fixes: #1252

Signed-off-by: Graham Whaley <graham.whaley@intel.com>

PR #1560, opened 2019-04-17
Firecracker: update to version v0.15.2

egernst:
Fixes: #1559

Signed-off-by: Eric Ernst <eric.ernst@intel.com>

PR #1614, opened 2019-05-01
firecracker: Add support for v0.15.x

mcastelino:
Add support for v0.15.x. Change the drive naming scheme to match
the requirement of v0.15.x

Fixes: #1598

### DIFF
--- a/virtcontainers/fc.go
+++ b/virtcontainers/fc.go
@@ -390,12 +390,16 @@ func (fc *firecracker) startSandbox(timeout int) error {
 	return fc.waitVMM(timeout)
 }
 
+func fcDriveIndexToID(i int) string {
+	return "drive_" + strconv.Itoa(i)
+}
+
 func (fc *firecracker) createDiskPool() error {
 	span, _ := fc.trace("createDiskPool")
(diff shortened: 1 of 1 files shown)


mcastelino:
Once this is merged we can also merge https://github.com/kata-containers/runtime/pull/1265 and/or https://github.com/kata-containers/runtime/pull/1560

mcastelino:
Should also fix https://github.com/kata-containers/runtime/pull/1265

virtcontainers/fc.go
```diff
@@ -395,7 +395,7 @@ func (fc *firecracker) createDiskPool() error {
 	defer span.Finish()
 
 	for i := 0; i < fcDiskPoolSize; i++ {
-		driveID := "drive-" + strconv.Itoa(i)
+		driveID := "drive_" + strconv.Itoa(i)
```
jodh-intel: A tiny change I know, but if we think this could change again, maybe a `func fcDriveIndexToID(i int) string` might be a good idea?

mcastelino: done

Issue #1604, opened 2019-04-29
namespace test failing with q35 on centos


jodh-intel:
CentOS q35 CI failing:

```
=== RUN   TestHostNetworkingRequested
--- FAIL: TestHostNetworkingRequested (0.01s)
    assertions.go:239: 
                          
	Error Trace:	network_test.go:99
        
	Error:      	Should be true
```

See https://github.com/kata-containers/runtime/pull/1587.

jodh-intel:
Also seen here: https://github.com/kata-containers/runtime/pull/1599#event-2306049751.

lifupan:
Hi @jodh-intel, 

I found the root cause is that the kernel in CentOS q35 is 3.10 which doesn't support the "nsfs",
and the "nsfs" isn't supported until  3.19 version. 
I'm curious when did we add the centos q35 test? Why this failed didn't appear until recently?

jodh-intel:
Thanks @lifupan.

 The Jenkins q35 jobs were added on https://github.com/kata-containers/ci/pull/144, but I'm not sure if the date on that PR is the same as the time we actually started using them. Maybe @GabyCT can provide further info?

Rather than ignoring on centos, I think this is another use-case for https://github.com/kata-containers/runtime/pull/1587. I'll add a new call to allow skipping a test if the kernel version is too old...

GabyCT:
@jodh-intel and @lifupan , yes the day that that PR landed the CI started running for PR, the master jenkins jobs for Q35 started working before that PR as I tested that the jobs were stable

jodh-intel:
This is a duplicate of #1517.

PR #1623, opened 2019-05-03
mounts: fix isSystemMount check for mountSharedDirMounts

### LINKED ISSUE #1591: Certain Dockerfile volumes preventing pod start
We've found certain volume paths defined in the Dockerfile prevent the Pod from starting with Kata. One such example is `/sys/fs/cgroup`.

The following Dockerfile can be used to replicate:
```Dockerfile
FROM ubuntu

VOLUME /sys/fs/cgroup
```

We receive the following errors:
```
Error: failed to create containerd task: Could not run process: container_linux.go:348: starting container process caused "process_linux.go:402: container init caused \"rootfs_linux.go:58: mounting \\\"/var/lib/containerd/io.containerd.grpc.v1.cri/containers/sleep/volumes/a45869c2ff90df3649c035b429e23e8b935804e90cd5cfd997778ff0d64d693c\\\" to rootfs \\\"/run/kata-containers/shared/containers/sleep\\\" at \\\"/sys/fs/cgroup\\\" caused \\\"stat /var/lib/containerd/io.containerd.grpc.v1.cri/containers/sleep/volumes/a45869c2ff90df3649c035b429e23e8b935804e90cd5cfd997778ff0d64d693c: no such file or directory\\\"\"": unknown
```

Unfortunately we do not have control over the contents of the Dockerfiles/images and need an elegant solution for handling this in Kata.

The following paths have been found to prevent the pods from starting:
 - `/sys/fs/cgroup`

(I will add more as I find them)

I think this issue is similar to https://github.com/kata-containers/runtime/issues/1561 as both are related to the volumes defined in Docker images/OCI image specs.

awprice:
This change updates the isSystemMount check for mountSharedDirMounts
when setting up shared directory mounts for the container and uses
the source of the mount instead of the destination for the check.

We want to exclude system mounts from the host side as they
shouldn't be mounted into the container.

We do however want to allow system mounts within the
container as denying them can prevent some containers from
running properly.

Fixes #1591

Signed-off-by: Alex Price <aprice@atlassian.com>

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -515,10 +515,11 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
 	var sharedDirMounts []Mount
 	var ignoredMounts []Mount
 	for idx, m := range c.mounts {
-		if isSystemMount(m.Destination) {
-			if !(IsDockerVolume(m.Source) || Isk8sHostEmptyDir(m.Source)) {
-				continue
-			}
+		// Skip mounting certain system paths from the source on the host side
+		// into the container as it does not make sense to do so.
+		// Example sources could be /sys/fs/cgroup etc.
+		if isSystemMount(m.Source) {
+			continue
 		}
 
 		if m.Type != "bind" {


virtcontainers/container.go
```diff
@@ -515,10 +515,11 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
...
 	var ignoredMounts []Mount
 	for idx, m := range c.mounts {
-		if isSystemMount(m.Destination) {
-			if !(IsDockerVolume(m.Source) || Isk8sHostEmptyDir(m.Source)) {
-				continue
-			}
+		// Skip mounting certain system paths from the source on the host side
+		// into the container as it does not make sense to do so.
+		// Example sources could be /sys/fs/cgroup etc.
+		if isSystemMount(m.Source) {
```
amshinde: @awprice Your comment makes sense, we do not mount certain system paths from the host side into the guest, however it makes sense to do so for system paths that are docker volumes and k8s empty dirs.

awprice: We are using containerd, maybe we need to update the logic to handle docker volumes, k8s empty dirs and containerd volumes?

I don't think containerd volumes have been factored in here and that's probably why we are running into this issue, as we don't use docker.

jodh-intel:
Only the single centos CI failure which is known (https://github.com/kata-containers/runtime/issues/1604), so merging...

Issue #1602, opened 2019-04-29
some question about kata shim v1/v2


Ace-Tang:
# Description of problem

Here are some question I see in my test, not sure should be fixed

# shim v1 interface
 use shim v1 interface,  after kill a task use ctr, container state pid should be -1
```
# ctr run -d docker.io/library/busybox:latest kk1 top
# kata-runtime list | grep kk1
kk1                                                                4992        running     /home/t4/pouch/containerd/state/io.containerd.runtime.v1.linux/default/kk1                                                                2019-04-28T15:01:51.342150578Z   #0
ctr t kill kk1
[root@daishu-test ~]# ./kata-runtime list | grep kk1
kk1                                                                4992        stopped     /home/t4/pouch/containerd/state/io.containerd.runtime.v1.linux/default/kk1                                                                2019-04-28T15:01:51.342150578Z   #1184 
```

# shim v1 interface
With shim v2,  after called kill , container status not changed, but task alreay stop, but kata-runtime list shows the container is still running
```
# ctr run --runtime kata.v2 -d docker.io/library/busybox:latest v2-k2 top
# ctr t ls | grep v2-k2
v2-k2                                                               5412     RUNNING
# ctr t kill v2-k2
# ctr t ls | grep v2-k2
v2-k2                                                               5412     STOPPED
# kata-runtime list | grep v2-k2
v2-k2                                                              -1          running     /home/t4/pouch/containerd/state/io.containerd.runtime.v2.task/default/v2-k2                                                               2019-04-28T15:20:46.735636968Z   #0

```

containerd-shim-v2 not set container status to stop when container exit in wait() goroutine, and in Cleanup interface, it will trigger one more signal to process, I think it can be avoid
```
    if oci.StateToOCIState(status.State) != oci.StateStopped {
        err := sandbox.KillContainer(cid, syscall.SIGKILL, true)
        if err != nil {
            logrus.WithError(err).WithField("container", cid).Warn("failed to kill container")
            return err 
        }   
    }   
```
The logic will not cause error but I think it is not correct

lifupan:
@Ace-Tang  Yes, I think we can improve it by moving the stop container logic into wait section, in which to stop the specific container once shimv2 monitored the process exited.

lifupan:
Hi @Ace-Tang , I think this issue can be fixed by pr https://github.com/kata-containers/runtime/pull/1723, can you have a test and close this issue?

lifupan:
[...]

PR #1625, opened 2019-05-05
shimv2: update container status when process exited

lifupan:
When a container process terminated, shimv2 should update
container's status internally, otherwise, ctr command would
get a wrong container status even the process had exited.

Fixes:#1602

### DIFF
--- a/containerd-shim-v2/service.go
+++ b/containerd-shim-v2/service.go
@@ -665,6 +665,18 @@ func (s *service) Kill(ctx context.Context, r *taskAPI.KillRequest) (_ *ptypes.E
 		return nil, err
 	}
 
+	// If the container has stopped, return directly when killed with
+	// SIGKILL or SIGTERM signal.
+	if signum == syscall.SIGKILL || signum == syscall.SIGTERM {
+		status, err := s.sandbox.StatusContainer(c.id)
+		if err != nil {
+			return nil, errors.Wrap(err, "kill container")
+		}
+		if status.State.State == types.StateStopped {
+			return empty, nil
+		}
+	}
+
 	processID := c.id
 	if r.ExecID != "" {
 		execs, err := c.getExec(r.ExecID)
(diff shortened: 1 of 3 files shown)


containerd-shim-v2/service.go
```diff
@@ -665,6 +665,18 @@ func (s *service) Kill(ctx context.Context, r *taskAPI.KillRequest) (_ *ptypes.E
 		return nil, err
 	}
 
+	// If the container has stopped, return directly when killed with
+	// SIGKILL or SIGTERM signal.
```
Ace-Tang: Why we still send other signal except SIGKILL and SIGTERM  when container is stop

lifupan: We cannot make sure when/what signals containerd would send, here only filter out the sigterm/sigkill signal.

Ace-Tang: I think the signal should be send to container, if the container init process are stopped, it won't deal the signal, I see kata agent will discard signal if container stop, so it also ok here.
https://github.com/kata-containers/agent/blob/master/grpc.go#L908

lifupan: Actually if the container has been stopped, the signal hadn't reached agent and would returned here https://github.com/kata-containers/runtime/blob/a7e2bbd31c5689255abe618730fe7b1b017aaf48/virtcontainers/container.go#L1075.

Here the filter would make sure even the container has been stopped, send SIGKILL/SIGTERM  signal wouldn't return an err.

containerd-shim-v2/service.go
```diff
@@ -665,6 +665,18 @@ func (s *service) Kill(ctx context.Context, r *taskAPI.KillRequest) (_ *ptypes.E
...
+	if signum == syscall.SIGKILL || signum == syscall.SIGTERM {
+		status, err := s.sandbox.StatusContainer(c.id)
+		if err != nil {
+			return nil, errors.Wrap(err, "kill container")
+		}
+		if status.State.State == types.StateStopped {
```
bergwolf: Why do we need to fake success here? runc shim doesn't do it either. It feels like container state between containerd and kata shimv2 mismatched.

lifupan:
Close this PR since another one https://github.com/kata-containers/runtime/pull/1723 would fix the same issue.

Issue #1627, opened 2019-05-07
containerd cri test failed with shimv2


lifupan:
# Description of problem

This commit https://github.com/kata-containers/runtime/pull/1436/commits/0e2be425141ad5c1c29247075cbe9245ba122bdf  updated the ttrpc module to latest which included the commit of https://github.com/containerd/ttrpc/commit/a364f44e55b1de969afc68a603408dde233a23d1, which will do context cancel after this service call return, and this cancel will propagated into kata sandbox's agent/hypervisor and resulted in the following calls canceled. 

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #1628, opened 2019-05-07
shimv2: fix the issue ttrpc server canceled context

lifupan:
This latest ttrpc vendor supports the feature of request timeout propgation.
this feature will do context cancel after a service call return, and this cancel
will propagated into kata sandbox's agent/hypervisor and resulted in the following
calls canceled. To fix this issue, pass the service's context instead of the service's
call's context to CreateSandbox(), and this context will live until the shim exited.

Fixes:#1627

### DIFF
--- a/containerd-shim-v2/service.go
+++ b/containerd-shim-v2/service.go
@@ -43,6 +43,10 @@ const (
 
 	chSize      = 128
 	exitCode255 = 255
+
+	// A time span used to wait for publish a containerd event,
+	// once it costs a longer time than timeOut, it will be canceld.
+	timeOut = 5 * time.Second
 )
 
 var (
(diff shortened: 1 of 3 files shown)


containerd-shim-v2/service.go
```diff
@@ -209,7 +214,10 @@ func (s *service) StartShim(ctx context.Context, id, containerdBinary, container
 
 func (s *service) forward(publisher events.Publisher) {
 	for e := range s.events {
-		if err := publisher.Publish(s.context, getTopic(s.context, e), e); err != nil {
+		ctx, cancel := context.WithTimeout(s.ctx, 5*time.Second)
```
devimc: Q. why 5?

lifupan: Hi @devimc, there is no specific reason for it, and it's just a time span that I think it's enough to publish an event to containerd.

jodh-intel: In which case, I'd consider making that a top-level `const` with a description rather than having it "buried" down here.

lifupan: @jodh-intel updated.

containerd-shim-v2/service.go
```diff
@@ -43,6 +43,8 @@ const (
 
 	chSize      = 128
 	exitCode255 = 255
+
+	pubTimeOut = 5 * time.Second
```
jodh-intel: Thanks. A couple of nits:

- It would be useful to explain "why 5?" in a comment here though.
- I'd consider calling this `publishTimeout` as "pub" is ambiguous - it might imply "public" or even "[public house](https://en.wikipedia.org/wiki/Pub)" if you live in certain parts of the world :laughing:

teawater:
jenkins-ci-centos-7-4-q-35 fail:
```
--- FAIL: TestHostNetworkingRequested (0.01s)
    assertions.go:239: 
                          
	Error Trace:	network_test.go:99
        
	Error:      	Should be true
```
This is a known issue #1604 .

Issue #1632, opened 2019-05-07
Firecracker: rootfs is mounted rw and also passed in rw to the VMM


mcastelino:
# Description of problem

In firecracker rootfs is mounted rw and also passed in rw to the VMM. The rootfs should not be modified as it is shared across multiple pods on the same system.

PR #1633, opened 2019-05-07
firecracker: Setup rootfs to be RO

mcastelino:
Setup rootfs to be RO both from the VMM point of view and the
VM point of view.

Fixes: #1632

### DIFF
--- a/virtcontainers/fc.go
+++ b/virtcontainers/fc.go
@@ -57,6 +57,8 @@ const (
 var fcKernelParams = []Param{
 	// The boot source is the first partition of the first block device added
 	{"root", "/dev/vda1"},
+	{"rootflags", "data=ordered,errors=remount-ro ro"},
+	{"rootfstype", "ext4"},
 	{"pci", "off"},
 	{"reboot", "k"},
 	{"panic", "1"},
@@ -299,7 +301,7 @@ func (fc *firecracker) fcSetVMRootfs(path string) error {
 	driveID := "rootfs"
 	driveParams := ops.NewPutGuestDriveByIDParams()
 	driveParams.SetDriveID(driveID)
-	isReadOnly := false
+	isReadOnly := true
 	//Add it as a regular block device
 	//This allows us to use a paritioned root block device
 	isRootDevice := false


virtcontainers/fc.go
```diff
@@ -57,6 +57,8 @@ const (
 var fcKernelParams = []Param{
 	// The boot source is the first partition of the first block device added
 	{"root", "/dev/vda1"},
+	{"rootflags", "data=ordered,errors=remount-ro ro"},
```
grahamwhaley: Why `data=ordered`?
Do we really want `errors=remount-ro` - that means afaict, if we hit an error, mount the fs `ro`. But, we should not hit any errors should we - and, would we not rather abort if we did?

jodh-intel: That's a fair question, but we do already use these flags for qemu, but only for some architectures:

- https://github.com/kata-containers/runtime/blob/master/virtcontainers/qemu_amd64.go#L40
- https://github.com/kata-containers/runtime/blob/master/virtcontainers/qemu_arm64.go#L48

But we don't set `rootflags` for ppc64 nor for s390x. It does seem like we have an opportunity to make this all consistent?

/cc @devimc, @amshinde, @sboeuf.

devimc: sure we can, unfortunately we still don't have a CI in place for all architectures hence we don't know if `ro` can break something. I think this change should be made by the architecture maintainer since he/she has the hw to test it.

cc @Pennyzct @nitkon @alicefr

mcastelino: We should make this consistent. I would not expect to see any failures as there is no need for us to write to the rootfs (and we should not from a security point of view).  If we see any failure we should root cause and fix them.

@grahamwhaley Regarding the ordered as we do not write I do not think we need it, but I chose not to remove to keep consistency.

About the remount I was not sure if we needed it, but it felt safer.

egernst:
Let's unify with other architectures in a follow on PR.

LGTM

mcastelino:
Added issue to track comments to closure

https://github.com/kata-containers/runtime/issues/1642

/cc @grahamwhaley @Pennyzct @nitkon @alicefr @devimc @egernst @jodh-intel

alicefr:
@devimc our daily build was successful this morning and it already included this commit. So, I think no issues for s390x

devimc:
@alicefr but this commit is for firecracker and AFAIK fc is not running on s390x

Pennyzct:
Hi~
@devimc @mcastelino thanks for the notification~ ;)
I'm always thinking that I should find a time to switch to `ro` for qemu part when you @devimc changed it for x86_64 in #1390. I will firstly try it on my local machine and fire a PR for it.
And since firecracker hasn't been ready for aarch64. maybe later in the future. ;)

Pennyzct:
Hi~ @devimc @mcastelino I have already raise the related PR #1660, ptal. ;)

PR #490, opened 2018-07-16
makefile: update PREFIX to remove redundant slash

egernst:
Fixes: #488

Signed-off-by: Eric Ernst <eric.ernst@intel.com>

PR #1634, opened 2019-05-07
dnm: dummy

egernst:
Depends-on: github.com/kata-containers/packaging/pull/491

Fixes: #490


ganeshmaharaj:
ARM is failing with qemu checkout failing cause of changed files and q35 is failing with the network_ns failure which are both consistent and i believe have patches en route to fix them both.

Issue #1368, opened 2019-03-14
Fail to stop kata container


Ace-Tang:
# Description of problem

When parallel run kata containers more than thousands times, will get some case cannot stop container, error is 
```
124 time="2019-03-13T11:34:04+08:00" level=info msg="{\"execute\":\"quit\"}" arch=amd64 command=kill container=d5f325b8ca2f83dd876aa7061f90f989bb58d221fb79063a610217a674376a7e name=kata-runtime pid=75093 sandbox=d5f325b8ca2f83dd876aa7061f90f989bb58d221fb79063a610217a674376a7e source=virtcontainers subsystem=qmp
125 time="2019-03-13T11:34:04+08:00" level=info msg="{\"timestamp\": {\"seconds\": 1552448044, \"microseconds\": 291115}, \"event\": \"SHUTDOWN\", \"data\": {\"guest\": true}}" arch=amd64 command=kill container=d5f325b8ca2f83dd876aa7061f90f989bb58d221fb79063a610217a674376a7e name=kata-runtime pid=75093 sandbox=d5f325b8ca2f83dd876aa7061f90f989bb58d221fb79063a610217a674376a7e     source=virtcontainers subsystem=qmp
126 time="2019-03-13T11:34:04+08:00" level=error msg="Fail to execute qmp QUIT" arch=amd64 command=kill container=d5f325b8ca2f83dd876aa7061f90f989bb58d221fb79063a610217a674376a7e error="unable to write command to qmp socket write unix @->/run/vc/vm/d5f325b8ca2f83dd876aa7061f90f989bb58d221fb79063a610217a674376a7e/qmp.sock: write: broken pipe" name=kata-runtime pid=75093 sand    box=d5f325b8ca2f83dd876aa7061f90f989bb58d221fb79063a610217a674376a7e source=virtcontainers subsystem=qemu
```

I assume if is qemu shutdown first than give response to `quit` request. (Just guess)

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

jodh-intel:
The container paths are unique so it seems unlikely this is directly related to running in parallel, but maybe more that you've managed to run a lot of containers and hit some sort of condition not normally seen.

What state is the qemu instance in (if it's still running)? Can you login with a debug console or get a stack trace of the qemu process?

[...]

PR #1635, opened 2019-05-08
qemu: receive qmp event when stop vm

Ace-Tang:
receive qmp event when start a qmp connect. This is useful for issue
1368, when send a quit qmp request to qemu, but vm shutdown first, and
not response for the quit qmp request. So receive the shutdown event to
judge whether vm is exited.

Fix: #1368

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -83,6 +83,8 @@ type qemu struct {
 	ctx context.Context
 
 	nvdimmCount int
+
+	isShutdown bool
 }
 
 const (
@@ -658,6 +660,14 @@ func (q *qemu) stopSandbox() error {
 
 	err = q.qmpMonitorCh.qmp.ExecuteQuit(q.qmpMonitorCh.ctx)
 	if err != nil {
+		for i := 0; i < 100; i++ {
+			if q.isShutdown {
+				q.Logger().Warnf("receive shutdown event but not get quit qmp response")
+				return nil
+			}
+
+			time.Sleep(time.Duration(50) * time.Millisecond)
+		}
 		q.Logger().WithError(err).Error("Fail to execute qmp QUIT")
 		return err
 	}
@@ -729,7 +739,9 @@ func (q *qemu) qmpSetup() error {
 		return nil
 	}
 
-	cfg := govmmQemu.QMPConfig{Logger: newQMPLogger()}
+	events := make(chan govmmQemu.QMPEvent)
+	cfg := govmmQemu.QMPConfig{Logger: newQMPLogger(),
+		EventCh: events}
 
 	// Auto-closed by QMPStart().
 	disconnectCh := make(chan struct{})
@@ -748,8 +760,18 @@ func (q *qemu) qmpSetup() error {
 	}
 	q.qmpMonitorCh.qmp = qmp
 	q.qmpMonitorCh.disconn = disconnectCh
+	go q.loopQMPEvent(events)
 
 	return nil
+}
+
+func (q *qemu) loopQMPEvent(event chan govmmQemu.QMPEvent) {
+	for e := range event {
+		q.Logger().WithField("event", e).Info("receive qemu event")
+		if strings.Contains(e.Name, "SHUTDOWN") {
+			q.isShutdown = true
+		}
+	}
 }
 
 func (q *qemu) qmpShutdown() {


Ace-Tang:
Hi, @devimc , could you help to see if the patch make sense ?

virtcontainers/qemu.go
```diff
@@ -748,10 +760,20 @@ func (q *qemu) qmpSetup() error {
 	}
 	q.qmpMonitorCh.qmp = qmp
 	q.qmpMonitorCh.disconn = disconnectCh
+	go q.loopQMPEvent(events)
 
 	return nil
 }
 
+func (q *qemu) loopQMPEvent(event chan govmmQemu.QMPEvent) {
```
lifupan: If the qemu is killed instead of exit by sending SHUTDOWN cmd from qmp, would this loopQMPEvent catch this case?

Ace-Tang: @lifupan , yes, it can be catched

lifupan: Hi @Ace-Tang 
Sorry, I cannot catch when and where the qmp shutdown event was created when qemu was killed?

Ace-Tang: :confused:   @lifupan , Can you describe how you test

lifupan: It's simple, run a pod and then kill the qemu process,  and I didn't see any shutdown events was logged.

Ace-Tang: I am not test with k8s, but only run with ctr run, and kill the qemu, then I can get event
```
time="2019-05-15T00:25:02.069044069+08:00" level=info msg="{\"timestamp\": {\"seconds\": 1557851102, \"microseconds\": 68970}, \"event\": \"SHUTDOWN\", \"data\": {\"guest\": false}}" ID=kill-k1 source=virtcontainers subsystem=qmp
time="2019-05-15T00:25:02.069167555+08:00" level=info msg="log event SHUTDOWN" ID=kill-k1 source=virtcontainers subsystem=qemu
time="2019-05-15T00:25:02.069678935+08:00" level=error msg="Wait for process failed" container=kill-k1 error="rpc error: code = Unavailable desc = transport is closing" pid=kill-k1
```

jcvenegas: @Ace-Tang  can you also get when you do a `kill -9` ?

virtcontainers/qemu.go
```diff
@@ -658,6 +660,14 @@ func (q *qemu) stopSandbox() error {
 
 	err = q.qmpMonitorCh.qmp.ExecuteQuit(q.qmpMonitorCh.ctx)
 	if err != nil {
+		for i := 0; i < 100; i++ {
+			if q.isShutdown {
```
jcvenegas: It is not wrong but looks a bit racy  if this start to be used in other places. 

I wonder if you try to check if  qemu process exist  or the a qemu socket (that is known exist) is a better way to identify if VM is running or not.

Ace-Tang: Make sense, if we do this, we can remove the logic about log qmp event

virtcontainers/qemu.go
Ace-Tang: No, so this method is not always work.

raravena80:
@Ace-Tang any updates? Thx!

Ace-Tang:
@raravena80 ， actually, I am not sure it is ok to make this change, code is not elegant

raravena80:
@Ace-Tang do you want to close this? and re-open later with a better change?

virtcontainers/qemu.go
```diff
@@ -658,6 +660,14 @@ func (q *qemu) stopSandbox() error {
 
 	err = q.qmpMonitorCh.qmp.ExecuteQuit(q.qmpMonitorCh.ctx)
```
amshinde: @Ace-Tang I am surprised that the earlier qmp.Setup() succeeds, so looks like for the qemu exits between qmpSetup and  qmp.ExecuteQuit() call.

virtcontainers/qemu.go
amshinde: @Ace-Tang I think its better if can check if the qmp.disconnectedCh is closed(indicating conneection has been lost) and check that the qmp.sock no longer exits to verify that the qemu instance has exited on receiving an error on `qmp.ExecuteQuit`. I dont think you can rely on SHUTDOWN msg, esp if you execute a `kill -9` on the qemu process. 

@markdryan What do you think?

markdryan:
> @Ace-Tang I think its better if can check if the qmp.disconnectedCh is closed(indicating conneection has been lost) and check that the qmp.sock no longer exits to verify that the qemu instance has exited on receiving an error on qmp.ExecuteQuit. I dont think you can rely on SHUTDOWN msg, esp if you execute a kill -9 on the qemu process.
> 
> @markdryan What do you think?

Yes this is correct.  Looking at govmm, [this doesn't seem to be documented very well](https://github.com/intel/govmm/blob/master/qemu/qmp.go#L656), which is probably what has led to the confusion.  However, qmp.disconnectedCh, is closed when the instance dies, either of its on accord or because it has been terminated using ExecuteQuit.  There is an example of how to do this.

https://github.com/intel/govmm/blob/master/qemu/examples_test.go#L84

disconnectedCh exists for this very reason.  It provides a race free way for the govmm client to know when a QEMU instance has actually quit.

caoruidong:
ping @Ace-Tang

Ace-Tang:
I will close it now, since I do not have enough time on fix this problem

Issue #1636, opened 2019-05-08
Hypervisor process may be left running if startSandbox fails


marcov:
I did some tests running a container with firecracker. I encountered a situation where kata-runtime would exit with an error message, yet the firecracker process was left running.

In my case the failure was due to a misconfiguration (the firecracker binary I used wasn't built with vsock support); yet in similar situations the hypervisor should not be left running.

PR #1637, opened 2019-05-08
virtcontainers: kill hypervisor if startSandbox fails

marcov:
Make sure the hypervisor is stopped if startSandbox does not succeed, by calling stopSandbox.
    
Fixes: #1636

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -967,11 +967,8 @@ func (s *Sandbox) startVM() (err error) {
 			if err != nil {
 				return err
 			}
-			err = vm.assignSandbox(s)
-			if err != nil {
-				return err
-			}
-			return nil
+
+			return vm.assignSandbox(s)
 		}
 
 		return s.hypervisor.startSandbox(vmStartTimeout)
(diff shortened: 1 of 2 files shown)


virtcontainers/sandbox.go
```diff
@@ -959,6 +959,12 @@ func (s *Sandbox) startVM() (err error) {
 
 	s.Logger().Info("Starting VM")
 
+	defer func() {
```
lifupan: In such a change, then if the error is triggered before the hypervisor startSandbox, would this cleanup tigger any other issues?  Is it better to do another cleanup in hypervisor.startSandbox() function?

jcvenegas: I agree

1. The defer should go after setup the network setup, in case that  `if err := s.network.Run(s.networkNS.NetNsPath, func() error {` fail the hypervisor is not started so the defer need to be a few lines after just before call `startSandbox()`

2.  I think that `startSandbox` should kill the hypervisor process in case of something goes wrong and not let the caller clean up if something fails. I see that actually it is killing the virtiofs process we should do something similar inside the function.

marcov:
@lifupan @jcvenegas Thanks for the review. I have moved the kill logic inside `startSandbox` to address your comments.

Issue #1640, opened 2019-05-08
netns leak issue


Ace-Tang:
# Description of problem

```
// SetupNetworkNamespace create a network namespace
func SetupNetworkNamespace(config *vc.NetworkConfig) error {
    if config.DisableNewNetNs {
        kataUtilsLogger.Info("DisableNewNetNs is on, shim and hypervisor are running in the host netns")
        return nil 
    }   

    if config.NetNSPath == "" {
        n, err := ns.NewNS()
        if err != nil {
            return err 
        }   

        config.NetNSPath = n.Path()
        config.NetNsCreated = true
        kataUtilsLogger.Debugf("Create a netns %q", n.Path())

        return nil 
    }   
```
When kata create a netns, and if container created fails, the netns will leak, it is easy to re-produce. 

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #1641, opened 2019-05-08
vc: fix netns leak

Ace-Tang:
when kata container start fails, netns will leak, add a defer func
to fix it

Fixes: #1640

### DIFF
--- a/pkg/katautils/create.go
+++ b/pkg/katautils/create.go
@@ -173,7 +173,7 @@ func SetEphemeralStorageType(ociSpec oci.CompatOCISpec) oci.CompatOCISpec {
 
 // CreateSandbox create a sandbox container
 func CreateSandbox(ctx context.Context, vci vc.VC, ociSpec oci.CompatOCISpec, runtimeConfig oci.RuntimeConfig, rootFs vc.RootFs,
-	containerID, bundlePath, console string, disableOutput, systemdCgroup, builtIn bool) (vc.VCSandbox, vc.Process, error) {
+	containerID, bundlePath, console string, disableOutput, systemdCgroup, builtIn bool) (_ vc.VCSandbox, _ vc.Process, err error) {
 	span, ctx := Trace(ctx, "createSandbox")
 	defer span.Finish()
 
@@ -203,6 +203,16 @@ func CreateSandbox(ctx context.Context, vci vc.VC, ociSpec oci.CompatOCISpec, ru
 	if err := SetupNetworkNamespace(&sandboxConfig.NetworkConfig); err != nil {
 		return nil, vc.Process{}, err
 	}
+
+	defer func() {
+		// cleanup netns if kata creates it
+		ns := sandboxConfig.NetworkConfig
+		if err != nil && ns.NetNsCreated {
+			if ex := cleanupNetNS(ns.NetNSPath); ex != nil {
+				kataUtilsLogger.WithField("path", ns.NetNSPath).WithError(ex).Warn("failed to cleanup netns")
+			}
+		}
+	}()
 
 	// Run pre-start OCI hooks.
 	err = EnterNetNS(sandboxConfig.NetworkConfig.NetNSPath, func() error {
--- a/pkg/katautils/network.go
+++ b/pkg/katautils/network.go
@@ -66,6 +66,7 @@ func SetupNetworkNamespace(config *vc.NetworkConfig) error {
 
 		config.NetNSPath = n.Path()
 		config.NetNsCreated = true
+		kataUtilsLogger.WithField("netns", n.Path()).Info("create netns")
 
 		return nil
 	}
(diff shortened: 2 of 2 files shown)


pkg/katautils/create.go
```diff
@@ -204,6 +204,16 @@ func CreateSandbox(ctx context.Context, vci vc.VC, ociSpec oci.CompatOCISpec, ru
 		return nil, vc.Process{}, err
 	}
 
+	defer func() {
+		// cleanup netns if kata creates it
+		ns := sandboxConfig.NetworkConfig
+		if err != nil && ns.NetNsCreated {
+			if ex := cleanupNetNS(ns.NetNSPath); ex != nil {
+				kataUtilsLogger.Warnf("Failed to cleanup netns %s", ns.NetNSPath)
```
devimc: please use `WithField` in all the logs to avoid using format specifiers. for example 

```go
kataUtilsLogger.WithField("path", ns.NetNSPath).Warn("Failed to cleanup netns")
```

pkg/katautils/network.go
```diff
@@ -179,3 +180,25 @@ func hostNetworkingRequested(configNetNs string) (bool, error) {
...
+// cleanupNetNS cleanup netns created by kata, trigger only create sandbox fails
+func cleanupNetNS(netNSPath string) error {
+	n, err := ns.GetNS(netNSPath)
+	if err != nil {
+		return err
+	}
+
+	err = n.Close()
+	if err != nil {
+		return err
```
amshinde: Can you return a more verbose message here : "Failed to close netns: %s"

pkg/katautils/create.go
Ace-Tang: Updated.

pkg/katautils/network.go
Ace-Tang: Updated.

pkg/katautils/create.go
```diff
@@ -204,6 +204,16 @@ func CreateSandbox(ctx context.Context, vci vc.VC, ociSpec oci.CompatOCISpec, ru
 		return nil, vc.Process{}, err
 	}
 
+	defer func() {
+		// cleanup netns if kata creates it
+		ns := sandboxConfig.NetworkConfig
+		if err != nil && ns.NetNsCreated {
+			if ex := cleanupNetNS(ns.NetNSPath); ex != nil {
+				kataUtilsLogger.WithField("path", ns.NetNSPath).Warn("Failed to cleanup netns")
```
jodh-intel: Could you tweak this slightly to also include the error:

```go
kataUtilsLogger.WithField("path", ns.NetNSPath).WithError(ex).Warn("Failed to cleanup netns")
```

grahamwhaley:
I've nudged a few CI rebuilds.
The metrics one just looked like something fell over.
The fedora and vsocks ones (which is also fedora) both had what looked like the same fail - I suspect it is not related to this PR, so respan. Here is the error for ref, in case it does re-occur:

```
• Failure [22.752 seconds]
state
/tmp/jenkins/workspace/kata-containers-runtime-fedora-PR/go/src/github.com/kata-containers/tests/functional/state_test.go:26
  container
  /tmp/jenkins/workspace/kata-containers-runtime-fedora-PR/go/src/github.com/kata-containers/tests/vendor/github.com/onsi/ginkgo/extensions/table/table.go:92
    with workload [true], timeWait 5 [It]
    /tmp/jenkins/workspace/kata-containers-runtime-fedora-PR/go/src/github.com/kata-containers/tests/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:46

    Expected
        <int>: 1
    to equal
        <int>: 0

    /tmp/jenkins/workspace/kata-containers-runtime-fedora-PR/go/src/github.com/kata-containers/tests/functional/state_test.go:45
SS

Summarizing 1 Failure:

[Fail] state container [It] with workload [true], timeWait 5 
```

pkg/katautils/create.go
Ace-Tang: Oh, yes , forget to log error

pkg/katautils/create.go
```diff
@@ -204,6 +204,16 @@ func CreateSandbox(ctx context.Context, vci vc.VC, ociSpec oci.CompatOCISpec, ru
 		return nil, vc.Process{}, err
 	}
 
+	defer func() {
+		// cleanup netns if kata creates it
+		ns := sandboxConfig.NetworkConfig
+		if err != nil && ns.NetNsCreated {
+			if ex := cleanupNetNS(ns.NetNSPath); ex != nil {
+				kataUtilsLogger.WithField("path", ns.NetNSPath).WithError(ex).Warn("Failed to cleanup netns")
```
jodh-intel: You might find that the CI linters complain as errors should start with a lower-case letter (same comment for errors below...)

Ace-Tang: I was vary confuse about this, if my memory is right, golang advised log should be start begin with lower-case. But kata use them both, that confused me, so I followed the upper-case way.
Make all log begin with lower-case?

jodh-intel: Hi @Ace-Tang - sorry! I meant to put this comment on the first of your new `fmt.Errorf()` calls as this clearly isn't an error, but a log call that's logging an error ;)

But yes, I think you are correct - we need to be more consistent and make all errors start with a lower-case letter at some point (not this PR).

But, for now, if the CI is happy, I'm happy :smile:

Ace-Tang: Hi, @jodh-intel ,do you mean 
```
n, err := ns.GetNS(netNSPath)
187     if err != nil { 
188         return fmt.Errorf("Failed to get netns %s: %v", netNSPath, err)   --> change Fail to fail, is it ?
189     }
```

jodh-intel: Yes. As I say, if the CI checks are happy, no need to change strictly. But I think this should probably be:

```go
return fmt.Errorf("failed to get netns %s: %v", netNSPath, err)
```

Ace-Tang: Update, @jodh-intel

jodh-intel:
Thanks @Ace-Tang.

/retest

lifupan:
Since the failed case "TestHostNetworkingRequested" is not related with the PR, so merged this PR.

Ace-Tang:
I will try to see if `TestHostNetworkingRequested` fail can be fixed

PR #1651, opened 2019-05-09
shimv2: create the rootfs dir if it doesn't exist

lifupan:
Latest containerd commit<c0f0b21314b93a1> had moved the
step of creating rootfs dir from creating bundle to container
creation; in order to support both of the old and latest
containerd, check the "rootfs" existed before creating it.

### DIFF
--- a/containerd-shim-v2/create.go
+++ b/containerd-shim-v2/create.go
@@ -210,6 +210,16 @@ func checkAndMount(s *service, r *taskAPI.CreateTaskRequest) error {
 }
 
 func doMount(mounts []*containerd_types.Mount, rootfs string) error {
+	if len(mounts) == 0 {
+		return nil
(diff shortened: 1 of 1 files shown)


containerd-shim-v2/create.go
```diff
@@ -210,6 +210,16 @@ func checkAndMount(s *service, r *taskAPI.CreateTaskRequest) error {
 }
 
 func doMount(mounts []*containerd_types.Mount, rootfs string) error {
+	if len(mounts) <= 0 {
```
jodh-intel: Can this really be `< 0`? Maybe just?:

```go
if len(mounts) == 0 {
    return nil
}
```

lifupan: Aha, you are right.

PR #1658, opened 2019-05-10
Use O_NONBLOCK for tty stdin.

xs3c:
our testing found that the client does not open
stdin with O_WRONLY, so the shim v2 was blocked forever
in open stdin with O_RDONLY. It's better to make it
opened with O_NONBLOCK, and do not block starting process
of container. and the containerd runc shim has done this
by bc1ff514 as well.

Signed-off-by: Yang, Wei <w90p710@gmail.com>
Signed-off-by: Yang, Wei <wei.yang1@linux.alibaba.com>

### DIFF
--- a/containerd-shim-v2/stream.go
+++ b/containerd-shim-v2/stream.go
@@ -56,7 +56,7 @@ func newTtyIO(ctx context.Context, stdin, stdout, stderr string, console bool) (
 	var err error
 
 	if stdin != "" {
-		in, err = fifo.OpenFifo(ctx, stdin, syscall.O_RDONLY, 0)
+		in, err = fifo.OpenFifo(ctx, stdin, syscall.O_RDONLY|syscall.O_NONBLOCK, 0)
 		if err != nil {
 			return nil, err
 		}


Ace-Tang:
@xs3c ， first you should create an issue for the PR, then add Fixes: # xxx in you commit content

containerd-shim-v2/stream.go
```diff
@@ -56,7 +56,7 @@ func newTtyIO(ctx context.Context, stdin, stdout, stderr string, console bool) (
 	var err error
 
 	if stdin != "" {
-		in, err = fifo.OpenFifo(ctx, stdin, syscall.O_RDONLY, 0)
+		in, err = fifo.OpenFifo(context.Background(), stdin, syscall.O_RDONLY|syscall.O_NONBLOCK, 0)
```
lifupan: Why create a new context here?

grahamwhaley:
@xs3c - when creating the matching Issue for this PR, please add details why you need this change :-) thx

containerd-shim-v2/stream.go
xs3c: The fifo non-blocking goroutine is not in the request scope, so I use a new context.

lifupan: In thus a case, you should use shimv2's service context.

xs3c: Okay.

bergwolf:
@xs3c You need to add proper description in your commit message to explain why the change is needed. Otherwise it looks good to me.

raravena80:
@xs3c any updates? Thx.  Can you add the details? 😄

xs3c:
@raravena80 @bergwolf @grahamwhaley sorry for late, I was busy with other works, I would add it this week.

jodh-intel:
Another CI fail due to the `mount` issue @amshinde:

```
[2]     should be empty [It]
[2]     /tmp/jenkins/workspace/kata-containers-runtime-ubuntu-18-04-PR/go/src/github.com/kata-containers/tests/integration/docker/run_test.go:295
[2] 
[2]     Expected
[2]         <string>: [    0.501821] systemd[71]: tmp.mount: Failed to execute command: No such file or directory
[2]         [    0.502146] systemd[71]: tmp.mount: Failed at step EXEC spawning /usr/bin/mount: No such file or directory
[2]     to be empty
[2] 
```

egernst:
restarting ci-vsocks and metrics.  I suspect these are just noise and this should be okay.

Issue #1664, opened 2019-05-11
Network namespace not available at hypervisor.createSandbox time


mcastelino:
# Description of problem

When the hypervisor createSandbox is called the network namespace is not properly populated.
virtcontainers/sandbox.go->createSandbox()->newSandbox()->s.hypervisor.createSandbox(ctx, s.id, s.networkNS, &sandboxConfig.HypervisorConfig, s.store)

At this point the network namespace is not populated. This leads to issues when running firecracker with jailer as the jailer needs a valid network namespace to move the firecracker process into the appropriate network namespace.

```
// createSandbox creates a sandbox from a sandbox description, the containers list, the hypervisor
// and the agent passed through the Config structure.
// It will create and store the sandbox structure, and then ask the hypervisor
// to physically create that sandbox i.e. starts a VM for that sandbox to eventually
// be started.
func createSandbox(ctx context.Context, sandboxConfig SandboxConfig, factory Factory) (*Sandbox, error) {
        span, ctx := trace(ctx, "createSandbox")
        defer span.Finish()

        if err := createAssets(ctx, &sandboxConfig); err != nil {
                return nil, err
        }

        s, err := newSandbox(ctx, sandboxConfig, factory)
        if err != nil {
                return nil, err
        }

        if len(s.config.Experimental) != 0 {
                s.Logger().WithField("features", s.config.Experimental).Infof("Enable experimental features")
        }

        // Fetch sandbox network to be able to access it from the sandbox structure.
        var networkNS NetworkNamespace
        if err := s.store.Load(store.Network, &networkNS); err == nil {
                s.networkNS = networkNS
        }

```

mcastelino:
This is due to the order in which we create network.

At `createSandbox` time the network is not yet available in the case of docker. So any hypervisor implementation needs to load it at `startSandbox` time.

```
func createSandboxFromConfig(ctx context.Context, sandboxConfig SandboxConfig, factory Factory) (*Sandbox, error) {
        span, ctx := trace(ctx, "createSandboxFromConfig")
        defer span.Finish()

        var err error

        // Create the sandbox.
        s, err := createSandbox(ctx, sandboxConfig, factory)
        if err != nil {
                return nil, err
        }

        // cleanup sandbox resources in case of any failure
        defer func() {
                if err != nil {
                        s.Delete()
                }
        }()

        // Create the sandbox network
        if err = s.createNetwork(); err != nil {
                return nil, err
        }

        // network rollback
        defer func() {
[...]

PR #1665, opened 2019-05-11
Networking: Ensure that network namespace is propagated

mcastelino:
Network namespace needs to be propagated if available at
createSandbox()

Fixes: #1664

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -543,6 +543,7 @@ func newSandbox(ctx context.Context, sandboxConfig SandboxConfig, factory Factor
 		shmSize:         sandboxConfig.ShmSize,
 		sharePidNs:      sandboxConfig.SharePidNs,
 		stateful:        sandboxConfig.Stateful,
+		networkNS:       NetworkNamespace{NetNsPath: sandboxConfig.NetworkConfig.NetNSPath},
 		ctx:             ctx,
 	}
 


virtcontainers/sandbox.go
```diff
@@ -543,6 +543,7 @@ func newSandbox(ctx context.Context, sandboxConfig SandboxConfig, factory Factor
 		shmSize:         sandboxConfig.ShmSize,
 		sharePidNs:      sandboxConfig.SharePidNs,
 		stateful:        sandboxConfig.Stateful,
+		networkNS:       NetworkNamespace{NetNsPath: sandboxConfig.NetworkConfig.NetNSPath},
```
sboeuf: @mcastelino Looking at your description of the issue #1664 there are actually 2 ways to fix this. The network config is stored inside `sandbox.config.NetworkConfig`, so you could find the same information from the config.
But this patch looks good to me too :)

bergwolf: The two are the same, -- `sandbox.config` is `sandboxConfig` as assigned above.

mcastelino:
Also when debugging this I find create sandbox being called multiple times during the course of the container creation. This is a bit unexpected. Hopefully the traces will indicate why.

virtcontainers/sandbox.go
mcastelino: @sboeuf yes the information is available within the store, and that is what I ended up doing. However it is wrong for us to pass in an empty value when there is a value present. So either we need to remove that parameter and always retrieve if from config or set it up properly.

jcvenegas:
Re-test - now that kernel issue is fixed

jodh-intel:
Merging as the sles CI is dead...

PR #1667, opened 2019-05-13
factory: Add new option disable_config_check

teawater:
In default mode, the runtime will check the VM config of VM factory before
get VM from it.
Disable VM factory config check.  Once enabled, the runtime will not check
whether the current VM config is same with the VM config of VM factory.
It helps speeding up new container creation.

Fixes: #1666

### DIFF
--- a/cli/config/configuration-qemu.toml.in
+++ b/cli/config/configuration-qemu.toml.in
@@ -232,6 +232,15 @@ enable_iothreads = @DEFENABLEIOTHREADS@
 #guest_hook_path = "/usr/share/oci/hooks"
 
 [factory]
+# In default mode, the runtime will check the VM config of VM factory before
+# get VM from it.
+# Disable VM factory config check.  Once enabled, the runtime will not check
+# whether the current VM config is same with the VM config of VM factory.
+# It helps speeding up new container creation.
+#
+# Default false
+#disable_config_check = true
+
 # VM templating support. Once enabled, new VMs are created from template
 # using vm cloning. They will share the same initial kernel, initramfs and
 # agent memory by mapping it readonly. It helps speeding up new container
--- a/pkg/katautils/config.go
+++ b/pkg/katautils/config.go
@@ -75,10 +75,11 @@ type tomlConfig struct {
 }
 
 type factory struct {
-	Template        bool   `toml:"enable_template"`
-	TemplatePath    string `toml:"template_path"`
-	VMCacheNumber   uint   `toml:"vm_cache_number"`
-	VMCacheEndpoint string `toml:"vm_cache_endpoint"`
+	DisableConfigCheck bool   `toml:"disable_config_check"`
+	Template           bool   `toml:"enable_template"`
+	TemplatePath       string `toml:"template_path"`
+	VMCacheNumber      uint   `toml:"vm_cache_number"`
+	VMCacheEndpoint    string `toml:"vm_cache_endpoint"`
 }
 
 type hypervisor struct {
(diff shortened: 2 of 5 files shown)


jodh-intel:
Hi @teawater - I can see what this PR is doing, but I don't understand *why* :smile: Can you provide more details of this new option in the commit please.

WeiZhang555:
It seems `apt-get update` fails a lot:
```
Summarizing 1 Failure:

[Fail] [Serial Test] package manager update test check apt-get update and upgrade [It] should not fail 
/tmp/jenkins/workspace/kata-containers-runtime-ubuntu-18-04-PR/go/src/github.com/kata-containers/tests/integration/docker/package_manager_test.go:61
```

grahamwhaley:
yeah, it feels like there was a bunch of 'repo unavailable' type errors across the CI last night :-( Let's see if there is a pattern so we can 'filter' the unrelated fails. /cc @chavafg

teawater:
> Hi @teawater - I can see what this PR is doing, but I don't understand _why_ 😄 Can you provide more details of this new option in the commit please.
@jodh-intel updated commit log and configuration-qemu.toml.in for it.

jodh-intel:
Thanks @teawater. Faster container creation sounds good, but doing less checking sounds potentially unsafe? Is that why the option is not enabled by default?

Can you share approximate figures for how much faster container creation is with this option set (a few percent or much faster)?

teawater:
> Can you share approximate figures for how much faster container creation is with this option set (a few percent or much faster)?

Set "disable_config_check = true" and "vm_cache_number = 1":
time for (( i = 0; i < 10; i++)); do docker run --rm --runtime=c busybox echo 1;  done
...
...
...
real    0m15.322s
user    0m0.218s
sys     0m0.138s

Just set "vm_cache_number = 1":
time for (( i = 0; i < 10; i++)); do docker run --rm --runtime=c busybox echo 1;  done
...
...
...
real    0m15.561s
user    0m0.223s
sys     0m0.144s

grahamwhaley:
Just to note, using https://github.com/kata-containers/tests/tree/master/metrics/report could give you a nice breakdown/view of the variance and where the time potentially went as well ;-) (yes, maybe I am biased ;-) ). You'd only need `grabdata.sh -t` to get the boot time test data.

grahamwhaley:
/test
We had some CI stability that I'm hoping got fixed overnight. re-spinning.

teawater:
[log.zip](https://github.com/kata-containers/runtime/files/3181784/log.zip)
1.json without any factory config:
| *F* | boot-times           |                0.50 |          1.22 | 0.70 | 33.3% | 1.17 | 1.28 | 9.7% | 1.8% | 100 |

2.json with "vm_cache_number = 1":
| *F* | boot-times           |                0.50 |          0.80 | 0.70 | 33.3% | 0.76 | 0.85 | 11.9% | 2.3% | 100 |

3.json with "vm_cache_number = 1" and "disable_config_check = true":
| *F* | boot-times           |                0.50 |          0.80 | 0.70 | 33.3% | 0.75 | 0.85 | 12.5% | 2.7% | 100 |

I removed the check "check_processes" in test because vmcache will start "kata-proxy" before test running.

grahamwhaley:
Thanks @teawater - summary then, with `vm_cache`, on your system, mean boot-to-workload time over 100 single (only) container launches for (I suspect) busybox went from 1.22s to 0.8s :-)

bergwolf:
> real 0m15.322s vs. real 0m15.561s
This is only slightly (1.556%) better and exposes potential risk of launching different VMs than expected. I'm not sure it worths the compromise.

jodh-intel:
Yeah, I don't like the idea of disabling any checks tbh.

PR #1668, opened 2019-05-13
persist: merge more files with `persist.json`

### LINKED ISSUE #803: [RFC] baseline persist data for live-upgrade

WeiZhang555:
Fixes #803

Merge more container storage files with `persist.json` including:
* devices.json
* mounts.json
* process.json

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -372,7 +372,10 @@ func (c *Container) GetPid() int {
 func (c *Container) SetPid(pid int) error {
 	c.process.Pid = pid
 
-	return c.storeProcess()
+	if !c.sandbox.supportNewStore() {
+		return c.storeProcess()
+	}
+	return nil
 }
 
 func (c *Container) setStateFstype(fstype string) error {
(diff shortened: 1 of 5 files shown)


WeiZhang555:
@jodh-intel The `newstore` feature is meant to replace old `VCStore`, current tests could be enough, we only need to guarantee that it doesn't break regression tests :)

Issue #1669, opened 2019-05-13
the unbinding of vfio device would be hang when using vhost and vfio


xs3c:
# Description of problem

the unbinding of vfio device would be hang when using vhost and vfio. 

I get this in dmesg:

> vfio-pci 0000:5c:00.0: Relaying device request to user

The vhost FD is opened by the containerd-shim-kata-v2 process and qemu

> lsof -p 8733 | grep vhost
> container 8733 root   18u      CHR             10,238      0t0   63842 /dev/vhost-net
> lsof -p 8751 | grep vhost
> qemu-kvm 8751 root    3u      CHR             10,238         0t0   63842 /dev/vhost-net

In the scenario, vhost net kernel thread tasks a reference to qemu's mm. and the reference
also includes the mmap regions on the vfio device file. so vhost kernel thread would be not
released when qemu is killed as the vhost file descriptor still is opened by shim v2 process, and vfio device is not released because there's still a reference to the mmap.

# Expected result

the kata container is stopped

# Actual result

I got a hang when unbinding network device from vfio-pci.
I get this in dmesg:

> vfio-pci 0000:5c:00.0: Relaying device request to user

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #1670, opened 2019-05-13
shim v2: Close vhostfd after vm get vhostfd

xs3c:
If kata containers is using vfio and vhost net，the unbinding
of vfio would be hang. In the scenario, vhost net kernel thread
takes a reference to the qemu's mm, and the reference also includes
the mmap regions on the vfio device file. so vhost kernel thread
would be not released when qemu is killed as the vhost file
descriptor still is opened by shim v2 process, and the vfio device
is not released because there's still a reference to the mmap.

Fixes: #1669

Signed-off-by: Yang, Wei <w90p710@gmail.com>

### DIFF
--- a/virtcontainers/network.go
+++ b/virtcontainers/network.go
@@ -1437,6 +1437,32 @@ func (n *Network) Add(ctx context.Context, config *NetworkConfig, hypervisor hyp
 	return endpoints, nil
 }
 
+func (n *Network) PostAdd(ctx context.Context, ns *NetworkNamespace, hotplug bool) error {
+	if hotplug {
+		return nil
+	}
+
+	if ns.Endpoints == nil {
+		return nil
+	}
+
+	endpoints := ns.Endpoints
+
+	for _, endpoint := range endpoints {
+		netPair := endpoint.NetworkPair()
+		if netPair == nil {
+			continue
+		}
+		if netPair.VhostFds != nil {
+			for _, VhostFd := range netPair.VhostFds {
+				VhostFd.Close()
+			}
+		}
+	}
+
+	return nil
+}
+
 // Remove network endpoints in the network namespace. It also deletes the network
 // namespace in case the namespace has been created by us.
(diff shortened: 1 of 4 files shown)


devimc:
cc @amshinde

virtcontainers/network.go
```diff
@@ -1437,6 +1437,21 @@ func (n *Network) Add(ctx context.Context, config *NetworkConfig, hypervisor hyp
 	return endpoints, nil
 }
 
+func (n *Network) PostAdd(ctx context.Context, ns *NetworkNamespace, hotplug bool) error {
+	if !hotplug && ns.Endpoints != nil {
```
jodh-intel: Please can you invert this logic to avoid the deep nesting:

```go
if hotplug {
    return nil
}

if ns.Endpoints == nil {
    return nil
}

endpoints := ns.Endpoints

for _, endpoint := range endpoints {
    // ...
}

```

xs3c: Done

jodh-intel:
Ahh, but the CI caught a small problem:

```
Found 1 commit between commit 01f1646fc06186101a7d0765e1f4abbda6a4bcf7 and branch master
ERROR: Commit 01f1646fc06186101a7d0765e1f4abbda6a4bcf7: Failed to find subsystem in subject: "shim v2: Close vhostfd after vm get vhostfd"
ERROR: checkcommits failed. See the document below for help on formatting
commits for the project.
```

That's arguably a bug with our commit checker but the solution is simple: change the "shim v2:" prefix to "shim-v2:" I think ;)

egernst:
/test

@xs3c - I ammended your commit header to make travis happy (to me, that static failure was lame, so I wanted to make your life easier).  Thx - LGTM.

lifupan:
Hi @xs3c , it seems you didn't do gofmt to the files you changed, that's why the CI failed.

xs3c:
@lifupan Thx fupan, I did gofmt these files and  then pushed it again. @egernst I used the commit log you modified ,so I also keep your SOF.

egernst:
still looks good, restarting what appears to be a flaky test failure on Ubuntu 18.04.

Issue #1489, opened 2019-04-08
fix cpu hotplug vcpu numbers calculation


justin-he:
# Description of problem
1.  There is two CPU number in q.config.NumVCPUs(from toml file) and in q.qemuConfig.SMP.CPUs (passed to govmm qemu).
In hotplugAddCPUs(), it didn't use it consistant VCPUs number calculation with resizeVCPUs().

2.  There are two limitations for vcpu number for cpu resource limitation:
a) In toml config file s.hypervisor.hypervisorConfig().NumVCPUs
b) utils.CalculateVCpusFromMilliCpus(mCPU) which is calculated from
cpu resource limitation such as "--cpus" , quota, period

IIUC, the correct logic should be:
if CalculateVCpusFromMilliCpus is not 0
        vcpu = CalculateVCpusFromMilliCpus
else
        vcpu = NumVCPUs

Without this patch, the cpu number is incorrect after
```
$docker run --runtime=kata-runtime -ti --cpus 2 debian bash
$lscpu
...
On-line CPU(s) list:   0-2
Off-line CPU(s) list:  3-7
...
```

PR #1677, opened 2019-05-15
virtcontainers: fix inconsistant VCPUs number

justin-he:
There are two VCPU numbers in q.config.NumVCPUs(from toml file) and
q.qemuConfig.SMP.CPUs (passed to govmm qemu).

Considered that config(the toml file) could change at any moment but
to the metadata we stored, it may be better to use q.qemuConfig.SMP.CPUs

Fixes: #1489
Signed-off-by: Jia He <justin.he@arm.com>


justin-he:
wrong and duplicated

Issue #1678, opened 2019-05-15
Got defunct kata-proxy after kata quit when vmcache is enabled


teawater:
```
root@teawater-virtual-machine:~# docker run --rm --runtime=c busybox echo 1
1
root@teawater-virtual-machine:~# ps -e | grep proxy
 27586 pts/13   00:00:00 kata-proxy <defunct>
 27681 pts/13   00:00:00 kata-proxy
```

PR #1679, opened 2019-05-15
kata_proxy: Open a special goroutine do cmd.Wait

teawater:
Got a defunct kata-proxy after kata quit when VMCache is enabled.
The reason is vmcache server opens kata-proxy but doesn't wait it.

If VMCache is disabled, kata-runtime will quit before kata-proxy.
So it will not meet the issue.

Open a special goroutine do cmd.Wait in kataProxy.start to handle
the isssue.

Fixes: #1678

### DIFF
--- a/virtcontainers/kata_proxy.go
+++ b/virtcontainers/kata_proxy.go
@@ -51,6 +51,8 @@ func (p *kataProxy) start(params proxyParams) (int, string, error) {
 		return -1, "", err
 	}
 
+	go cmd.Wait()
+
 	return cmd.Process.Pid, proxyURL, nil
 }
 


virtcontainers/kata_proxy.go
```diff
@@ -51,6 +51,8 @@ func (p *kataProxy) start(params proxyParams) (int, string, error) {
 		return -1, "", err
 	}
 
+	go cmd.Wait()
```
devimc: but.. but.. someone has to wait for this new routine, :confused: :question:

teawater: > but.. but.. someone has to wait for this new routine, 😕 ❓

Goroutine is a thread.  So when kata-runtime quit, it will quit with it.

PR #1685, opened 2019-05-15
Backport 1.7

egernst:
- include nemu configuration adjustments
- network namespace propagation fix


Makefile
```diff
@@ -163,9 +163,9 @@ DEFENTROPYSOURCE := /dev/urandom
 DEFDISABLEBLOCK := false
 DEFSHAREDFS := virtio-9p
 DEFSHAREDFS_NEMU := virtio-fs
-DEFVIRTIOFSDAEMON := $(VIRTIOFSDBINDIR)/virtiofsd
+DEFVIRTIOFSDAEMON := $(VIRTIOFSDBINDIR)/virtiofsd-x86_64
```
mcastelino: @ganeshmaharaj @egernst did we not add support for file backed non huge page for virtiofsd.
Do you plan to add that as a subsequent PR

https://github.com/kata-containers/runtime/pull/1657

egernst: I wish we had that now, but it isn't ready yet for 1.7.  TBH, I'm not sure if this would be a fair candidate for stable releases of 1.7, but definitely will enhance the usability.  Perhaps since this is an experimental  feature...

ganeshmaharaj: Yup. That is the next thing in our pipe. Given this feature is experimental and also the fact that file backend seems to have too many criterion to satisfy, it might be better to run it through rigorous testing before we land it.

PR #1687, opened 2019-05-15
pkg/katautils: improve boot time disabling systemd units

### LINKED ISSUE #1686: systemd: disable unnedded systemd units
boot time can be improved if some systemd units are disabled, for example `systemd-udevd-control.socket` and `systemd-udevd-kernel.socket `

devimc:
There are still some systemd units that fail to start for different
reasons. Disable these systemd units to improve boot time.

fixes #1686

cc @grahamwhaley 

### DIFF
--- a/pkg/katautils/create.go
+++ b/pkg/katautils/create.go
@@ -52,6 +52,10 @@ var noTraceKernelParam = []vc.Param{
 		Key:   "systemd.mask",
 		Value: "systemd-journal-flush.service",
 	},
+	{
+		Key:   "systemd.mask",
+		Value: "systemd-journald-dev-log.socket",
+	},
 	// No udev events: agent implements udev events
 	{
 		Key:   "systemd.mask",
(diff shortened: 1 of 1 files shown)


grahamwhaley:
oh, and /cc @jcvenegas for thoughts on removing the units from the rootfs itself indeed @bergwolf

jodh-intel:
https://github.com/kata-containers/osbuilder/issues/289 raised so we don't forget about this.

devimc:
@grahamwhaley 

> Is this related to the debian dmesg -l err issue I saw (which looks like I maybe only reported on slack, as I don't see an Issue open :-( )
Do you know what circumstances this does happen in, if it doesn't happen all the time?

please try again, I found this in system 241

grahamwhaley:
@devimc  ? do you mean `systemd` version 241? And, does that mean a recent rebuild of our rootfs image will be updated to/beyond that because the rolling Clear Linux has bumped their version?

devimc:
@grahamwhaley 

> do you mean systemd version 241? 

yes

> And, does that mean a recent rebuild of our rootfs image will be updated to/beyond that because the rolling Clear Linux has bumped their version?

Yes, the latest version of CL has systemd 241, but I don't know what version of CL we are using to generate our images. cc @chavafg  @jcvenegas

jcvenegas:
@devimc we use always the latest release

PR #1691, opened 2019-05-16
Add virtiofsd log and fix qemu hang due to virtiofsd vq setup failure

bergwolf:
When hugepage is not enabled, virtiofsd fails to setup vq with qemu and it causes qemu process to hang forever. waitSandbox only checks qmp so cannot detect such failure. Add virtiofsd debug log and stop qemu if virtiofsd quits.

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -645,6 +645,7 @@ func (q *qemu) startSandbox(timeout int) error {
 				}
 			}
 			q.Logger().Info("virtiofsd quits")
+			q.stopSandbox()
 		}()
 		timeoutDuration := time.Duration(timeout) * time.Second
 		select {
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -598,10 +598,16 @@ func (q *qemu) startSandbox(timeout int) error {
 		// connection with QEMU closes.  Therefore we do not keep track
 		// of this child process after returning from this function.
 		sourcePath := filepath.Join(kataHostSharedDir, q.id)
-		cmd := exec.Command(q.config.VirtioFSDaemon,
-			"-o", "vhost_user_socket="+sockPath,
-			"-o", "source="+sourcePath,
-			"-o", "cache="+q.config.VirtioFSCache)
+		args := []string{
+			"-o", "vhost_user_socket=" + sockPath,
+			"-o", "source=" + sourcePath,
+			"-o", "cache=" + q.config.VirtioFSCache}
+		if q.config.Debug {
+			args = append(args, "-d")
+		} else {
+			args = append(args, "-f")
+		}
+		cmd := exec.Command(q.config.VirtioFSDaemon, args...)
 		stderr, err := cmd.StderrPipe()
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


virtcontainers/qemu.go
```diff
@@ -598,10 +598,14 @@ func (q *qemu) startSandbox(timeout int) error {
...
-		cmd := exec.Command(q.config.VirtioFSDaemon,
-			"-o", "vhost_user_socket="+sockPath,
-			"-o", "source="+sourcePath,
-			"-o", "cache="+q.config.VirtioFSCache)
+		args := []string{
+			"-o", "vhost_user_socket=" + sockPath,
+			"-o", "source=" + sourcePath,
+			"-o", "cache=" + q.config.VirtioFSCache}
+		if q.config.Debug {
+			args = append(args, "-d")
```
jodh-intel: I'm not sure I understand this fully - don't we want to add this `-d` option always to detect the scenario you are describing?

bergwolf: Good point! We can use `-f` flag to avoid getting too many debug outputs.

bergwolf: updated. ptal.

virtcontainers/qemu.go
```diff
@@ -621,16 +627,24 @@ func (q *qemu) startSandbox(timeout int) error {
...
+			var sent bool
 			for scanner.Scan() {
-				if strings.Contains(scanner.Text(), "Waiting for vhost-user socket connection...") {
+				if q.config.Debug {
+					q.Logger().WithField("source", "virtiofsd").Debug(scanner.Text())
+				}
+				if !sent && strings.Contains(scanner.Text(), "Waiting for QEMU socket connection...") {
 					sockReady <- nil
-					return
+					sent = true
```
stefanha: I'm not sure it's useful to keep running the loop after sockReady <- nil since the kata-runtime process will terminate and this goroutine will disappear along with it.  Continuing to log from a goroutine makes sense in the v2 shim model, but I don't think it works in the classic kata-runtime OCI runtime model.

virtcontainers/qemu.go
```diff
@@ -621,16 +627,24 @@ func (q *qemu) startSandbox(timeout int) error {
 		timeStart := time.Now()
 		go func() {
 			scanner := bufio.NewScanner(stderr)
+			var sent bool
 			for scanner.Scan() {
-				if strings.Contains(scanner.Text(), "Waiting for vhost-user socket connection...") {
+				if q.config.Debug {
+					q.Logger().WithField("source", "virtiofsd").Debug(scanner.Text())
+				}
+				if !sent && strings.Contains(scanner.Text(), "Waiting for QEMU socket connection...") {
```
stefanha: s/QEMU/vhost-user/

bergwolf: Nice catch! Yeah, I need to update my virtiofsd as well.

virtcontainers/qemu.go
bergwolf: Yes, indeed. It can only be a best effort in the OCI command line case. And it works better for shimv2.

virtcontainers/qemu.go
bergwolf: fixed. ptal.

virtcontainers/qemu.go
stefanha: Looks good.

Issue #753, opened 2018-09-19
support config options through annotations


bergwolf:
As we were discussing the agent hook pr (https://github.com/kata-containers/agent/pull/365#discussion_r218652335), we might want to support specifying all kata configs through labels or annotations, and then there is no dependency on a fixed cli configuration and each pod can be customized. The issue is created to track the feature/discussion.

A simple suggestion can be that we define a format like `io.kata-containers.config.<subsystem>.<key>`. Then we can have labels/annotations:
1. `io.kata-containers.config.hypervisor.qemu.path`: the qemu path in `hypervisor.qemu` section
2. `io.kata-containers.config.shim.kata.enable_debug`: the `enable_debug` option in `shim.kata` section

etc.

We can still have a default config in `configuration.toml`, and the per pod annotation configs takes a higher priority and override the values in `configuration.toml`.

The use case is to support very flexible controllers that can customize any config options we support on a per pod basis.

/cc @jodh-intel @amshinde @eguzman3 @egernst

jodh-intel:
This is in fact an [old topic](https://github.com/moby/moby/issues/21324). The problem is that docker labels are not passed to the container and (atleast the last time I looked) docker doesn't provide a way to inject [OCI annotations](https://github.com/opencontainers/runtime-spec/blob/master/config.md#annotations) into the `config.json` it generates.

The OCI spec of course does [now](https://github.com/opencontainers/runtime-spec/pull/949) support the specification of a hypervisor, kernel and image as [first-class objects](https://github.com/opencontainers/runtime-spec/blob/master/config-vm.md).

Also, virtcontainers already provides a way to specify those same details as annotations:

- https://github.com/kata-containers/runtime/blob/master/virtcontainers/asset.go

If we want to add kata config as annotations it would need a prefix like `io.katacontainers.${subsystem}.${option_key}` or we could just stuff everything into a single annotation:

```
io.katacontainers.config = {
    "version": "1.2.3",
    "runtime": {
        "enable_debug": true
            :
    },
    "shim": {
    },
    :
}
```

If we are going to consider adding another set of config via annotations we need to:

- Think very carefully about which configuration sources we want to maintain.
- Ideally simplify what we already support (might require deprecation, rather than immediate removal).
- Determine how users can provide such config when using docker/k8s/etc.
- Document the ordering and priority that such "configuration sources" will take extremely carefully.
- Test all possible combinations extremely carefully.
[...]

PR #1695, opened 2019-05-16
[RFC][WIP]: Sandbox Annotations: Add new annotations for sandbox setup

mcastelino:
Kata-runtime supports multiple hypervisors. Not all hypervisor
support hotplug of resources. Hence all resources required by
the containers in the sandbox need to be setup at VM creation
time.

Introduce support for sandbox annotations which allow the
orchestrator to inject resource requirements at the pod level.
These annotations will typically be added to the pod spec
by an admission controller or an operator.

Initially add support for configuring memory and cpu.
Longer term this can be enhanced to include devices and other
resources.

### DIFF
--- a/virtcontainers/pkg/annotations/annotations.go
+++ b/virtcontainers/pkg/annotations/annotations.go
@@ -8,6 +8,15 @@ package annotations
 const (
 	vcAnnotationsPrefix = "com.github.containers.virtcontainers."
 
+	// SandboxMemory is a sandbox annotation for passing a per sandbox memory limit
+	// This detemines the memory size of the virtual machine
+	// Typically an admission controller would set this. This allows hypervisors that do not support
+	// memory hotplug to have a VM that is sized to match the sum of the container resources
+	SandboxMemory = vcAnnotationsPrefix + "SandboxMemory"
+
+	// SandboxCpus is a sandbox annotation for passing a per sandbox CPU limit
+	SandboxCpus = vcAnnotationsPrefix + "SandboxCpus"
+
 	// KernelPath is a sandbox annotation for passing a per container path pointing at the kernel needed to boot the container VM.
 	KernelPath = vcAnnotationsPrefix + "KernelPath"
 


bergwolf:
Again I would like to advocate the idea of allowing to set any kata config options via annotations. For example, the `hypervisor.qemu.path` config can be set via annotation `com.github.kata-containers.config.hypervisor.qemu.path`. This way, we can support very flexible controllers that can customize **any** config options on a per-pod basis.

The idea was raised before at #753.

mcastelino:
I agree. I will start with these two and build a template and we can target all the config options we care about either in this PR itself or follow on. This allows controllers to optimally configure kata based on pod, workload or a namespace based policy.

bergwolf:
Thanks @mcastelino . Please parse annotation config with kata defined config types so that
1. we only maintain supported configs in one place,
2. unsupported/obsolete annotation configs are always identified

virtcontainers/pkg/annotations/annotations.go
```diff
@@ -8,6 +8,15 @@ package annotations
...
 	vcAnnotationsPrefix = "com.github.containers.virtcontainers."
 
+	// SandboxMemory is a sandbox annotation for passing a per sandbox memory limit
+	// This detemines the memory size of the virtual machine
+	// Typically an admission controller would set this. This allows hypervisors that do not support
+	// memory hotplug to have a VM that is sized to match the sum of the container resources
+	SandboxMemory = vcAnnotationsPrefix + "SandboxMemory"
+
+	// SandboxCpus is a sandbox annotation for passing a per sandbox CPU limit
+	SandboxCpus = vcAnnotationsPrefix + "SandboxCpus"
```
jodh-intel: I think these should be:

```
com.github.containers.virtcontainers.sandbox.memory
com.github.containers.virtcontainers.sandbox.cpus
```

or, from #753:

```
io.kata-containers.config.virtcontainers.sandbox.memory
io.kata-containers.config.virtcontainers.sandbox.cpus
```

mcastelino: @jodh-intel based on @bergwolf  original proposal we should be directly able to map to the configuration.toml variables right. So should it be

```
io.kata-containers.config.hypervisor.default_vcpus
io.kata-containers.config.hypervisor.default_memory
io.kata-containers.config.hypervisor.enable_debug

```
or
```
io.kata-containers.config.hypervisor.firecracker.default_vcpus
io.kata-containers.config.hypervisor.firecracker.default_memory
io.kata-containers.config.hypervisor.firecracker.enable_debug
```

The hypervisor vs hypervisor.firecracker is a bit confusing. Which one should we use.
I like `io.kata-containers.config.hypervisor.default_vcpus` but then we lose the direct mapping.

That way we have a full and direct mapping for each and every config option?

Same with

`proxy.kata` vs `proxy` and `shim.kata` vs `shim`

jodh-intel: My only concern was not wanting `Foo${blah}` but instead `foo.${blah}`, so this sounds good to me.

mcastelino: @bergwolf @jodh-intel Which one

```
io.kata-containers.config.hypervisor.default_vcpus
```

or 
```
io.kata-containers.config.hypervisor.firecracker.default_vcpus
io.kata-containers.config.hypervisor.qemu.default_vcpus
```

I prefer the first method as it is but more generic and also due to the fact that virtcontainers/hypervisor.go etc are generic, but the second method is an exact mapping.

egernst: My preference is the first:
> ```io.kata-containers.config.hypervisor.default_vcpus```

bergwolf: @egernst I tend to agree with you but we need to define `CommonHypervisorConfig` first otherwise it will be pain to manually convert each field to a hypervisor type specific one.

jodh-intel: At the risk of over-engineering, we could support a generic default:

```
io.kata-containers.config.hypervisor.default.default_vcpus
```
(that "double default" is ugly so maybe `io.kata-containers.config.hypervisor.default.vcpus` or `io.kata-containers.config.hypervisor.generic.default_vcpus`).

... and a hypervisor-specific value which, if set, takes priority over `default`:

```
io.kata-containers.config.hypervisor.firecracker.vcpus
io.kata-containers.config.hypervisor.qemu.vcpus
```

As long as there are no hypervisors called `default`, that should work ;)

jodh-intel: And of course whichever strategy we decide upon needs to be documented: I think in an actual `README.md` (not just the code).

raravena80:
@mcastelino any updates? Thx!

virtcontainers/pkg/annotations/annotations.go
mcastelino: @jodh-intel I like default. Let me create a first cut implementation with the same.

mcastelino:
> @mcastelino any updates? Thx!

@raravena80 I think I have enough feedback to start with the implementation. Time now to do the actual implementation

egernst:
this is getting stale.  This would be very beneficial to get in, as it'll improve usability of Firecracker with Kata greatly.

amshinde:
@mythi Are you interested in picking this up?  I know you were keen on adding annotations for kernel-command-line options earlier.

virtcontainers/pkg/annotations/annotations.go
jodh-intel: Any update @mcastelino?

egernst:
this one needs help -- I think this is one of the more important languishing/needs help PRs out there.  Any takers?

grahamwhaley:
@jcvenegas was asking about current annotation state yesterday... was at least going to clean up the docs on what we have today I think?

caoruidong:
ping @ganeshmaharaj Do you take this one?

virtcontainers/pkg/annotations/annotations.go
amshinde: I am revisiting this feature to push an implementation for this. Looking at how things are implemented in Kata today, here are my thoughts:
The main areas that the end-user would need custom configurations through annotations are Hypervisor, Agent and Runtime(maybe in that order).
For hypervisors, we have a generic structure in the code today called `HypervisorConfig` for the hypervisor configuration, instead of a per hypervisor config structures specific to say qemu, firecracker etc.
So, I think we can start with supporting hypervisor annotations in the form 
`io.kata-containers.config.hypervisor.default_vcpus` 
that would apply to the chosen hypervisor for that workload. This would fit it pretty well with the existing code as well.

If we do want to support overrides in the form of per hypervisor specific annotations  and we see that is something that the users do want, we can consider adding hypervisor specific annotations of the form that @jodh-intel suggested:
`io.kata-containers.config.hypervisor.firecracker.vcpus`
I propose we tackle this in the second stage of this feature, if we really do see value in this.

While, we do want to support all the configurations that we currently have in our toml file, I am going to propose to tackle a subset of the configurations for the first pass.
I will document what these are, when I push the code for the initial implementation.

In case of the agent, it would make sense to follow the same generic approach, although if we look at the code for the agent config, we do not have a generic `AgentConfig`, but a specific `KataAgentConfig`, which is ok since technically we support just the kata-agent today.

I would also like to add that historically, we have supported a limited subset of annotations for custom kernel, rootfs etc of the form: 
```
com.github.containers.virtcontainers.KernelPath: "/boot/vmlinuz-custom-container"
```
See the existing annotation prefix `com.github.containers.virtcontainers`:
https://github.com/kata-containers/runtime/blob/2ed94cbd9d9090ae85fa2dc8b39cfaa46ddf8f29/virtcontainers/pkg/annotations/annotations.go#L9

I plan to drop that prefix and replace that with the one discussed here and make sure the existing annotations comply with the hierarchical format discussed here.
I think it should be ok to drop backward compatibility in this case, as this would mean a simple change in the pod yaml.
Let me know what others think. 
cc @bergwolf @jodh-intel @egernst

jcvenegas:
@amshinde  can we close this now?

amshinde:
@jcvenegas Yes closing this.

Issue #1700, opened 2019-05-17
agent debug console is broken


bergwolf:
With commit 582f20f489e5341171104d74384b1e58f2389aaa, guest console socket is read by the shim process. It works well until we want to connect to the agent debug console with socat to debug the guest. Then socat reads nothing as all data is read by the shim.

PR #1701, opened 2019-05-17
agent: fix agent debug console

bergwolf:
We should not let shim read guest console socket if agent
debug console is set. Otherwise when we connect to it with
socat, it reads nothing since all output is read by the shim.

Fixes: #1700

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -620,7 +620,10 @@ func (k *kataAgent) startProxy(sandbox *Sandbox) error {
 		agentURL:   agentURL,
 		consoleURL: consoleURL,
 		logger:     k.Logger().WithField("sandbox", sandbox.id),
-		debug:      sandbox.config.ProxyConfig.Debug,
+		// Disable debug so proxy doesn't read console if we want to
+		// debug the agent console ourselves.
+		debug: sandbox.config.ProxyConfig.Debug &&
+			!k.hasAgentDebugConsole(sandbox),
 	}
 
 	// Start the proxy here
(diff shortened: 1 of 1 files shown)


virtcontainers/kata_agent.go
```diff
@@ -1111,6 +1111,15 @@ func (k *kataAgent) buildContainerRootfs(sandbox *Sandbox, c *Container, rootPat
 	return nil, nil
 }
 
+func (k *kataAgent) hasAgentDebugConsole(sandbox *Sandbox) bool {
+	for _, p := range sandbox.config.HypervisorConfig.KernelParams {
+		if p.Key == "agent.debug_console" {
```
jodh-intel: Just a thought... although this is perfectly correct, what if we enhanced the capabilities of this flag (`agent.debug_console=false`, `agent.debug_console=bitmask` or something)? It is a little paranoid maybe, but it might be worth guarding against that by also ensuring `p.Value == ""`?

virtcontainers/kata_agent.go
```diff
@@ -1230,8 +1239,11 @@ func (k *kataAgent) createContainer(sandbox *Sandbox, c *Container) (p *Process,
 	}
 
 	// Ask to the shim to print the agent logs, if it's the process who monitors the sandbox and use_vsock is true (no proxy)
+	// Don't read the console socket if agent debug console is enabled.
 	var consoleURL string
-	if sandbox.config.HypervisorConfig.UseVSock && c.GetAnnotations()[vcAnnotations.ContainerTypeKey] == string(PodSandbox) {
+	if sandbox.config.HypervisorConfig.UseVSock &&
+		c.GetAnnotations()[vcAnnotations.ContainerTypeKey] == string(PodSandbox) &&
+		!k.hasAgentDebugConsole(sandbox) {
```
jodh-intel: Another random thought: might be good to add a log call is `k.hasAgentDebugConsole()` returns `true` so it's clear someone is already connected?

virtcontainers/kata_agent.go
bergwolf: `agent.debug_console` is the current agent ABI. If we enhance it with `agent.debug_console=false` we break the ABI and it is not necessary. For bitmask, I would suggest just adding `agent.debug_console.bitmask`. Either case, it doesn't preserve compatibility by ensuring `p.Value == ""` here, considering `p.Value == true` case ;) We will have to update both runtime and agent if we change the ABI.

virtcontainers/kata_agent.go
bergwolf: Yes, indeed. While at it, I found that the proxy also might be reading the debug console. I've updated the PR to fix it as well. PTAL.

Issue #1712, opened 2019-05-20
vfio error cause failed to start privileged kata container


zhabinecho:
# Description of problem

When I create the POD with network interface passthrough to guest, it fails:

> time="2019-05-18T16:52:11.569715264+08:00" level=error msg="failed to hotplug VFIO device" ID=396f4418692cb203212f3c1ef479f14e2a1911b946676dfcae168da4de749a85 error="QMP command failed: vfio error: 0000:5c:00.0: device is already attached" sandbox=396f4418692cb203212f3c1ef479f14e2a1911b946676dfcae168da4de749a85 source=virtcontainers subsystem=sandbox vfio-device-BDF="5c:00.0" vfio-device-ID=vfio-f20ea15af640067e0

> time="2019-05-18T16:52:11.569748867+08:00" level=error msg="Failed to add device" ID=396f4418692cb203212f3c1ef479f14e2a1911b946676dfcae168da4de749a85 error="QMP command failed: vfio error: 0000:5c:00.0: device is already attached" source=virtcontainers subsystem=device

The pci device 5c:00.0 has already pass to guest through qemu, the qemu command as follow

>  -device vfio-pci,host=0000:5c:00.0

Maybe the vfio in container devices don't need hotplug when start privileged kata container?

# Expected result

# Actual result

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

jodh-intel:
Thanks for raising @zhabinecho.

/cc @amshinde, @devimc.

egernst:
Interesting. /cc @amshinde @bergwolf 

When executing privileged, all the device nodes which exist on the host and exist in the guest are passe into the container.  I believe in your scenario, you hotplug the VFIO device into the virtual machine.  Because the device now exists in the host and the guest, it is automatically passed into the container namespace.  I think the error then occurs since you are trying to pass the device from guest to the container in the guest, as part of the hotplug in the agent code path, but it was already done via priv handling.

Archana, can you PTAL and comment?

liwei:
> error="QMP command failed: vfio error: 0000:5c:00.0: device is already attached" 

It seems the error message came from qemu qmp, I guess the root cause is we attach a vfio device to qemu twice(one as NIC passthrough, the other from privileged device list).

As we already have filterDevices() to filter out devices that have no meaning to pass to the guest, I suggest we make something like a blacklist or whitelist to feed to filterDevices()

[...]

PR #1714, opened 2019-05-21
virtcontainers: skip vfio device hotplug for privileged

zhabinecho:
Now the container will hotplug all the block device
and vfio devices in privileged. There is one case is that
one vfio device has already been hotplug to guest, so will
cause duplicate.

Fixes: #1712

Signed-off-by: Zha Bin <zhabin@linux.alibaba.com>

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -646,6 +646,14 @@ func filterDevices(c *Container, devices []ContainerDevice) (ret []ContainerDevi
 			continue
 		}
 
+		devType := c.sandbox.devManager.GetDeviceByID(dev.ID).DeviceType()
+		if devType == config.DeviceVFIO {
+			c.Logger().WithFields(logrus.Fields{
+				"device": dev.ContainerPath,
+			}).Info("Not attach device because it is a VFIO")
+			continue
+		}
+
 		ret = append(ret, dev)
 	}
 	return


virtcontainers/container.go
```diff
@@ -646,6 +646,14 @@ func filterDevices(c *Container, devices []ContainerDevice) (ret []ContainerDevi
 			continue
 		}
 
+		devType := c.sandbox.devManager.GetDeviceByID(dev.ID).DeviceType()
+		if devType == config.DeviceVFIO {
```
bergwolf: deviceManager.createDevice() is supposed to avoid adding the same device twice. Any reason that is not working? /cc @WeiZhang555 

Also we should just filter already added devices rather than skipping VFIO devices all together.

zhabinecho: Thanks,  @bergwolf.  I have read the deviceManager.createDevice.  The function dm.findDeviceByMajorMinor can avoid adding the same device. Maybe if the devices use dm.NewDevice to create, it can be working. Now the vfio device is been added to Qemu command line directly.  Please correct me if I'm wrong, I will check again, Thanks.

Yes, I will optimize the filter vfio devices :-)

bergwolf: @zhabinecho whatever you filter in filterDevices() must have gone through c.sandbox.devManager.NewDevice()->dm.createDevice(). So I don't see why you end up plugging the same vfio device obviously. One possibility is that the persistent data is inconsistent. Could you check `/var/run/vc/sbs/<sandbox-id>/devices.json` and see if the failed vfio device is listed there? You can add a long sleep where the qmp command fails and then check the devices.json file.

amshinde: > Now the vfio device is been added to Qemu command line directly.

@zhabinecho All vfio devices should be hotplugged and not passed on the command line. Are you seeing otherwise?
As @bergwolf pointed out, this likely seems an issue with persisting data.

amshinde: @zhabinecho Is this a network device? And are you using an sriov network plugin?

zhabinecho: @amshinde Yes, It is network device which is physical endpoint created by CNI network plugin. @bergwolf is right, the vfio device hasn't been persisted. I haven't found the right way to persist it in the physical endpoint attach.

caoruidong:
ping @zhabinecho

raravena80:
ping @zhabinecho

jodh-intel:
Restarted fedora + nemu CI's which had transitory failures.

amshinde:
This PR still needs some rework based on @bergwolf 's suggestion. 
@zhabinecho Can you address the comments?

zhabinecho:
> This PR still needs some rework based on @bergwolf 's suggestion.
> @zhabinecho Can you address the comments?

Sorry, I think that PhysicalEndpoint Attach may need use device manager to maintain the vfio device according to the  suggestion
But there was not appropriate way to get the device manager in PhysicalEndpoint Attach.

amshinde:
Because we cannot differentiate between priveleged and non-priveleged mode, with this change, I dont think we can pass vfio devices to a container any longer:
`sudo docker run -it --device=/dev/vfio/1 debian sh`

We really need to fix this by having PhysicalEndpoint Attach use the device manager to keep track if the device has been attached before.

Taking a broader look, I can now see how this would become an issue when you have a vfio device on your system and you try to run several priveleged containers on your system. Since all devices are passed to the container, each container will try to pass the vfio device to its respective VM causing failures.
cc @bergwolf

grahamwhaley:
I'm going to put a DNM label on this, just to make it clear I think the discussion is still on-going. @amshinde @bergwolf @zhabinecho - feel free to drop the label once everything is settled and resolved.

raravena80:
@zhabinecho any updates on this PR? Thx!

raravena80:
@zhabinecho any updates on this PR?

Your weekly Kata herder

raravena80:
@zhabinecho ping 😄

raravena80:
@zhabinecho any updates? Thanks!

raravena80:
@zhabinecho this is old, should we close this and move the work to 2.0?

fidencio:
As this PR is marked as `do-not-merge`, we're assuming it won't make for the final 1.13.0 release, thus we're closing it.
In case you feel this has been wrongly closed, please, feel free to re-open it.

Issue #1692, opened 2019-05-16
Miss use containerd namespace as netns ?


Ace-Tang:
# Description of problem

Code is locate at `containerd-shim-v2/service.go` 
```

// Create a new sandbox or container with the underlying OCI runtime
func (s *service) Create(ctx context.Context, r *taskAPI.CreateTaskRequest) (_ *taskAPI.CreateTaskResponse, err error) {
    s.mu.Lock()
    defer s.mu.Unlock()

//the network namespace created by cni plugin                                              
    netns, err :=                                             
    if err != nil {                                                                            
        return nil, errors.Wrap(err, "create namespace")                                       
    }                                                                                          
```         
namespaces.NamespaceRequired(ctx) , this namespace concept is containerd namespace, should not used as netns, the way get container netns is weird.
                          
I think this logic should be removed                    

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #1722, opened 2019-05-22
shimv2: remove use containerd ns as netns

Ace-Tang:
```
//the network namespace created by cni plugin
netns, err = namespaces.NamespaceRequired(ctx)
if err != nil {
        return nil, errors.Wrap(err, "create namespace")
}
```

the netns is a containerd namespace concept, it not netns, event a cni
set netns for this, this is a tricky way, so remove the logic.

Fixes: #1692

### DIFF
--- a/containerd-shim-v2/create.go
+++ b/containerd-shim-v2/create.go
@@ -10,12 +10,13 @@ package containerdshim
 import (
 	"context"
 	"fmt"
+	"os"
+	"path/filepath"
+
 	"github.com/containerd/typeurl"
 	vc "github.com/kata-containers/runtime/virtcontainers"
 	"github.com/kata-containers/runtime/virtcontainers/pkg/oci"
 	"github.com/pkg/errors"
-	"os"
-	"path/filepath"
 
 	taskAPI "github.com/containerd/containerd/runtime/v2/task"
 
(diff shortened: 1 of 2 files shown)


containerd-shim-v2/create.go
```diff
@@ -136,23 +137,6 @@ func loadSpec(r *taskAPI.CreateTaskRequest, netns string) (*oci.CompatOCISpec, s
...
-	//networkNamespace has been ignored by
-	//kata-agent in sandbox.
-
-	for _, n := range ociSpec.Linux.Namespaces {
-		if n.Type != specs.NetworkNamespace {
-			continue
-		}
-
-		if n.Path == "" {
-			n.Path = netns
```
Ace-Tang: Just say，this code is wrong, n is a temporary value， so n.Path is not changed

lifupan:
@Ace-Tang Thanks for the fix, I'm fine with this change once you  fixed the travis failure.

Ace-Tang:
Updated, `make` not take effect since containerd-shim-kata-v2 binary exist

containerd-shim-v2/create.go
jodh-intel: Good catch @Ace-Tang !!

PR #1723, opened 2019-05-22
fix the issue of hypervisor process is killed by kubelet 

lifupan:
For k8s, once all of the containers in a pod terminated, the pod status would be set "Succeeded" or "Failed" according the containers exited status, in any case, kubelet would consider the pod has terminated, and would try to cleanup the pod cgroup resources, which means kubelet would kill all of the processes in the pod cgroups, thus the qemu process would be killed by accident, at a result, kata shimv2 wouldn't communicate with qemu to do some sandbox cleanup and left the shimv2 process there.

This patch will try to fix this issue by moving "stop sandbox" from deleting podsandbox container stage to the time when podsandbox container exited and before kubelet detect the podsandbox container exited, thus kubelet wouldn't kill the hypervisor process.

### DIFF
--- a/containerd-shim-v2/container.go
+++ b/containerd-shim-v2/container.go
@@ -6,7 +6,6 @@
 package containerdshim
 
 import (
-	"sync"
 	"time"
 
 	"github.com/containerd/containerd/api/types/task"
@@ -31,7 +30,6 @@ type container struct {
 	stderr   string
 	bundle   string
 	cType    vc.ContainerType
-	mu       sync.Mutex
 	exit     uint32
 	status   task.Status
 	terminal bool
--- a/containerd-shim-v2/delete.go
+++ b/containerd-shim-v2/delete.go
@@ -17,21 +17,21 @@ import (
 )
 
 func deleteContainer(ctx context.Context, s *service, c *container) error {
-
-	status, err := s.sandbox.StatusContainer(c.id)
-	if err != nil {
-		return err
-	}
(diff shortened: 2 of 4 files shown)


nitkon:
@jodh-intel : Power CI errored 'cannot remove workspace' due to file permission issues. @grahamwhaley fixed the Jenkins configuration to make sure that the whole workspace of the previous job is owned by the Jenkins user so it can then delete it for the next run.

grahamwhaley:
@egernst @lifupan @gnawux @jodh-intel - is this maybe a stable backport candidate? Sounds like a real bug that really needed fixing?

lifupan:
> @egernst @lifupan @gnawux @jodh-intel - is this maybe a stable backport candidate? Sounds like a real bug that really needed fixing?

@grahamwhaley , yes, I think so.

grahamwhaley:
@lifupan - great, I've added the `stable-candidate` label so we don't forget it. Would you be able to PR the stable patches? /cc @jcvenegas @ganeshmaharaj 
BTW, **great work** on tracking down and fixing this :-)

lifupan:
> @lifupan - great, I've added the `stable-candidate` label so we don't forget it. Would you be able to PR the stable patches? /cc @jcvenegas @ganeshmaharaj
> BTW, **great work** on tracking down and fixing this :-)

Hi, @grahamwhaley  I had send three PR for stable-1.5 to stable-1.7.

PR #1741, opened 2019-05-26
runtime : delete redundant code in CreateContainer

### LINKED ISSUE #1740: Some redundant code in CreateContainer
# Description of problem

https://github.com/kata-containers/runtime/blob/360ffdca37866f82622c29dca1bf7a53fc3be50c/pkg/katautils/create.go#L228

here we get some strange codes^_^

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

jshachm:
Here we have done with logger and container ID map
Just delete these code.
fixes #1740

Signed-off-by: Haomin Tsai <caihaomin@huawei.com>


jshachm:
ping @lifupan @jodh-intel

PR #1755, opened 2019-06-03
stable: Catch up 1.7.0 tag

ganeshmaharaj:
It seems that stable-1.7 branch was created on master but the release tag `1.7.0` was forked off `1.7.0-rc1` tag with some changes and got pushed to master and tagged. Since then stable-1.7 has recieved a couple of backports which ends us with a fork between a release and it's branch. This merge is to bring them back onto a single tree for further maintenance and release of 1.7.x .


ganeshmaharaj:
The branch got fixed. Change not needed anymore.

Issue #1761, opened 2019-06-04
Firecracker toml result in failure to boot


egernst:
# Description of problem

Firecracker  toml result in failure to boot. Since proxy.kata section exists in resulting toml, our config handling will try to resolve path, regardless of vsock utilization.

See:
```go
if err := updateRuntimeConfig(resolved, tomlConf, &config, builtIn); err != nil {
		return "", config, err
	}

	config.DisableGuestSeccomp = tomlConf.Runtime.DisableGuestSeccomp

	// use no proxy if HypervisorConfig.UseVSock is true
	if config.HypervisorConfig.UseVSock {
		kataUtilsLogger.Info("VSOCK supported, configure to not use proxy")
		config.ProxyType = vc.NoProxyType
		config.ProxyConfig = vc.ProxyConfig{}
	}
```

https://github.com/kata-containers/runtime/blob/master/pkg/katautils/config.go#L942-L953

# Expected result

proxy shouldn't be considered, since vsock is being used

# Actual result

We check for proxy, based on existence of the table in toml, and fail

# Suggestion
We can either augment `updateRuntimeConfig` to only check for proxy if !vsock, or update the default toml to remove the proxy section.  I think just removing the proxy section in the toml.in makes most sense.

egernst:
Runtime equivalent to this issue: https://github.com/kata-containers/packaging/issues/543

PR #1762, opened 2019-06-04
fc-toml: remove proxy section in config

egernst:
proxy will never be use with the Firecracker VMM. Keeping this header
will result in runtime failures, since the configuration will be parsed
on the path searched for.

Since vsock will always be used, remove the proxy section.

Fixes: #1761

### DIFF
--- a/cli/config/configuration-fc.toml.in
+++ b/cli/config/configuration-fc.toml.in
@@ -201,8 +201,6 @@ use_vsock = true
 # Default false
 #enable_template = true
 
-[proxy.@PROJECT_TYPE@]
-
 [shim.@PROJECT_TYPE@]
 path = "@SHIMPATH@"
 


egernst:
rekicking ubuntu CI...

ganeshmaharaj:
ARM build seems to be failing with this.
```
10:19:39 Running command '/usr/local/bin/kata-runtime [kata-runtime --log=/tmp/bundle118142353/log run --bundle=/tmp/bundle118142353 --console= --pid-file=/tmp/bundle118142353/pid --detach iT1jFZZxSMw0qqeS7V0d]'
10:19:40 command failed error 'exit status 1'
10:19:40 [kata-runtime --log=/tmp/bundle118142353/log run --bundle=/tmp/bundle118142353 --console= --pid-file=/tmp/bundle118142353/pid --detach iT1jFZZxSMw0qqeS7V0d]
10:19:40 Timeout: 120 seconds
10:19:40 Exit Code: 1
10:19:40 Stdout: 
10:19:40 Stderr: qemu-system-aarch64: This host does not support 42b IPA: maxram/slots options not usable
10:19:40 
10:19:40 
```
Not sure how we can control which machine on packet this PR might land.

grahamwhaley:
@ganeshmaharaj @Pennyzct @chavafg - if we now have two different packet ARM Jenkins slaves with different feature sets, then we probably cannot, and do not want to, dictate which slave the job lands on... the code in the repo should handle and work on both imo, no?
But, having said that, I am concerned that we might land code that passes the CI on one of the slaves, but would have failed on the other. Maybe having disparate machines with different features means we should really run parallel ARM CI jobs to test both machine types for each PR? :-(

PR #1764, opened 2019-06-04
fc-toml: remove proxy section in config

egernst:
proxy will never be use with the Firecracker VMM. Keeping this header
will result in runtime failures, since the configuration will be parsed
on the path searched for.

Since vsock will always be used, remove the proxy section.

Fixes: #1761

### DIFF
--- a/cli/config/configuration-fc.toml.in
+++ b/cli/config/configuration-fc.toml.in
@@ -201,8 +201,6 @@ use_vsock = true
 # Default false
 #enable_template = true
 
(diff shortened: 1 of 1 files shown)


amshinde:
@egernst @ganeshmaharaj Seeing a bunch of go lint errors here.

ganeshmaharaj:
@amshinde yeah.. and does not match with what the changes are. @egernst might be a side-effect of us fixing the branch yesterday?

egernst:
did the test repository change?

PR #1774, opened 2019-06-06
factory: make vm templating work with vsock

bergwolf:
As virtio v1.1 spec states:
The guest_cid configuration field MUST be fetched to determine the current CID when a VIRTIO_VSOCK_EVENT_TRANSPORT_RESET event is received.
Existing connections MUST be shut down when a VIRTIO_VSOCK_EVENT_TRANSPORT_RESET event is received.
Listen connections MUST remain operational with the current CID when a VIRTIO_VSOCK_EVENT_TRANSPORT_RESET event is received.

We should be able to use vm templating together with vsock easily, as
qemu already sends VIRTIO_VSOCK_EVENT_TRANSPORT_RESET event to guest.

### DIFF
--- a/virtcontainers/factory/factory.go
+++ b/virtcontainers/factory/factory.go
@@ -112,7 +112,6 @@ func resetHypervisorConfig(config *vc.VMConfig) {
 	config.HypervisorConfig.BootFromTemplate = false
 	config.HypervisorConfig.MemoryPath = ""
 	config.HypervisorConfig.DevicesStatePath = ""
-	config.ProxyType = vc.NoopProxyType
 	config.ProxyConfig = vc.ProxyConfig{}
 }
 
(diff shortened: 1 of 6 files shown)


virtcontainers/factory/factory.go
```diff
@@ -140,7 +139,7 @@ func checkVMConfig(config1, config2 vc.VMConfig) error {
 func (f *factory) checkConfig(config vc.VMConfig) error {
 	baseConfig := f.base.Config()
 
-	return checkVMConfig(config, baseConfig)
+	return checkVMConfig(baseConfig, config)
```
jodh-intel: This change seems to have no effect so do we really need to do this?

bergwolf: It's a clean up that the log otherwise what the log says new is actually old base config right now.

jodh-intel: ack - thanks @bergwolf.

bergwolf:
@jodh-intel Please see https://github.com/kata-containers/tests/pull/1675

Issue #1775, opened 2019-06-06
Disable default hugepages for NEMU as virtio-fs can now use file-based backend


ganeshmaharaj:
NEMU needed file-based backend with sharing on. Kata did not have that feature when virtiofs was enabled and the workaround was to enable hugepages by default. Now that we have file based backend enabled and that is used by virtio-fs always, we can remove this change.

PR #1776, opened 2019-06-06
Disable default hugepages enabling for virtio-fs

ganeshmaharaj:
hugepages were enbled by default on NEMU to allow use of virtio-fs. kata
now has a change where virtio-fs will default to use /dev/shm as the
shared memory file backing location. With that, we should be able to
disable default hugepages for NEMU

Fixes: #1775


mcastelino:
LGTM

We need this as we should not use huge pages unless we have to.

mcastelino:
LGTM

Issue #1786, opened 2019-06-12
share file not ignore after copy to guest


Ace-Tang:
# Description of problem

share file should be ignored after copy to guest successfully

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

egernst:
@Ace-Tang @bergwolf is this still an issue? Saw the PR closed, should we close this as well as it works as expected?

Ace-Tang:
Yes, sorry to close the issue

PR #1787, opened 2019-06-12
virtcontainers: fix share file not be ignore

Ace-Tang:
share file should be ignored after copy to guest successfully, or it
still mount to host

Fixes: #1786

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -500,6 +500,8 @@ func (c *Container) shareFiles(m Mount, idx int, hostSharedDir, guestSharedDir s
 		if err := c.sandbox.agent.copyFile(m.Source, guestDest); err != nil {
 			return "", false, err
 		}
+
+		return "", true, nil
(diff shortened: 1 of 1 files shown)


Ace-Tang:
From the logic, I think the fix is right, but not sure, please help to review it :smile:

lifupan:
> From the logic, I think the fix is right, but not sure, please help to review it 😄

Even the files are copied to guest, I think we still cannot ignore them, since those files are copied into a shared guest dir, and they are still needed to be mounted into container  dir  in guest. If you ignored them, kata-agent will not do the mount and you cannot access them in container.

Ace-Tang:
@lifupan ， Actually, I want to ignore them since they have copied to the guest, and copy to the directory under rootfs, so kata-agent not need to mount again. But now I am not sure my test is right, I will check it again.

Ace-Tang:
Oh, I test was wrong

Issue #1788, opened 2019-06-12
shim v2 fail to run with cnm network


Ace-Tang:
# Description of problem

Fail to run shim v2 with bridged network, network use cnm mode.
```
# pouch run  -d --runtime=kata-runtime docker.io/library/busybox:latest
Error: failed to run container de3522: {"message":"failed to create container(de35222f6e3c4288be95e8b78ddfd82ed71f5457093c9ed3bc3a0ac497330d2f) on containerd: failed to create task for container(de35222f6e3c4288be95e8b78ddfd82ed71f5457093c9ed3bc3a0ac497330d2f): exit status 1: stdout: , stderr: time=\"2019-06-12T15:47:20+08:00\" level=info msg=\"external key /proc/14245/ns/net\" \ntime=\"2019-06-12T15:47:20+08:00\" level=fatal msg=\"failed to add interface vethf6bce45 to sandbox: error renaming interface \"vethf6bce45\" to \"eth0\": file exists\" \n: unknown"}
```

From read libnetwork/kata code, and with help of @lifupan , finally I found the reason:

1. kata create a new netns
2. with `EnterNS`, kata change netns to the created one.
3. when use cnm network mode, in pre-start hook, kata will re-exec libnetwork process `libnetwork-setkey`, and send self pid to it. libnetwork use `/proc/pid/ns/net` to find the netns kata use , and set veth into the netns

v1/v2 shim use the same way to create network, v1 can successful because `EnterNS` changed both current thread and main thread 's netns. But use v2 shim , only changed current thread netns, main thread still use host netns, so it fails. So if we pass `syscall.Gettid()` to libnetwork, it will work.

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

PR #1789, opened 2019-06-12
katautils: fix shim v2 fail to work with libnetwork

Ace-Tang:
detail how kata work with libnetwork
1. kata create a new netns
2. with EnterNS, kata change netns to the created one.
3. in pre-start hook, kata will re-exec libnetwork process
libnetwork-setkey, and send self pid to it. libnetwork use
/proc/pid/ns/net to find the netns kata use, and set veth into the netns.

v1/v2 shim use the same way to create network, v1 can successful
because EnterNS changed both current thread and main thread's netns.
But use v2 shim, only changed current thread netns, main thread still
use host netns, so it fails. Looks like v1 just lucky to be successful.
In kata, `state.Pid` should be tid.

Fixes: #1788

### DIFF
--- a/pkg/katautils/hook.go
+++ b/pkg/katautils/hook.go
@@ -11,7 +11,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"os"
 	"os/exec"
 	"strings"
 	"syscall"
@@ -39,7 +38,7 @@ func runHook(ctx context.Context, hook specs.Hook, cid, bundlePath string) error
 		log.String("hook-args", strings.Join(hook.Args, " ")))
 
 	state := specs.State{
-		Pid:    os.Getpid(),
+		Pid:    syscall.Gettid(),
 		Bundle: bundlePath,
 		ID:     cid,
 	}


egernst:
@Ace-Tang can you please submit this against stable-1.6 and stable-1.7 branches?

amshinde:
@Ace-Tang How does this fix the problem? you mentioned that for the network namespace created, the tid is differerent from the pid. 
But passing the tid to libnetwork would cause it to use the tid in entering the network namespace using `/proc/%d/ns/net`. How would that work? We are substituting a thread id here instead of process id.

I think with shimv1 we explicitly used LockOSThread() to make sure the netns is created on the main thread so that pid and tid are the same. @sboeuf had worked on this iirc.
We should probably look at why we see this beahviour in case of shimv2 where the tid ends up being different.

egernst:
@amshinde - the thread is associated with the correct netns.

That path, using thread id, will resolve correctly to the /proc/pid/task..../ns/net

amshinde:
@egernst Thanks for the explanation,I guess the thread entries are hidden when you do a `ls -l /proc`

Now that this is merged, we should also have a test for this, I dont think we test shimv2 today with CNM.
@Ace-Tang Since you went through the process recently, can you list out the setup steps for shimv2 with CNM.

Ace-Tang:
> I think with shimv1 we explicitly used LockOSThread() to make sure the netns is created on the main thread so that pid and tid are the same. @sboeuf had worked on this iirc.
We should probably look at why we see this beahviour in case of shimv2 where the tid ends up being different.

`this beahviour in case of shimv2 where the tid ends up being different` , this is what I not figure out , I pause the v2 shim process after it enter netns, I saw that pid from `getpid()` still keep host netns, but only `gettid()` changed the netns

Ace-Tang:
@amshinde, as I posts in issue, I test this with pouch, since docker not support shim v2, but pouch does and pouch use libnetwork. So I simply run with pouch.

bergwolf:
goroutine within LockOSThread()/UnlockOSThread() is executed in a different os thread. And grpc server uses goroutines to serve requests. Therefore we cannot rely on `LockOSThread()` to ensure all grpc request handlers run in the same kernel thread with shimv2. That is also the why tids ends up being different with shimv2.

Issue #1792, opened 2019-06-12
Fix typos and formatting


jodh-intel:
Correct spelling mistakes and formatting.

PR #1793, opened 2019-06-12
docs: Fix typos and formatting

jodh-intel:
Correct spelling mistakes and formatting issues.

Fixes: #1792.


jodh-intel:
@egernst - Regarding "config / configs" - again, we *could* add these to the custom dictionary. I decided that typing "config" was simply laziness on the part of the author but wdyt? ftr, we can add descriptions against our custom dictionary words which may allow us to "bend the rules" a bit further than maybe we should... :smile:

PR #1795, opened 2019-06-13
[cherry-pick 1.7]: katautils: fix shim v2 fail to work with libnetwork

Ace-Tang:
detail how kata work with libnetwork
1. kata create a new netns
2. with EnterNS, kata change netns to the created one.
3. in pre-start hook, kata will re-exec libnetwork process
libnetwork-setkey, and send self pid to it. libnetwork use
/proc/pid/ns/net to find the netns kata use, and set veth into the netns.

v1/v2 shim use the same way to create network, v1 can successful
because EnterNS changed both current thread and main thread's netns.
But use v2 shim, only changed current thread netns, main thread still
use host netns, so it fails. Looks like v1 just lucky to be successful.
In kata, `state.Pid` should be tid.

Fixes: #1788

### DIFF
--- a/pkg/katautils/hook.go
+++ b/pkg/katautils/hook.go
@@ -11,7 +11,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"os"
 	"os/exec"
 	"strings"
 	"syscall"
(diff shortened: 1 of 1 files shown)


Ace-Tang:
```
pkg/katatestutils/README.md
README.md
ERROR: Invalid URL 'https://godoc.org/github.com/kata-containers/runtime/pkg/katatestutils' found in the following files:
pkg/katatestutils/README.md
ERROR: Invalid URL 'https://godoc.org/github.com/kata-containers/runtime?status.svg' found in the following files:
```

Is it caused by network ?

grahamwhaley:
@jodh-intel - query on doc url checker and a badge I think?

jodh-intel:
@Ace-Tang - yes, it's a transitory network issue.

PR #1797, opened 2019-06-14
kata-check: add kvm extension check on aarch64

Pennyzct:
Auger Eric's latest patches about [ARM virt: Initial RAM expansion and extended memory map](https://patchwork.kernel.org/cover/10835377/) paves the way to device memory, which is the foundation for NVDIMM and memory hotplug.
This new feature on qemu kinds of depends on host kernel's new feature
on [dynamic IPA range](https://lwn.net/Articles/750176/).
The availability of this feature is advertised by a new kvm cap `KVM_CAP_ARM_VM_IPA_SIZE`. When supported, this capability returns the maximum IPA shift supported by the host. 
The supported IPA size on a host could be different from the system's PARange indicated
by the CPUs (e.g, kernel limit on the PA size).
This is how my local arm64 machine shows:
```
INFO[0000] kvm extension is supported                    arch=arm64 description="Maximum IPA shift supported by the host" id=165 name=KVM_CAP_ARM_VM_IPA_SIZE pid=98555 source=runtime type="kvm extension"
INFO[0000] IPA limit size: 44 bits.                      arch=arm64 name=KVM_CAP_ARM_VM_IPA_SIZE pid=98555 source=runtime type="kvm extension"
```

### DIFF
--- a/cli/kata-check.go
+++ b/cli/kata-check.go
@@ -11,6 +11,7 @@ package main
 #include <linux/kvm.h>
 
 const int ioctl_KVM_CREATE_VM = KVM_CREATE_VM;
+const int ioctl_KVM_CHECK_EXTENSION = KVM_CHECK_EXTENSION;
 */
 import "C"
 
@@ -38,6 +39,15 @@ type kernelModule struct {
 
 	// if it is definitely required
 	required bool
+}
+
+// nolint: structcheck, unused, deadcode
+type kvmExtension struct {
+	// description
+	desc string
+
+	// extension identifier
+	id int
 }
 
 type vmContainerCapableDetails struct {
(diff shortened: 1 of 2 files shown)


Pennyzct:
emmmmm, ARM CI will keep failing until PR [osbuilder/#265](https://github.com/kata-containers/osbuilder/pull/265) gets landed. 😭

Pennyzct:
anyone would like to have a peek? @grahamwhaley @devimc @bergwolf (～￣▽￣)～

cli/kata-check.go
```diff
@@ -388,3 +398,41 @@ func genericKvmIsUsable() error {
...
+		}
+
+		ret, _, errno := syscall.Syscall(syscall.SYS_IOCTL,
+			uintptr(kvm),
+			uintptr(C.ioctl_KVM_CHECK_EXTENSION),
+			uintptr(extension.id))
+
+		results[name] = int(ret)
+
+		if errno != 0 || ret <= 0 {
```
bergwolf: Does ARM syscall calling convention return 0 upon error? On x86_64, ret is either 0 or -1 so it seems this always fails.

cli/kata-check.go
```diff
@@ -388,3 +398,41 @@ func genericKvmIsUsable() error {
...
+			"description": extension.desc,
+			"id":          extension.id,
+		}
+
+		ret, _, errno := syscall.Syscall(syscall.SYS_IOCTL,
+			uintptr(kvm),
+			uintptr(C.ioctl_KVM_CHECK_EXTENSION),
+			uintptr(extension.id))
+
+		results[name] = int(ret)
```
bergwolf: You might want to move the line to after the error check so that we don't pollute successful results.

cli/kata-check.go
Pennyzct: Hi~ @bergwolf 
my neglect. ARM will use return value to illustrate some hardware feature, just like this cap `KVM_CAP_ARM_VM_IPA_SIZE` , it will return 0 upon error, and specific IPA limit this host supports.
I'm referring the kvm extension check in qemu source code, it should be applied to all archs.
```
int kvm_check_extension(KVMState *s, unsigned int extension)
{
    int ret;

    ret = kvm_ioctl(s, KVM_CHECK_EXTENSION, extension);
    if (ret < 0) {
        ret = 0;
    }

    return ret;
}
```
So here, I think I should only return < 0 upon error. ;)

cli/kata-check.go
Pennyzct: ok. thanks.  I will move it. ;)

cli/kata-check.go
bergwolf: I see. Thanks for the explanation. It's interesting to know that kvm ioctl can return zero upon failure.

Pennyzct:
@bergwolf  updated! ;) ptal.<(￣︶￣)↗[GO!]

cli/kata-check.go
egernst: @Pennyzct -- if it isn't completely clear to someone as smart as @bergwolf, I think it warrants a comment to clarify the behavior in the code.  WDYT?

cli/kata-check.go
```diff
@@ -388,3 +398,40 @@ func genericKvmIsUsable() error {
 
 	return nil
 }
+
+// genericCheckKVMExtension allows to query about the specifc kvm extensions
```
amshinde: nit: /specifc/specific/

cli/kata-check.go
Pennyzct: can't agree more.🤭, I will add detailed comments asap.

cli/kata-check.go
Pennyzct: good catch, thanks! update asap. 👻

Pennyzct:
@egernst @amshinde updated! ptal. 🧐

PR #1799, opened 2019-06-14
qemu: use x-ignore-shared to implement vm template

bergwolf:
As we want to move to upstream qemu, we should drop as many private qemu patches as possible. One big change we carry with qemu-lite is the migration bypass-shared-memory patch. Qemu 4.0 already has a x-ignore-shared migration capability. With a small change to it, we can make it work similar to bypass-shared-memory and thus get rid of the private qemu patch. Then we can use x-ignore-shared instead of bypass-shared-memory to implement the VM template feature.

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -421,8 +421,7 @@ func (q *qemu) setupTemplate(knobs *govmmQemu.Knobs, memory *govmmQemu.Memory) g
 		}
 
 		if q.config.BootFromTemplate {
-			incoming.MigrationType = govmmQemu.MigrationExec
-			incoming.Exec = "cat " + q.config.DevicesStatePath
+			incoming.MigrationType = govmmQemu.MigrationDefer
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


devimc:
@bergwolf govmm pr was merged

bergwolf:
@devimc Thanks. It still needs to wait for the qemu patch which is being reviewed on the qemu mailing list though.

raravena80:
@bergwolf ping, any updates on the qemu patch?

bergwolf:
@raravena80 I asked Dave and he said the patch will be included in the next qemu migration pull request. Once that is done, I'll move forward the kata side changes.

bergwolf:
Dropping `dnm` as the qemu patch has been merged upstream.

virtcontainers/qemu.go
```diff
@@ -585,6 +584,85 @@ func (q *qemu) vhostFSSocketPath(id string) (string, error) {
 	return utils.BuildSocketPath(store.RunVMStoragePath, id, vhostFSSocket)
 }
 
+func (q *qemu) setupVirtiofsd(timeout int) (remain int, err error) {
```
jodh-intel: What if `timeout < 0`?

virtcontainers/qemu.go
```diff
@@ -585,6 +584,85 @@ func (q *qemu) vhostFSSocketPath(id string) (string, error) {
...
+	sockPath, err := q.vhostFSSocketPath(q.id)
+	if err != nil {
+		return 0, err
+	}
+
+	// The daemon will terminate when the vhost-user socket
+	// connection with QEMU closes.  Therefore we do not keep track
+	// of this child process after returning from this function.
+	sourcePath := filepath.Join(kataHostSharedDir, q.id)
+	args := []string{
```
jodh-intel: This function does three things:

1. constructs arguments.
1. goroutine and channel logic.
1. adjusts the timeout.

If you broke it into sub-functions for (atleast) (1) and (3), you could add some simple unit tests for those.

virtcontainers/qemu.go
bergwolf: `timeout` is internal and never changed before reaching here. I'm not against validating it but what is the place to do the validation along the call stack? That said, we do make sure it is non-negative when changing it below.

virtcontainers/qemu.go
bergwolf: Sounds good. I'll add some UTs for it.

bergwolf: I've split the function and added UTs. PTAL.

chavafg:
Found this error on the nemu job:
```
08:01:03 [2] run container with docker 
08:01:03 [2]   in background, interactive and with a tty
08:01:03 [2]   /tmp/jenkins/workspace/kata-containers-runtime-ubuntu-nemu/go/src/github.com/kata-containers/tests/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:46
08:01:03 [2] Running command '/usr/bin/docker [docker run --cidfile /tmp/cid577675508/StgN2Yo5CpBsjk77tAl8OrYVzfbSs2 --runtime kata-runtime --name StgN2Yo5CpBsjk77tAl8OrYVzfbSs2 -dit busybox sh]'
08:01:43 [2] command failed error 'exit status 125'
08:01:43 [2] [docker run --cidfile /tmp/cid577675508/StgN2Yo5CpBsjk77tAl8OrYVzfbSs2 --runtime kata-runtime --name StgN2Yo5CpBsjk77tAl8OrYVzfbSs2 -dit busybox sh]
08:01:43 [2] Timeout: 120 seconds
08:01:43 [2] Exit Code: 125
08:01:43 [2] Stdout: 8aa0465c28a1c978c76de03840ffe08b48e360215857ef715ef8ecf61f0dc973
08:01:43 [2] 
08:01:43 [2] Stderr: docker: Error response from daemon: OCI runtime create failed: nemu-system-x86_64: -device vhost-user-fs-pci,chardev=char-07b3b750745d5889,tag=kataShared,cache-size=1024M: Failed to read msg header. Read 0 instead of 12. Original request 1.
08:01:43 [2] nemu-system-x86_64: -device vhost-user-fs-pci,chardev=char-07b3b750745d5889,tag=kataShared,cache-size=1024M: vhost_dev_init failed: Operation not permitted: unknown.
```
restarted job

Issue #1603, opened 2019-04-29
sandbox dns is not set


bergwolf:
kubernetes cri does pass the dns config in `RunPodSandboxRequest.PodSandboxConfig.DNSConfig` and containerd saves it as `resolv.conf` file [in sandbox root dir](https://github.com/containerd/containerd/blob/master/vendor/github.com/containerd/cri/pkg/server/sandbox_run.go#L495). We should pick it up and pass it to the guest via `CreateSandboxRequest` agent grpc.

harche:
@bergwolf thanks, I will send the PR to fix it.

amshinde:
@bergwolf Just to clarify when you say copied into the sandbox root dir, this is actually the root dir of the infra/pod container. So if I understand correctly, currently nothing is broken.
I believe you have created this issue just to support the use case of allowing network access from the agent itself to allow it to download files before any container is started that @harche and @nitkon are interested in. Please confirm. 
cc @egernst

bergwolf:
> Just to clarify when you say copied into the sandbox root dir, this is actually the root dir of the infra/pod container

@amshinde I think it is [the sandbox dir.](https://github.com/containerd/containerd/blob/master/vendor/github.com/containerd/cri/pkg/server/helpers.go#L199). Check files under `/var/lib/containerd/io.containerd.grpc.v1.cri/sandboxes/<sandboxid>/`. We need to see how runc shim makes use of them and copy the behavior.

jcvenegas:
Hey just comming from https://github.com/kata-containers/osbuilder/pull/346 that is 
needed  by https://github.com/kata-containers/agent/pull/625.

@bergwolf I'd like to make clear if setting the DNS in the sandbox adds some value.

Talking with @nitkon  I see the only value for now in https://github.com/kata-containers/agent/pull/625 is just for completeness.

To make this possible, the approach suggested by @nitkon  is to set the `/etc/resolv.conf` in the guest. Because /etc is read mounted, seems that they the file is required to exist from the rootfs (and a PR is in place for it https://github.com/kata-containers/osbuilder/pull/346).

So we are adding as a requirement that the filesystem has a file `/etc/resolv.conf` to be able to modified it, also this require that the `/etc/resolv.conf` remounted to allow write.

My question is we are adding two changes but no value, if the completeness of the sanbox network is an strong reason lets go for it.  If not just lets  close the issue or find a less complex solution.

[...]

PR #1825, opened 2019-06-24
virtcontainers: Set sandbox dns 

nitkon:
Inorder to set VM hostname, dns, hosts file, we can copy
these files from the host to VM rootfs.

Depends-on: github.com/kata-containers/agent#590

Fixes: #1603

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -6,11 +6,13 @@
 package virtcontainers
 
 import (
+	"bytes"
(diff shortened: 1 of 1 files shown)


amshinde:
@nitkon I have asked for some clarification from @bergwolf on the issue that this PR fixes.
Please feel free to pitch in.
This looks good to me, but I believe you were running into issues with accessing the network, last time we talked even with this PR. Is that resolved?

Issue #1358, opened 2019-03-12
Evaluate Kata + Podman


egernst:
I'm confident integration of Kata into podman is feasible, and we should have a "how-to" guide and some level of automated testing available to ensure the 'how-to' guides' functionality.

Futher, it would be good to evaluate this in the context of running kata rootless.  This should be feasible with overlay2, though there would likely be complications wrt devicemapper.

egernst:
Any progress on this front? @gabibeyer can you summarize your findings thus far?

gabibeyer:
kata seems to work pretty seamlessly with podman as long as selinux is disabled (which already seems to be a reported [issue](https://github.com/kata-containers/runtime/issues/784) within kata). I have gotten it running in fedora 28, trying ubuntu bionic currently just to see if there are differences.

mcastelino:
@gabibeyer can you describe your setup. Also the final goal is to run Kata as rootless with podman. 

We need to ensure we do not create files with root uid/gid on the host side when using overlay. Also ensure that the other files like resolv.conf etc do not end up being root when modified.

gabibeyer:
@mcastelino right, that is not yet rootless,. Should I still describe my setup if it is not yet the rootless implementation?

gabibeyer:
https://gist.github.com/gabibeyer/ae3f8ee5e7ebc4e1db9d9c26119ca861

gabibeyer:
Modifications to [podman](https://github.com/containers/libpod/issues/2897) and [slirp4netns](https://github.com/rootless-containers/slirp4netns/issues/84) are in progress for a rootless podman + kata integration

gabibeyer:
- [ ] Configurable paths to be at rootless or default location
- [ ] Make usage of cgroups configurable
- [ ] Create tap device structure intended for tun/tap devices

alicefr:
Any plans to add tests for podman?

jodh-intel:
@alicefr - we're hoping to convert https://github.com/kata-containers/documentation/pull/565 into both a document and a test (by "executing" the document commands). However, if you have cycles to help write some more comprehensive tests, that would be great! :smile:

alicefr:
@jodh-intel thx!Nice! There are some tests in the podman repo at https://github.com/containers/libpod/tree/master/test. I'm not sure if they can be extended to be used with kata. I'm trying to check if podman+kata works on s390x. I'll have a look after to their testsuite.

PR #1828, opened 2019-06-25
cli: update paths to be configurable for rootless execution

gabibeyer:
Before using the default ctrsMapTrePath, check whether the runtime
is being ran rootlessly, and if so set the ctrsMapTreePath to the
rootlessRuntimeDir configured by the libpod rootless library.

Fixes: #1358

Signed-off-by: Gabi Beyer <gabrielle.n.beyer@intel.com>

### DIFF
--- a/pkg/katautils/oci.go
+++ b/pkg/katautils/oci.go
@@ -11,6 +11,8 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+
+	"github.com/kata-containers/runtime/pkg/rootless"
 )
 
 const ctrsMappingDirMode = os.FileMode(0750)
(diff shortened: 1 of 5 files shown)


jodh-intel:
We now have https://github.com/kata-containers/tests/blob/master/Unit-Test-Advice.md as a guide ;)

pkg/katautils/oci.go
```diff
@@ -11,6 +11,9 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+
+	"github.com/containers/libpod/pkg/rootless"
```
amshinde: @gabibeyer I dont see the vendoring for this. Are you planning to add the vendoring for this? 
In any case, I see that you are using just a couple of functions from the package. It is better to copy just the required function if possible in our own repo, rather than vendor the entire package.

jodh-intel: For reference, see:

- https://github.com/kata-containers/community/blob/master/VENDORING.md#vendor-a-newly-imported-package
- https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#subsystem

amshinde:
@gabibeyer I see you opened #1875 to address similar issue, is this PR still relevant?
If not, can we close this one?

gabibeyer:
@amshinde I was doing multiple separate PRs for the different components: path changes, rootless package, networking, etc. with different commits in each PR. Should I have 1 PR with all of the commits instead?

Issue #1846, opened 2019-07-01
borken url: Clear Linux staless docs


jcvenegas:
# Description of problem

https://clearlinux.org/features/stateless is not longer working, lets update the docs

# Expected result

Point to as probably a more stable URL.
 https://github.com/clearlinux/clr-man-pages/blob/master/stateless.7.rst

jodh-intel:
:heart:  it - borken is my favourite word of the week! :laughing:

gnawux:
🤔 why did people change their URLs in this week together?

jodh-intel:
I have no idea :smile: It's good to know the URL checker is actually working though.

Ideally, we'd have a daily / weekly CI run on `master` to identify these sorts of issues faster as it would be checking all the docs in a repo, not just those that a PR changed.

Related: https://github.com/kata-containers/ci/issues/101.

PR #1849, opened 2019-07-02
docs: Fix broken URL

jodh-intel:
Update broken Clear Linux stateless URL.

Fixes: #1846.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ The file contains comments explaining all options.
 > You might need to modify this file if you have specialist needs.
 
 Since the runtime supports a
-[stateless system](https://clearlinux.org/features/stateless),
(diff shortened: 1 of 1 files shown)


jodh-intel:
Hi @gnawux,

> looks there's no anchor for the title Stateless Concept: File-Level Separation, right?

Unfortunately not :sob: 

I thought it would make sense to link to the official about page which lists the features rather than the man page on github since:

- the about page (we hope) should continue to exist, being a "well-known" sort of URL that most sites provide.
- the man page looks "less official" (even though it contains all the information) as it's on a different web site to the official one.

jodh-intel:
> clagged

I just learnt a new word! I hit urbandictionary before wiktionary. Heavy hint: don't do that y'all! :smile:

PR #1857, opened 2019-07-03
shimv2: Add missing page size to Hugetlb Stat

### LINKED ISSUE #1856: Page size is missing in hugetlb Metric
# Description of problem

Page size is missing when converting to libcontainer/cgroups hugetlb metric.

https://github.com/kata-containers/runtime/blob/master/containerd-shim-v2/metrics.go#L37

# Expected result

`pageSize` should be a part of the new struct.

# Actual result

`pageSize`  is not a part of the new struct.

odinuge:
This fixes #1856

Signed-off-by: Odin Ugedal <odin@ugedal.com>


odinuge:
Hi @egernst !

Sorry, totally forgot that :upside_down_face: You do indeed have awesome contribution docs, way better than most open source projects! I really like having proper consistency in stuff like this, so keep up the good work.

Should be good to go now.

Issue #1835, opened 2019-06-29
Run `go test -v` in virtcontainers directory get error 


woshijpf:
# Description of problem

kata-version: 1.7

Since I want to run all testcases in virtcontainers directory, so I execute the following the command:
```
# cd virtcontainers
# make
# make install //install some mock binary files
# go test -v
......
=== RUN   TestCalculateSandboxMem/3-mix-constraints
=== RUN   TestCalculateSandboxMem/3-constrained
--- PASS: TestCalculateSandboxMem (0.00s)
    --- PASS: TestCalculateSandboxMem/1-unconstrained (0.00s)
    --- PASS: TestCalculateSandboxMem/2-unconstrained (0.00s)
    --- PASS: TestCalculateSandboxMem/1-constrained (0.00s)
    --- PASS: TestCalculateSandboxMem/2-constrained (0.00s)
    --- PASS: TestCalculateSandboxMem/3-mix-constraints (0.00s)
    --- PASS: TestCalculateSandboxMem/3-constrained (0.00s)
=== RUN   TestCreateSandboxEmptyID
--- PASS: TestCreateSandboxEmptyID (0.00s)
=== RUN   TestSandboxStateReadyRunning
--- PASS: TestSandboxStateReadyRunning (0.00s)
=== RUN   TestSandboxStateRunningPaused
--- PASS: TestSandboxStateRunningPaused (0.00s)
=== RUN   TestSandboxStatePausedRunning
--- PASS: TestSandboxStatePausedRunning (0.00s)
=== RUN   TestSandboxStatePausedStopped
--- PASS: TestSandboxStatePausedStopped (0.00s)
=== RUN   TestSandboxStateRunningStopped
--- PASS: TestSandboxStateRunningStopped (0.00s)
=== RUN   TestSandboxStateReadyPaused
ERRO[0001] Unable to launch /tmp/vc-tmp-526112270/hypervisor: fork/exec /tmp/vc-tmp-526112270/hypervisor: permission denied  source=virtcontainers subsystem=qmp
ERRO[0001]                                               source=virtcontainers subsystem=qmp
ok  	github.com/kata-containers/runtime/virtcontainers	1.561s
```

And then I execute the `TestSandboxStateReadyPaused` testcase alone, however it passes:
```
go test -v -run TestSandboxStateReadyPaused
INFO: Creating virtcontainers test directory /tmp/vc-tmp-785259853
=== RUN   TestSandboxStateReadyPaused
[...]

PR #1859, opened 2019-07-04
Fix fork exec test errors

jodh-intel:
Fixed `TestSandboxCreationFromConfigRollbackFromCreateSandbox` which requires that the hypervisor does not exist. Unfortunately, it does exist (as a fake test binary), but isn't executable meaning although the test failed (since an error is expected), rather than the expected `ENOENT` error, the test was logging a message similar to the following since the fake hypervisor exists with non-executable permissions:

```
Unable to launch /tmp/vc-tmp-526112270/hypervisor: fork/exec /tmp/vc-tmp-526112270/hypervisor: permission denied
```

Update the `TestQemuAddDeviceKataVSOCK` test so that it:

- Doesn't hard-code the file descriptor number.
- Cleans up after itself.

The latter issue was causing an odd error similar to the following in the test output:

```
Unable to launch /tmp/vc-tmp-526112270/hypervisor: fork/exec /tmp/vc-tmp-526112270/hypervisor: permission denied
```

Fixes: #1835.

### DIFF
--- a/virtcontainers/qemu_test.go
+++ b/virtcontainers/qemu_test.go
@@ -280,22 +280,33 @@ func TestQemuAddDeviceSerialPortDev(t *testing.T) {
 }
 
 func TestQemuAddDeviceKataVSOCK(t *testing.T) {
+	assert := assert.New(t)
+
+	dir, err := ioutil.TempDir("", "")
+	assert.NoError(err)
+	defer os.RemoveAll(dir)
+
+	vsockFilename := filepath.Join(dir, "vsock")
+
 	contextID := uint64(3)
 	port := uint32(1024)
-	vHostFD := os.NewFile(1, "vsock")
+
+	vsockFile, err := os.Create(vsockFilename)
+	assert.NoError(err)
+	defer vsockFile.Close()
 
 	expectedOut := []govmmQemu.Device{
 		govmmQemu.VSOCKDevice{
 			ID:        fmt.Sprintf("vsock-%d", contextID),
 			ContextID: contextID,
-			VHostFD:   vHostFD,
+			VHostFD:   vsockFile,
 		},
 	}
 
 	vsock := kataVSOCK{
 		contextID: contextID,
 		port:      port,
-		vhostFd:   vHostFD,
+		vhostFd:   vsockFile,
 	}
 
 	testQemuAddDevice(t, vsock, vSockPCIDev, expectedOut)
--- a/virtcontainers/sandbox_test.go
+++ b/virtcontainers/sandbox_test.go
@@ -1530,6 +1530,10 @@ func TestSandboxCreationFromConfigRollbackFromCreateSandbox(t *testing.T) {
 		Volumes:          nil,
 		Containers:       nil,
 	}
+
+	// Ensure hypervisor doesn't exist
+	assert.NoError(os.Remove(hConf.HypervisorPath))
+
 	_, err := createSandboxFromConfig(ctx, sConf, nil)
 	// Fail at createSandbox: QEMU path does not exist, it is expected. Then rollback is called
 	assert.Error(err)


jodh-intel:
Restarted the vsocks CI which fell over on a k8s issue:

```
NAME                     STATUS     ROLES    AGE   VERSION
ubuntu1804-azure011321   NotReady   master   8s    v1.14.1
No resources found.
clusterrole.rbac.authorization.k8s.io/flannel created
clusterrolebinding.rbac.authorization.k8s.io/flannel created
serviceaccount/flannel created
configmap/kube-flannel-cfg created
daemonset.extensions/kube-flannel-ds-amd64 created
daemonset.extensions/kube-flannel-ds-arm64 created
daemonset.extensions/kube-flannel-ds-arm created
daemonset.extensions/kube-flannel-ds-ppc64le created
daemonset.extensions/kube-flannel-ds-s390x created
No resources found.
   :
The connection to the server 10.0.0.25:6443 was refused - did you specify the right host or port?
   :
Failed at 88: return 1
Failed at 34: ./init.sh
Makefile:118: recipe for target 'kubernetes' failed
```

jodh-intel:
Please can we get another review on this PR? To add a little more detail, this fix resolves the immediate issue. But there is a deeper problem which looks like a bug with `golang` to me:


| environment | Number of `virtcontainers` tests run | exit code |
|-|-|-|
| *without* this PR | `299` | `0` |
| *with* this PR | `405` | `0` |

Short-term, we need this fix landed to re-enable the `106` tests that appear to be "skipped" silently by `golang`.

virtcontainers/qemu_test.go
```diff
@@ -280,22 +280,34 @@ func TestQemuAddDeviceSerialPortDev(t *testing.T) {
...
+	dir, err := ioutil.TempDir("", "")
+	assert.NoError(err)
+	defer os.RemoveAll(dir)
+
+	vsockFilename := filepath.Join(dir, "vsock")
+
 	contextID := uint64(3)
 	port := uint32(1024)
-	vHostFD := os.NewFile(1, "vsock")
+	assert.NoError(err)
```
grahamwhaley: this assert on err looks redundant - am I missing something subtle maybe?

virtcontainers/sandbox_test.go
```diff
@@ -1530,6 +1530,10 @@ func TestSandboxCreationFromConfigRollbackFromCreateSandbox(t *testing.T) {
 		Volumes:          nil,
 		Containers:       nil,
 	}
+
+	// Ensure hypervisor doesn't exist
+	assert.NoError(os.Remove(hConf.HypervisorPath))
+
 	_, err := createSandboxFromConfig(ctx, sConf, nil)
```
grahamwhaley: Given we previously failed(correctly), but with the wrong error - should we be checking for the actual error we fail with as well iyswim - to avoid such 'failed, but for the wrong reason' problems recurring?

virtcontainers/qemu_test.go
jodh-intel: Good catch - a vim bounce issue I think ;) Updated.

virtcontainers/sandbox_test.go
jodh-intel: Yeah, I do swym, but that would require string matching on an error string which is going to be fragile. The only place we do this afaik is in the mock package to assert we got the correct "type" of error:

- https://github.com/kata-containers/runtime/blob/master/virtcontainers/pkg/vcmock/utils.go#L26

For example,

- https://github.com/kata-containers/runtime/blob/master/virtcontainers/pkg/vcmock/mock_test.go#L153..L155

But you are right, we should consider in future updating the entire codebase to return a typed error, not just a simple string error as we can check for the specific error types in the tests. That would be a huge job though ;)

Alternatively, we could do something crazy like my "structured errors" idea: https://github.com/kata-containers/kata-containers/issues/16 ;)

jodh-intel:
Correction to https://github.com/kata-containers/runtime/pull/1859#issuecomment-508702367 - the existing behaviour doesn't stop those `106` tests from running, it just hides their output since the `File` created in `TestQemuAddDeviceKataVSOCK()` eventually gets closed... and since it wraps the `stdout` fd, `go test` doesn't show any further output until the final `ok`!

I don't think therefore this is a bug with `golang`, although it might be a nice feature if it were to ensure that each test run was provided with a sane set of standard streams. Ultimately though, it's the tests responsibility to maintain the original sane set :smile:

PR #1875, opened 2019-07-10
rootless: add rootless to kata

gabibeyer:
Run kata without sudo privileges. To do this various steps were taken:

1. Add capability to check whether execution is rootless or not, and provide a rootless runtime directory for rootless user read/write access.
2. Make paths to various locations configurable to point to the rootless directory prefix, so data is local to user, and permission issues do not occur.
3. Set up rootless networking, and construct new tuntap interface
4. Cgroup usage configurable depending on rootless or not

Includes README.md on configurations user must make in order to run seamlessly

Fixes: #1358


pkg/rootless/rootless.go
```diff
@@ -0,0 +1,117 @@
...
+// if unable to access or create directory fallback to the userpath.
+func setRootlessDir() error {
+	// TODO homedirpath doesn't work.
+	//homedirPath := "~/.local/kata-rootless"
+	//if err := setupDir(homedirPath); err == nil {
+	//	rootlessDir = homedirPath
+	//	return nil
+	//}
+
+	userPath := fmt.Sprintf("/run/user/%d", hostUID)
```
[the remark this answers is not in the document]
AkihiroSuda: this should use `$XDG_RUNTIME_DIR`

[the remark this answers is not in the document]
gabibeyer: Thank you @AkihiroSuda, that cleaned up a lot of code

gabibeyer:
Thank you @jodh-intel for all of the feedback and help; working on making those changes :)

gabibeyer:
@jodh-intel @ganeshmaharaj @GabyCT @amshinde @jcvenegas 
Do any of you know how to skip a golangci-lint test, or maybe fix the conversion?
```
virtcontainers/tuntap_endpoint.go:133:15: S1025: should use String() instead of fmt.Sprintf (gosimple)
				HardAddr: fmt.Sprintf("%s", hwName),
```
hwName is a []byte. The fmt does an uninterpreted print, and `string(hwName)` returns `:` as `\x`, so addresses look like: `\xea\x8e\xac\x8d\x84\xa4` instead of `ea:4f:fc:8f:84:a4` (the latter is required).

jodh-intel:
Hi @gabibeyer - you need to annotate with a magic `// nolint` comment:

- https://github.com/golangci/golangci-lint#nolint

gabibeyer:
@jodh-intel Thank you so much!! :open_hands: :dancer:

pkg/rootless/rootless.go
```diff
@@ -0,0 +1,79 @@
...
+		if line == nil {
+			return nil
+		}
+
+		// if the container id (id[0]) is 0 (root inside the container)
+		// has a mapping to the host id (id[1]) that is not root, then
+		// it can be determined that the host user is running rootless
+		ids := strings.Fields(string(line))
+		if ids[0] == "0" && ids[1] != "0" {
+			rootlessLog.Warnf("Running as rootless")
```
jodh-intel: Nit: You can make this a `Warn()` as you haven't specified any formats in the string. But should this be a warning or an `Info()` message?

pkg/rootless/rootless_test.go
```diff
@@ -0,0 +1,84 @@
...
+//
+
+package rootless
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	//"github.com/sirupsen/logrus"
```
jodh-intel: Commented out code.

pkg/rootless/rootless_test.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// TestSetRootlessRootlessUID1000 tests that isRootless is set to
+// true when a host UID is not 0
+func TestSetRootlessRootlessUID1000(t *testing.T) {
+	// by default isRootless should be set to false initially
+	assert.False(t, IsRootless())
+
+	uidMapPathStore := uidMapPath
+	uidMapPath = "/tmp/testUIDMapFile"
```
jodh-intel: It would be safer to avoid hard-coding a path like this as it might already exist. I'd use something like:

```go
assert := assert.New(t)
tmpDir, err := ioutil.TempDir("", "")  
assert.NoError(err)
defer os.RemoveAll(tmpDir)

uidMapPath := filepath.Join(tmpDir, "testUIDMapFile")
```

pkg/rootless/rootless_test.go
```diff
@@ -0,0 +1,84 @@
...
+		uidMapPath = uidMapPathStore
+		os.RemoveAll(uidMapPath)
+		isRootless = false
+	}()
+
+	mapping := "\t0\t1000\t5555"
+	err := createTestUIDMapFile(mapping)
+	if err != nil {
+		fmt.Printf("Could not create test file: %v", err)
+		os.Exit(1)
```
jodh-intel: Tests shouldn't really exit - just use assert:

```go
assert.NoError(err)
```

gabibeyer: even if it technically not a test, but instead a setup for a test?

pkg/rootless/rootless_test.go
gabibeyer: ahh thank you!

pkg/rootless/rootless_test.go
jodh-intel: As far as go is concerned, this is a test as the function name starts with `Test`. If you want to perform test setup, either create a `func init()` to do that, or create a [`TestMain()`](https://golang.org/pkg/testing/#hdr-Main) function like:

- https://github.com/kata-containers/runtime/blob/master/virtcontainers/virtcontainers_test.go#L87

virtcontainers/container.go
```diff
@@ -873,8 +874,11 @@ func (c *Container) create() (err error) {
 	}
 	c.process = *process
 
-	if err = c.newCgroups(); err != nil {
-		return
+	// cgroups are not yet supported with rootless execution
```
amshinde: So is this a system/kernel limitation?

virtcontainers/network.go
```diff
@@ -1378,12 +1394,15 @@ func createEndpoint(netInfo NetworkInfo, idx int, model NetInterworkingModel) (E
 		} else if netInfo.Iface.Type == "tap" {
 			networkLogger().Info("tap interface found")
 			endpoint, err = createTapNetworkEndpoint(idx, netInfo.Iface.Name)
+		} else if netInfo.Iface.Type == "tuntap" {
```
amshinde: You also need to check the Mode for the interface to determine is this is a `tap` or `tun` device:
https://github.com/vishvananda/netlink/blob/df0136952828fc27513ab7b03b3423e80f53e8cb/link.go#L310
Note: if you are seeing the Mode as 0, then it means the Mode was not determined correctly.
Reason being you are running a kernel < 4.13, and so the Mode was not determined through netlink. In that case, you may need to mount /sys.

virtcontainers/container.go
gabibeyer: I think it depends on if kata supports [cgroups v2](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34a9304a96d6351c2d35dcdc9293258378fc0bd8), and if so would need to work with podman to get that working. It looks like they have an issue open on getting that working with runc:
https://github.com/containers/libpod/issues/1429

virtcontainers/store/filesystem_backend.go
```diff
@@ -65,15 +66,33 @@ const VMPathSuffix = "vm"
 
 // ConfigStoragePath is the sandbox configuration directory.
 // It will contain one config.json file for each created sandbox.
-var ConfigStoragePath = filepath.Join("/var/lib", StoragePathSuffix, SandboxPathSuffix)
+var ConfigStoragePath = func() string {
```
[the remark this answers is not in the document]
gabibeyer: I did it this way, so that in the unit tests I could set a different function to modify the return - sort of a mocking the function call. Is there a better way of doing that in unit tests?!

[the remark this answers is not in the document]
amshinde: @gabibeyer Didnt realise you were using them in the unit tests, we have done this differently in the past, but this looks good to me as well. Can you just add a comment here, that the function was declared this way for mocking in the unit tests for future reference.

jcvenegas:
@gabibeyer  ping PR needs rebase.

jodh-intel:
@gabibeyer - any update? This seems to have gone stale so please can you rebase to resolve the conflict and address the remaining comments.

gabibeyer:
Been working on getting the podman side of things resolved first. Will get to this ASAP

virtcontainers/network.go
gabibeyer: I'm a little confused about how the `Mode` will determine whether its a `tap` or `tun`, and how to get the `Mode` since all we have is the `NetworkInfo`. I will add a check for if the mode is 0, and then try mounting `/sys/` for /`sys/class/net`

egernst:
@gabibeyer it looks like you didn't have "permissions" to kick off the tests.
1) we should fix that but adding you,
2) we should probably report an error when user doesn't have permission.
@chavafg is this feasible with the existing plugin?

virtcontainers/container.go
egernst: Interesting change.  Does podman manage any cgroups that we execute in?

/cc @jcvenegas

virtcontainers/network.go
egernst: @amshinde can you clarify this further?

gabibeyer: @amshinde I ran on kernel 4.12 fedora 27, and it did not require the mode to get the links being passed. I haven't found a moment when mode 0 is being reached. Would you still like me to add the logic to mount /sys if that occurs?

gabibeyer: I should mention that the Mode is 0 for the older kernels, but if I don't check the mode and instead check the type it works by creating the tuntap endpoint as designed.

amshinde: @gabibeyer The Link.Type() is just going to tell you that this is a Tuntap device, the mode for the Tuntap struct tells you specifically whether this is a tun device or tap. We need to make that distinction.
I had tried this on 4.13.0 kernel  on Ubuntu where netlink did not give the information.
In that case, it is necessary to get that information from sysfs. However, it is the upper layer that should be mounting this instead of kata.
I know for sure rootless docker does this, is this is missing from podman that it should be added there.

gabibeyer: https://github.com/containers/libpod/pull/3831

gabibeyer:
@kata-containers/architecture-committee this PR is ready to be re-evaluated. I believe I resolved all conversations. I was hoping the Podman changes would be in place before this merge, but that is taking longer than anticipated, and the ordering shouldn't really matter. Thank you in advance

virtcontainers/kata_agent.go
```diff
@@ -132,6 +129,40 @@ const (
 	grpcStopTracingRequest       = "grpc.StopTracingRequest"
 )
 
+// The function is declared this way for mocking in unit tests
+var kataHostSharedDir = func() string {
+	path := "/run/kata-containers/shared/sandboxes/"
```
jodh-intel: I think we need to find another solution here as we shouldn't be hard-coding paths like this in non-test code. Maybe add some sort of test setup function? Same comment for functions below.

pkg/rootless/rootless.go
```diff
@@ -0,0 +1,103 @@
...
+func SetLogger(ctx context.Context, logger *logrus.Entry) {
+	fields := rootlessLog.Data
+	rootlessLog = logger.WithFields(fields)
+}
+
+// setRootless reads a uid_map file, compares the UID of the
+// user inside the container vs on the host. If the host UID
+// is not root, but the container ID is, it can be determined
+// the user is running rootlessly.
+func setRootless() error {
```
jodh-intel: I think this function would benefit from a set of [table-based](https://github.com/kata-containers/tests/blob/master/Unit-Test-Advice.md#table-driven-tests) unit tests to exercise:

- corrupt / invalid `uidMapPath` file.
- all sorts of interesting host+container uid combos (particularly focusing on boundary values [`0`, `1`, max value, `max-1` *etc*]).

virtcontainers/tuntap_endpoint.go
```diff
@@ -0,0 +1,213 @@
+// Copyright (c) 2018 Huawei Corporation
```
jodh-intel: - Is this copyright year still correct?
- Shouldn't we add a SoB for the original author in the commit?

marcov:
Hi @gabibeyer, this need a rebase (again)

marcov:
The **jenkins-ci-ubuntu-nemu** build history has plenty of failure for all recent PRs.

When running docker intergration tests, we get this:
```
Stderr: docker: Error response from daemon: OCI runtime create failed: rpc error: code = Internal desc = Could not mount none to /run/kata-containers/shared/containers/: no such device: unknown.
```

Wondering what we can do, or if this can be ignored :thinking:

marcov:
@amshinde @egernst could have another review at this, now that all tests are passing?

virtcontainers/tuntap_endpoint.go
egernst: agreed....

@marcov FYI.

virtcontainers/kata_agent.go
egernst: It seems that today its hardcoded in this same file already, right? See https://github.com/kata-containers/runtime/blob/master/virtcontainers/kata_agent.go#L59-L62

WDYT @jodh-intel ?


How are these handled/specified today?  Seems like the initial path assignment is problematic, and we should be using

pkg/rootless/rootless.go
```diff
@@ -0,0 +1,79 @@
...
+			rootlessLog.Infof("Running as rootless")
+			isRootless = true
+			return nil
+		}
+	}
+}
+
+// IsRootless states whether kata is being ran with root or not
+func IsRootless() bool {
+	return isRootless
```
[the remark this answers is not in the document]
egernst: Addressed? @amshinde ACK?

pkg/rootless/rootless.go
egernst: Looks like a couple of tests exist in this PR, but we could use a case sto test the uidMapPath being invalid.

@marcov is this something you can address?

egernst:
There are just a couple small/quick issues that are still outstanding. @marcov if you are willing to address, that'd be great.  Otherwise I can see if we can find someone...

virtcontainers/tuntap_endpoint.go
marcov: This is a derived work of 
https://github.com/kata-containers/runtime/blob/master/virtcontainers/tap_endpoint.go

Can we leave `2018 Huawei` and add `2019 Intel...` ?

marcov:
@egernst sure, I can start working on that.

virtcontainers/kata_agent.go
jodh-intel: Yes, we already have `kataHostSharedDir` and `kataGuestSharedDir` globals for these paths, so why can't the mocking code either use or change those variables for testing purposes? I don't understand the need for the duplication - added to which it looks very odd to me.

Any thoughts @marcov?

virtcontainers/tuntap_endpoint.go
jodh-intel: Sounds reasonable to me.

virtcontainers/kata_agent.go
marcov: The comment on the variables is wrong. The reason to have functions in place of const is to have those directories placed in a path accessible by regular users when running rootless.

As alternative, we can leave the const declaration as before, and wherever they are used do something like:
```golang
rootless.ResolvePath(kataHostSharedDir)
```
But I like more how it is done now.

marcov:
Added a few more tests since coverage went too low

/test

marcov:
@egernst , @jodh-intel : can you please take another look?
codecov is still red, and if that's gating I'll try to add a few more tests.

PR #1887, opened 2019-07-18
qemu: use x-ignore-shared to implement vm template

bergwolf:
backport master commit d14968b66a7beca87e9c778a522e44fd2988d0cc to stable-1.8

qemu upstream has x-ignore-shared that works similar
to our private bypass-shared-memory. We can use it to
implement the vm template feature.


egernst:
this will be reliant on the other QEMU patches which we are backporting.

Let's close this PR and I'll include the commit in my stable 1.8 backport PR.

PR #1889, opened 2019-07-18
persist: manage "hypervisor.json" with new store

### LINKED ISSUE #803: [RFC] baseline persist data for live-upgrade

WeiZhang555:
Fixes #803

This PR moved 3 files to new store, now new store can manage more parts:

* hypervisor.json
* agent.json
* network.json


virtcontainers/qemu.go
```diff
@@ -1837,3 +1849,57 @@ func (q *qemu) toGrpc() ([]byte, error) {
...
+	for _, cpu := range q.state.HotpluggedVCPUs {
+		s.HotpluggedVCPUs = append(s.HotpluggedVCPUs, persistapi.CPUDevice{
+			ID: cpu.ID,
+		})
+	}
+	return
+}
+
+func (q *qemu) load(s persistapi.HypervisorState) {
+	q.state.UUID = s.UUID
```
bergwolf: qemu pid is not restored. Then why do we need to persist it?

virtcontainers/sandbox.go
```diff
@@ -574,8 +572,20 @@ func newSandbox(ctx context.Context, sandboxConfig SandboxConfig, factory Factor
 		}
```
bergwolf: nit: the defer function should be combined with the above one.

virtcontainers/sandbox.go
```diff
@@ -574,8 +572,20 @@ func newSandbox(ctx context.Context, sandboxConfig SandboxConfig, factory Factor
...
 	}()
 
-	if err = s.hypervisor.createSandbox(ctx, s.id, s.networkNS, &sandboxConfig.HypervisorConfig, s.store); err != nil {
-		return nil, err
+	if s.supportNewStore() {
+		s.devManager = deviceManager.NewDeviceManager(sandboxConfig.HypervisorConfig.BlockDeviceDriver, nil)
+
+		// Ignore the error. Restore can fail for a new sandbox
+		s.Restore()
```
bergwolf: Why is it safe to ignore errors here? The old store relies on restore to determine if it should really create new guest. What is the workflow of newstore here?

virtcontainers/qemu.go
WeiZhang555: fc and ACRN use it.

virtcontainers/sandbox.go
WeiZhang555: Yep, nice catch!

virtcontainers/sandbox.go
WeiZhang555: They're almost same. `s.Restore()` always fail when it's first sandbox.

https://github.com/kata-containers/runtime/pull/1889/files#diff-bca2c2b5b04e0e74533b492fd23a3cc8R478 

this part will check if `s.state.State==""`, if state is empty, then we need to create a new sandbox; else this is an existing sandbox.

WeiZhang555:
/test

ping @kata-containers/runtime @bergwolf @sboeuf @egernst 

Please help review it, before I make the PR too huge...

virtcontainers/bridgedmacvlan_endpoint.go
```diff
@@ -115,3 +116,52 @@ func (endpoint *BridgedMacvlanEndpoint) HotAttach(h hypervisor) error {
...
+
+func (endpoint *BridgedMacvlanEndpoint) save() (s persistapi.NetworkEndpoint) {
+	s.Type = string(endpoint.Type())
+	s.BridgedMacvlan = &persistapi.BridgedMacvlanEndpoint{
+		NetPair: persistapi.NetworkInterfacePair{
+			TapInterface: persistapi.TapInterface{
+				ID:   endpoint.NetPair.TapInterface.ID,
+				Name: endpoint.NetPair.TapInterface.Name,
+				TAPIface: persistapi.NetworkInterface{
+					Name:     endpoint.NetPair.TapInterface.TAPIface.Name,
```
jodh-intel: It would be clearer if you create a variable to refer to `endpoint.NetPair` (and maybe others for referring to `endpoint.NetPair.TapInterface` and `endpoint.NetPair.VirtIface`) I think.

virtcontainers/bridgedmacvlan_endpoint.go
```diff
@@ -115,3 +116,52 @@ func (endpoint *BridgedMacvlanEndpoint) HotAttach(h hypervisor) error {
 func (endpoint *BridgedMacvlanEndpoint) HotDetach(h hypervisor, netNsCreated bool, netNsPath string) error {
 	return fmt.Errorf("BridgedMacvlanEndpoint does not support Hot detach")
 }
+
+func (endpoint *BridgedMacvlanEndpoint) save() (s persistapi.NetworkEndpoint) {
+	s.Type = string(endpoint.Type())
+	s.BridgedMacvlan = &persistapi.BridgedMacvlanEndpoint{
```
jodh-intel: Nit: Would be clearer to add a blank line after `s.Type = ...`.

virtcontainers/kata_agent.go
```diff
@@ -2070,3 +2075,14 @@ func (k *kataAgent) cleanup(id string) {
 		k.Logger().WithError(err).Errorf("failed to cleanup vm share path %s", path)
 	}
 }
+
+func (k *kataAgent) save() (s persistapi.AgentState) {
+	s.ProxyPid = k.state.ProxyPid
+	s.URL = k.state.URL
+	return
```
jodh-intel: This is of course correct, but would look more "conventional" as:

```go
return persistapi.AgentState{
    ProxyPid: k.state.ProxyPid,
    URL: k.state.URL,
}
```

Same comment for `load()` below.

virtcontainers/macvtap_endpoint.go
```diff
@@ -102,3 +104,17 @@ func (endpoint *MacvtapEndpoint) SetPciAddr(pciAddr string) {
 func (endpoint *MacvtapEndpoint) NetworkPair() *NetworkInterfacePair {
 	return nil
 }
+
+func (endpoint *MacvtapEndpoint) save() (s persistapi.NetworkEndpoint) {
+	s.Type = string(endpoint.Type())
+	s.Macvtap = &persistapi.MacvtapEndpoint{
```
jodh-intel: Again, maybe just:

```go
return &persistapi.MacvtapEndpoint{
  // ...
}
```

virtcontainers/persist.go
```diff
@@ -229,6 +260,38 @@ func (c *Container) loadContProcess(cs persistapi.ContainerState) {
...
+		case BridgedMacvlanEndpointType:
+			ep = &BridgedMacvlanEndpoint{}
+		case MacvtapEndpointType:
+			ep = &MacvtapEndpoint{}
+		case TapEndpointType:
+			ep = &TapEndpoint{}
+		case IPVlanEndpointType:
+			ep = &IPVlanEndpoint{}
+		default:
+			continue
```
jodh-intel: I wonder if it's worth adding atleast a warning log message here as this scenario shouldn't occur I think?

virtcontainers/persist/api/hypervisor.go
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2016 Intel Corporation
```
jodh-intel: Incorrect copyright header?

virtcontainers/qemu.go
```diff
@@ -252,7 +254,17 @@ func (q *qemu) setup(id string, hypervisorConfig *HypervisorConfig, vcStore *sto
...
-	if err = q.store.Load(store.Hypervisor, &q.state); err != nil {
+	var create bool
+	if q.store != nil { //use old store
+		if err := q.store.Load(store.Hypervisor, &q.state); err != nil {
+			// hypervisor doesn't exist, create new one
+			create = true
+		}
+	} else if q.state.UUID == "" { // new store
+		create = true
+	}
```
jodh-intel: This set of tets could do with a unit test if possible?

virtcontainers/sandbox.go
jodh-intel: How about capturing the error and adding it to a `logger.WithError(err).Debug("found expected restore error")`?

WeiZhang555: @jodh-intel 
Actually I think it can be expected or unexpected, we can't know so we can't print debug message saying "it's expected"

virtcontainers/bridgedmacvlan_endpoint.go
WeiZhang555: ok

virtcontainers/kata_agent.go
WeiZhang555: Nice, let me do this.

virtcontainers/macvtap_endpoint.go
WeiZhang555: ok

virtcontainers/persist.go
WeiZhang555: Yep, it's good idea.

virtcontainers/qemu.go
WeiZhang555: Not sure how, let me take a shot.

virtcontainers/bridgedmacvlan_endpoint.go
WeiZhang555: The copy is deep so it's hard to be clear, I don't think separate some variables would help 😂

WeiZhang555: I've separated common parts to several functions, and added unit tests.
Please take another look. @jodh-intel

virtcontainers/persist.go
jodh-intel: Thanks. This should probably be a `Warn()` though since we can recover from it (by ignoring it ;)

virtcontainers/sandbox.go
jodh-intel: Ack ;)

WeiZhang555:
Thanks! @egernst @jodh-intel @bergwolf

PR #1893, opened 2019-07-18
# Kata Containers 1.7.4

egernst:
- runtime: Enable file based backend
- backport: Verify and block merge if other projects are not updated

e61e8d7 runtime: Enable file based backend
064cb4b ci: Verify and block merge if other projects are not updated


egernst:
Is the initrd failure familiar?  /cc @GabyCT

PR #1899, opened 2019-07-19
Fix UT failures with non-root

bergwolf:
1. 
```
=== RUN   TestStartNetworkMonitor
--- FAIL: TestStartNetworkMonitor (0.00s)
        Error Trace:    sandbox_test.go:1481
        Error:          Expected nil, but got: &errors.errorString{s:"Error switching to ns /proc/6648/task/6651/ns/net: operation not permitted"}
```
2.
```
=== RUN   TestBindUnmountContainerRootfsENOENTNotError
--- FAIL: TestBindUnmountContainerRootfsENOENTNotError (0.00s)
        Error Trace:    mount_test.go:360
        Error:          Expected nil, but got: 0x1
```

They should both be skipped with non-root.

### DIFF
--- a/virtcontainers/mount_test.go
+++ b/virtcontainers/mount_test.go
@@ -344,6 +344,9 @@ func TestIsEphemeralStorage(t *testing.T) {
 // or directory attempting to be unmounted doesn't exist, then it
 // is not considered an error
 func TestBindUnmountContainerRootfsENOENTNotError(t *testing.T) {
+	if os.Getuid() != 0 {
+		t.Skip("Test disabled as requires root user")
+	}
 	testMnt := "/tmp/test_mount"
 	sID := "sandIDTest"
 	cID := "contIDTest"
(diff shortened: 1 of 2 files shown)


virtcontainers/mount_test.go
```diff
@@ -344,6 +344,9 @@ func TestIsEphemeralStorage(t *testing.T) {
 // or directory attempting to be unmounted doesn't exist, then it
 // is not considered an error
 func TestBindUnmountContainerRootfsENOENTNotError(t *testing.T) {
+	if os.Getuid() != 0 {
+		t.Skip("Test disabled as requires root user")
+	}
```
jodh-intel: This could be the more general:

```go
if tc.NotValid(ktu.NeedRoot()) {
    t.Skip(ktu.TestDisabledNeedRoot)         
}
```

___

Thought: In fact, we could simplify the constraints API to call `t.Skip()` itself:

```go
func (tc *TestConstraint) SkipIfNotValid(t *testing.T, constraints ...Constraint) {
    if tc.NotValid(t, constraints...) {
        t.Skip()
   }
}
```

But that would require a bit more rework to auto-generate the message passed to `t.Skip()`.

PR #1904, opened 2019-07-22
ci: Allow travis to use go install script

### LINKED ISSUE #1903: Allow travis to use go install script
Allow travis to use the go install script instead of having a hard coded golang version at travis.yml

GabyCT:
This allows travis to use the go install script instead of having a
hard coded golang version at travis.yml

Fixes #1903

Signed-off-by: Gabriela Cervantes <gabriela.cervantes.tellez@intel.com>

### DIFF
--- a/.ci/install_go.sh
+++ b/.ci/install_go.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+cidir=$(dirname "$0")
+source "${cidir}/lib.sh"
+
+clone_tests_repo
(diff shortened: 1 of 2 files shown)


.ci/install_go.sh
```diff
@@ -0,0 +1,16 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+cidir=$(dirname "$0")
+source "${cidir}/lib.sh"
+
+clone_tests_repo
+
+pushd "${tests_repo_dir}"
```
grahamwhaley: nit: I guess we could combine the push/run/pop into a single full path run (that is, drop the push/pop and just prepend the `${tests_repo_dir}` to the command.

PR #1910, opened 2019-07-24
Remove nested vendor dir

### LINKED ISSUE #1909: Get rid of nested vendor directories
dep's recommended approach is not to have nested vendor directories:
https://github.com/golang/dep/issues/985

amshinde:
It is not really recommended to have nested vendor directories.
   dep does not work well with nested directories:
  https://github.com/golang/dep/issues/985
  Recommendation is to use flatten the vendor directories.
  Hence remove any nested vendor directories.
    
   Fixes #1909


amshinde:
@ganeshmaharaj @jodh-intel Ptal.
If this looks ok, we can do the same for the tests repo as well.

PR #1924, opened 2019-08-01
virtiofs: fix virtiofs crash when cache=none

ganeshmaharaj:
When virtio_fs_cache is set to none, the mount options for the folder
inside the guest should not contain the dax option else it leads to
invalid address errors and a crash of the daemon on the host.

Fixes: #1907

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -66,24 +66,26 @@ var (
 	kataGuestSandboxDir   = "/run/kata-containers/sandbox/"
 	type9pFs              = "9p"
 	typeVirtioFS          = "virtio_fs"
+	typeVirtioFSNoCache   = "none"
 	vsockSocketScheme     = "vsock"
 	// port numbers below 1024 are called privileged ports. Only a process with
 	// CAP_NET_BIND_SERVICE capability may bind to these port numbers.
-	vSockPort                = 1024
-	kata9pDevType            = "9p"
-	kataMmioBlkDevType       = "mmioblk"
-	kataBlkDevType           = "blk"
-	kataSCSIDevType          = "scsi"
-	kataNvdimmDevType        = "nvdimm"
-	kataVirtioFSDevType      = "virtio-fs"
-	sharedDir9pOptions       = []string{"trans=virtio,version=9p2000.L,cache=mmap", "nodev"}
-	sharedDirVirtioFSOptions = []string{"default_permissions,allow_other,rootmode=040000,user_id=0,group_id=0,dax,tag=" + mountGuest9pTag, "nodev"}
-	shmDir                   = "shm"
-	kataEphemeralDevType     = "ephemeral"
-	ephemeralPath            = filepath.Join(kataGuestSandboxDir, kataEphemeralDevType)
-	grpcMaxDataSize          = int64(1024 * 1024)
-	localDirOptions          = []string{"mode=0777"}
-	maxHostnameLen           = 64
+	vSockPort                   = 1024
+	kata9pDevType               = "9p"
+	kataMmioBlkDevType          = "mmioblk"
+	kataBlkDevType              = "blk"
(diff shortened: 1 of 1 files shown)


virtcontainers/kata_agent.go
```diff
@@ -66,24 +66,26 @@ var (
...
+	vSockPort                   = 1024
+	kata9pDevType               = "9p"
+	kataMmioBlkDevType          = "mmioblk"
+	kataBlkDevType              = "blk"
+	kataSCSIDevType             = "scsi"
+	kataNvdimmDevType           = "nvdimm"
+	kataVirtioFSDevType         = "virtio-fs"
+	sharedDir9pOptions          = []string{"trans=virtio,version=9p2000.L,cache=mmap", "nodev"}
+	sharedDirVirtioFSOptions    = []string{"default_permissions,allow_other,rootmode=040000,user_id=0,group_id=0,tag=" + mountGuest9pTag, "nodev"}
+	sharedDirVirtioFSDaxOptions = "dax"
```
devimc: really hard to identify what was the change here

virtcontainers/kata_agent.go
```diff
@@ -754,6 +756,9 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
 		// (resolv.conf, etc...) and potentially all container
 		// rootfs will reside.
 		if sandbox.config.HypervisorConfig.SharedFS == config.VirtioFS {
+			if sandbox.config.HypervisorConfig.VirtioFSCache != typeVirtioFSNoCache {
+				sharedDirVirtioFSOptions = append(sharedDirVirtioFSOptions, sharedDirVirtioFSDaxOptions)
```
devimc: could you please add a comment here explaining why?

virtcontainers/kata_agent.go
ganeshmaharaj: I know. I wish golang wouldn't keep moving every line to match every other line. The change is an one line addition.
`sharedDirVirtioFSDaxOptions = "dax"`

devimc: and remove `dax` from `sharedDirVirtioFSOptions`

ganeshmaharaj:
@jcvenegas @devimc the initrd fails is the one a revert is going to fix, right?

virtcontainers/kata_agent.go
```diff
@@ -754,6 +756,14 @@ func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
 		// (resolv.conf, etc...) and potentially all container
 		// rootfs will reside.
 		if sandbox.config.HypervisorConfig.SharedFS == config.VirtioFS {
+			// If virtio-fs uses either of the two cache options 'auto, always',
+			// the guest directory can be mounted with option 'dax' allowing it to
+			// directly map contents from the host. When set to 'none', the mount
+			// options should not contain 'dax' lest the virtio-fs daemon crashing
+			// with an invalid address reference.
```
amshinde: @ganeshmaharaj Do we know why we get the invalid address reference?

ganeshmaharaj: @amshinde cause the guest thinks the memory region is shared between the host daemon and the guest and expects the host to know memory mapping to that region but the host process was started with `cache=none` and does not understand the translation.

ganeshmaharaj:
metrics CI seems to be failing with below error.
```
08:06:18 Error: Error downloading packages:
08:06:18   Curl error (16): Error in the HTTP2 framing layer for https://mirrors.fedoraproject.org/metalink?repo=updates-released-f30&arch=x86_64 []
08:06:18 The command '/bin/sh -c dnf install -y qemu-img parted gdisk e2fsprogs gcc xfsprogs findutils' returned a non-zero code: 1
```
@chavafg any chance you have seen this in the recent past?

chavafg:
not really, maybe a network issue in the infra? I sent a restart, hopefully it doesn't fail this time

Issue #1929, opened 2019-08-05
kata-shim-v2 leak when hypervisor exit unexpectly


Ace-Tang:
# Description of problem

Use `ctr` to re-produce, if hypervisor(qemu here) exit unexpectly, shim will leak

```
1. ctr -a /run/containerd.sock  run -d  --runtime=kata.v2  docker.io/library/busybox:latest test-kata-2  
2. kill 119490   -> kill qemu
```
kata shim v2 will leak

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

PR #1930, opened 2019-08-05
shim-v2: fix shim leak when hypervisor exit unexpectly

Ace-Tang:
Fixes: #1929

in containerd-kata-v2, container can only be deleted in Delete
interface, or other shim operates(like kill/delete) all fails
since can not get container info.

### DIFF
--- a/containerd-shim-v2/wait.go
+++ b/containerd-shim-v2/wait.go
@@ -117,7 +117,6 @@ func watchSandbox(s *service) {
 			}
 		}
 	}
-	s.containers = make(map[string]*container)
 
 	// Existing container/exec will be cleaned up by its waiters.
 	// No need to send async events here.


devimc:
thanks @Ace-Tang 

cc @lifupan

lifupan:
Hi @Ace-Tang 
I can't figure out why this patch could resolve your named "shim" leak.  In my opinion it's containerd's responsible to send shutdown rpc call to stop the shimv2, but containerd didn't.

Ace-Tang:
@lifupan , when containerd call delete shim, since container has be deleted in `watchAgent`, the function Delete can not get container, then delete ops was error

lifupan:
> @lifupan , when containerd call delete shim, since container has be deleted in `watchSandbox`, the function Delete can not get container, then delete ops was error

Did/n't  https://github.com/kata-containers/runtime/blob/8ad6b0ea9312ac10bff4b6f887109657c2ac4835/containerd-shim-v2/service.go#L669 check  the container's status and return directly?

Ace-Tang:
Seems no place update container status `container.status`

Ace-Tang:
But it can be add, if you think that will better

lifupan:
> Seems no place update container status `container.status`

In theory this line https://github.com/kata-containers/runtime/blob/00e0aaa6e43cec27077484ef6700e131202c874c/containerd-shim-v2/wait.go#L68  should change the container status once the qemu process terminated, and BTW, in your code, since the sandbox has been deleted by watchSandbox, then you shouldn't get the container status from sandbox, isn' t it?

Ace-Tang:
I add set containers status in `watchSandbox`, would you check again

devimc:
restarting centos CI

Issue #1932, opened 2019-08-06
For no_proxy type, it also needs to watch hypervisor's console 


lifupan:
# Description of problem

For no proxy type such as used vsock, we also needs the options of 
watch hypervisor's console for debug. 

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

PR #1933, opened 2019-08-06
add watchconsole for no_proxy type

lifupan:
For no proxy type, we also need the feature
of watch hypervisor's console to help debug.

Fixes:#1932

### DIFF
--- a/virtcontainers/proxy.go
+++ b/virtcontainers/proxy.go
@@ -6,12 +6,22 @@
 package virtcontainers
 
 import (
+	"bufio"
 	"fmt"
+	"io"
+	"net"
 	"path/filepath"
 
 	"github.com/kata-containers/runtime/virtcontainers/store"
 	"github.com/sirupsen/logrus"
 )
+
+var buildinProxyConsoleProto = consoleProtoUnix
+
+type proxyBuiltin struct {
+	sandboxID string
+	conn      net.Conn
+}
 
 // ProxyConfig is a structure storing information needed from any
 // proxy in order to be properly initialized.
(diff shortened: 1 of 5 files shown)


virtcontainers/proxy.go
```diff
@@ -161,3 +171,90 @@ type proxy interface {
...
+		conn    net.Conn
+	)
+
+	switch proto {
+	case consoleProtoUnix:
+		conn, err = net.Dial("unix", console)
+		if err != nil {
+			return err
+		}
+		// TODO: add pty console support for kvmtools
```
devimc: could you please file an issue and mention it here?

lifupan: Hi @devimc Done.

devimc: thanks

virtcontainers/proxy.go
```diff
@@ -161,3 +171,91 @@ type proxy interface {
...
+func (p *proxyBuiltin) start(params proxyParams) (int, string, error) {
+	if params.logger == nil {
+		return -1, "", fmt.Errorf("Invalid proxy parameter: proxy logger is not set")
+	}
+
+	if p.consoleWatched() {
+		return -1, "", fmt.Errorf("kata proxy running for sandbox %s", params.id)
+	}
+
+	params.logger.Debug("Starting kata proxy")
```
amshinde: @lifupan This will be logged for the the noProxy.start() as well which is not desired.
Similarly I would suggest changing the above Error message as well to something like "Error setting up console" for the noProxy case, else the user may think that an actual proxy is running looking at the logs/errors.

lifupan: Hi @amshinde, Updated!

devimc:
restarting ubuntu and firecracker CI

Issue #1811, opened 2019-06-18
Multus CNI attempts to overwrite device name for multiple interfaces


shrmrf:
### Description of problem
I followed the [multus-cni/quickstart.md#what-if-i-want-more-interfaces](https://github.com/intel/multus-cni/blob/master/doc/quickstart.md#what-if-i-want-more-interfaces) guide. 

This configuration fails:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: samplepod
  annotations:
    k8s.v1.cni.cncf.io/networks: macvlan-conf,macvlan-conf
spec:
  containers:
  - name: samplepod
    command: ["/bin/bash", "-c", "trap : TERM INT; sleep infinity & wait"]
    image: dougbtv/centos-network
```

### Expected result
Like a normal pod, I expected to see 2 network interfaces created (perhaps called `net1` and `net2`).

### Actual result
Pod never gets created
```
...
Events:
  Type     Reason                  Age               From                Message
  ----     ------                  ----              ----                -------
  Normal   Scheduled               59s               default-scheduler   Successfully assigned default/samplepod to kadasd620
  Warning  FailedCreatePodSandBox  2s (x8 over 52s)  kubelet, kadasd620  Failed create pod sandbox: rpc error: code = Unknown desc = container create failed: rpc error: code = Internal desc = Could not add route
 dest(192.168.1.0/24)/gw()/dev(net2): file exists
```

shrmrf:
BTW, 2 seperate custom resources work fine.

_i.e, thie following comment remains valid_
>If you were to create another custom resource with the name `foo` you could use that such as: `k8s.v1.cni.cncf.io/networks: foo,macvlan-conf`, and use any number of attachments.

amshinde:
@shrmrf Was able to reproduce this.
I have raised https://github.com/kata-containers/runtime/pull/1936 and https://github.com/kata-containers/agent/pull/624 to address this issue.

shrmrf:
Hey thanks guys... sorry for the inactivity in the past two weeks.

PR #1936, opened 2019-08-06
network: Ignore routes with proto as "kernel"

amshinde:
Routes with proto "kernel" are routes that are automatically added
by the kernel.
It is a route added automatically when you assign an address to an
interface which is not /32.
With this commit, these routes are ignored. The guest kernel
would add these routes on the guest side. A corresponding commit on the
agent side would no longer delete these routes while updating them.

Without this commit, netlink gives an error complaining that a route
already exists when you try to add a route with the same dest subnet.

Something like:
dest: 192.168.1.0/24 device:net1 source:192.168.1.217 scope:253
dest: 192.168.1.0/24 device:net2 source:192.168.1.218 scope:253

Fixes: #1811

### DIFF
--- a/netmon/netmon.go
+++ b/netmon/netmon.go
@@ -307,6 +307,11 @@ func convertRoutes(netRoutes []netlink.Route) []vcTypes.Route {
 	// by Kata yet.
 	for _, netRoute := range netRoutes {
 		dst := ""
+
+		if netRoute.Protocol == unix.RTPROT_KERNEL {
+			continue
+		}
+
 		if netRoute.Dst != nil {
 			if netRoute.Dst.IP.To4() != nil {
 				dst = netRoute.Dst.String()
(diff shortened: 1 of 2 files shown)


amshinde:
@chavafg @GabyCT 
Seeing this on the debian CI:
```
11:36:18 INFO: Checking Jaeger status (and logging traces to /var/log/kata-tests/logs/traces/kata-runtime-traces.json)
11:36:18 Failed at 171: status=$(get_jaeger_status "$service")
11:36:21 jaeger
11:36:21 jaeger
11:36:21 INFO: Disabling kata tracing on /usr/share/defaults/kata-containers/configuration.toml
11:36:21 make: *** [Makefile:177: tracing] Error 56
11:36:21 Failed at 40: sudo -E PATH="$PATH" bash -c "make test"
11:36:21 Build step 'Execute shell' marked build as failure
```

Is this something that we have seen before. Going to rerun it.

Issue #1945, opened 2019-08-08
qemu vm do not exit after sending qmp quit command


xs3c:
# Description of problem

In production env, we encounter that the qemu process does  not exit after send qmp quit command, 
in especial, if VF NIC device is passed through to vm, the vm process does not exit, and unbind VF to vfio-pci, the host kernel always is in trouble.

# Expected result

The stopSandbox function guarantee that the qemu process must be dead.

# Actual result

qemu process is not dead

PR #1946, opened 2019-08-08
Try to kill qemu process and verify it has exited

xs3c:
The code is stolen from libvirt, it makes sure that
the qemu process has exited.

fixes: #1945

Signed-off-by: Yang, Wei <wei.yang1@linux.alibaba.com>

PR #1949, opened 2019-08-09
sandbox: make sure qemu process not exit after stop

Ace-Tang:
We found qemu still living for a while after stopVM successful, this
will cause a problem when we use passthrough network. In stop
container/pod, network plugin try bind vf to host kernel driver, if qemu
still running, kernel got panic

bind vf to virtio-pci driver
```
0xffffffffa02f6d50 : __vfio_group_unset_container+0x0/0xc0 [vfio]
0xffffffffa02f6e39 : vfio_group_try_dissolve_container+0x29/0x30 [vfio]
0xffffffffa02f7192 : vfio_group_put_external_user+0x12/0x70 [vfio]
0xffffffffa056aff3 : kvm_vfio_group_put_external_user+0x23/0x40 [kvm]
0xffffffffa056b10b : kvm_vfio_destroy+0x3b/0x90 [kvm]
0xffffffffa0563421 : kvm_put_kvm+0x161/0x1e0 [kvm]
0xffffffffa0563511 : kvm_vm_release+0x21/0x30 [kvm]
```

kvm release container fd

```
0xffffffffa0045900 : virtnet_probe+0x0/0x0 [virtio_net]
0xffffffffa007d6e4 : virtio_dev_probe+0x144/0x1e0 [virtio]
0xffffffff814e2e37 : driver_probe_device+0x1c7/0x410 [kernel]
0xffffffff814e31fc : __device_attach_driver+0x8c/0x100 [kernel]
0xffffffff814e0957 : bus_for_each_drv+0x67/0xb0 [kernel]
0xffffffff814e2aed : __device_attach+0xdd/0x160 [kernel]
0xffffffff814e32b3 : device_initial_probe+0x13/0x20 [kernel]
0xffffffff814e1d22 : bus_probe_device+0x92/0xa0 [kernel]
0xffffffff814df889 : device_add+0x3c9/0x620 [kernel]
```

Fixes: #1945

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/kata-containers/agent/protocols/grpc"
@@ -39,6 +40,11 @@ const (
 	// vmStartTimeout represents the time in seconds a sandbox can wait before
 	// to consider the VM starting operation failed.
 	vmStartTimeout = 10
+	// waitRetry represents retry times wait for vm exit
+	waitRetry = 50
+	// waitVMExitTime represents times wait for vm exit in one loop
+	// the unit of value is millisecond
+	waitVMExitTime = 200
 )
 
 // SandboxStatus describes a sandbox status.
@@ -1047,7 +1053,7 @@ func (s *Sandbox) startVM() (err error) {
 }
 
 // stopVM: stop the sandbox's VM
-func (s *Sandbox) stopVM() error {
+func (s *Sandbox) stopVM() (err error) {
 	span, _ := s.trace("stopVM")
 	defer span.Finish()
 
@@ -1062,6 +1068,30 @@ func (s *Sandbox) stopVM() error {
 		s.Logger().Info("Not stopping VM")
 		return nil
 	}
+	pid := s.hypervisor.pid()
+
+	defer func() {
+		if running, _ := isProcRunning(pid); !running {
+			err = nil
+			return
+		}
+		if err != nil {
+			networkLogger().WithError(err).WithField("pid", pid).Warn("Fail to stop vm")
+		}
+		syscall.Kill(pid, syscall.SIGKILL)
+		for i := 0; i < waitRetry; i++ {
+			if running, _ := isProcRunning(pid); !running {
+				err = nil
+				return
+			}
+			time.Sleep(waitVMExitTime * time.Millisecond)
+		}
+		waitTime := (waitRetry * waitVMExitTime) / (time.Second / time.Millisecond)
+		networkLogger().WithField("pid", pid).
+			Warnf("NOTE: hypervisor not exited after wait %ds", waitTime)
+
+		err = fmt.Errorf("hypervisor not exited after wait %ds", waitTime)
+	}()
 
 	s.Logger().Info("Stopping VM")
 	return s.hypervisor.stopSandbox()
--- a/virtcontainers/shim.go
+++ b/virtcontainers/shim.go
@@ -227,7 +227,7 @@ func startShim(args []string, params ShimParams) (int, error) {
 	return cmd.Process.Pid, nil
 }
 
-func isShimRunning(pid int) (bool, error) {
+func isProcRunning(pid int) (bool, error) {
 	if pid <= 0 {
 		return false, nil
 	}
@@ -253,7 +253,7 @@ func waitForShim(pid int) error {
 
 	tInit := time.Now()
 	for {
-		running, err := isShimRunning(pid)
+		running, err := isProcRunning(pid)
 		if err != nil {
 			return err
 		}
--- a/virtcontainers/api.go
+++ b/virtcontainers/api.go
@@ -584,7 +584,7 @@ func statusContainer(sandbox *Sandbox, containerID string) (ContainerStatus, err
 				container.state.State == types.StatePaused) &&
 				container.process.Pid > 0 {
 
-				running, err := isShimRunning(container.process.Pid)
+				running, err := isProcRunning(container.process.Pid)
 				if err != nil {
 					return ContainerStatus{}, err
 				}
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -981,7 +981,7 @@ func (c *Container) stop(force bool) error {
 
 		// If shim is still running something went wrong
 		// Make sure we stop the shim process
-		if running, _ := isShimRunning(c.process.Pid); running {
+		if running, _ := isProcRunning(c.process.Pid); running {
 			l := c.Logger()
 			l.Error("Failed to stop container so stopping dangling shim")
 			if err := stopShim(c.process.Pid); err != nil {
--- a/virtcontainers/shim_test.go
+++ b/virtcontainers/shim_test.go
@@ -190,7 +190,7 @@ func TestStopShimSuccessfulProcessRunning(t *testing.T) {
 
 func testIsShimRunning(t *testing.T, pid int, expected bool) {
 	assert := assert.New(t)
-	running, err := isShimRunning(pid)
+	running, err := isProcRunning(pid)
 	assert.NoError(err)
 	assert.Equal(running, expected)
 }


Ace-Tang:
Try close #1946. /cc @xs3c @gnawux @devimc 

After change wait time from 1s to 10s, the patch has run for half a month and work good.

virtcontainers/sandbox.go
```diff
@@ -1062,6 +1063,26 @@ func (s *Sandbox) stopVM() error {
 		s.Logger().Info("Not stopping VM")
 		return nil
 	}
+	pid := s.hypervisor.pid()
+
+	defer func() {
+		// if use PhysicalEndpoint, wait for hypervisor exited
+		// https://work.aone.alibaba-inc.com/issue/20349397
```
devimc: is this an internal link? I can't access it

virtcontainers/sandbox.go
```diff
@@ -1062,6 +1063,26 @@ func (s *Sandbox) stopVM() error {
...
+			if running, _ := isShimRunning(pid); !running {
+				exited = true
+				break
+			}
+			syscall.Kill(pid, syscall.SIGKILL)
+			networkLogger().Warnf("NOTE: hypervisor still running after stop %d", pid)
+			time.Sleep(200 * time.Millisecond)
+		}
+		if !exited {
+			networkLogger().Infof("NOTE: hypervisor %d not exited after 1s", pid)
```
devimc: 1s?  `50 * 200 milliseconds  = 10s` , right?

virtcontainers/sandbox.go
```diff
@@ -1062,6 +1063,26 @@ func (s *Sandbox) stopVM() error {
...
 		return nil
 	}
+	pid := s.hypervisor.pid()
+
+	defer func() {
+		// if use PhysicalEndpoint, wait for hypervisor exited
+		// https://work.aone.alibaba-inc.com/issue/20349397
+		exited := false
+		for i := 0; i < 50; i++ {
+			if running, _ := isShimRunning(pid); !running {
```
devimc: here we are relying that the `shim` dies when the hypervisor dies, this is prone to bugs

virtcontainers/sandbox.go
```diff
@@ -1062,6 +1063,26 @@ func (s *Sandbox) stopVM() error {
 		s.Logger().Info("Not stopping VM")
 		return nil
 	}
+	pid := s.hypervisor.pid()
+
+	defer func() {
+		// if use PhysicalEndpoint, wait for hypervisor exited
+		// https://work.aone.alibaba-inc.com/issue/20349397
+		exited := false
+		for i := 0; i < 50; i++ {
```
devimc: magic number

virtcontainers/sandbox.go
```diff
@@ -1062,6 +1063,26 @@ func (s *Sandbox) stopVM() error {
 		s.Logger().Info("Not stopping VM")
 		return nil
 	}
+	pid := s.hypervisor.pid()
+
+	defer func() {
+		// if use PhysicalEndpoint, wait for hypervisor exited
```
devimc: missing condition?  `if PhysicalEndpoint {`

virtcontainers/sandbox.go
```diff
@@ -1062,6 +1063,26 @@ func (s *Sandbox) stopVM() error {
...
+		// https://work.aone.alibaba-inc.com/issue/20349397
+		exited := false
+		for i := 0; i < 50; i++ {
+			if running, _ := isShimRunning(pid); !running {
+				exited = true
+				break
+			}
+			syscall.Kill(pid, syscall.SIGKILL)
+			networkLogger().Warnf("NOTE: hypervisor still running after stop %d", pid)
```
devimc: Use `WithField`

virtcontainers/sandbox.go
devimc: Use `WithField`

virtcontainers/sandbox.go
Ace-Tang: Here I want add a explain, not condition.

Ace-Tang:
Updated !

virtcontainers/sandbox.go
```diff
@@ -1062,6 +1067,24 @@ func (s *Sandbox) stopVM() error {
 		s.Logger().Info("Not stopping VM")
 		return nil
 	}
+	pid := s.hypervisor.pid()
+
+	defer func() {
+		exited := false
```
devimc: I think this variable is not needed

virtcontainers/sandbox.go
```diff
@@ -1062,6 +1067,24 @@ func (s *Sandbox) stopVM() error {
...
+		for i := 0; i < waitRetry; i++ {
+			if running, _ := isProcRunning(pid); !running {
+				exited = true
```
devimc: just `return`

virtcontainers/sandbox.go
```diff
@@ -1062,6 +1067,24 @@ func (s *Sandbox) stopVM() error {
...
+				break
+			}
+			syscall.Kill(pid, syscall.SIGKILL)
+			networkLogger().WithField("pid", pid).Warn("NOTE: hypervisor still running after stopVM")
+			time.Sleep(waitVMExitTime * time.Millisecond)
+		}
+		if !exited {
```
devimc: and just log the error

virtcontainers/sandbox.go
```diff
@@ -39,6 +40,10 @@ const (
 	// vmStartTimeout represents the time in seconds a sandbox can wait before
 	// to consider the VM starting operation failed.
 	vmStartTimeout = 10
+	// waitRetry represents retry times wait for vm exit
+	waitRetry = 50
+	// waitVMExit represents times wait for vm exit in one loop
+	waitVMExitTime = 200
```
devimc: since this is time, please convert it to milliseconds `200 * time.Millisecond` or mention that this value is in milliseconds

virtcontainers/sandbox.go
```diff
@@ -1062,6 +1067,24 @@ func (s *Sandbox) stopVM() error {
...
+	defer func() {
+		exited := false
+		for i := 0; i < waitRetry; i++ {
+			if running, _ := isProcRunning(pid); !running {
+				exited = true
+				break
+			}
+			syscall.Kill(pid, syscall.SIGKILL)
+			networkLogger().WithField("pid", pid).Warn("NOTE: hypervisor still running after stopVM")
+			time.Sleep(waitVMExitTime * time.Millisecond)
```
devimc: `time.Sleep(waitVMExitTime)`

virtcontainers/sandbox.go
```diff
@@ -1062,6 +1067,24 @@ func (s *Sandbox) stopVM() error {
...
+		}
+		if !exited {
+			networkLogger().WithField("pid", pid).Info("NOTE: hypervisor not exited after 10s")
```
devimc: `10s` can change depending of `waitRetry` and `waitVMExitTime`, could you calculate this value dynamically, something like `((waitRetry * waitVMExitTime) / time.Millisecond)`

virtcontainers/shim.go
```diff
@@ -227,7 +227,7 @@ func startShim(args []string, params ShimParams) (int, error) {
 	return cmd.Process.Pid, nil
 }
 
-func isShimRunning(pid int) (bool, error) {
+func isProcRunning(pid int) (bool, error) {
```
devimc: Thanks, but this change should be part of a different commit

Ace-Tang:
Hi, @devimc , I update. Should we skip the check if `s.hypervisor.stopSandbox()` get error ?

virtcontainers/sandbox.go
```diff
@@ -1062,6 +1068,19 @@ func (s *Sandbox) stopVM() error {
...
+	defer func() {
+		for i := 0; i < waitRetry; i++ {
+			if running, _ := isProcRunning(pid); !running {
+				return
+			}
+			syscall.Kill(pid, syscall.SIGKILL)
+			time.Sleep(waitVMExitTime * time.Millisecond)
+		}
+		networkLogger().WithField("pid", pid).
+			Warnf("NOTE: hypervisor not exited after wait %ds", (waitRetry*waitVMExitTime)/(time.Second/time.Millisecond))
```
devimc: I think this calculation is wrong `(50 * 200)÷(1 / 1000000) = 10000000000` isn't it?

virtcontainers/sandbox.go
```diff
@@ -1062,6 +1068,19 @@ func (s *Sandbox) stopVM() error {
...
+	}()
 
 	s.Logger().Info("Stopping VM")
 	return s.hypervisor.stopSandbox()
```
devimc: @Ace-Tang 
 >  I update. Should we skip the check if s.hypervisor.stopSandbox() get error ?

good catch!  I think we shouldn't skip that check, instead you can modify the signature of this function: 
```go
(s *Sandbox) stopVM() (err error) {
``` 
Log the error if `s.hypervisor.stopSandbox()` fails: 
```go
if err = s.hypervisor.stopSandbox(); err != nil {
   // Log error
}
```
 and finally modify the defer function

```go
defer func() {
    if err == nil {
        return
    }
    for i := 0; i < waitRetry; i++ {
         if running, _ := isProcRunning(pid); !running {
               err = nil
               return
          }
          .......
     // convert this is an error
     // networkLogger().WithField("pid", pid)...
    return fmt.Errorf("hypervisor not exited after ......")
```

what do you think ?

Ace-Tang: Even `s.hypervisor.stopSandbox()` return nil, the qemu may still living for a while, so I think `err
== nil`, the check is need, but if `s.hypervisor.stopSandbox()` return error, do we still need to check ? like

```
defer func() {
    if err != nil {
        // log error
        return 
    }
    for i := 0; i < waitRetry; i++ {
         if running, _ := isProcRunning(pid); !running {
               err = nil
               return
          }
....
     return fmt.Errorf("hypervisor not exited after ......")

```

devimc: @Ace-Tang 

> Even s.hypervisor.stopSandbox() return nil, the qemu may still living for a while, so I think err == nil, the check is need,

does `qmp` not shutdown the vm at all and return nil? 
if so, then `err == nil` condition is not needed, the error should be logged and the hypervisor killed, I think

```go
defer func() {
    if err != nil {
        // just log the error
    }
    for i := 0; i < waitRetry; i++ {
         if running, _ := isProcRunning(pid); !running {
               err = nil
               return
          }
.....
```

amshinde: @Ace-Tang Yes, if its the hypervisor taking some time to exit, without qmp returning any error, then we need to just log the error in case we get one and wait for the hypervisor to stop in both the error/non-error case.

Ace-Tang: > does qmp not shutdown the vm at all and return nil?

Yes, this is reason cause kernel panic, qemu still get vfio, and iommu try to unbind vf from vfio.
And Updated.

virtcontainers/sandbox.go
```diff
@@ -1062,6 +1068,26 @@ func (s *Sandbox) stopVM() error {
...
+	defer func() {
+		if err != nil {
+			networkLogger().WithError(err).WithField("pid", pid).Warn("Fail to stop vm")
+		}
+		for i := 0; i < waitRetry; i++ {
+			if running, _ := isProcRunning(pid); !running {
+				err = nil
+				return
+			}
+			syscall.Kill(pid, syscall.SIGKILL)
```
bergwolf: SIGKILL is not catchable. Please only send it once.

devimc: > Please only send it once.
 
@Ace-Tang what do you think? Have you seen cases where the first signal `SIGKILL` doesn't kill the hypervisor?

Ace-Tang: I am not seen qemu not be killed by SIGKILL

Ace-Tang: Updated, @devimc @bergwolf

virtcontainers/shim.go
devimc: @Ace-Tang could you include this change in a different commit?

Ace-Tang:
@devimc , I do split it to two commit, there are mistake here.  Sorry, :cry:  updated.

devimc:
/test

@Ace-Tang thanks

devimc:
@Ace-Tang please rebase this pr

greg-bock:
Is this similar to #1946? The patch I linked to in this [comment](https://github.com/kata-containers/runtime/pull/1946#issuecomment-521018138) cleared up similar issues we had with sriov.

devimc:
@greg-bock 

> Is this similar to #1946? 

both fix the same issue

> The patch I linked to in this comment cleared up similar issues we had with sriov.

Thanks for the patch, In your patch you are waiting for qemu to finish, but there are some scenarios where qemu hangs or takes too long to finish, maybe the `kill -9 $PID` approach is better for these cases, what do you think?

greg-bock:
> In your patch you are waiting for qemu to finish, but there are some scenarios where qemu hangs or takes too long to finish, maybe the `kill -9 $PID` approach is better for these cases, what do you think?

It has been a while since we first patched this. I'm not sure if we ever saw qemu not eventually exit with our issue, but we've never seen it hang after we applied our patch.

The use of using `new_id` and `remove_id` driver override method in  `virtcontainers/device/drivers/vfio.go`is [non-deterministic](https://www.redhat.com/archives/libvir-list/2014-April/msg00222.html) and might also cause odd issues when using sriov. I [patched](https://gist.github.com/greg-bock/f4aa76058da5ba276efbac370e9e91f5) this to use the `driver_override` method which may have removed some other corner cases when using sriov. Note that this change is likely not safe, or at the very least could leave cruft around, for an in place upgrade unless you have another CNI plugin that can do the cleanup for the old `new_id` and `remove_id` style.

devimc:
>  I patched this to use the driver_override

nice patch!, we can need it if we face odd issues with sriov

virtcontainers/sandbox.go
```diff
@@ -1062,6 +1068,30 @@ func (s *Sandbox) stopVM() error {
...
+		for i := 0; i < waitRetry; i++ {
+			if running, _ := isProcRunning(pid); !running {
+				err = nil
+				return
+			}
+			time.Sleep(waitVMExitTime * time.Millisecond)
+		}
+		waitTime := (waitRetry * waitVMExitTime) / (time.Second / time.Millisecond)
+		networkLogger().WithField("pid", pid).
+			Warnf("NOTE: hypervisor not exited after wait %ds", waitTime)
```
jodh-intel: Nit: You could specify the wait time as a structured logger field:

```go
networkLogger().WithFields(logrus.Fields{"pid": pid, "wait-time": waitTime}).Warn("hypervisor still running after wait time")
```

jodh-intel:
@Ace-Tang - All the CI's are failing this PR - could you tal?

raravena80:
@Ace-Tang any updates on this PR? Thx!

Ace-Tang:
closed， thanks for review

Issue #1958, opened 2019-08-13
do not stop a guest multiple times


bergwolf:
# Description of problem

With shimv2 + sandbox monitor, we might call stopsandbox multiple times. Only the first time it can succeed. For later stop, we can just ignore them and return success.

PR #1959, opened 2019-08-13
qemu: do not try to stop qemu multiple times

bergwolf:
We've cleaned it up the first time. Future stop will
only fail.

Fixes: #1958


devimc:
cc @Ace-Tang @xs3c

Issue #1952, opened 2019-08-12
shim hang because of grpc request to agent hang


Ace-Tang:
# Description of problem

```
/go/src/github.com/kata-containers/runtime/virtcontainers/kata_agent.go:1716 +0x94
github.com/kata-containers/runtime/virtcontainers.(*kataAgent).sendReq(0xc00047d200, 0xeb1320, 0xc000160540, 0x0, 0x0, 0x0, 0x0)
/go/src/github.com/kata-containers/runtime/virtcontainers/kata_agent.go:1812 +0x60f
github.com/kata-containers/runtime/virtcontainers.(*kataAgent).createContainer(0xc00047d200, 0xc0000ac840, 0xc0002fa820, 0x0, 0x0, 0x0)
/go/src/github.com/kata-containers/runtime/virtcontainers/kata_agent.go:1242 +0xa7f
github.com/kata-containers/runtime/virtcontainers.(*Container).create(0xc0002fa820, 0x0, 0x0)
/go/src/github.com/kata-containers/runtime/virtcontainers/container.go:860 +0xf1\
ithub.com/kata-containers/runtime/virtcontainers.(*Sandbox).CreateContainer(0xc0000ac840, 0xc00044a380, 0x40, 0xc000036050, 0xa, 0x0, 0x0, 0xc0002bc390, 0x4, 0x0, ...)
/go/src/github.com/kata-containers/runtime/virtcontainers/sandbox.go:1146 +0x14c
github.com/kata-containers/runtime/pkg/katautils.CreateContainer(0x10840e0, 0xc000504d80, 0x10a22e0, 0x19f9720, 0x10a3220, 0xc0000ac840, 0xc0002bc410, 0x9, 0x0, 0xc00042cf40, ...)
/go/src/github.com/kata-containers/runtime/pkg/katautils/create.go:214 +0x2fc
github.com/kata-containers/runtime/containerd-shim-v2.create(0x1084020, 0xc0000358c0, 0xc00047d800, 0xc0000330e0, 0x0, 0x0, 0x0)
/go/src/github.com/kata-containers/runtime/containerd-shim-v2/create.go:114 +0x998
github.com/kata-containers/runtime/containerd-shim-v2.(*service).Create(0xc00047d800, 0x1084020, 0xc0000358c0, 0xc0000330e0, 0x0, 0x0, 0x0)
[...]

PR #1962, opened 2019-08-13
agent: add default timeout for grpc requests

bergwolf:
If guest is malfunctioning, we need a way to bail out. Add
a default timeout for most of the grpc requests so that the
runtime does not wait indefinitely.

Fixes: #1952

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -54,6 +54,7 @@ const (
 
 var (
 	checkRequestTimeout   = 30 * time.Second
+	defaultRequestTimeout = 60 * time.Second
 	defaultKataSocketName = "kata.sock"
 	defaultKataChannel    = "agent.channel.0"
 	defaultKataDeviceID   = "channel0"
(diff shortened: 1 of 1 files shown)


devimc:
weird error in nemu-ci

```
[3] Stderr: docker: Error response from daemon: OCI runtime create failed: nemu-system-x86_64: -device vhost-user-fs-pci,chardev=char-244fa09c3e4ac8e7,tag=kataShared,cache-size=1024M: Failed to read msg header. Read 0 instead of 12. Original request 1.
[3] nemu-system-x86_64: -device vhost-user-fs-pci,chardev=char-244fa09c3e4ac8e7,tag=kataShared,cache-size=1024M: vhost_dev_init failed: Operation not permitted: unknown. 
```

bergwolf:
@devimc That one seems to be a virtiofsd bug and is fixed by patch "virtiofsd: simplify setup_remount_slave" that I just sent to the virtiofs list today.

devimc:
@bergwolf 
> @devimc That one seems to be a virtiofsd bug and is fixed by patch "virtiofsd: simplify setup_remount_slave" that I just sent to the virtiofs list today.

cool

PR #1964, opened 2019-08-13
qemu interaction improvements

bergwolf:
The PR improves qemu hypervisor interaction in following ways:

1. add qemu logfile so that we can check why qemu fails
2. kata's memory memory preallocation option is just a property for different memory backends. We should make it compatible with hugepage/file-backed-memroy/memory-backend-ram
3. When guest panics or stops with unexpected internal error, qemu process might still be running but we can find out such situation with qmp. Then monitor can still report such failures to watchers.

The PR has several modifications to govmm. I'll rebase once the govmm change is merged.


virtcontainers/qemu.go
```diff
@@ -721,6 +723,11 @@ func (q *qemu) startSandbox(timeout int) error {
 	if err != nil {
 		return err
 	}
+	// append logfile only on debug
+	if q.config.Debug {
+		q.qemuConfig.LogFile = filepath.Join(vmPath, "qemu.log")
```
ganeshmaharaj: If we create the log file in the vmPath and if the QEMU crashes for some reason, wouldn't the logs also get removed as part of the sandbox cleanup? Would it help if we can leave it elsewhere? (I understand that poses an issue if orphaned files on the host created by Kata, but would help for debug later wouldn't it?)

bergwolf: Good point! Another option is to read the log file and output it to the logger system.

devimc: +1 for logging to the logger system

ganeshmaharaj: That would work for sure.

vendor/github.com/intel/govmm/qemu/qmp.go
```diff
@@ -1509,3 +1515,23 @@ func (q *QMP) ExecuteMigrationIncoming(ctx context.Context, uri string) error {
...
+// ExecuteQueryStatus queries guest status
+func (q *QMP) ExecuteQueryStatus(ctx context.Context) (StatusInfo, error) {
+	response, err := q.executeCommandWithResponse(ctx, "query-status", nil, nil, nil)
+	if err != nil {
+		return StatusInfo{}, err
+	}
+
+	data, err := json.Marshal(response)
+	if err != nil {
+		return StatusInfo{}, fmt.Errorf("unable to extract migrate status information: %v", err)
```
amshinde: @bergwolf Should this be "migrate status"?

vendor/github.com/intel/govmm/qemu/qemu.go
```diff
@@ -1456,8 +1456,8 @@ type Knobs struct {
 	// be set.
 	FileBackedMem bool
 
-	// FileBackedMemShared will set the FileBackedMem device as shared.
```
amshinde: I assume you will be properly vendoring this?

virtcontainers/qemu.go
bergwolf: I've added code to print it to the logger system when debug is on. I'm not sure if it useful to give users options to configure the log file location though,  because this is only for debug.

vendor/github.com/intel/govmm/qemu/qmp.go
bergwolf: oops, that one should be fixed in govmm instead.

PR #1973, opened 2019-08-19
add virtiofsd to sandbox cgroup

bergwolf:
Right now virtiofsd daemon is not in sandbox cgroup. We should add it to the same cgroup as the qemu main process.

Before fix:
```
$sudo cat /proc/101318/cgroup
12:pids:/system.slice/containerd.service
11:rdma:/
10:net_cls,net_prio:/
9:memory:/system.slice/containerd.service
8:freezer:/
7:blkio:/system.slice/containerd.service
6:cpu,cpuacct:/system.slice/containerd.service
5:perf_event:/
4:devices:/system.slice/containerd.service
3:cpuset:/
2:hugetlb:/
1:name=systemd:/system.slice/containerd.service
0::/system.slice/containerd.service
```
After fix:
```
$sudo cat /proc/99841/cgroup
12:pids:/system.slice/containerd.service
11:rdma:/
10:net_cls,net_prio:/
9:memory:/kata/docker/kata_84b5062407c3b2a8dde2bf51391f553362fb5f81a8ac1d7d891dcdd61c832fb2
8:freezer:/
7:blkio:/system.slice/containerd.service
6:cpu,cpuacct:/system.slice/containerd.service
5:perf_event:/
4:devices:/system.slice/containerd.service
3:cpuset:/
2:hugetlb:/
1:name=systemd:/system.slice/containerd.service
0::/system.slice/containerd.service
```

### DIFF
--- a/virtcontainers/cgroups.go
+++ b/virtcontainers/cgroups.go
@@ -209,9 +209,9 @@ func (s *Sandbox) deleteCgroups() error {
 }
 
 func (s *Sandbox) constrainHypervisor(cgroup cgroups.Cgroup) error {
-	pid := s.hypervisor.pid()
-	if pid <= 0 {
-		return fmt.Errorf("Invalid hypervisor PID: %d", pid)
+	pids := s.hypervisor.getPids()
+	if len(pids) == 0 || pids[0] == 0 {
+		return fmt.Errorf("Invalid hypervisor PID: %+v", pids)
 	}
 
 	// Move hypervisor into cgroups without constraints,
@@ -222,9 +222,14 @@ func (s *Sandbox) constrainHypervisor(cgroup cgroups.Cgroup) error {
 	if err != nil {
 		return fmt.Errorf("Could not create cgroup %v: %v", path, err)
 	}
-
-	if err := noConstraintsCgroup.Add(cgroups.Process{Pid: pid}); err != nil {
-		return fmt.Errorf("Could not add hypervisor PID %d to cgroup %v: %v", pid, path, err)
+	for _, pid := range pids {
+		if pid <= 0 {
+			s.Logger().Warnf("Invalid hypervisor pid: %d", pid)
+			continue
+		}
+		if err := noConstraintsCgroup.Add(cgroups.Process{Pid: pid}); err != nil {
+			return fmt.Errorf("Could not add hypervisor PID %d to cgroup %v: %v", pid, path, err)
+		}
 	}
 
 	// when new container joins, new CPU could be hotplugged, so we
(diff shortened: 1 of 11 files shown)


virtcontainers/cgroups.go
```diff
@@ -209,9 +209,9 @@ func (s *Sandbox) deleteCgroups() error {
 }
 
 func (s *Sandbox) constrainHypervisor(cgroup cgroups.Cgroup) error {
-	pid := s.hypervisor.pid()
-	if pid <= 0 {
-		return fmt.Errorf("Invalid hypervisor PID: %d", pid)
+	pids := s.hypervisor.pid()
+	if len(pids) == 0 || pids[0] == 0 {
```
jodh-intel: But what if `len(pids) > 1 && pid[n] == 0`?

If the code requires that the "main" PID be at index zero, we probably need a comment explaining that (and maybe a unit test? :smile:)

virtcontainers/hypervisor.go
```diff
@@ -668,7 +668,7 @@ type hypervisor interface {
 	hypervisorConfig() HypervisorConfig
 	getThreadIDs() (vcpuThreadIDs, error)
 	cleanup() error
-	pid() int
+	pid() []int
```
jodh-intel: Can you document the expected PID ordering here (is `pid[0]` "special" even if there are `>1` PIDs)?

stefanha: It would be nice to choose a name for pid() that clearly communicates that this method can return multiple pids.  I see there is also a getThreadIDs() method.  How about getPids() or pids() instead of pid()?

virtcontainers/vm.go
```diff
@@ -475,7 +475,7 @@ func (v *VM) ToGrpc(config VMConfig) (*pb.GrpcVM, error) {
 
 func (v *VM) GetVMStatus() *pb.GrpcVMStatus {
 	return &pb.GrpcVMStatus{
-		Pid:    int64(v.hypervisor.pid()),
+		Pid:    int64(v.hypervisor.pid()[0]),
```
jcvenegas: needs to check if len more than 0

jcvenegas: We need to document that the array ensure the  pid of hypervisor is on index 0.
If we are planing to break the API I wonder if we could do something additional. 

- virtiofsd is not depedent of qemu we potencially use in other hypervisors.
- we may have or not additional process that associated with a sandbox that work together with the VMM.

- We have the option to add `AdditionalPids() -> Map String:int`  to the hypervisor  interface. 
And then iterate over that map without need to break the current semantic of `pid()`
 OR make `pid()` return a map just in case we want the caller need an specific field (like this case).

I prefer `AdditionalPids()`

bergwolf: @jcvenegas I'm following @stefanha's suggestion to rename it getPids(). The purpose is quite similar so let's not split it.

Issue #1974, opened 2019-08-19
vc: bind destination should be removed when unmounting


darfux:
# Description of problem

`virtcontainers.ensureDestinationExists` will create the bind destination directory/file, which should be removed properly when unmounting.

# Expected result
When a container is stopped, the created bind destinations are removed as well.

# Actual result
The files/directories still leaves in `kataHostSharedDir`

PR #1975, opened 2019-08-19
vc: Remove bind destination when unmounting

darfux:
`virtcontainers.ensureDestinationExists` will create the bind destination directory/file, which should be removed properly when unmounting.

Fixes: https://github.com/kata-containers/runtime/issues/1974

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -24,6 +24,7 @@ import (
 	"github.com/kata-containers/runtime/virtcontainers/utils"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 
@@ -610,6 +611,20 @@ func (c *Container) unmountHostMounts() error {
 					"error":     err,
 				}).Warn("Could not umount")
 				return err
+			}
+
+			if m.Type == "bind" {
+				s, err := os.Stat(m.HostPath)
+				if err != nil {
+					return errors.Wrapf(err, "Could not stat host-path %v", m.HostPath)
+				}
+				// Remove the empty file or directory
+				if s.Mode().IsRegular() && s.Size() == 0 {
+					os.Remove(m.HostPath)
+				}
+				if s.Mode().IsDir() {
+					syscall.Rmdir(m.HostPath)
+				}
 			}
 
 			span.Finish()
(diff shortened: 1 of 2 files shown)


virtcontainers/container.go
```diff
@@ -612,6 +612,24 @@ func (c *Container) unmountHostMounts() error {
 				return err
 			}
 
+			if m.Type == "bind" {
+				s, err := os.Stat(m.HostPath)
+				if err != nil {
+					c.Logger().WithFields(logrus.Fields{
+						"host-path": m.HostPath,
+						"error":     err,
```
jodh-intel: The error should be handled like this:

```go
c.Logger().WithError(err).WithFields(...)
```

But is this log call even required if the caller logs the error (I haven't checked)?

virtcontainers/container.go
```diff
@@ -612,6 +612,24 @@ func (c *Container) unmountHostMounts() error {
 				return err
 			}
 
+			if m.Type == "bind" {
```
jodh-intel: It would be great if you could add a unit test for this scenario to ` virtcontainers/container_test.go`.

darfux: Thanks for the suggestion. I'll work on it 😀

virtcontainers/container.go
darfux: > The error should be handled like this

I copied this log line from the prev codes in the func:
```golang
if err := syscall.Unmount(m.HostPath, syscall.MNT_DETACH); err != nil {
  c.Logger().WithFields(logrus.Fields{
    "host-path": m.HostPath,
    "error":     err,
  }).Warn("Could not umount")
  return err
```
So is it an old style and should be `WithError(err)` now?

> is this log call even required

I also wonder why does `syscall.Unmount` do such log here... maybe it includes the `m.HostPath` which isn't carried by the error?

darfux:
@jodh-intel I have used `errors.Wrapf` instead of the logger and added a unit test, PTAL :)

darfux:
Hi @jodh-intel, how can I restart the CI test? I have tried re-pushing and commenting with /test but it seems that both don't work..

grahamwhaley:
thanks @nitkon . @darfux - you might not have enought 'rights' on the github org to re-trigger. There are maybe three options to re-trigger:
- issue a `/ t e s t` or `/ r e t e s t`
- issue a `/ t e s t - u b u n t u` or similar if you need to trigger just one CI, and it has an individual trigger phrase
- if re-building a single failed test, follow the *details* link to the Jenkins page for that job, log into Jenkins (it uses your github acct details), and click 'rebuild' for that job. This also requires you have enough 'rights' in the kata github organisation etc.

darfux:
@grahamwhaley Thanks for the explanation 😀

PR #1982, opened 2019-08-20
Fix the issues for monitor

lifupan:


virtcontainers/monitor.go
```diff
@@ -22,22 +27,25 @@ type monitor struct {
...
 
 func (m *monitor) newWatcher() (chan error, error) {
 	m.Lock()
 	defer m.Unlock()
 
-	watcher := make(chan error, 1)
+	// In case more than one events occur, the watcher's channel
+	// should have an enough capacity to write, otherwise it would be
+	// blocked by writing to channel.
+	watcher := make(chan error, eventNum)
```
bergwolf: How about just make the channel large enough (e.g., 64 or 128) and handle channel full failure? I know @Ace-Tang has a similar fix too. PTAL.

Ace-Tang: Yes

lifupan: > How about just make the channel large enough (e.g., 64 or 128) and handle channel full failure? I know @Ace-Tang has a similar fix too. PTAL.

If we deal with those notify events as a fatal message, I think it's enough to send only one message instead of sending multi one since we didn't different the notify events in receiver; If we do want to send multi messages, then I think it's better to define different events and let the watcher to register the specified type of event watcher which they interested in.

Ace-Tang: Even we split different event to different notify, we will still get more than one message. 

> define different events and let the watcher to register the specified type of event watcher which they interested in.

This is a good idea

bergwolf: We can have well-defined error types, send whatever we get, and let users choose if they want to handle it or not.

lifupan: > Even we split different event to different notify, we will still get more than one message.
> 
No, since the different event will has a corresponding class of notify channel, thus a specified notifier will only be notified by only one type of event.

lifupan:
close this PR and there is another one fixed the issues https://github.com/kata-containers/runtime/pull/1984

Issue #1985, opened 2019-08-20
kvm kernel modules not set as required for amd64


marcov:
This results in `kata-runtime kata-check` reporting the system as capable to run Kata containers:

```
ERRO[0000] kernel property not found                     arch=amd64 description="Intel KVM" name=kvm_intel pid=20288 source=runtime type=module
WARN[0000] modprobe insert module failed: modprobe: ERROR: could not insert 'kvm': Unknown symbol in module, or unknown parameter (see dmesg)
  arch=amd64 error="exit status 1" module=kvm name=kata-runtime pid=20288 source=runtime
ERRO[0000] kernel property not found                     arch=amd64 description="Kernel-based Virtual Machine" name=kvm pid=20288 source=runtime type=module
INFO[0000] kernel property found                         arch=amd64 description="Host kernel accelerator for virtio" name=vhost pid=20288 source=runtime type=module
INFO[0000] kernel property found                         arch=amd64 description="Host kernel accelerator for virtio network" name=vhost_net pid=20288 source=runtime type=module
INFO[0000] System is capable of running Kata Containers  arch=amd64 name=kata-runtime pid=20288 source=runtime
```

PR #1986, opened 2019-08-20
kata-check: require kvm modules for amd64

marcov:
KVM modules are required when using QEMU or firecracker.

Fixes: #1985

Signed-off-by: Marco Vedovati <mvedovati@suse.com>

### DIFF
--- a/cli/kata-check_amd64.go
+++ b/cli/kata-check_amd64.go
@@ -116,17 +116,21 @@ func setCPUtype(hypervisorType vc.HypervisorType) error {
 			}
 			archRequiredKernelModules = map[string]kernelModule{
 				kernelModkvm: {
-					desc: msgKernelVM,
+					desc:     msgKernelVM,
+					required: true,
 				},
 				kernelModkvmintel: {
 					desc:       "Intel KVM",
 					parameters: kvmIntelParams,
+					required:   true,
 				},
 				kernelModvhost: {
-					desc: msgKernelVirtio,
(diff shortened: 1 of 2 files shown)


cli/kata-check_amd64.go
```diff
@@ -116,21 +116,24 @@ func setCPUtype(hypervisorType vc.HypervisorType) error {
...
+					required: true,
 				},
 				kernelModvhostnet: {
-					desc: msgKernelVirtioNet,
+					desc:     msgKernelVirtioNet,
+					required: true,
 				},
 				kernelModvhostvsock: {
-					desc:     msgKernelVirtioVhostVsock,
-					required: false,
```
devimc: missing `required: true,` ?

marcov: vosck is not required, and by default required is false, so I just removed the line as it's redundant.

devimc: I didn't know that by default it's `false`, can we keep it explicit?

marcov: That's how [Go works](https://tour.golang.org/basics/12), but I am happy to set it explicitly to `false` if this gives more clarity.

devimc: thanks @marcov

marcov:
All tests are good, but we need a pair of reviews :-)

Calling in the usual folks :grin: @jodh-intel @grahamwhaley @devimc @jcvenegas

Issue #1988, opened 2019-08-21
vsock: set the right ioctlVhostVsockSetGuestCid for ppc64le


nitkon:
ioctlVhostVsockSetGuestCid depends on the arch you are running on. Set it right if kata running on ppc64le. Else kata containers with vsock enabled fails to start.

Issue #1943, opened 2019-08-07
vsock: Kata fails to start with use_vsock = true on ppc64le


nitkon:
# Description of problem

Set `use_vsock = true` in configuration.toml and enabled `+CONFIG_BINFMT_MISC=y
 ; +CONFIG_VSOCKETS=y` in the kata kernel. 

```
root@soe13:/home/nitesh# docker run -ti --runtime kata-runtime busybox sh
<hangs>
```

# Expected result

Kata container starts without any issue

# Actual result

Fails to start. Hangs/no response after starting the container.

journalctl -f 
```
Aug 07 21:50:48 soe13 kata-runtime[77350]: time="2019-08-07T21:50:48.262428432+05:30" level=debug msg="Restricting maxmem to 32GB as Qemu Version < 2.10, Assumption: Kernel Version >= 4.11" arch=ppc64le command=create container=eb639d5a3585dbc77d2f7ec8e18f49ba123b145f845bfcd77cd29a8f3daac2f5 name=kata-runtime pid=77350 source=virtcontainers subsystem=qemu
Aug 07 21:50:48 soe13 kata-runtime[77350]: time="2019-08-07T21:50:48.262512883+05:30" level=debug msg="agent: Using vsock VM socket endpoint" arch=ppc64le command=create container=eb639d5a3585dbc77d2f7ec8e18f49ba123b145f845bfcd77cd29a8f3daac2f5 name=kata-runtime pid=77350 source=virtcontainers subsystem=kata_agent
Aug 07 21:50:48 soe13 kata-runtime[77350]: time="2019-08-07T21:50:48.262555357+05:30" level=debug msg="Could not retrieve anything from storage" arch=ppc64le command=create container=eb639d5a3585dbc77d2f7ec8e18f49ba123b145f845bfcd77cd29a8f3daac2f5 name=kata-runtime pid=77350 source=virtcontainers subsystem=kata_agent
[...]

PR #1989, opened 2019-08-21
vsock: set VHOST_VSOCK_SET_GUEST_CID for ppc64le

nitkon:
Set the right VHOST_VSOCK_SET_GUEST_CID when
running on ppc64le arch

Fixes: #1988

### DIFF
--- a/virtcontainers/utils/utils_linux.go
+++ b/virtcontainers/utils/utils_linux.go
@@ -16,10 +16,6 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-// from <linux/vhost.h>
-// VHOST_VSOCK_SET_GUEST_CID = _IOW(VHOST_VIRTIO, 0x60, __u64)
-const ioctlVhostVsockSetGuestCid = 0x4008AF60
-
 var ioctlFunc = Ioctl
 
 // maxUInt represents the maximum valid value for the context ID.
@@ -78,6 +74,7 @@ func FindContextID() (*os.File, uint64, error) {
 		}
 	}
 
+	ioctlVhostVsockSetGuestCid := getIoctlVhostVsockGuestCid()
 	// Last chance to get a free context ID.
 	for cid := contextID - 1; cid >= firstContextID; cid-- {
 		if err = ioctlFunc(vsockFd.Fd(), ioctlVhostVsockSetGuestCid, uintptr(unsafe.Pointer(&cid))); err == nil {
--- a/virtcontainers/utils/utils_linux_ppc64le.go
+++ b/virtcontainers/utils/utils_linux_ppc64le.go
@@ -1,0 +1,18 @@
+// Copyright (c) 2019 IBM
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package utils
+
+// from <linux/vhost.h>
+// VHOST_VSOCK_SET_GUEST_CID = _IOW(VHOST_VIRTIO, 0x60, __u64)
+
+// _IOC_WRITE is 1 for arch generic and 4 for powerpc
+// Code reference: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/powerpc/include/uapi/asm/ioctl.h
+// Explanation: https://github.com/kata-containers/runtime/pull/1989#issuecomment-525993135
+const ioctlVhostVsockSetGuestCid = 0x8008AF60
+
+func getIoctlVhostVsockGuestCid() uintptr {
+	return ioctlVhostVsockSetGuestCid
+}
--- a/virtcontainers/utils/utils_linux_generic.go
+++ b/virtcontainers/utils/utils_linux_generic.go
@@ -1,0 +1,16 @@
+// +build amd64 arm64 s390x !ppc64le
+
+// Copyright (c) 2019 IBM
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package utils
+
+// from <linux/vhost.h>
+// VHOST_VSOCK_SET_GUEST_CID = _IOW(VHOST_VIRTIO, 0x60, __u64)
+const ioctlVhostVsockSetGuestCid = 0x4008AF60
+
+func getIoctlVhostVsockGuestCid() uintptr {
+	return ioctlVhostVsockSetGuestCid
+}


virtcontainers/utils/utils_linux.go
```diff
@@ -71,6 +72,12 @@ func FindContextID() (*os.File, uint64, error) {
 		return nil, 0, err
 	}
 
+	if runtime.GOARCH == "ppc64le" {
```
grahamwhaley: @alicefr  @justin-he @jongwu (/cc @Pennyzct ) - I guess we should check one of these (the default??) is correct for the other arch's as well..

virtcontainers/utils/utils_linux.go
```diff
@@ -71,6 +72,12 @@ func FindContextID() (*os.File, uint64, error) {
...
+		ioctlVhostVsockSetGuestCid = 0x8008AF60
```
jodh-intel: I don't understand why this value is different, just for ppc64le...? Can you point to the kernel source which shows why this is required as looking at `linux/vhost.h` from the comment above suggests it's a generic value.

nitkon: @jodh-intel : More info here: https://github.com/kata-containers/runtime/issues/1943#issuecomment-521159458

virtcontainers/utils/utils_linux.go
nitkon: @grahamwhaley: I guess its the same for s390x and x86. This [code](https://github.com/kata-containers/runtime/issues/1943#issuecomment-521117461) suggested its different for power.  cc @alicefr @jodh-intel

grahamwhaley: @nitkon - I think what @jodh-intel is asking for (and is the definitive 'source'), would be for you to show us exactly where in the Linux kernel source tree the difference is between power and all the other architectures.... somewhere maybe in https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/powerpc and show the x86 equivalent .... I think we need to see the root source of where the difference is coming from (just for everbodies info, and so they can then check the other arch's).

jongwu: @grahamwhaley I'm not familiar with it, spare me some time to check it on arm64.

alicefr: where is the prefix "8" instead of the "4" coming from?

justin-he: > @alicefr @justin-he @jongwu (/cc @Pennyzct ) - I guess we should check one of these (the default??) is correct for the other arch's as well..

IMO, the macro definition _IOW on arm64 is arch generic. i.e. VHOST_VSOCK_SET_GUEST_CID=4008af60, 
I search the kernel codes, seems only Powerpc has different definition of _IOC_WRITE.
At least, for arm arch, there is no need to add a specific code path

virtcontainers/utils/utils_linux.go
justin-he: > I don't understand why this value is different, just for ppc64le...? Can you point to the kernel source which shows why this is required as looking at `linux/vhost.h` from the comment above suggests it's a generic value.

Please see the kernel codes "#define _IOC_WRITE	4U"  under powerpc arch.

virtcontainers/utils/utils_linux.go
alicefr: I think it's the same for s390x. No need

alicefr: @nitkon have ever vsocket worked on power?

nitkon: @alicefr : Yes it did. Please check this [comment](https://github.com/kata-containers/runtime/issues/1943#issuecomment-523350547)

alicefr: mmh, did they change the `_IOC_WRITE ` for power in the mean time or how did it worked before?

nitkon: @alicefr : Kata with vsock was never tested before on Power. This is the first time we are enabling it.

nitkon:
CI failure on Ubuntu 1804 ppc64le is being tracked here: #1995

nitkon:
Since the codecov base report is missing `
Missing base report.
Unable to compare commits because the base of the compare did not upload a coverage report.
Learn more here.
`
is there anyway to just run re-run just that? Or should I go with a `/retest`

jodh-intel:
Hi @nitkon - please add a link to the Linux kernel sources (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/ or https://github.com/torvalds/linux or similar) showing where/why this value is different for ppc.

It's not obvious why this change is only required for power, hence it's good practise to add a reference so that interested parties can look deeper to understand this superficially simple change.

Also -- and although we do now have two existing instances of it -- it's worth pointing out that we try to avoid the `if arch { do_arch_specific_stuff() }` approach. Yes, it's simpler to add such code but at the expense of complicating the general understanding of the program logic (which we try to keep "clean" by factoring out architecture-specifics).

Instead using build tags to encapsulate architecture specific code in separate files is the preferred approach. This could be handled here by creating:

- ` virtcontainers/utils/utils_linux_ppc64le.go`:

  ```go
  // Comment explaining why
  const ioctlVhostVsockSetGuestCid = 0x8008AF60
  ```

nitkon:
Thanks @jodh-intel for review. Ptal ...

mpe:
The definitive source for `_IOC_WRITE` being 4 on powerpc is:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/powerpc/include/uapi/asm/ioctl.h

The reason it's 4 seems to be because it was copied from the alpha port, which took the value from OSF/1 which I think took the value from BSD 4.3 ... in short, ancient UNIX history.

The reason it's 1 on x86 seems to be because Linus just picked a value when he wrote the original code. Other arches that were added later, eg. arm64, then used the x86 value.

To get to the actual 32-bit value you need a few more #defines, the full expansion is:

```
#define _IOC_SIZEBITS	13

#define _IOC_WRITE	4U

#define _IOC_NRSHIFT	0
#define _IOC_TYPESHIFT	(_IOC_NRSHIFT+_IOC_NRBITS)
#define _IOC_SIZESHIFT	(_IOC_TYPESHIFT+_IOC_TYPEBITS)
#define _IOC_DIRSHIFT	(_IOC_SIZESHIFT+_IOC_SIZEBITS)

#define _IOC_NRBITS	8
#define _IOC_TYPEBITS	8

val = _IOC_WRITE << _IOC_DIRSHIFT
    = 4 << (_IOC_SIZESHIFT + _IOC_SIZEBITS)
    = 4 << ((_IOC_TYPESHIFT + _IOC_TYPEBITS) + _IOC_SIZEBITS)
    = 4 << (((_IOC_NRSHIFT + _IOC_NRBITS) + _IOC_TYPEBITS) + _IOC_SIZEBITS)
    = 4 << (((0 + 8) + 8) + 13)
    = 4 << 29
    = 0x80000000
```

It would obviously be preferable if there was some way for those values to be generated from the C headers and then made available to Go code, but I'm sure I'm not the first person to say that.

virtcontainers/utils/utils_linux_ppc64le.go
```diff
@@ -0,0 +1,95 @@
...
+// to find a context ID available.
+// On success vhost file and a context ID greater or equal than 3 are returned, otherwise 0 and an error are returned.
+// vhost file can be used to send vhost file decriptor to QEMU. It's the caller's responsibility to
+// close vhost file descriptor.
+//
+// Benefits of using random context IDs:
+// - Reduce the probability of a *DoS attack*, since other processes don't know whatis the initial context ID
+//   used by findContextID to find a context ID available
+//
+func FindContextID() (*os.File, uint64, error) {
```
jodh-intel: This function is the same across all architectures so it would be good to find a way to avoid this arch-specific copy.

jodh-intel:
Thanks for the excellent explanation @mpe !

virtcontainers/utils/utils_linux_ppc64le.go
nitkon: @jodh-intel: Updated my PR. Ptal ...

nitkon:
@jodh-intel: Updated my PR. Please take a look ...

nitkon:
All tests passed. `jenkins-ci-opensuse-15` is failing due to permission denied error when adding conmon to cgroupfs sandbox cgroup.

nitkon:
Can this be merged or we wait for opensuse CI fix?

chavafg:
I think we can merge as opensuse CI issue is also failing on other PRs.

nitkon:
@chavafg : Cool then ...

Issue #1970, opened 2019-08-16
firecracker CI failing with: `dial unix: missing address: unknown.`


chavafg:
# Description of problem

Firecracker CI constantly failing with the next error:

```
[docker run --cidfile /tmp/cid852224458/SANdxNxgiIowa9ttfTxUfMcjOnOdJ1 --runtime kata-runtime --rm --name SANdxNxgiIowa9ttfTxUfMcjOnOdJ1 debian bash -c for f in /sys/fs/cgroup/*/*; do echo 100 > $f && exit 1; done; exit 0]
Timeout: 120 seconds
Exit Code: 125
Stdout: 
Stderr: docker: Error response from daemon: OCI runtime create failed: dial unix: missing address: unknown.
time="2019-08-15T18:11:27Z" level=error msg="error waiting for container: context canceled"
```
taken from: http://jenkins.katacontainers.io/job/kata-containers-runtime-ubuntu-1804-PR-firecracker/962/consoleFull 

other jobs from different repos now showing this error.

searching through the jobs, seems that the issue started to happen after merging #1933, which shows that the firecracker CI failed. 

/cc @lifupan can you ptal?

lifupan:
@chavafg , NP

lifupan:
Hi @chavafg , based on @bergwolf issue https://github.com/kata-containers/tests/issues/1927,  it would be a CI issue.

bergwolf:
@lifupan I created #1933 kata-containers/tests#1927 to ask if we have modified FC CI setting to test serial port. If not, #1933 is still suspicious.

PR #1990, opened 2019-08-21
virtcontainers: Fix the issue of watching console for firecracker

lifupan:
Since firecracker hasn't support console watching by now, so skip
watching console if the consoleURL is empty.

Fixes: #1970


devimc:
restarting ARM CI

Issue #1997, opened 2019-08-22
didn't cleanup the container config from sandbox when create container failed


lifupan:
# Description of problem

When create container failed for sandbox, we should do cleanup of the container config from
sandbox.config.Containers, otherwise, the following new create container would update the resources included this previous failed container resources.

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

PR #1998, opened 2019-08-22
Fix the issue of update resources wrong

lifupan:
When create container failed, it should delete the container
config from sandbox, otherwise, the following new creating container
 would get a wrong resources caculating which would contain the previous
 failed container resources such as memory and cpu.

Fixes: #1997


lifupan:
> lgtm, but ideally would have a unit test to assert this behaviour - is that possible do you think?

Hi @jodh-intel , Had added some unit tests.

jodh-intel:
Thanks @lifupan !

devimc:
restarting nemu and initrd

Issue #2001, opened 2019-08-23
Kill a failed exec process will cause the container being killed


darfux:
# Description of problem
The commit https://github.com/containerd/cri/commit/d257c16dbc144a167780b9b3cee5ee6a71d2b64b let containerd to kill the exec process when context is canceled. If you run a wrong program which is not in the `$PATH` through cri, such as exec bash on busybox image, the cri will return a proper error but the container is killed unexpectedly.

# Expected result
Nothing happens due to the process don't exist

# Actual result
The whole container is killed

PR #2002, opened 2019-08-23
v2: Prevent killing all container processes when exec is failed

darfux:
If an exec is failed(such as executable file not found in $PATH), the `execs.id` will be empty. This leads to all the container processes being killed when calling `Kill` on such exec id.

Fixes: #2001


devimc:
would be great to have an integration test for this, @darfux can you add it or show us the commands to reproduce this?

darfux:
@devimc I have no idea about how to do such test now, sorry🤣 I have updated the description of the issue and it can be reproduced with the recent version of containerd. Start a busybox container and use command such as `crictl exec -it <cid> bash` will reproduce it.

devimc:
@darfux don't worry  I can add an integration test later :wink:

PR #2010, opened 2019-08-29
Stable 1.8 backports

egernst:
179485f9 v2: Prevent killing all container processes when exec is failed
7e77b754 qemu: fix error message miss
05c8b02f virtcontainers: cleanup the container config once failed
a85252e8 virtcontainers: remove the redundant sandbox config store
5737f803 qemu: do not try to stop qemu multiple times
09036394 agent: add default timeout for grpc requests
6af47f62 network: always cold unplug network devices
7a690b0e container: do not pause a StateReady container
e1be3263 virtcontainers: fix hotplug pci devices execeed max capacity bug
f43e18be versions: kernel: update to 4.19.65
1cf69cd5 network: fix failed to remove network
83bb7b30 pkg/katautils: Do not set `init` in the kernel command line
4179bf7e ut: skip TestBindUnmountContainerRootfsENOENTNotError for non-root
50f48bb8 ut: skip TestStartNetworkMonitor for non-root


jodh-intel:
Hi @egernst - thanks for raising but it looks like this PR is targetted at the wrong branch (should be `stable-1.8` I think?) Marking with dnm...

egernst:
Ah, correct.

PR #2022, opened 2019-09-03
CompatOCISpec: limit usage of CompatOCISpec

WeiZhang555:
Fixes: #2023 

CompatOCISpec is used to gurantee backward compatbility for old runtime
specs, after we convert CompatOCISpec to standard specs.Spec, we should
use specs.Spec instead of CompatOCISpec, and CompatOCISpec should be
useless from then.

Spread usage of CompatOCISpec can make code structure confusing and making
the runtime spec usage non-standard. Besides, this can be the very first
step of removing CompatOCISpec from config's Annotations field.

### DIFF
--- a/virtcontainers/pkg/oci/utils.go
+++ b/virtcontainers/pkg/oci/utils.go
@@ -18,7 +18,7 @@ import (
 
 	criContainerdAnnotations "github.com/containerd/cri-containerd/pkg/annotations"
 	crioAnnotations "github.com/cri-o/cri-o/pkg/annotations"
-	spec "github.com/opencontainers/runtime-spec/specs-go"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/sirupsen/logrus"
 
 	vc "github.com/kata-containers/runtime/virtcontainers"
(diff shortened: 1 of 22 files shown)


virtcontainers/pkg/oci/utils.go
```diff
@@ -323,14 +323,14 @@ func containerCapabilities(s CompatOCISpec) (types.LinuxCapabilities, error) {
 }
 
 // ContainerCapabilities return a LinuxCapabilities for virtcontainer
-func ContainerCapabilities(s CompatOCISpec) (types.LinuxCapabilities, error) {
+func ContainerCapabilities(s CompatOCISpec) (specs.LinuxCapabilities, error) {
```
jodh-intel: I think this (public) function could be replaced entirely with the existing private `containerCapabilities()` one (just add the `s.Process` check to it). That would confine the use of `CompatOCISpec` to `utils.go` entirely.

virtcontainers/pkg/oci/utils.go
```diff
@@ -75,23 +75,23 @@ const (
...
 
-// CompatOCISpec is a structure inheriting from spec.Spec defined
+// CompatOCISpec is a structure inheriting from specs.Spec defined
 // in runtime-spec/specs-go package. It relies on the CompatOCIProcess
 // structure declared above, in order to be compatible with both
 // v1.0.0-rc4 and v1.0.0-rc5.
 // Refer to: https://github.com/opencontainers/runtime-spec/commit/37391fb
 type CompatOCISpec struct {
-	spec.Spec
+	specs.Spec
```
jodh-intel: I *think* this `struct` can be made private now? (`compatOCISpec`).

Note that there is a comment in `virtcontainers/persist/api/container.go` about it still though:

```go
// BundlePath saves container OCI config.json, which can be unmarshaled
// and translated to "CompatOCISpec"
BundlePath string
```

Is that comment still valid?

virtcontainers/pkg/oci/utils.go
WeiZhang555: That sounds great. Hmm, why I didn't think of it....Will modify.

virtcontainers/pkg/oci/utils.go
WeiZhang555: The comment is still valid, besides there will be another PR about persist data doing the config file refactoring.

And for the private change, I'll need some more investigation.

jodh-intel: Thanks @WeiZhang555! And thanks for tackling the overall issue on this PR - great refactor! ;)

WeiZhang555: I checked the code and made the "CompatOCISpec" private structure, same for "ContainerCapabilities()". Now they're "compatOCISpec" and "containerCapabilities()".

Thanks @jodh-intel for your great suggestion!

WeiZhang555:
ping @jodh-intel @bergwolf @sboeuf

jodh-intel:
This is ready to land but let's just see if @lifupan or @bergwolf have any comments...

egernst:
@bergwolf @lifupan please merge when you are happy with this!

Issue #2023, opened 2019-09-03
Limit usage of CompatOCISpec


WeiZhang555:
CompatOCISpec is used to gurantee backward compatbility for old runtime
specs, after we convert CompatOCISpec to standard specs.Spec, we should
use specs.Spec instead of CompatOCISpec, and CompatOCISpec should be
useless from then.

Spread usage of CompatOCISpec can make code structure confusing and making
the runtime spec usage non-standard. Besides, this can be the very first
step of removing CompatOCISpec from config's Annotations field.

PR #2028, opened 2019-09-04
Remove annotation config json key

WeiZhang555:
This is a WIP PR trying to remove ConfigJSONKey annotation.

Currently we have several places saving the OCI spec in configurations which bing a lot of redundancy and confusion, the plan is removing them all and only keep "bundle/config.json" from upper level components.

Fixes: #2023

We can get OCI spec config from bundle instead of annotations, so this
field isn't necessary.

### DIFF
--- a/virtcontainers/pkg/annotations/annotations.go
+++ b/virtcontainers/pkg/annotations/annotations.go
@@ -47,9 +47,6 @@ const (
 	// AssetHashType is the hash type used for assets verification
 	AssetHashType = vcAnnotationsPrefix + "AssetHashType"
 
-	// ConfigJSONKey is the annotation key to fetch the OCI configuration.
-	ConfigJSONKey = vcAnnotationsPrefix + "pkg.oci.config"
-
 	// BundlePathKey is the annotation key to fetch the OCI configuration file path.
 	BundlePathKey = vcAnnotationsPrefix + "pkg.oci.bundle_path"
 
--- a/cli/create.go
+++ b/cli/create.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/kata-containers/runtime/pkg/katautils"
 	vc "github.com/kata-containers/runtime/virtcontainers"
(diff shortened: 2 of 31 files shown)


WeiZhang555:
depends-on: #2022 

Waiting for it's merging.

virtcontainers/pkg/annotations/annotations.go
```diff
@@ -47,9 +47,6 @@ const (
 	// AssetHashType is the hash type used for assets verification
 	AssetHashType = vcAnnotationsPrefix + "AssetHashType"
 
-	// ConfigJSONKey is the annotation key to fetch the OCI configuration.
-	ConfigJSONKey = vcAnnotationsPrefix + "pkg.oci.config"
```
jodh-intel: Since this is an API change, we need to think about how we're going to manage its removal - if this PR is merged and we make a new release, some users may be surprised the feature has disappeared :smile:

Related: https://github.com/kata-containers/runtime/issues/1113#issuecomment-473825027

@egernst, @gnawux - any thoughts on this? Raise this issue in the AC meeting and announce the plan on the mailing list?

WeiZhang555: @jodh-intel I don't think it's an API change. This annotation is used for storing OCI spec in kata-runtime itself's configuration. It's not exposed to user and user won't have any way to use it, so this is an internal change  and it's definitely safe to remove :smile:

WeiZhang555:
/test

Rebased on upstream.

virtcontainers/pkg/annotations/annotations.go
jodh-intel: Sorry @WeiZhang555 ! You are of course correct - the lower-case `pkg.oci` prefix should have been the clue to me :smile:

WeiZhang555:
/test

Rebase on upstream master.

ping @lifupan @bergwolf @sboeuf @devimc

jshachm:
LGTM

Issue #2033, opened 2019-09-05
s390x: add appendVSock with devno


alicefr:
The function `appendVSock` needs to be reimplemented in order to be able to assign the devno to the vsock device.

PR #2034, opened 2019-09-05
s390x: add appendVSock with devno

alicefr:
Reimplementation of appendVSock in order to assign the devno to the vsock device.

Fixes: #2033

### DIFF
--- a/virtcontainers/qemu_s390x.go
+++ b/virtcontainers/qemu_s390x.go
@@ -7,11 +7,10 @@ package virtcontainers
 
 import (
 	"fmt"
-	"time"
-
 	govmmQemu "github.com/intel/govmm/qemu"
 	"github.com/kata-containers/runtime/virtcontainers/device/config"
 	"github.com/kata-containers/runtime/virtcontainers/types"
+	"time"
 )
 
 type qemuS390x struct {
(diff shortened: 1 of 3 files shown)


virtcontainers/qemu_s390x.go
```diff
@@ -257,3 +257,30 @@ func (q *qemuS390x) appendSCSIController(devices []govmmQemu.Device, enableIOThr
 	devices = append(devices, d)
 	return devices, t
 }
+
+func (q *qemuS390x) appendVSock(devices []govmmQemu.Device, vsock kataVSOCK) []govmmQemu.Device {
+	var devno string
+	id := fmt.Sprintf("vsock-%d", vsock.contextID)
+	addr, b, err := q.addDeviceToBridge(id, types.CCW)
+        if err != nil {
+                logrus.Errorf("Failed to append VSock: %v", err)
```
jodh-intel: The two error scenarios here imply to me that `appendVSock()` should return an error (now):

```go
appendVSock(devices []govmmQemu.Device, vsock kataVSOCK) ([]govmmQemu.Device error)
```

alicefr: @jodh-intel good point! I updated the PR. Actually, we need to add an error for all the append function introduced by 5f2c193d . Is it ok if I open another PR to fix that?

jodh-intel: Hi @alicefr - sure :smile:

PR #2050, opened 2019-09-13
Support Firecracker 0.18

### LINKED ISSUE #2027: Fix support for latest vsock version from Firecracker
The latest Firecracker version merged a brand new hybrid implementation of vsock, which comes with some special constraints compared to the regular vhost-vsock version.

Here are the steps which needs to be addressed in order to reach the proper implementation:
1. Update Firecracker client API with `swagger`
2. Fix `fc.go` to build with the new API
3. Move from `UseVsock` being a boolean into an actual choice about the type of socket connection being used. Ideally, we should have something like: `unix`, `vsock`, `vsock-hybrid`. Along with this change, we need to introduce a new type `kataVSockHybrid` that will have different fields than `kataVSOCK`, and in particular a path to the socket that needs to be chosen by Kata runtime as it needs to be unique across multiple instances.
4. Implement `kata_agent.go` to implement the case where `vsock-hybrid` is picked up. This means it's not needed to rely on `/dev/vhost-vsock` in this case to retrieve an appropriate CID.
5. Update `fc.go` to rely on the path decided earlier in the code, and use it as `UdsPath`. No need to close the `vhostFd` since there will be no `vhostFd` in this case.
6. Now that `kata_agent.go` returns an URL with `vsock-hybrid://` instead of `vsock://`, we need to update `kata-containers/agent/protocols/client/client.go`, function `NewAgentClient()` to handle the case of a third possibility. This client wrapper should handle both `unix`, `vsock` and `vsock-hybrid`. In details, the hybrid version needs to be handled by sending a special message right after the connection to the unix socket.
7. The other consumers of this URL, that is `kata-proxy` and `kata-shim` should not need to be modified since they will receive a URL that they will use through the client wrapper code directly. For this reason, both `kata-proxy` and `kata-shim` should simply need an update of the vendored version from the `kata-containers/agent` repository.

The agent code doesn't require any change since the guest still sees an `AF_VSOCK` socket.

devimc:
Support Firecracker 0.18

Depends-on: github.com/kata-containers/tests#1960
    
fixes #2027

### DIFF
--- a/virtcontainers/fc.go
+++ b/virtcontainers/fc.go
@@ -742,13 +742,15 @@ func (fc *firecracker) fcAddVsock(vs kataVSOCK) error {
 	vsockParams := ops.NewPutGuestVsockByIDParams()
 	vsockID := "root"
 	ctxID := int64(vs.contextID)
+	udsPath := ""
 	vsock := &models.Vsock{
 		GuestCid: &ctxID,
-		ID:       &vsockID,
+		UdsPath: &udsPath,
+		VsockID: &vsockID,
 	}
 	vsockParams.SetID(vsockID)
 	vsockParams.SetBody(vsock)
-	_, _, err := fc.client().Operations.PutGuestVsockByID(vsockParams)
+	_, err := fc.client().Operations.PutGuestVsockByID(vsockParams)
 	if err != nil {
 		return err
 	}
--- a/virtcontainers/fc.go
+++ b/virtcontainers/fc.go
@@ -58,6 +58,12 @@ const (
 	// We attach a pool of placeholder drives before the guest has started, and then
 	// patch the replace placeholder drives with drives with actual contents.
 	fcDiskPoolSize = 8
+
+	defaultHybridVSocketName = "kata.hvsock"
+
+	// This is the first usable vsock context ID. All the vsocks can use the same
+	// ID, since it's only used in the guest.
+	defaultGuestVSockCID = int64(0x3)
 )
 
 var fcKernelParams = append(commonVirtioblkKernelRootParams, []Param{
@@ -737,30 +743,33 @@ func (fc *firecracker) resumeSandbox() error {
 	return nil
 }
 
-func (fc *firecracker) fcAddVsock(vs kataVSOCK) error {
+func (fc *firecracker) fcAddVsock(hvs types.HybridVSock) error {
 	span, _ := fc.trace("fcAddVsock")
 	defer span.Finish()
 
+	udsPath := hvs.UdsPath
+	if fc.jailed {
+		udsPath = filepath.Join("/", defaultHybridVSocketName)
+	}
+
 	vsockParams := ops.NewPutGuestVsockByIDParams()
 	vsockID := "root"
-	ctxID := int64(vs.contextID)
-	udsPath := ""
+	ctxID := defaultGuestVSockCID
 	vsock := &models.Vsock{
 		GuestCid: &ctxID,
-		UdsPath: &udsPath,
-		VsockID: &vsockID,
+		UdsPath:  &udsPath,
+		VsockID:  &vsockID,
 	}
 	vsockParams.SetID(vsockID)
 	vsockParams.SetBody(vsock)
+
 	_, err := fc.client().Operations.PutGuestVsockByID(vsockParams)
 	if err != nil {
 		return err
 	}
-	//Still racy. There is no way to send an fd to the firecracker
-	//REST API. We could release this just before we start the instance
-	//but even that will not eliminate the race
-	vs.vhostFd.Close()
+
 	return nil
+
 }
 
 func (fc *firecracker) fcAddNetDevice(endpoint Endpoint) error {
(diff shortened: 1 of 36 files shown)


virtcontainers/fc.go
```diff
@@ -737,30 +739,33 @@ func (fc *firecracker) resumeSandbox() error {
...
+	udsPath := hvs.UdsPath
+	if fc.jailed {
+		udsPath = filepath.Join("/", defaultHybridVSocketName)
+	}
+
 	vsockParams := ops.NewPutGuestVsockByIDParams()
 	vsockID := "root"
-	ctxID := int64(vs.contextID)
-	udsPath := ""
+	ctxID := int64(3)
```
egernst: Can you tell me about the magic number 3?

Thanks @devimc!

virtcontainers/hypervisor.go
```diff
@@ -104,6 +104,10 @@ const (
 
 	// memoryDevice is memory device type
 	memoryDev
+
+	// hybridVSockDev is a hybrid virtio-vsock device supported
+	// only on certain hypervisors, like firecracker.
+	hybridVSockDev
```
amshinde: How about renaming this and below names `hybridVirtioVsockDev` to be more explicit?

virtcontainers/types/sandbox.go
```diff
@@ -161,6 +163,19 @@ func (v *Volumes) String() string {
 	return strings.Join(volSlice, " ")
 }
 
+// HybridVSock defines a hybrid vsocket to communicate between
+// the host and any process inside the VM.
```
amshinde: I would add a brief explanation stating that this is a virtio-vsock implementation based on AF_VSOCK on the guest side and multiple AF_UNIX sockets on the host side iiuc.

virtcontainers/fc.go
```diff
@@ -737,30 +739,33 @@ func (fc *firecracker) resumeSandbox() error {
 	return nil
 }
 
-func (fc *firecracker) fcAddVsock(vs kataVSOCK) error {
-	span, _ := fc.trace("fcAddVsock")
+func (fc *firecracker) fcAddVsock(hvs types.HybridVSock) error {
+	span, _ := fc.trace("fcAddHybridVsock")
```
amshinde: Just thinking out loud here, do we want this change to be backward compatible, ie use the correct vsock type based on the version of firecracker? At the very least do we check beforehand that the hypervisor supports virtio-vsock and throw a friendly error message asking the user to upgrade version of firecracker?

virtcontainers/fc.go
amshinde: Yes, please add comments here.

virtcontainers/fc.go
```diff
@@ -737,30 +739,33 @@ func (fc *firecracker) resumeSandbox() error {
...
 	vsockID := "root"
-	ctxID := int64(vs.contextID)
-	udsPath := ""
+	ctxID := int64(3)
 	vsock := &models.Vsock{
 		GuestCid: &ctxID,
-		UdsPath: &udsPath,
-		VsockID: &vsockID,
+		UdsPath:  &udsPath,
+		VsockID:  &vsockID,
```
amshinde: what am I missing here, just a gofmt change? Is the new format correct?

virtcontainers/fc.go
```diff
@@ -737,30 +739,33 @@ func (fc *firecracker) resumeSandbox() error {
...
 	}
 	vsockParams.SetID(vsockID)
 	vsockParams.SetBody(vsock)
-	_, err := fc.client().Operations.PutGuestVsockByID(vsockParams)
+
+	e, err := fc.client().Operations.PutGuestVsockByID(vsockParams)
 	if err != nil {
-		return err
+		return fmt.Errorf("%v: %v", e.Error(), err)
```
amshinde: Was this just for debugging? e.Error() should not be required as it simply returns the error string.

virtcontainers/fc.go
devimc: Good question, IMO it'd be a pain maintaining backward compatibility with all the versions of a project like FC that may introduce changes like this in each release. WDYT?

cc @sboeuf @egernst

virtcontainers/fc.go
devimc: yes, alignment with `GuestCid: &ctxID`

virtcontainers/fc.go
devimc: ouch! good catch , fixing right away

virtcontainers/hypervisor.go
devimc: ok, sounds good

virtcontainers/types/sandbox.go
devimc: yeah, thx

virtcontainers/fc.go
sboeuf: No I don't think so. This new implementation of vsock was waited for a long time from the FC folks, and now they have it, they won't move back. For that reason, we shouldn't bother with backward compatibility here.

devimc:
@amshinde @egernst changes applied, thanks

virtcontainers/fc.go
amshinde: @sboeuf I was just referring to this in the context of having an updated runtime, but an older version of firecracker on a system. Maybe this is something that may happen on a dev machine, the runtime could  just check that we are running an older version of firecracker and throw a friendly error message.

sboeuf: @amshinde that sounds like a great proposal :+1:

virtcontainers/fc.go
```diff
@@ -288,6 +298,23 @@ func (fc *firecracker) vmRunning() bool {
 	}
 }
 
+func (fc *firecracker) checkVersion(vmmInfo *fcmodels.InstanceInfo) error {
```
devimc: @amshinde  ptal

amshinde: thanks, looks good.

Issue #2042, opened 2019-09-11
qemu: No qemu output on crash if debug enabled


jschintag:
If debug is enabled in the configuration, all qemu output is written to `qemu.log` in the sandbox folder. When the launch of qemu fails, govmm returns an empty string as error output from qemu, resulting in `level=error msg="fail to launch qemu: exit status 1, error messages from qemu log: "`. The output from qemu is in the file `qemu.log`, which is cleaned up together with the sandbox on failure. This means that the output from qemu is lost.

PR #2058, opened 2019-09-16
qemu: Ensure error output is written to log with debug=true

jschintag:
If launching qemu fails and debug output is enabled, read the content of
qemu.log and write it into the runtime log.

Fixes: #2042

Signed-off-by: Jan Schintag <jan.schintag@de.ibm.com>


lifupan:
it's duplicated with https://github.com/kata-containers/runtime/pull/2055

jschintag:
Hmm #2055 does look nicer

Issue #2048, opened 2019-09-12
NEMU not working with docker


GabyCT:
It seems that NEMU is failing while running docker integration tests, we have the following issue
```
Stderr: docker: Error response from daemon: OCI runtime create failed: rpc error: code = Internal desc = Could not mount none to /run/kata-containers/shared/containers/: no such device: unknown.
```
This issue we get it when trying to run
```
$ docker run  --runtime kata-runtime -td --name pNORtcc4V30Ju0jN4MIeasaZlEweA1 busybox sh
```
This can be seen at 
http://jenkins.katacontainers.io/job/kata-containers-runtime-ubuntu-nemu/865/console

These are the logs
## Runtime logs

Recent runtime problems found in system journal:
```
time="2019-09-12T09:19:46.770612625Z" level=debug msg="Could not retrieve anything from storage" arch=amd64 command=create container=e75a3b0980ed6b26af539c5dd11176a23387303e222098109cde7c9919c0cf72 name=kata-runtime pid=65752 source=virtcontainers subsystem=kata_agent
time="2019-09-12T09:19:46.770744725Z" level=warning msg="load sandbox devices failed" arch=amd64 command=create container=e75a3b0980ed6b26af539c5dd11176a23387303e222098109cde7c9919c0cf72 error="open /run/vc/sbs/e75a3b0980ed6b26af539c5dd11176a23387303e222098109cde7c9919c0cf72/devices.json: no such file or directory" name=kata-runtime pid=65752 sandbox=e75a3b0980ed6b26af539c5dd11176a23387303e222098109cde7c9919c0cf72 sandboxid=e75a3b0980ed6b26af539c5dd11176a23387303e222098109cde7c9919c0cf72 source=virtcontainers subsystem=sandbox
[...]

PR #2066, opened 2019-09-17
config: use 9p as default shared filesystem for nemu

chavafg:
currently virtiofs support is being reworked to use qemu
and kernel from virtio-fs gitlab repo. The current support
that we have for virtio-fs has been disabled as required
patches for the 4.19 kernel were removed in [1].

[1] https://github.com/kata-containers/packaging/pull/703

Fixes: #2048.


Makefile
```diff
@@ -169,7 +169,7 @@ DEFENTROPYSOURCE := /dev/urandom
 
 DEFDISABLEBLOCK := false
 DEFSHAREDFS := virtio-9p
-DEFSHAREDFS_NEMU := virtio-fs
+DEFSHAREDFS_NEMU := virtio-9p
```
GabyCT: @chavafg so it seems that this https://github.com/kata-containers/runtime/pull/2054/files#diff-ebce02844da8831ba833219d9a0d0d36R103 will be affected by this change, why we do not use `DEFSHAREDFS` at the configuration.toml just to not have the same variable with the same value

chavafg: I could use `DEFSHAREDFS`, but I think `DEFSHAREDFS_NEMU` should only be used by `configuration-nemu.toml`. nemu related variables will be removed once we deprecate nemu, so I think we should use another variable for `virtio-fs`.
@jodh-intel @egernst any opinion on this?

chavafg:
@GabyCT ptal, I added a new variable `DEFSHAREDFS_QEMU_VIRTIOFS` for the `configuration-qemu-virtiofs.toml`, so that it doesn't affect the nemu change.

PR #2073, opened 2019-09-19
FC: extract error info from firecracker built-in log and metrics scheme

Pennyzct:
Hi~ I've recently been working on enabling firecracker on kata containers.
If Firecracker exits abnormally after `InstantStart` request, the error output you could get is only timeout related, nothing useful.
```
docker: Error response from daemon: OCI runtime create failed: Failed to check if grpc server is working: context deadline exceeded: unknown.
```
Firecracker has its own logging scheme, involving two fifo files. For better debugging, we should
extract those error info and re-play them in kata-runtime log.

### DIFF
--- a/virtcontainers/fc.go
+++ b/virtcontainers/fc.go
@@ -6,6 +6,7 @@
 package virtcontainers
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"net"
@@ -19,6 +20,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/containerd/fifo"
 	httptransport "github.com/go-openapi/runtime/client"
 	"github.com/go-openapi/strfmt"
 	kataclient "github.com/kata-containers/agent/protocols/client"
@@ -59,13 +61,16 @@ const (
 	// firecracker guest VM.
 	// We attach a pool of placeholder drives before the guest has started, and then
 	// patch the replace placeholder drives with drives with actual contents.
-	fcDiskPoolSize = 8
-
+	fcDiskPoolSize           = 8
 	defaultHybridVSocketName = "kata.hvsock"
 
 	// This is the first usable vsock context ID. All the vsocks can use the same
 	// ID, since it's only used in the guest.
 	defaultGuestVSockCID = int64(0x3)
+
+	// This is related to firecracker logging scheme
+	fcLogFifo     = "logs.fifo"
+	fcMetricsFifo = "metrics.fifo"
 )
 
 // Specify the minimum version of firecracker supported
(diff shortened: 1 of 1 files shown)


Pennyzct:
@jodh-intel @devimc plz review. ;)

virtcontainers/fc.go
```diff
@@ -546,6 +555,77 @@ func (fc *firecracker) fcSetVMBaseConfig(mem int64, vcpus int64, htEnabled bool)
...
+		return err
+	}
+
+	// listen on log fifo file and transfer error info
+	go func() {
+		scanner := bufio.NewScanner(fc.logFifo)
+		for scanner.Scan() {
+			fc.Logger().WithFields(logrus.Fields{
+				"level":    "log",
+				"contents": scanner.Text()}).Error("firecracker failed")
```
amshinde: @Pennyzct Why are you logging these as errors? Are these generic log messages or errors?

virtcontainers/fc.go
```diff
@@ -546,6 +555,77 @@ func (fc *firecracker) fcSetVMBaseConfig(mem int64, vcpus int64, htEnabled bool)
...
+		return err
+	}
+
+	// listen on metrics file and transfer error info
+	go func() {
+		scanner := bufio.NewScanner(fc.metricsFifo)
+		for scanner.Scan() {
+			fc.Logger().WithFields(logrus.Fields{
+				"level":    "metrics",
+				"contents": scanner.Text()}).Error("firecracker failed")
```
amshinde: Same question as above. Why are metrics logged as errors as well.

virtcontainers/fc.go
```diff
@@ -546,6 +555,85 @@ func (fc *firecracker) fcSetVMBaseConfig(mem int64, vcpus int64, htEnabled bool)
...
+		}
+
+		if err := scanner.Err(); err != nil {
+			fc.Logger().WithError(err).Error("Failed reading firecracker metrics from fifo file")
+		}
+
+		if err := fc.metricsFifo.Close(); err != nil {
+			fc.Logger().WithError(err).Error("Failed closing firecracker metrics fifo file")
+		}
+	}()
```
devimc: I think you can factorize a lot of code here, how about creating a new function to open and read the pipe and call it with different parameters depending on the pipe type

virtcontainers/fc.go
Pennyzct: Hi @amshinde Since it's the error info extracted from firecracker built-in logging scheme. It will only have outputs when firecracker failed setting up vmm. I have already set the logging level as `Error` here.
```
fcLogLevel := "Error"
```

virtcontainers/fc.go
Pennyzct: yep, metrics data will be more about performance, but it also has data structure like `drive_fails`,  `machine_cfg_fails`, ... etc for users to spot root cause of failure. And Like I said last comment, I have set the level as `Error`, it will only output when firecracker fail.

virtcontainers/fc.go
Pennyzct: Hi @devimc I have already used the API from [containerd/fifo](https://github.com/containerd/fifo) to open, read,..., close pipe files. 
And you're right, reading from l`ogs.fifo` and `metrics.fifo` share a lot same logic and I should create one new function to summarize. ;)

virtcontainers/fc.go
amshinde: thanks @Pennyzct for the explanation.

Pennyzct:
re-base and /test

jcvenegas:
@Pennyzct sorry a re-re-rebase is required here , btw some test were failing

virtcontainers/fc.go
```diff
@@ -579,6 +584,68 @@ func (fc *firecracker) fcSetVMBaseConfig(mem int64, vcpus int64, htEnabled bool)
 	return err
 }
 
+func (fc *firecracker) fcSetLogger() error {
```
jodh-intel: I don't like the `fc` prefix as it's redundant but I see there is already precedent for this in the codebase.

Pennyzct:
Hi @amshinde 
So sorry for the delayed response, just got back from our national holiday. ;)
I will re-re-base asap. For the failing part, with arm CI, it's because the missing aarch64 entry in the latest image `centos:latest`. I have addressed specific details in [kata-contains/ci:#222](https://github.com/kata-containers/ci/issues/222).
And for the other failing part, I will track down and try to find out how it's related. 😁

virtcontainers/fc.go
```diff
@@ -602,6 +607,68 @@ func (fc *firecracker) fcSetVMBaseConfig(mem int64, vcpus int64, htEnabled bool)
...
+	}
+
+	param := ops.NewPutLoggerParams()
+	cfg := &models.Logger{
+		Level:       &fcLogLevel,
+		LogFifo:     &jailedLogFifo,
+		MetricsFifo: &jailedMetricsFifo,
+	}
+	param.SetBody(cfg)
+	_, err = fc.client().Operations.PutLogger(param)
```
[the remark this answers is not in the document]
Pennyzct: I'm afraid not, it should belong to the configuration part, and should before the `InstanceStart` command. 
And if you are using `curl` to reproduce, it works well. 😂
```
 curl --unix-socket api.socket -i \
         -X PUT "http://localhost/logger" \
         -H "accept: application/json" \
         -H "Content-Type: application/json" \
         -d "{
                  \"log_fifo\": \"/root/firecracker/log.fifo\",
                  \"metrics_fifo\": \"/root/firecracker/metrics.fifo\",
                   \"level\": \"Info\"
              }"
```

Pennyzct:
I have tried a few "plausible" reason:
```
The connection has already been closed. But when client initiates another request before it actually
detects the close, and the EOF that it reads get interpreted as an error for this new request rather
than a close that occurred prior to the request.
```
Given that, I've given sleeping time between requests to give it time to detect the close, but, still,
the error occured.

virtcontainers/fc.go
```diff
@@ -602,6 +607,68 @@ func (fc *firecracker) fcSetVMBaseConfig(mem int64, vcpus int64, htEnabled bool)
...
+
+	return err
+}
+
+func (fc *firecracker) fcListenToFifo(fifoName string) (string, error) {
+	fcFifoPath := filepath.Join(fc.vmPath, fifoName)
+	fcFifo, err := fifo.OpenFifo(context.Background(), fcFifoPath, syscall.O_CREAT|syscall.O_RDONLY|syscall.O_NONBLOCK, 0)
```
devimc: @Pennyzct  have you tried with a blocking io?

lifupan:
Hi @Pennyzct 

I spend some time investigating on this issue, and found that the root cause located in firecracker, which would be crashed when setting logger with options as null.
I had opened a defect to firecracker: https://github.com/firecracker-microvm/firecracker/issues/1375 .

I think you can workaround this issue  in kata side by applying the following diff to your commits:

```
diff --git a/virtcontainers/fc.go b/virtcontainers/fc.go
index 1d8e9f39..b3acdf56 100644
--- a/virtcontainers/fc.go
+++ b/virtcontainers/fc.go
@@ -630,6 +630,7 @@ func (fc *firecracker) fcSetLogger() error {
                Level:       &fcLogLevel,
                LogFifo:     &jailedLogFifo,
                MetricsFifo: &jailedMetricsFifo,
+               Options: []string{},
        }
        param.SetBody(cfg)
        _, err = fc.client().Operations.PutLogger(param)
```

Pennyzct:
Hi @lifupan 
Thanks three thousand times!!!! update asap. ;).

PR #2081, opened 2019-09-24
Add annotations to provide custom configs

### LINKED ISSUE #1655: Add kernel_params as K8s annotation in addition to hypervisor, kernel, and image.
At https://github.com/kata-containers/runtime/blob/master/virtcontainers/pkg/annotations/annotations.go there is no annotation to customize the kernel boot parameters.  For an Intel(r) QuickAssist card there are multiple versions out there each with their own kernel modules. It would be convenient to build one kata rootfs with all the QAT modules but use a kubernetes yaml file to specify which kernel modules to load on boot that is dependent on which QAT card is used. This could be done with a kermel_params value where the modules to load are specified.

amshinde:
annotations: Support annotations to customise kata config
    
    Add support for annotations that allow us to custimise a subset
    of the configurations provided in kata conf toml file.
    This initial commit adds support for customising vcpus, default max
    vcpus, memory and the kernel command line passed as Hypervisor
    config.
    
    Replaces #1695
    Fixes #1655

### DIFF
--- a/virtcontainers/pkg/annotations/annotations.go
+++ b/virtcontainers/pkg/annotations/annotations.go
@@ -6,52 +6,56 @@
 package annotations
 
 const (
-	vcAnnotationsPrefix = "com.github.containers.virtcontainers."
+	kataAnnotationsPrefix     = "io.kata-containers."
+	kataConfAnnotationsPrefix = kataAnnotationsPrefix + "config."
+	kataAnnotHypervisorPrefix = kataConfAnnotationsPrefix + "hypervisor."
+	kataAnnotAgentPrefix      = kataConfAnnotationsPrefix + "agent."
+	kataAnnotRuntimePrefix    = kataConfAnnotationsPrefix + "runtime." // nolint: unused
 
 	// KernelPath is a sandbox annotation for passing a per container path pointing at the kernel needed to boot the container VM.
-	KernelPath = vcAnnotationsPrefix + "KernelPath"
+	KernelPath = kataAnnotHypervisorPrefix + "kernel"
 
 	// ImagePath is a sandbox annotation for passing a per container path pointing at the guest image that will run in the container VM.
-	ImagePath = vcAnnotationsPrefix + "ImagePath"
+	ImagePath = kataAnnotHypervisorPrefix + "image"
 
 	// InitrdPath is a sandbox annotation for passing a per container path pointing at the guest initrd image that will run in the container VM.
-	InitrdPath = vcAnnotationsPrefix + "InitrdPath"
+	InitrdPath = kataAnnotHypervisorPrefix + "initrd"
 
 	// HypervisorPath is a sandbox annotation for passing a per container path pointing at the hypervisor that will run the container VM.
-	HypervisorPath = vcAnnotationsPrefix + "HypervisorPath"
+	HypervisorPath = kataAnnotHypervisorPrefix + "path"
 
 	// JailerPath is a sandbox annotation for passing a per container path pointing at the jailer that will constrain the container VM.
-	JailerPath = vcAnnotationsPrefix + "JailerPath"
+	JailerPath = kataAnnotHypervisorPrefix + "jailer_path"
 
 	// FirmwarePath is a sandbox annotation for passing a per container path pointing at the guest firmware that will run the container VM.
-	FirmwarePath = vcAnnotationsPrefix + "FirmwarePath"
+	FirmwarePath = kataAnnotHypervisorPrefix + "firmware"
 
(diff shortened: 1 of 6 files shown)


virtcontainers/pkg/oci/utils.go
```diff
@@ -342,14 +355,65 @@ func addAssetAnnotations(ocispec specs.Spec, config *vc.SandboxConfig) {
...
+		vcpus, err := strconv.ParseUint(value, 10, 32)
+		if err != nil {
+			return fmt.Errorf("Error encountered parsing annotation for default_vcpus: %v, please specify numeric value", err)
+		}
+
+		numCPUs := goruntime.NumCPU()
+
+		// Assign the override only if value is valid, else fallback to what is provided by the config
+		if uint32(vcpus) < uint32(numCPUs) {
+			config.HypervisorConfig.NumVCPUs = uint32(vcpus)
```
jcvenegas: A warning may be helpful here in logs

virtcontainers/pkg/oci/utils.go
```diff
@@ -342,14 +355,65 @@ func addAssetAnnotations(ocispec specs.Spec, config *vc.SandboxConfig) {
...
+
+	if value, ok := ocispec.Annotations[vcAnnotations.DefaultMaxVCPUs]; ok {
+		maxVCPUs, err := strconv.ParseUint(value, 10, 32)
+		if err != nil {
+			return fmt.Errorf("Error encountered parsing annotation for default_maxvcpus: %v, please specify positive numeric value", err)
+		}
+
+		numCPUs := goruntime.NumCPU()
+		max := uint32(maxVCPUs)
+		if max < uint32(numCPUs) && max < vc.MaxQemuVCPUs() {
```
jcvenegas: `MaxQemuVCPUs`? I think annotations are agnostic to hypervisor implementation

jcvenegas: Again a warning saying that the annotation could not be applied would be nice

virtcontainers/pkg/oci/utils.go
jcvenegas: Or just fail is good as well

virtcontainers/pkg/oci/utils.go
```diff
@@ -342,14 +355,65 @@ func addAssetAnnotations(ocispec specs.Spec, config *vc.SandboxConfig) {
...
+			config.HypervisorConfig.DefaultMaxVCPUs = max
+		}
+	}
+
+	if value, ok := ocispec.Annotations[vcAnnotations.DefaultMemory]; ok {
+		memorySz, err := strconv.ParseUint(value, 10, 32)
+		if err != nil || memorySz < 8 {
```
jcvenegas: I think is a good moment to define a minimal memory for kata and have it defined in virtcontainers.
```go
if err != nil || memorySz < vc.minMemory {
return fmt.Errorf("Error encountered parsing annotation for default_memory: %v, please specify positive numeric value greater than vc.minMemory", err)
		}
}
```

virtcontainers/pkg/annotations/annotations.go
```diff
@@ -51,36 +66,135 @@ const (
...
+
+	// BlockDeviceDriver specifies the driver to be used for block device either VirtioSCSI or VirtioBlock
+	BlockDeviceDriver = kataAnnotHypervisorPrefix + "block_device_driver"
+
+	// DisableBlockDeviceUse  is a sandbox annotation that disallows a block device from being used.
+	DisableBlockDeviceUse = kataAnnotHypervisorPrefix + "disable_block_device_use"
+
+	// EnableIOThreads is a sandbox annotation to enable IO to be processed in a separate thread.
+	// Supported currently for virtio-scsi driver.
+	EnableIOThreads = kataAnnotHypervisorPrefix + "enable_iothreads"
```
jcvenegas: It would be nice that golang could use a macro so enable_iothreads cold be reused  so we dont have to modify this in case a config name changes :(

virtcontainers/pkg/oci/utils.go
amshinde: Yeah, I was thinking what the behaviour should be here while implementing this, was leaning towards just erroring out here myself. But ended up implementing behaviour similar to what we have for config toml file, where we fallback to defaults we have in the runtime.
I can certainly error out here. We should probably change how we handle configuration from the config.toml file to keep things consistent.
Another thing to consider maybe we could just assign the max number of CPUs on the system when the number requested exceeds that, especially if you have a heterogeneous cluster.
cc @jcvenegas @egernst

amshinde:
@jcvenegas @egernst Please take a look.

mythi:
@amshinde thanks!

Issue #2088, opened 2019-09-25
QEMU: do not require nvdimm machine option with initrd


marcov:
Do not add the "nvdimm" machine option to QEMU when the config specifies
a initrd file.
For arm64, this allows using a vanilla QEMU, where "virt" machine does
not support the "nvdimm" option.

PR #2089, opened 2019-09-25
QEMU: do not require nvdimm machine option with initrd

marcov:
Do not add the "nvdimm" machine option to QEMU when the config specifies
a initrd file.
For arm64, this allows using a vanilla QEMU, where "virt" machine does
not support the "nvdimm" option.

Fixed: #2088

Signed-off-by: Marco Vedovati <mvedovati@suse.com>


jcvenegas:
@Pennyzct PTAL

Issue #2091, opened 2019-09-27
shimv2 missed watching the hypervisor's console after re-FetchSandbox


lifupan:
# Description of problem

Both of the KataBuiltInProxyType and NoProxyType have the
same feature of can watch the hypervisor's console in
shimv2.

When do the reloading sandbox in shimv2, it's needed to
rewatch the hypervisor's console when debug enabled.

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

PR #2092, opened 2019-09-27
virtcontainers: fix the issue of missing watchConsole

lifupan:
When do the reloading sandbox in shimv2, it's needed to
rewatch the hypervisor's console when debug enabled.

Fixes:#2091

### DIFF
--- a/virtcontainers/agent.go
+++ b/virtcontainers/agent.go
@@ -126,6 +126,9 @@ type agent interface {
 	// check will check the agent liveness
 	check() error
 
+	// tell whether the agent is long  live connected or not
+	longLiveConn() bool
+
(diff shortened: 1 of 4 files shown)


devimc:
this patch is not working with vsocks, may be because the kata-shim can't watch the console ?

lifupan:
> this patch is not working with vsocks, may be because the kata-shim can't watch the console ?

Oh, yeah, I got it. For the kata cli mode, it cannot  watch the console, but for the shimv2, it can. 
I need to figure out a way to distinguish these two cases.

devimc:
thanks @lifupan

lifupan:
Ping @bergwolf @devimc @jodh-intel

PR #2122, opened 2019-10-09
v2: Change the event and error behavior of pause/resume

### LINKED ISSUE #2121: v2: pause and resume behavior is strange
The pause and resume behavior is strange from the following aspects:
1. The first call of pause/resume will not trigger an event.
2. Calling pause/resume on the same container multiple times will not cause errors, even if the container has been paused/resumed. And the event will be sent as well.

darfux:
1. Send the event when the container is paused/resumed successfully
2. Return the error of the pause/resume function rather than `getContainerStatus`.

Fixes #2121

### DIFF
--- a/containerd-shim-v2/service.go
+++ b/containerd-shim-v2/service.go
@@ -592,17 +592,17 @@ func (s *service) Pause(ctx context.Context, r *taskAPI.PauseRequest) (_ *ptypes
 	err = s.sandbox.PauseContainer(r.ID)
 	if err == nil {
 		c.status = task.StatusPaused
+		s.send(&eventstypes.TaskPaused{
+			ContainerID: c.id,
+		})
 		return empty, nil
 	}
 
-	c.status, err = s.getContainerStatus(c.id)
-	if err != nil {
+	if status, err := s.getContainerStatus(c.id); err != nil {
 		c.status = task.StatusUnknown
+	} else {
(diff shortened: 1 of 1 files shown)


devimc:
thanks @darfux 

cc @lifupan

containerd-shim-v2/service.go
```diff
@@ -592,18 +592,18 @@ func (s *service) Pause(ctx context.Context, r *taskAPI.PauseRequest) (_ *ptypes
 	err = s.sandbox.PauseContainer(r.ID)
 	if err == nil {
 		c.status = task.StatusPaused
+		s.send(&eventstypes.TaskPaused{
+			ContainerID: c.id,
+		})
 		return empty, nil
```
lifupan: What about removing this "return" line and do a double check container status to make sure it's changed to "Paused" status, and then send the event, otherwise, return an error info?  

The following "resume" is the same case.

darfux: IMO the `PauseContainer` will guarantee the container is paused successfully when there's no error, as it will set the container status to `StatePaused` before return (https://github.com/kata-containers/runtime/blob/stable-1.9/virtcontainers/container.go#L1261). So it should have been checked by CI rather than at runtime?

lifupan: yes, you are right.

PR #2124, opened 2019-10-09
virtcontainers/annotations: use right domain name for kata annotations

### LINKED ISSUE #2123: annotations: kata domain name
currently the prefix used for the kata annotations is `io.kata-containers`, but I think the right suffix is `io.katacontainers`, since the domain name of kata containers is `katacontainer.io`  not `kata-containers.io`

devimc:
The domain name should be used as prefix for the annotations, for
kata containers the domain name is katacontainers.io, not kata-containers.io

fixes #2123

### DIFF
--- a/virtcontainers/pkg/annotations/annotations.go
+++ b/virtcontainers/pkg/annotations/annotations.go
@@ -6,7 +6,7 @@
 package annotations
 
 const (
-	kataAnnotationsPrefix     = "io.kata-containers."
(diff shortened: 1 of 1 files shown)


amshinde:
@devimc This makes sense although we name all our repos as "kata-containers" instead of "katacontaainers". I am not opposed to this change.
@egernst @bergwolf Wdyt?

amshinde:
Lets wait for to see if we get some input on this.

Issue #2129, opened 2019-10-11
Cannot update sandbox‘s CPU config


fuxiao511:
# Description of problem

[root@localhost runtime]# cat /root/updatetxt
{
  "memory": {
    "limit": 0,
    "reservation": 0,
    "swap": 0,
    "kernel": 0,
    "kernelTCP": 0
  },
  "cpu": {
    "shares": 0,
    "quota": 200000,
    "period": 100000,
    "realtimeRuntime": 0,
    "realtimePeriod": 0,
    "cpus": "",
    "mems": ""
  },
  "blockIO": {
    "weight": 0
  },
  "pids": {
    "limit": 0
  }
}

[root@localhost runtime]# kata-runtime --log=/dev/stdout update -r /root/updatetxt aaa
INFO[0000] loaded configuration                          arch=amd64 command=update file=/etc/kata-containers/configuration.toml format=TOML name=kata-runtime pid=16102 source=katautils
DEBU[0000]                                               arch=amd64 command=update default-kernel-parameters="systemd.unit=kata-containers.target systemd.mask=systemd-networkd.service systemd.mask=systemd-networkd.socket" name=kata-runtime pid=16102 source=katautils
INFO[0000]                                               arch=amd64 arguments="\"update -r /root/updatetxt aaa\"" command=update commit=c7b4c5eab9f950dd0f933fbfbd7205a62a717242 name=kata-runtime pid=16102 source=runtime version=1.9.0-rc0
ERRO[0000] Unable to determine if running rootless       arch=amd64 command=update container=aaa error="Failed to parse uid map file /proc/self/uid_map" name=kata-runtime pid=16102 source=rootless
DEBU[0000] New filesystem store backend for /var/lib/vc/sbs/aaa  arch=amd64 backend=filesystem command=update container=aaa name=kata-runtime path=/var/lib/vc/sbs/aaa pid=16102 source=virtcontainers/store subsystem=store
DEBU[0000] New filesystem store backend for /run/vc/sbs/aaa  arch=amd64 backend=filesystem command=update container=aaa name=kata-runtime path=/run/vc/sbs/aaa pid=16102 source=virtcontainers/store subsystem=store
DEBU[0000] New filesystem store backend for /var/lib/vc/uuid  arch=amd64 backend=filesystem command=update container=aaa name=kata-runtime path=/var/lib/vc/uuid pid=16102 source=virtcontainers/store subsystem=store
[...]

PR #2130, opened 2019-10-11
virtcontainers/sandbox: calculate container's CPU from sandbox.contai…

fuxiao511:
…ners

sandbox.config.Containers is not up to date, so use container.config directly

Fixes: #2129

Signed-off-by: Wang Liang <wangliangzz@inspur.com>

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -717,7 +717,7 @@ func (c *Container) createBlockDevices() error {
 }
 
 // newContainer creates a Container structure from a sandbox and a container configuration.
-func newContainer(sandbox *Sandbox, contConfig ContainerConfig) (*Container, error) {
+func newContainer(sandbox *Sandbox, contConfig *ContainerConfig) (*Container, error) {
 	span, _ := sandbox.trace("newContainer")
 	defer span.Finish()
 
@@ -729,7 +729,7 @@ func newContainer(sandbox *Sandbox, contConfig ContainerConfig) (*Container, err
 		id:            contConfig.ID,
 		sandboxID:     sandbox.id,
 		rootFs:        contConfig.RootFs,
-		config:        &contConfig,
+		config:        contConfig,
 		sandbox:       sandbox,
 		runPath:       store.ContainerRuntimeRootPath(sandbox.id, contConfig.ID),
 		configPath:    store.ContainerConfigurationRootPath(sandbox.id, contConfig.ID),
@@ -812,7 +812,7 @@ func (c *Container) createMounts() error {
 	return nil
 }
 
-func (c *Container) createDevices(contConfig ContainerConfig) error {
+func (c *Container) createDevices(contConfig *ContainerConfig) error {
 	// If sandbox supports "newstore", only newly created container can reach this function,
 	// so we don't call restore when `supportNewStore` is true
 	if !c.sandbox.supportNewStore() {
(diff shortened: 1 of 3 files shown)


virtcontainers/sandbox.go
```diff
@@ -1873,8 +1873,8 @@ func (s *Sandbox) calculateSandboxMemory() int64 {
 func (s *Sandbox) calculateSandboxCPUs() uint32 {
 	mCPU := uint32(0)
 
-	for _, c := range s.config.Containers {
-		if cpu := c.Resources.CPU; cpu != nil {
+	for _, c := range s.containers {
```
lifupan: Here I think you can't simply using sandbox containers, since when creating a container in sandbox, the s.containers is updated before calculating resources, thus you could lost accounting the created container's resources. 
IMO the root cause is that the container's ContainerConfig pointer doesn't point to the instance  located in sandbox's SandboxConfig's  map of ContainerConfig, which caused the configure's not sync.

lifupan: Yes, I think there are issues  for both of the cpu and memory resource updating.

fuxiao511: @lifupan Yes, the root cause is that the container's ContainerConfig pointer doesn't point to the instance located in sandbox's SandboxConfig's map of ContainerConfig.

If we want to fix this way, we need to modify code in fetchSandbox->fetchContainers->newContainer, pass by reference, and more tests. 

Or we can just update sandbox's SandboxConfig's map of ContainerConfig at container.go:update(), which need a add method such like 
func (s *Sandbox) **updateConfig**() error

Which way do you perfer?

lifupan: > @lifupan Yes, the root cause is that the container's ContainerConfig pointer doesn't point to the instance located in sandbox's SandboxConfig's map of ContainerConfig.
> 
> If we want to fix this way, we need to modify code in fetchSandbox->fetchContainers->newContainer, pass by reference, and more tests.
>
I'd prefer this one, thus we can keep only one instance of container's config in memory and changed it at one place would be saw in other place.

fuxiao511: OK，I'll try to code it.

virtcontainers/container.go
```diff
@@ -785,7 +785,7 @@ func newContainer(sandbox *Sandbox, contConfig ContainerConfig) (*Container, err
 		return nil, err
 	}
 
-	if err = c.createDevices(contConfig); err != nil {
+	if err = c.createDevices(*contConfig); err != nil {
```
lifupan: For c.createDevice function,  I think you can also changed it's parameter from instance to pointer.

fuxiao511: Yes, we can change it in code. But I don't think we can change its semantics, we cannot change contConfig at **create**Devices procedure.

lifupan: What I mean is change it in code, since there is no need to pass an config instance here.

fuxiao511: OK, I'll change it.
Should I change and force-push all this two commits again, or just add a new commit?
which one is better?

lifupan: It's better to change them in a single commit and do a force push

fuxiao511:
@bergwolf There are two tests failed.  They seem to have nothing to do with this commit.
Can I do something, or should I just wait it to be fixed by others?

egernst:
virtio-fs CI failure is the dnf issue which is known.

Issue #2139, opened 2019-10-17
Race condition in qmpSetup


YvesChan:
**Which feature do you think can be improved?**

Currently, containerd-shim-kata-v2 maintain a persistent connection to qemu via qmp socket. But when we changed to ephemeral connection for some reason(e.g. debug), we found the `qmpSetup()` may cause a severe race condition, which leads to shim process not responding. It's because there's two qmp mainLoop goroutines exists at the same time.

**How can it be improved?**

Add a mutex in qmpSetup

PR #2140, opened 2019-10-17
vc/qemu: add mutex to qmp monitor channel in qmpSetup()

YvesChan:
Solve possible race condition in qmpSetup()

Fixes: #2139

Signed-off-by: Yves Chan <shanks.cyp@gmail.com>

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -19,6 +19,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 	"unsafe"
@@ -48,6 +49,7 @@ const romFile = ""
 const defaultDisableModern = false
 
 type qmpChannel struct {
+	sync.Mutex
 	ctx     context.Context
 	path    string
 	qmp     *govmmQemu.QMP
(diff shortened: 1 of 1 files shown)


devimc:
/test

thanks @YvesChan

YvesChan:
> Thanks @YvesChan!
> 
> Is there any way we can add a test (either a unit test here, or a test in https://github.com/kata-containers/tests) to give us some confidence the race has been fixed?

As I mentioned above, now the containerd-shim-kata-v2 process will hold the qmp.socket all the time, so there's no chance to reproduce the race condition in current runtime version. But it's not very convenient for user to debug the qemu issue if they cannot access qmp.socket since it's held by shim. So it's just better to make qmpSetup() more robust. What if we change the qmp connection to ephemeral(free after used) in the future? who knows :)

virtcontainers/qemu.go
```diff
@@ -921,6 +923,9 @@ func (q *qemu) togglePauseSandbox(pause bool) error {
 }
 
 func (q *qemu) qmpSetup() error {
+	q.qmpMonitorCh.Lock()
+	defer q.qmpMonitorCh.Unlock()
+
```
bergwolf: If we get parallel qmp setup, is it possible to get parallel qmp shutdown as well?

bergwolf: ping. The commit message says "Solve possible race condition in qmpSetup() and qmpShutdown()" but the shutdown part seems missing?

YvesChan: odd... I have already pushed that commit. Anyway, I do it again.  Thanks :)

jodh-intel:
This is now ready to land but I'll let @bergwolf click the lovely green button as an ack to the comment about shutdown.

Issue #2173, opened 2019-11-08
rootless:  rangeUID is parsed with wrong index


darfux:
The `rangeUID` should be parsed as ids[2] while ids[1] is the `hostUID`.

Issue #2132, opened 2019-10-11
Fix running rootless


jodh-intel:
Issue colllecting what needs to be done to fix running rootless with podman for example:

- https://github.com/kata-containers/documentation/pull/565

# Tasks

- Fix range index

  https://github.com/kata-containers/runtime/blob/master/pkg/rootless/rootless.go#L93

  This should be `ids[2]` fwics.

devimc:
# Tasks

make [`VCStoreUUIDPath`](https://github.com/kata-containers/runtime/blob/master/virtcontainers/store/vc.go#L273) rootless. uuid file shouldn't be created at `/var/` if `rootless`

patch #2134

egernst:
@alicefr - FYI

amshinde:
Related issue: https://github.com/kata-containers/runtime/issues/2177

amshinde:
@jodh-intel I have raised https://github.com/kata-containers/runtime/pull/2178 to fix an issue with rootless. 
With this in place and using the latest podman(1.6.2-dev), I was able to run rootless containers on Ubuntu.

c3d:
Tracked in Fedora as https://bugzilla.redhat.com/show_bug.cgi?id=1775164

amshinde:
# Tasks

Fix running rootless with net=none
Issue: https://github.com/kata-containers/runtime/issues/2319

PR #2174, opened 2019-11-08
rootless: Fix rangeUID parsing

darfux:
`rangeUID` should be parsed as ids[2] rather than ids[1]

Fixes: #2173


devimc:
related to https://github.com/kata-containers/runtime/issues/2132

cc @jodh-intel

PR #2178, opened 2019-11-09
rootless: Fix cgroup creation logic for rootless

### LINKED ISSUE #2177: Fix bug in cgroup creation logic for rootless containers with podman
We should not be creating cgroups for rootless case. But the current logic in place does not implement this correctly.

amshinde:
We do not want to create cgroups in case of rootless.
Fix the logic to implement this.

Fixes #2177


egernst:
@amshinde can you please send this fix over to stable-1.9 as well?

PR #2196, opened 2019-11-13
Remove nemu

egernst:
Deprecation was announced in 1.8, and we no longer support the NEMU VMM, as it is not actively maintained and the benefits of NEMU are now mostly realized by QEMU, Firecracker and Cloud Hypervisor VMMs.

This PR removes it from the runtime.


egernst:
image

PR #2200, opened 2019-11-14
FC: introduce `--config-file` to bypass API ready state

Pennyzct:
New command-line parameter for firecracker v0.19.0, named `--config-file`, which represents the path to a file that contains a JSON which can be used for configuring and starting a microVM without sending any API requests.
If we are using this new flag, we could bypass the whole API ready state.
I have wrote a draft on how it may work, And I use the variable `enabledconfigOption` to distinguish two processes.
So wdyt, do we need to hold two different boot-up processes for users to choose, or just turn to use the new `--config-file` option?

### DIFF
--- a/virtcontainers/fc.go
+++ b/virtcontainers/fc.go
@@ -36,7 +36,6 @@ import (
 	"github.com/blang/semver"
 	"github.com/containerd/console"
 	"github.com/kata-containers/runtime/virtcontainers/device/config"
-	fcmodels "github.com/kata-containers/runtime/virtcontainers/pkg/firecracker/client/models"
 	"github.com/kata-containers/runtime/virtcontainers/store"
 	"github.com/kata-containers/runtime/virtcontainers/types"
 	"github.com/kata-containers/runtime/virtcontainers/utils"
@@ -108,7 +107,8 @@ func (s vmmState) String() string {
 // FirecrackerInfo contains information related to the hypervisor that we
 // want to store on disk
 type FirecrackerInfo struct {
-	PID int
+	PID     int
+	Version string
 }
 
 type firecrackerState struct {
@@ -311,12 +311,27 @@ func (fc *firecracker) vmRunning() bool {
 	}
 }
 
-func (fc *firecracker) checkVersion(vmmInfo *fcmodels.InstanceInfo) error {
-	if vmmInfo == nil || vmmInfo.VmmVersion == nil {
-		return fmt.Errorf("Unknown firecracker version")
+func (fc *firecracker) getVersionNumber() (string, error) {
+	args := []string{"--version"}
+	checkCMD := exec.Command(fc.config.HypervisorPath, args...)
+
+	data, err := checkCMD.Output()
+	if err != nil {
+		return "", fmt.Errorf("Running checking FC version command failed: %v", err)
 	}
 
-	v, err := semver.Make(*vmmInfo.VmmVersion)
+	var version string
+	fields := strings.Split(string(data), " ")
+	if len(fields) > 1 {
+		version = strings.TrimSpace(fields[1])
+		return version, nil
+	}
+
+	return "", errors.New("getting FC version failed, the output is malformed")
+}
+
+func (fc *firecracker) checkVersion(version string) error {
+	v, err := semver.Make(version)
 	if err != nil {
 		return fmt.Errorf("Malformed firecracker version: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


Pennyzct:
@lifupan @bergwolf @gnawux @devimc

lifupan:
IMO, I 'd prefer using --config-file to replace the  configure API requests instead of keep two options which would make it much complex.

virtcontainers/types/fcConfig.go
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2018 ARM Limited
```
devimc: 2019 ?

Pennyzct:
Hi guys @lifupan @jodh-intel @devimc 
Since on one is objecting that we use `--config-file` to replace API configure request.
I've updated code and ptal. ;). BTW, firecracker CI is working well.

virtcontainers/fc.go
```diff
@@ -698,22 +663,64 @@ func (fc *firecracker) startSandbox(timeout int) error {
...
+func (fc *firecracker) startSandbox(timeout int) error {
+	span, _ := fc.trace("startSandbox")
+	defer span.Finish()
+
+	if err := fc.fcInitConfiguration(); err != nil {
 		return err
 	}
 
-	return fc.waitVMM(timeout)
+	data, errJSON := json.MarshalIndent(fc.fcConfig, "", "\t")
```
devimc: `MarshalIndent` why not just `Marshal` ?

virtcontainers/fc.go
```diff
@@ -384,8 +399,8 @@ func (fc *firecracker) fcInit(timeout int) error {
 		}
 	}()
 
-	var args []string
 	var cmd *exec.Cmd
+	args := []string{"--config-file", fc.fcConfigPath}
```
devimc: may be worth to validate version greater than 0.19 before starting fc, what do you think?

virtcontainers/fc.go
```diff
@@ -307,12 +316,21 @@ func (fc *firecracker) vmRunning() bool {
 	}
 }
 
-func (fc *firecracker) checkVersion(vmmInfo *fcmodels.InstanceInfo) error {
-	if vmmInfo == nil || vmmInfo.VmmVersion == nil {
-		return fmt.Errorf("Unknown firecracker version")
+func (fc *firecracker) checkVersion() error {
+	args := []string{"--version"}
```
devimc: this function is called several times in `waitVMMRunning`, may be we should get the version once (at `init()` ? ) instead of executing `firecracker` several times, what do you think?

virtcontainers/fc.go
Pennyzct: yep. you're right. should be more careful~~😋

virtcontainers/fc.go
Pennyzct: I'm not sure that we could get version number in `init()`, which is called during the first importing package time. The path of  `firecracker` binary may not be tracked back to that early time.
But you're right, we should store it somewhere for later frequent use. ;).

virtcontainers/fc.go
Pennyzct: Just want to make the JSON text output more like what we used to see? 😜
```
The difference between json.Marshal() and json.MarshalIndent() is that the former does a compact output without indentation, the latter applies (somewhat customizable) indent. 
````
You could see [the example here](https://play.golang.org/p/iMCmfkQARRk) for better declaration . ;).

**Comment by Pennyzct** [comment]
rebase and update. ;)
/test-fc

**Comment by devimc** [comment]
thanks @Pennyzct 

/test-ubuntu

**Comment by Pennyzct** [comment]
Hi guys
I got all green here. 🎉🎊

**Comment by Pennyzct** [comment]
Hi~
need one more reviewer here~~ @lifupan care to take a look? 😁

File: virtcontainers/fc.go [hunk 5]
```diff
@@ -311,12 +321,27 @@ func (fc *firecracker) vmRunning() bool {
...
+	data, err := checkCMD.Output()
+	if err != nil {
+		return "", fmt.Errorf("Running checking FC version command failed: %v", err)
+	}
+
+	var version string
+	fields := strings.Split(string(data), " ")
+	if len(fields) > 1 {
+		version = strings.TrimSpace(fields[1])
+		return version, nil
```
tedyu: How about storing the version in a field so that subsequent calls can return the version directly ?

virtcontainers/fc.go
```diff
@@ -108,7 +111,8 @@ func (s vmmState) String() string {
 // FirecrackerInfo contains information related to the hypervisor that we
 // want to store on disk
 type FirecrackerInfo struct {
-	PID int
+	PID     int
+	Version string
```
Pennyzct: Hi @tedyu 
I've store the FC version here~ ;)

Issue #2201, opened 2019-11-14
log out the firecracker's console for debugging when hypervisor debug enabled


lifupan:
# Description of problem
We should afford a method to log out the FC's console
for debugging the system panic issues.

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

PR #2202, opened 2019-11-14
FC: log out the firecracker's console when debug enabled

lifupan:
When hypervisor's debug option set, log out the firecracker's
console output which contains the kernel boot logs; thus it
would be easy for system panic debugging.

When agent debug was enabled by passing "agent.log=debug" to
kernel parameter, it will also log out the agent logs from
the console output.

Fixes: #2201


lifupan:
ping @kata-containers/runtime

PR #2208, opened 2019-11-18
Respect containerd's debug config

### LINKED ISSUE #2209: Too many noisy logs from containerd-shim-kata-v2
# Description of problem
Because kubelet is using  static cpu manager policy, it will call `UpdateContainerResources` every 10 seconds. There is a lot of logs like this
```
time="2019-11-18T22:25:19.636579269+08:00" level=info msg="{\"execute\":\"query-cpus\"}" ID=f91e5c366d715c2fe4dd936c2ae6b8dcf4e7217e13d65019d291b08457879357 source=virtcontainers subsystem=qmp
time="2019-11-18T22:25:19.636993728+08:00" level=info msg="{\"return\": [{\"arch\": \"x86\", \"current\": true, \"props\": {\"core-id\": 0, \"thread-id\": 0, \"node-id\": 0, \"socket-id\": 0}, \"CPU\": 0, \"qom_path\": \"/machine/unattached/device[0]\", \"pc\": -2123138173, \"halted\": true, \"thread_id\": 49823}, {\"arch\": \"x86\", \"current\": false, \"props\": {\"core-id\": 0, \"thread-id\": 0, \"node-id\": 0, \"socket-id\": 1}, \"CPU\": 1, \"qom_path\": \"/machine/unattached/device[2]\", \"pc\": -2123138173, \"halted\": true, \"thread_id\": 49824}, {\"arch\": \"x86\", \"current\": false, \"props\": {\"core-id\": 0, \"thread-id\": 0, \"node-id\": 0, \"socket-id\": 2}, \"CPU\": 2, \"qom_path\": \"/machine/unattached/device[3]\", \"pc\": -2123138173, \"halted\": true, \"thread_id\": 49825}, {\"arch\": \"x86\", \"current\": false, \"props\": {\"core-id\": 0, \"thread-id\": 0, \"node-id\": 0, \"socket-id\": 55}, \"CPU\": 55, \"qom_path\": \"/machine/peripheral/cpu-0\", \"pc\": -2123138173, \"halted\": true, \"thread_id\": 49881}]}" ID=f91e5c366d715c2fe4dd936c2ae6b8dcf4e7217e13d65019d291b08457879357 source=virtcontainers subsystem=qmp
time="2019-11-18T22:25:19.638305092+08:00" level=info msg="UpdateContainerResources for "b863809bf6ffb68c0911c892b6a82b2f39152cf1cf65a1337c29d98fbbd9dd54" returns successfully" 
```

(replace this text with the list of steps you followed)

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

choury:
Currently, There are too many logs from kata shim in containerd's log. We should handle it.

Fix #2209

### DIFF
--- a/containerd-shim-v2/service.go
+++ b/containerd-shim-v2/service.go
@@ -65,6 +65,10 @@ func New(ctx context.Context, id string, publisher events.Publisher) (cdshim.Shi
 	// it will output into stdio, from which containerd would like
 	// to get the shim's socket address.
 	logrus.SetOutput(ioutil.Discard)
+	opts := ctx.Value(cdshim.OptsKey{}).(cdshim.Opts)
+	if !opts.Debug {
+		logrus.SetLevel(logrus.WarnLevel)
+	}
 	vci.SetLogger(ctx, logger)
 	katautils.SetLogger(ctx, logger, logger.Logger.Level)
 
@@ -141,7 +145,10 @@ func newCommand(ctx context.Context, containerdBinary, id, containerdAddress str
 		"-address", containerdAddress,
 		"-publish-binary", containerdBinary,
 		"-id", id,
-		"-debug",
+	}
+	opts := ctx.Value(cdshim.OptsKey{}).(cdshim.Opts)
+	if opts.Debug {
+		args = append(args, "-debug")
 	}
 	cmd := sysexec.Command(self, args...)
 	cmd.Dir = cwd


grahamwhaley:
Hi @choury - thanks for the PR. Looks sane to me, but I will let others fully review....
you will however need to open an Issue and use that Issue number in your commit for the CI to be happy - see the https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#patch-format for details....
Can I ask, when you open the Issue, can you add details (snippets of the logs), showing the before (too much noise) and after (less noise) examples?

containerd-shim-v2/service.go
```diff
@@ -65,6 +65,12 @@ func New(ctx context.Context, id string, publisher events.Publisher) (cdshim.Shi
 	// it will output into stdio, from which containerd would like
 	// to get the shim's socket address.
 	logrus.SetOutput(ioutil.Discard)
+	opts := ctx.Value(cdshim.OptsKey{}).(cdshim.Opts)
```
lifupan: Is this setting needed? Since the log has been discarded by "logrus.SetOutput(ioutil.Discard)" previously.

choury: Yes,  the followed ` katautils.SetLogger` will use the level.

lifupan: I mean since the log's output has been discarded, thus whatever the log level you set, you wouldn't get what you want. Thus, is better to remove the setting here?

choury: They are not the same `logger`, maybe I change the SetLevel line to 
```
logger.Logger.SetLevel(logrus.DebugLevel)
```
will make it clearly?

lifupan: In my opinion, they are the same logger which is the standard logger, and it will be reset the level here https://github.com/containerd/containerd/blob/master/runtime/v2/shim/shim.go#L133, thus I think there is no need 
to set the level here.

choury: [startShim](https://github.com/containerd/containerd/blob/master/runtime/v2/shim/shim.go#L211) will start a new process  [here](https://github.com/kata-containers/runtime/blob/master/containerd-shim-v2/service.go#L201), do you think the level will effect new process? You can build and try my patch, and you will trust me .

choury: If `debugFlag` isn't set, the level will not be reset. You can build and test my patch, and you will trust me.

lifupan: If the debugFlag isn't set, then it would use the default level InfoLevel. So here you want to change it to WarnLevel
if the debugFlag isn't set?

choury: Yes, too many INFO logs are generated by shim. If anyone need more detail logs, he can just turn debug on.

lifupan: In that case, is it better to change it as below:

if !opts.Debug {
       logrus.SetLevel(logrus.WarnLevel)
}

PR #2211, opened 2019-11-18
Continue parsing PrimaryGroup when parsedUser[1] fails parsing

### LINKED ISSUE #2210: Continue parsing PrimaryGroup when parsedUser[1] fails parsing
# Description of problem
In cmdToKataProcess :
```
		i, err = strconv.ParseUint(parsedUser[1], 10, grpcUserBits)
		if err != nil {
			return nil, err
```
However, in normal flow (no error), cmd.PrimaryGroup would be parsed next. The parsing result would be stored in gid as well.

# Expected result
I think when parsedUser[1] fails parsing, we should continue parsing cmd.PrimaryGroup.

# Actual result

tedyu:
Fixes #2210

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -451,7 +451,7 @@ func cmdToKataProcess(cmd types.Cmd) (process *grpc.Process, err error) {
 	var gid uint32
 	if len(parsedUser) > 1 {
 		i, err = strconv.ParseUint(parsedUser[1], 10, grpcUserBits)
-		if err != nil {
+		if err != nil && if len(cmd.PrimaryGroup) == 0 {
 			return nil, err
 		}
 


virtcontainers/kata_agent.go
```diff
@@ -451,7 +451,7 @@ func cmdToKataProcess(cmd types.Cmd) (process *grpc.Process, err error) {
 	var gid uint32
 	if len(parsedUser) > 1 {
 		i, err = strconv.ParseUint(parsedUser[1], 10, grpcUserBits)
-		if err != nil {
+		if err != nil && if len(cmd.PrimaryGroup) == 0 {
```
amshinde: @tedyu Can you explain the reason for doing this?
We expect user to be of the format `uid` or `uid:gid`, so if we receive an unxpected format we rightfully error out.

tedyu: What if cmd. PrimaryGroup is also present and parses correctly ?

amshinde: I prefer erroring out here in the case where we have got incorrect input.
cc @jodh-intel @grahamwhaley

PR #2239, opened 2019-11-21
Land experimental "newstore" as formal feature

### LINKED ISSUE #803: [RFC] baseline persist data for live-upgrade

WeiZhang555:
Fixes #803 

The "newstore" feature has had been a "experimental" feature for long time, it aims to re-organize the persistent data on disk and merging separate files into one "persist.json", and also better guarantee the compatibility.

I have been working on this for a long time and preparing its landing before Kata 2.0, and now I think it's the time. Let's land it as a formal feature!

I'm not sure when is the exact date of release 2.0, but since I can't put enough energy on this work, I hope to merge it ASAP in case nobody else would like to pick up.

You can see I removed lots of code, the "newstore" will make code clearer and easier to read, I hope everyone likes it. :-)

### DIFF
--- a/virtcontainers/acrn.go
+++ b/virtcontainers/acrn.go
@@ -7,6 +7,7 @@ package virtcontainers
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"os/exec"
@@ -20,6 +21,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
+	"github.com/kata-containers/runtime/pkg/rootless"
 	"github.com/kata-containers/runtime/virtcontainers/device/config"
 	persistapi "github.com/kata-containers/runtime/virtcontainers/persist/api"
 	"github.com/kata-containers/runtime/virtcontainers/pkg/uuid"
@@ -27,6 +29,25 @@ import (
 	"github.com/kata-containers/runtime/virtcontainers/types"
 	"github.com/kata-containers/runtime/virtcontainers/utils"
 )
+
+// Since ACRN is using the store in a quite abnormal way, let's first draw it back from store to here
+
+// UUIDPathSuffix is the suffix used for uuid storage
+const (
+	UUIDPathSuffix = "uuid"
+	uuidFile       = "uuid.json"
+)
+
+// VMUUIDStoragePath is the uuid directory.
+// It will contain all uuid info used by guest vm.
+var VMUUIDStoragePath = func() string {
+	path := filepath.Join("/run/vc", UUIDPathSuffix)
+	if rootless.IsRootless() {
+		return filepath.Join(rootless.GetRootlessDir(), path)
+	}
+	return path
+
+}
 
 // ACRN currently supports only known UUIDs for security
 // reasons (FuSa). When launching VM, only these pre-defined
@@ -73,7 +94,6 @@ type AcrnState struct {
 // Acrn is an Hypervisor interface implementation for the Linux acrn hypervisor.
 type Acrn struct {
 	id         string
-	store      *store.VCStore
 	config     HypervisorConfig
 	acrnConfig Config
 	state      AcrnState
@@ -276,7 +296,7 @@ func (a *Acrn) buildDevices(imagePath string) ([]Device, error) {
 }
 
 // setup sets the Acrn structure up.
-func (a *Acrn) setup(id string, hypervisorConfig *HypervisorConfig, vcStore *store.VCStore) error {
+func (a *Acrn) setup(id string, hypervisorConfig *HypervisorConfig) error {
 	span, _ := a.trace("setup")
 	defer span.Finish()
 
@@ -286,24 +306,19 @@ func (a *Acrn) setup(id string, hypervisorConfig *HypervisorConfig, vcStore *sto
 	}
 
 	a.id = id
-	a.store = vcStore
 	a.config = *hypervisorConfig
 	a.arch = newAcrnArch(a.config)
 
 	var create bool
 	var uuid string
 
-	if a.store != nil { //use old store
-		if err = a.store.Load(store.Hypervisor, &a.state); err != nil {
-			create = true
-		}
-	} else if a.state.UUID == "" { // new store
+	if a.state.UUID == "" {
 		create = true
 	}
 
 	if create {
 		a.Logger().Debug("Setting UUID")
-		if uuid, err = a.GetNextAvailableUUID(nil); err != nil {
+		if uuid, err = a.GetNextAvailableUUID(); err != nil {
 			return err
 		}
 		a.state.UUID = uuid
@@ -313,10 +328,6 @@ func (a *Acrn) setup(id string, hypervisorConfig *HypervisorConfig, vcStore *sto
 		// The path might already exist, but in case of VM templating,
 		// we have to create it since the sandbox has not created it yet.
 		if err = os.MkdirAll(store.SandboxRuntimeRootPath(id), store.DirMode); err != nil {
-			return err
-		}
-
-		if err = a.storeState(); err != nil {
 			return err
 		}
 
@@ -348,14 +359,14 @@ func (a *Acrn) createDummyVirtioBlkDev(devices []Device) ([]Device, error) {
 }
 
 // createSandbox is the Hypervisor sandbox creation.
-func (a *Acrn) createSandbox(ctx context.Context, id string, networkNS NetworkNamespace, hypervisorConfig *HypervisorConfig, store *store.VCStore, stateful bool) error {
+func (a *Acrn) createSandbox(ctx context.Context, id string, networkNS NetworkNamespace, hypervisorConfig *HypervisorConfig, stateful bool) error {
 	// Save the tracing context
 	a.ctx = ctx
 
 	span, _ := a.trace("createSandbox")
 	defer span.Finish()
 
-	if err := a.setup(id, hypervisorConfig, store); err != nil {
+	if err := a.setup(id, hypervisorConfig); err != nil {
 		return err
 	}
 
@@ -458,11 +469,6 @@ func (a *Acrn) startSandbox(timeoutSecs int) error {
 		return err
 	}
 
-	//Store VMM information
-	if err = a.storeState(); err != nil {
-		return err
-	}
-
 	return nil
 }
 
@@ -499,7 +505,7 @@ func (a *Acrn) stopSandbox() (err error) {
 	uuid := a.state.UUID
 	Idx := acrnUUIDsToIdx[uuid]
 
-	if err = a.store.Load(store.UUID, &a.info); err != nil {
+	if err = a.loadInfo(); err != nil {
 		a.Logger().Info("Failed to load UUID availabiity info")
 		return err
 	}
@@ -698,7 +704,7 @@ func (a *Acrn) getPids() []int {
 	return []int{a.state.PID}
 }
 
-func (a *Acrn) fromGrpc(ctx context.Context, hypervisorConfig *HypervisorConfig, store *store.VCStore, j []byte) error {
+func (a *Acrn) fromGrpc(ctx context.Context, hypervisorConfig *HypervisorConfig, j []byte) error {
 	return errors.New("acrn is not supported by VM cache")
 }
 
@@ -737,20 +743,14 @@ func (a *Acrn) GetACRNUUIDBytes(uid string) (uuid.UUID, error) {
 
 // GetNextAvailableUUID returns next available UUID VM creation
 // If no validl UUIDs are available it returns err.
-func (a *Acrn) GetNextAvailableUUID(uuidstore *store.VCStore) (string, error) {
+func (a *Acrn) GetNextAvailableUUID() (string, error) {
 	var MaxVMSupported uint8
 	var Idx uint8
 	var uuidStr string
 	var err error
 
-	if uuidstore == nil {
-		uuidstore = a.store
-	}
-
-	if uuidstore != nil { //use old store
-		if err = uuidstore.Load(store.UUID, &a.info); err != nil {
-			a.Logger().Infof("Load UUID store failed")
-		}
+	if err = a.loadInfo(); err != nil {
+		a.Logger().Infof("Load UUID store failed")
 	}
 
 	if MaxVMSupported, err = a.GetMaxSupportedACRNVM(); err != nil {
@@ -795,22 +795,79 @@ func (a *Acrn) GetMaxSupportedACRNVM() (uint8, error) {
 	return platformInfo.maxKataContainers, nil
 }
 
-func (a *Acrn) storeState() error {
-	if a.store != nil {
-		if err := a.store.Store(store.Hypervisor, a.state); err != nil {
-			a.Logger().WithError(err).Error("failed to store acrn state")
+func (a *Acrn) storeInfo() error {
+	dirPath := VMUUIDStoragePath()
+
+	_, err := os.Stat(dirPath)
+	if os.IsNotExist(err) {
+		// Root directory
+		a.Logger().WithField("path", dirPath).Debugf("Creating UUID directory")
+		if err := os.MkdirAll(dirPath, store.DirMode); err != nil {
 			return err
 		}
+	} else if err != nil {
+		return err
 	}
+
+	dirf, err := os.Open(dirPath)
+	if err != nil {
+		return err
+	}
+	defer dirf.Close()
+
+	if err := syscall.Flock(int(dirf.Fd()), syscall.LOCK_EX|syscall.LOCK_NB); err != nil {
+		return err
+	}
+
+	// write data
+	f, err := os.OpenFile(filepath.Join(dirPath, uuidFile), os.O_RDWR|os.O_CREATE, 0755)
+	if err != nil {
+		return fmt.Errorf("failed to store information into uuid.json: %v", err)
+	}
+	defer f.Close()
+
+	jsonOut, err := json.Marshal(a.info)
+	if err != nil {
+		return fmt.Errorf("Could not marshall data: %s", err)
+	}
+	f.Write(jsonOut)
 	return nil
 }
 
-func (a *Acrn) storeInfo() error {
-	if a.store != nil {
-		if err := a.store.Store(store.UUID, a.info); err != nil {
-			a.Logger().WithError(err).Error("failed to store acrn info")
-			return err
-		}
+func (a *Acrn) loadInfo() error {
+	dirPath := VMUUIDStoragePath()
+
+	_, err := os.Stat(dirPath)
+	if err != nil {
+		return fmt.Errorf("failed to load ACRN information: %v", err)
+	}
+
+	dirf, err := os.Open(dirPath)
+	if err != nil {
+		return err
+	}
+
+	if err := syscall.Flock(int(dirf.Fd()), syscall.LOCK_SH|syscall.LOCK_NB); err != nil {
+		dirf.Close()
+		return err
+	}
+
+	defer dirf.Close()
+
+	// write data
+	f, err := os.Open(filepath.Join(dirPath, uuidFile))
+	if err != nil {
+		return fmt.Errorf("failed to load information into uuid.json: %v", err)
+	}
+
+	dec := json.NewDecoder(f)
+	if dec != nil {
+		return fmt.Errorf("failed to create json decoder")
+	}
+
+	err = dec.Decode(&a.info)
+	if err != nil {
+		return fmt.Errorf("could not decode data: %v", err)
 	}
 	return nil
 }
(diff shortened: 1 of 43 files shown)


WeiZhang555:
Depends on: https://github.com/kata-containers/tests/pull/2113

CI failed since I remove "/var/lib/vc" dir but test cases always check this dir for active pod info.

"/var/lib/vc" isn't necessary, we can put all persist informations in "/var/run". Suppose the server meets a reboot, all qemu and kata processes will be dead, but garbage information will be left in "/var/lib/vc". Since "/run/vc" is tmpfs, it will be automatically cleaned up after reboot.

devimc:
@bergwolf  @lifupan @WeiZhang555 @egernst @chavafg this PR breaks backward compatibility, I ran a container with master, then applied this patch and tried to stop the container, but I got the following error:

```
Error response from daemon: cannot stop container: 8be8ef9e3058: Cannot kill container 8be8ef9e3058228dda9ace31f0ae1942d1c72b7eb400aed484bd18ffa8e61461: OCI runtime state failed: open /run/vc/sbs/8be8ef9e3058228dda9ace31f0ae1942d1c72b7eb400aed484bd18ffa8e61461/persist.json: no such file or directory: unknown
```

I'm working in a test to check backward compatibility in the CI

WeiZhang555:
@devimc Yes, this will break backward compatibility, AS expected. This PR will finally remove all VCStore and replace it with experimental  "newstore" feature which changed persistent data structure totally.

And I'm happy you're adding the test, we should add backward compatibility test cases from long time ago.

virtcontainers/acrn.go
```diff
@@ -28,6 +30,25 @@ import (
...
+// UUIDPathSuffix is the suffix used for uuid storage
+const (
+	UUIDPathSuffix = "uuid"
+	uuidFile       = "uuid.json"
+)
+
+// VMUUIDStoragePath is the uuid directory.
+// It will contain all uuid info used by guest vm.
+var VMUUIDStoragePath = func() string {
+	path := filepath.Join("/run/vc", UUIDPathSuffix)
```
jodh-intel: I think it would be better if this path were not hard-coded here - ideally `/var/vc/` would be defined in `pkg/katautils/config.go` and then passed to each hypervisor somehow.

virtcontainers/acrn.go
```diff
@@ -737,20 +743,14 @@ func (a *Acrn) GetACRNUUIDBytes(uid string) (uuid.UUID, error) {
...
-	if uuidstore == nil {
-		uuidstore = a.store
-	}
-
-	if uuidstore != nil { //use old store
-		if err = uuidstore.Load(store.UUID, &a.info); err != nil {
-			a.Logger().Infof("Load UUID store failed")
-		}
+	if err = a.loadInfo(); err != nil {
+		a.Logger().Infof("Load UUID store failed")
```
jodh-intel: - Probably worth logging the error here (may not be `ENOENT` but something else?
- I see the logic is similar to what was there before, but it might be worth adding a comment explaining why this is not an error scenario (`err` is not being returned).

virtcontainers/acrn.go
```diff
@@ -737,20 +743,14 @@ func (a *Acrn) GetACRNUUIDBytes(uid string) (uuid.UUID, error) {
 
 // GetNextAvailableUUID returns next available UUID VM creation
 // If no validl UUIDs are available it returns err.
-func (a *Acrn) GetNextAvailableUUID(uuidstore *store.VCStore) (string, error) {
+func (a *Acrn) GetNextAvailableUUID() (string, error) {
```
jodh-intel: Not part of this PR, but there is a typo in the function header comment ("validl" => "valid").

virtcontainers/acrn.go
```diff
@@ -795,22 +795,79 @@ func (a *Acrn) GetMaxSupportedACRNVM() (uint8, error) {
...
-	if a.store != nil {
-		if err := a.store.Store(store.Hypervisor, a.state); err != nil {
-			a.Logger().WithError(err).Error("failed to store acrn state")
+func (a *Acrn) storeInfo() error {
+	dirPath := VMUUIDStoragePath()
+
+	_, err := os.Stat(dirPath)
+	if os.IsNotExist(err) {
+		// Root directory
+		a.Logger().WithField("path", dirPath).Debugf("Creating UUID directory")
```
jodh-intel: Nit: You can use `Debug()` here rather than `Debugf()`.

virtcontainers/acrn.go
```diff
@@ -795,22 +795,79 @@ func (a *Acrn) GetMaxSupportedACRNVM() (uint8, error) {
...
+		return err
+	}
+	defer dirf.Close()
+
+	if err := syscall.Flock(int(dirf.Fd()), syscall.LOCK_EX|syscall.LOCK_NB); err != nil {
+		return err
+	}
+
+	// write data
+	f, err := os.OpenFile(filepath.Join(dirPath, uuidFile), os.O_RDWR|os.O_CREATE, 0755)
```
jodh-intel: Why `0755`? Can we use `0600`?

virtcontainers/acrn.go
```diff
@@ -795,22 +795,79 @@ func (a *Acrn) GetMaxSupportedACRNVM() (uint8, error) {
...
+	if err != nil {
+		return fmt.Errorf("failed to store information into uuid.json: %v", err)
```
jodh-intel: Can you use `uuidFile` rather than hard-coding the name of the file in the error (here and below).

virtcontainers/acrn.go
```diff
@@ -795,22 +795,79 @@ func (a *Acrn) GetMaxSupportedACRNVM() (uint8, error) {
...
+	}
+	defer f.Close()
+
+	jsonOut, err := json.Marshal(a.info)
+	if err != nil {
+		return fmt.Errorf("Could not marshall data: %s", err)
 	}
+	f.Write(jsonOut)
```
jodh-intel: `Write()` can fail so we should return the error value.

virtcontainers/persist.go
```diff
@@ -443,12 +443,7 @@ func (c *Container) Restore() error {
 }
 
 func (s *Sandbox) supportNewStore() bool {
-	for _, f := range s.config.Experimental {
-		if f == persist.NewStoreFeature && exp.Get("newstore") != nil {
-			return true
-		}
-	}
-	return false
+	return true
```
jodh-intel: Is this function actually needed now?

virtcontainers/persist/fs/fs.go
```diff
@@ -77,21 +75,21 @@ func Init() (persistapi.PersistDriver, error) {
...
-	return filepath.Join(RunStoragePath(), id), nil
+func (fs *FS) sandboxDir(sandboxID string) (string, error) {
+	return filepath.Join(RunStoragePath(), sandboxID), nil
 }
 
 // ToDisk sandboxState and containerState to disk
 func (fs *FS) ToDisk(ss persistapi.SandboxState, cs map[string]persistapi.ContainerState) (retErr error) {
+	id := ss.SandboxContainer
+	if id == "" {
+		return fmt.Errorf("sandbox container id required")
```
jodh-intel: Nit: You can use `errors.New()` as there are no formats (same comment below).

virtcontainers/sandbox.go
```diff
@@ -568,38 +568,14 @@ func newSandbox(ctx context.Context, sandboxConfig SandboxConfig, factory Factor
...
-			return nil, err
-		}
-	} else {
-		// Fetch sandbox network to be able to access it from the sandbox structure.
-		var networkNS NetworkNamespace
-		if err = s.store.Load(store.Network, &networkNS); err == nil {
-			s.networkNS = networkNS
-		}
+	// Ignore the error. Restore can fail for a new sandbox
+	s.Restore()
```
jodh-intel: How about logging it in debug mode though?

devimc:
@WeiZhang555 ok, afaik until now all previous releases are backward compatible, right? so if this PR lands, the next release version should be `2.0.0` and include in the release notes that **this release is not backward compatible** ?

cc @egernst @gnawux @jshachm  @kata-containers/architecture-committee

WeiZhang555:
/test

@devimc

>ok, afaik until now all previous releases are backward compatible, right? 

I hope so, but I guess it's not...This PR is trying to standardize the persist files for better guarantee the backward compatibility, without this PR, the backward compatibility can be broken so easily without any notice.

WeiZhang555:
/test

Remove WIP flag, it's ready for review now! ping @kata-containers/runtime @kata-containers/architecture-committee 

This is supposed to be landed in release 2.0, but I hope we can get this in as soon as possible since I have little free time on this and this PR is easy to conflict with others, and rebasing can bring failing CIs which is hard to debug for such a big change. 

Thanks!

virtcontainers/clh.go
```diff
@@ -20,12 +20,12 @@ import (
 	"time"
 
 	persistapi "github.com/kata-containers/runtime/virtcontainers/persist/api"
+	"github.com/kata-containers/runtime/virtcontainers/persist/fs"
```
devimc: @ericooper  can you please help us reviewing this file?

virtcontainers/api_test.go
```diff
@@ -1412,7 +1422,7 @@ func TestProcessListContainer(t *testing.T) {
 
 	pImpl, ok := p.(*Sandbox)
 	assert.True(ok)
-	defer store.DeleteAll()
+	// defer store.DeleteAll()
```
devimc: please remove this line

bergwolf:
I think we'd want more features than just `newstore` to cut kata 2.0. @gnawux and @egernst have been discussing 2.0 roadmap for sometime and might have more thoughts to share.

WeiZhang555:
/test

@devimc @jodh-intel comments are addressed.

This PR is trying to solve the problem of "backward compatibility is so easy to break without notice", the more earlier we get this in, the more benefits we can get. So I would say, we should merge this ASAP, even the next release isn't 2.0.0.

virtcontainers/acrn.go
```diff
@@ -795,22 +796,39 @@ func (a *Acrn) GetMaxSupportedACRNVM() (uint8, error) {
...
+func (a *Acrn) storeInfo() error {
+	store, err := persist.GetDriver("fs")
+	if err != nil {
+		return err
+	}
+	relPath := filepath.Join(UUIDPathSuffix, uuidFile)
+
+	jsonOut, err := json.Marshal(a.info)
+	if err != nil {
+		return fmt.Errorf("Could not marshall data: %s", err)
```
jodh-intel: Nit: s/marshall/marshal/

virtcontainers/acrn.go
```diff
@@ -795,22 +795,79 @@ func (a *Acrn) GetMaxSupportedACRNVM() (uint8, error) {
...
+	}
+
+	defer dirf.Close()
+
+	// write data
+	f, err := os.Open(filepath.Join(dirPath, uuidFile))
+	if err != nil {
+		return fmt.Errorf("failed to load information into uuid.json: %v", err)
+	}
+
```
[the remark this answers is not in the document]
jodh-intel: Looks like this comment no longer applies so can be marked as resolved?

virtcontainers/acrn.go
WeiZhang555: @jodh-intel I don't know why this didn't throw an error out, guess this is a bug? It's better to let ACRN guys take a look. I think it should return the error.

virtcontainers/persist/fs/fs.go
WeiZhang555: I think "fmt.Errorf" has the same effect with "errors.New()" so we can choose either one. 😄

virtcontainers/acrn.go
```diff
@@ -20,14 +21,34 @@ import (
...
 	"github.com/kata-containers/runtime/virtcontainers/device/config"
 	persistapi "github.com/kata-containers/runtime/virtcontainers/persist/api"
+	"github.com/kata-containers/runtime/virtcontainers/persist/fs"
 	"github.com/kata-containers/runtime/virtcontainers/pkg/uuid"
-	"github.com/kata-containers/runtime/virtcontainers/store"
 	"github.com/kata-containers/runtime/virtcontainers/types"
 	"github.com/kata-containers/runtime/virtcontainers/utils"
 )
 
+// Since ACRN is using the store in a quite abnormal way, let's first draw it back from store to here
```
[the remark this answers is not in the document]
vijaydhanraj: sure, let me check the changes in my environment and get back.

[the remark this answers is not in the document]
vijaydhanraj: I was able to verify the changes in my local setup and things look fine at my end.

WeiZhang555:
/test

Rebased.

virtcontainers/acrn.go
WeiZhang555: @vijaydhanraj Do you think we should return the error here? I guess this is a bug but not sure.

WeiZhang555:
/test

Rebased.

virtcontainers/acrn.go
vijaydhanraj: @WeiZhang555 , The intent here is to log a message when UUID.json file is not available. This is not a bug, as we proceed and create UUID.json file to store UUID availability.

WeiZhang555:
/test

Rebased

WeiZhang555:
ping @kata-containers/architecture-committee @gnawux @lifupan @sboeuf @devimc

devimc:
@WeiZhang555 I'd like to merge this, but I agree with @jodh-intel 

> Might be worth raising this as a discussion topic for the Architecture Committee meeting?

WeiZhang555:
@kata-containers/architecture-committee 

Add one commit https://github.com/kata-containers/runtime/pull/2239/commits/2270c135062839d0ed7a5ac6cde9e641d12f56ff for fixing the backward compatibility with some redundant codes, now it should be safe to move from old store to new store, it won't break any running workloads any more.

devimc:
@WeiZhang555 compatibility test is ready https://github.com/kata-containers/tests/pull/2135
I tested this PR manually and it's not backward compatible,
do we really want this PR backward compatible? next release can be 2.0

WeiZhang555:
@devimc It's not? I tried in my local machine, new kata-runtime can clean old containers started with old kata-runtime.

My opinion is the new store is added for better keep backward compatibility, because it's so easy to break with old store, so the earlier we merge this, the more benefit we can enjoy.

I did a lots of cleanup work to remove old storage, but if we want to keep the compatibility, we have to add them back, this could make the code base kind of messy and hard to maintain.

WeiZhang555:
@devimc I can see there is foreward compatibility test in your PR, it's good but I have no idea how to pass it, because that means you need to let old kata-runtime recognize new persist storage structure. I believe it will fail for this one at least.

One solution I can find is that we add the feature in one release but not enable it, then enable it in the following release...But I hope we can get this in earlier :-(

egernst:
Forward compat. is a bit harder to handle, and I'm not sure in a production environment when you'd realistically hit this.  Based on your local tests, I'm happy with this @WeiZhang555, but I want to wait on an ACK from @devimc  (Julio - thanks for putting these tests in).

devimc:
@WeiZhang555 @egernst I'm using the latest version of this PR

```
INFO: Running backward compatibility test
+ runtime_path=/usr/local/bin/kata-runtime
+ master_runtime_path=/tmp/kata/master/bin/kata-runtime
+ runtime_backup_path=/usr/local/bin/kata-runtime.backup
+ sudo cp /usr/local/bin/kata-runtime /usr/local/bin/kata-runtime.backup
+ cont_name=backward_test
+ docker run -d --name backward_test --runtime=kata-runtime busybox tail -f /dev/null
Unable to find image 'busybox:latest' locally
latest: Pulling from library/busybox
bdbbaa22dec6: Pull complete 
Digest: sha256:6915be4043561d64e0ab0f8f098dc2ac48e077fe23f488ac24b665166898115a
Status: Downloaded newer image for busybox:latest
e01741b12998f9088e6cebae7d4e266023f5f76bc4f2ad9e54b471c6f16d9754
+ sudo cp -a /tmp/kata/master/bin/kata-runtime /usr/local/bin/kata-runtime
+ sync
+ docker exec backward_test true
OCI runtime state failed: open /run/vc/sbs/e01741b12998f9088e6cebae7d4e266023f5f76bc4f2ad9e54b471c6f16d9754/lock: no such file or directory: unknown
++ handle_error 75
++ local exit_code=126
++ local line_number=75
++ echo 'Failed at 75: docker exec "${cont_name}" true'
Failed at 75: docker exec "${cont_name}" true
```

WeiZhang555:
@devimc I think you're testing the backward compatibility in wrong direction(correct me if I'm wrong).

The backward testing process should be:
1. start container with "master" kata-runtime
2. replace kata-runtime from PR
3. `docker exec` and `docker rm` the container with PR kata-runtime to check if it breaks master.

I think your "backward compatibility" test is actually "forward compatibility" and vice versa.

I modified the test case according to my environment and run it locally, the script:

```
#!/bin/bash

test_backward_compatibility() {
	echo "Running backward compatibility test"

	runtime_path="$1"
	master_runtime_path="$2"

	# start container with old runtime
	cp $runtime_path kata-runtime && sync

	# run a container with the current runtime
	cont_name="backward_test"
	docker run -d --name "${cont_name}" --net none --runtime=kata busybox tail -f /dev/null

	# debug info
	# print kata-runtime version
	kata-runtime version

	# print dir content
	sudo tree /run/vc/sbs/

	# switch to master runtime
	cp $master_runtime_path kata-runtime && sync
	# print kata-runtime version
	kata-runtime version

	# exec
	docker exec "${cont_name}" true

	# stop and remove container
	docker rm -f "${cont_name}"

	# start new container  and check dir content
	docker run -d --name "${cont_name}" --net none --runtime=kata busybox tail -f /dev/null
	sudo tree /run/vc/sbs/
	docker rm -f "${cont_name}"
}

test_backward_compatibility ./kata-runtime-old ./kata-runtime-new

```

result:

```
Running backward compatibility test
8a3260687b5b3ddf032d0bd92e3ada321eefaeb0c3e563ca57217e1aff6cf2cd
kata-runtime  : 1.10.0-rc0
   commit   : 3ea3d3201b3a0b6fc35dea48192e73a021b4734b
   OCI specs: 1.0.1-dev
/run/vc/sbs/
└── 8a3260687b5b3ddf032d0bd92e3ada321eefaeb0c3e563ca57217e1aff6cf2cd
    ├── 8a3260687b5b3ddf032d0bd92e3ada321eefaeb0c3e563ca57217e1aff6cf2cd
    │   ├── devices.json
    │   ├── lock
    │   ├── mounts.json
    │   ├── process.json
    │   ├── raw
    │   └── state.json
    ├── agent.json
    ├── devices.json
    ├── hypervisor.json
    ├── lock
    ├── network.json
    ├── proxy.sock
    ├── raw
    └── state.json

4 directories, 12 files
kata-runtime  : 1.10.0-rc0
   commit   : 290339da6b9d094a8c73e1252ccafa42ca75c6b1
   OCI specs: 1.0.1-dev
backward_test
2e5df935b811a68313791814863eb777f79119836251b6d2dd8ff52c3775c8a9
/run/vc/sbs/
└── 2e5df935b811a68313791814863eb777f79119836251b6d2dd8ff52c3775c8a9
    ├── 2e5df935b811a68313791814863eb777f79119836251b6d2dd8ff52c3775c8a9
    │   └── persist.json
    ├── persist.json
    └── proxy.sock

2 directories, 3 files
backward_test

```

PR kata-runtime `290339da6b9d094a8c73e1252ccafa42ca75c6b1` is backward compatible to master `3ea3d3201b3a0b6fc35dea48192e73a021b4734b`

devimc:
@WeiZhang555 

> I think your "backward compatibility" test is actually "forward compatibility" and vice versa.

yes, but what about forward compatibility? should kata support this? what happen when users run containers with latest kata but they rollback because something went wrong?

egernst:
If its passing backward, i'm okay with this.  Handling the forward upgrade is a very niche problem, and if we're fighting that, we're in a more serious issue imo.

If it passes backward, let's plan on merging this.  @devimc can we update the test for just backward compat (at least for now)?

devimc:
@WeiZhang555 @egernst  test fixed, ubuntu CI is passing but FC and initrd CIs are failing

WeiZhang555:
@devimc I think initrd and firecracker CI are unstable, for FC test, it fails quite randomly, besides I don't have a complete env testing and debugging this. 

Is there anyone can help?

bergwolf:
Nice work @WeiZhang555 !

WeiZhang555:
Wow 🎆 👏 

Thanks everyone for help review!

egernst:
WEEEEEE IT MADE IT!

PR #2248, opened 2019-11-22
vc: Restore state if setting container state fails for some container

### LINKED ISSUE #2247: Restore state if setting container state fails for some container
# Description of problem

This applies to Sandbox#pauseSetStates and Sandbox#resumeSetStates.

If setting container state fails for some container (let's call it C1), we should restore state for the containers whose state is changed prior to setting state for C1.

# Expected result

We should restore state for the containers whose state is changed prior to setting state for C1.

# Actual result

tedyu:
Fixes #2247

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -1664,7 +1664,7 @@ func (s *Sandbox) setSandboxState(state types.StateString) error {
 func (s *Sandbox) pauseSetStates() error {
 	// XXX: When a sandbox is paused, all its containers are forcibly
 	// paused too.
-	if err := s.setContainersState(types.StatePaused); err != nil {
+	if err := s.setContainersState(types.StatePaused, types.StateRunning); err != nil {
 		return err
 	}
 
@@ -1674,7 +1674,7 @@ func (s *Sandbox) pauseSetStates() error {
 func (s *Sandbox) resumeSetStates() error {
 	// XXX: Resuming a paused sandbox puts all containers back into the
 	// running state.
-	if err := s.setContainersState(types.StateRunning); err != nil {
+	if err := s.setContainersState(types.StateRunning, types.StatePaused); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


virtcontainers/sandbox.go
```diff
@@ -1717,15 +1717,27 @@ func (s *Sandbox) decrementSandboxBlockIndex() error {
 	return nil
 }
 
-func (s *Sandbox) setContainersState(state types.StateString) error {
+func (s *Sandbox) setContainersState(state types.StateString, oringinal types.StateString) error {
```
jodh-intel: Typo: `original`

tedyu:
@jodh-intel 
I looked at virtcontainers/sandbox_test.go but didn't see error injection.

I will dig some more.

jodh-intel:
Hi @tedyu - @WeiZhang555 may have some thoughts on this [1] since I think this PR "overlaps" with #2239.

---

[1] - "this" being (a) whether we need tests for your PR, (b) how to write those tests, and (c) whether those tests needs to be written and landed before #2239 lands or after :smile:

tedyu:
@jodh-intel 
I looked at @WeiZhang555 's changes in sandbox.go - there was no overlap. His changes are for store functions.

I think this can go in while I continue to figure out how error injection can be done.

what do you think ?

jodh-intel:
@tedyu - I was talking about `Container.setContainerState()` in `container.go` - your changes on this PR modify `Sandbox.setContainersState()` which calls `Container.setContainerState()`. So, to add a test for your changes, we need a test for `Sandbox.setContainersState()`. But since that function really only calls `Container.setContainerState()`, you might want to get input from @WeiZhang555 since his PR changes `Container.setContainerState()`.

My personal view: "small" changes like this absolutely must have unit tests to assert the expected behaviour. Sometimes these sorts of fixes look "so obvious" that tests seem superfluous. But we've been bitten by these sorts of changes before where a tiny change changed the behaviour sufficiently to introduce some very subtle and difficult to debug bugs. In summary **all** PRs should contain unit tests where possible.

tedyu:
To test the change in Sandbox.setContainersState(), we would need at least two containers.
The setting of the first container state passes while the setting of the second container state fails.

I went over virtcontainers/container_test.go and virtcontainers/sandbox_test.go but haven't found a suitable test to extrapolate.

It seems the test would require creating a TestContainer struct which wraps Container struct. Specifcally overriding setContainerState() func.

tedyu:
It seems we can create TestSandbox struct, embedding Sandbox struct.
TestSandbox would override Sandbox#addContainer where TestContainer is recorded in s.containers.
TestContainer has a new func :
```
func (c *Container) setPosition(int pos)
```
which is called by TestSandbox before Container#setContainerState() is called (so that fault can be injected).

@jodh-intel 
What do you think ?

bergwolf:
We have abandoned sandbox level pause functionality. Is it time for a little code cleanup and just drop PauseSandbox() vc api?

tedyu:
I can send out a PR.
We can drop the resume vc api as well ?

tedyu:
@bergwolf 
Please confirm you were referring to Sandbox#Pause (there is no PauseSandbox).

bergwolf:
@tedyu yes

tedyu:
@bergwolf 
When removing these two func's, I found that togglePauseSandbox() calls them.
togglePauseSandbox() is called in api.go

I will remove these two func's from api.go:
```
func PauseSandbox(ctx context.Context, sandboxID string) (VCSandbox, error) {

func ResumeSandbox(ctx context.Context, sandboxID string) (VCSandbox, error) {
```

PR #2251, opened 2019-11-22
k8s: fix wrong number cpus after killing a container

### LINKED ISSUE #2186: k8s: wrong number of cpus after killing a container
# Description of problem

the number of vpus increases after killing a container with cpus constrains, **if the container is killed 99 times, the same number of vpus are hot plugged !!**

**pod-cpu.yaml**
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: cpu-test
spec:
  runtimeClassName: kata
  containers:
  - name: c1
    image: busybox
    command:
      - sh
    resources:
      limits:
        cpu: "1"
    tty: true
  - name: c2
    image: busybox
    command:
      - sh
    resources:
      limits:
        cpu: "1"
    tty: true
```

```
$ kubectl apply -f pod-cpu.yaml
$ kubectl get pods
NAME       READY   STATUS    RESTARTS   AGE
cpu-test   2/2     Running   0          67s
$ kubectl exec -ti pod/cpu-test -c c1 nproc
3
# the number of vcpus is correct: 1 (c1) + 1 (c2) + 1 (default_vcpus) = 3

# Attach to c2 and kill it
$ kubectl attach -ti pod/cpu-test -c c2
If you don't see a command prompt, try pressing enter.
/ # exit

# wait for c2, it's restarted
$ kubectl get pods
NAME       READY   STATUS    RESTARTS   AGE
cpu-test   2/2     Running   0          67s

# check again the number of vcpus
$ kubectl exec -ti pod/cpu-test -c c1 nproc
4
# Now the number of vcpus is 4, the number of vpus increases after every restart of c2
```

# Expected result

below command should always return 3

```
$ kubectl exec -ti pod/cpu-test -c c1 nproc
3
```

# Actual result

```
$ kubectl exec -ti pod/cpu-test -c c1 nproc
4
# kill c2
$ kubectl exec -ti pod/cpu-test -c c1 nproc
5
# kill c2 again
$ kubectl exec -ti pod/cpu-test -c c1 nproc
6
# kill c2 again
$ kubectl exec -ti pod/cpu-test -c c1 nproc
7
```

might be this issue was introduced here https://github.com/kata-containers/runtime/pull/1998 ?

cc @lifupan @egernst @jcvenegas

devimc:
Don't hot add again non-running container resources to avoid having extra
and useless resources

fixes #2186

NOTE: this PR won't break backward compatibility

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -1176,12 +1176,6 @@ func (s *Sandbox) CreateContainer(contConfig ContainerConfig) (VCContainer, erro
 		}
 	}()
 
-	// Sandbox is reponsable to update VM resources needed by Containers
-	err = s.updateResources()
-	if err != nil {
-		return nil, err
-	}
-
 	err = c.create()
 	if err != nil {
 		return nil, err
@@ -1189,6 +1183,14 @@ func (s *Sandbox) CreateContainer(contConfig ContainerConfig) (VCContainer, erro
 
 	// Add the container to the containers list in the sandbox.
 	if err = s.addContainer(c); err != nil {
+		return nil, err
+	}
+
+	// Sandbox is reponsable to update VM resources needed by Containers
+	// Update resources after having added containers to the sandbox, since
+	// container status is requiered to know if more resources should be added.
+	err = s.updateResources()
+	if err != nil {
 		return nil, err
 	}
 
@@ -1228,7 +1230,13 @@ func (s *Sandbox) StartContainer(containerID string) (VCContainer, error) {
 	}
 
 	s.Logger().Info("Container is started")
-	//Fixme Container delete from sandbox, need to update resources
+
+	// Update sandbox resources in case a stopped container
+	// is started
+	err = s.updateResources()
+	if err != nil {
+		return nil, err
+	}
 
 	return c, nil
 }
@@ -1487,10 +1495,6 @@ func (s *Sandbox) createContainers() error {
 	span, _ := s.trace("createContainers")
 	defer span.Finish()
 
-	if err := s.updateResources(); err != nil {
-		return err
-	}
-
 	for _, contConfig := range s.config.Containers {
 
 		c, err := newContainer(s, &contConfig)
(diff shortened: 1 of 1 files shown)


virtcontainers/sandbox.go
```diff
@@ -1951,6 +1955,12 @@ func (s *Sandbox) updateResources() error {
 func (s *Sandbox) calculateSandboxMemory() int64 {
 	memorySandbox := int64(0)
 	for _, c := range s.config.Containers {
+		// Do not hot add again non-running containers resources
+		if cont, ok := s.containers[c.ID]; ok && cont.state.State == types.StateStopped {
```
bergwolf: Then you need to recalculate resources in `s.StartContainer()`, right?

lifupan:
Hi @devimc, I have a question about this issue the PR fixed: once the container process was killed, the contaienrd/cri would get a task exit event and would try to delete the container from https://github.com/containerd/cri/blob/master/pkg/server/events.go#L320, which would trigger kata shimv2 to delete the container from the sandbox. But from your fix, it seem that the killed container was still existed in the sandbox, Im puzzled with why the stopped container wasn't deleted from sandbox or am I missed some thing?

devimc:
@jodh-intel there is already an integration test -  https://github.com/kata-containers/tests/blob/master/integration/kubernetes/k8s-number-cpus.bats

@lifupan yes, you are right, the container status is `stopped` but `delete` is never called, instead k8s creates a new container. Something weird, if this new container is killed, its status change to `stopped` and `delete` is called, I guess because k8s don't want to delete the *original* containers? :thinking:

virtcontainers/sandbox.go
devimc: good point!, let me do it and run the tests

devimc:
@bergwolf change applied, thanks

/test

lifupan:
> @lifupan yes, you are right, the container status is `stopped` but `delete` is never called, instead k8s creates a new container. Something weird, if this new container is killed, its status change to `stopped` and `delete` is called, I guess because k8s don't want to delete the _original_ containers? 🤔

Actually k8s didn't involved into the stopped container's keep/delete in runtime, it's containerd/cri's  responsibility. As you said it's weird, maybe there would be another bug embedded somewhere.

devimc:
can we merge this?

amshinde:
Did you test this with crio or containerd? Do you see the same behaviour with both? Would be good to track down any cri error.

devimc:
@amshinde no, I just tested with crio

amshinde:
@devimc Looks like a candidate for backporting. Can you open a PR to do so.

devimc:
@amshinde  done https://github.com/kata-containers/runtime/pull/2313

PR #2253, opened 2019-11-22
virtcontainers: rename GetOCISpec to GetPatchedOCISpec

### LINKED ISSUE #2252: Rename GetOCISpec
When I was debugging an issue in kata-runtime, it was difficult to realize that `GetOCISpec` method returns a patched version of the original spec, to avoid confusions we should rename it to something less generic

devimc:
GetOCISpec returns a patched version of the original OCI spec, it was modified
to support:
* capabilities
* Ephemeral storage
* k8s empty dir

In order to avoid consusions and make api clear, rename GetOCISpec
to GetPatchedOCISpec.

fixes #2252

### DIFF
--- a/virtcontainers/pkg/oci/utils.go
+++ b/virtcontainers/pkg/oci/utils.go
@@ -861,7 +861,10 @@ func ContainerConfig(ocispec specs.Spec, bundlePath, cid, console string, detach
 		Mounts:      containerMounts(ocispec),
 		DeviceInfos: deviceInfos,
 		Resources:   *ocispec.Linux.Resources,
-		Spec:        &ocispec,
+
+		// This is a custom OCI spec modified at SetEphemeralStorageType()
+		// to support ephemeral storage and k8s empty dir.
+		CustomSpec: &ocispec,
 	}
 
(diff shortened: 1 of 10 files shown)


virtcontainers/pkg/oci/utils.go
```diff
@@ -861,7 +861,10 @@ func ContainerConfig(ocispec specs.Spec, bundlePath, cid, console string, detach
 		Mounts:      containerMounts(ocispec),
 		DeviceInfos: deviceInfos,
 		Resources:   *ocispec.Linux.Resources,
-		Spec:        &ocispec,
+
+		// This is a custom OCI spec modified at SetEphemeralStorageType()
+		// to support ephemeral storage and k8s empty dir.
+		Spec: &ocispec,
```
jodh-intel: Idea: Could we also rename this to make it "stand out" a little more? (`CustomSpec: &ocispec,`?)

devimc: done, thanks

jodh-intel:
Thanks @devimc

PR #2257, opened 2019-11-22
virtcontainers: honour ContainerConfig struct comment and don't save OCI spec

### LINKED ISSUE #2256: Don't save container config Spec to disk
the following code block indicates that `Spec` won't be saved to disk

https://github.com/kata-containers/runtime/blob/8c7a83b936f747a32ad4b1cb534d9a54503f1c53/virtcontainers/container.go#L254-L255

unfortunately it's not honoured because in golang to specify that field should be omitted from the `json marshalling` it should has the json tag `-` not `_`

devimc:
Currently kata-runtime saves the Container OCI Spec even when it's not needed
and a comment in `ContainerConfig struct` specifically indicates that
it won't be saved to disk.
Use '-' as json tag instead of '_' to indicates that `Spec` field shouldn't
be saved to disk.

fixes #2256

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -252,7 +252,7 @@ type ContainerConfig struct {
 	Resources specs.LinuxResources
 
 	// Raw OCI specification, it won't be saved to disk.
-	Spec *specs.Spec `json:"_"`
+	Spec *specs.Spec `json:"-"`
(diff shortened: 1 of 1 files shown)


WeiZhang555:
Good catch!  @devimc

PR #2259, opened 2019-11-25
fc.go: Open firecracker log if hypervisor.enable_debug is true

teawater:
Create logs.fifo in fc.vmPath and open firecracker log in this file
if hypervisor.enable_debug is true.

Fixes: #2258


teawater:
https://github.com/kata-containers/runtime/pull/2073

PR #2245, opened 2019-11-21
vc: Sandbox#getAndSetSandboxBlockIndex shouldn't alter BlockIndex in case of error

tedyu:
Fixes #2244

Signed-off-by: Ted Yu <yuzhihong@gmail.com>

PR #2266, opened 2019-11-25
vc: Sandbox#getAndSetSandboxBlockIndex shouldn't alter BlockIndex in case of error

### LINKED ISSUE #2244: Sandbox#getAndSetSandboxBlockIndex shouldn't alter BlockIndex in case of error
# Description of problem

Currently Sandbox#getAndSetSandboxBlockIndex increments BlockIndex up front.
Later on if store.Store() returns error, the index increment is left behind.

# Expected result

BlockIndex should not increment in case of error.

Similar issue applies to decrementSandboxBlockIndex where the symptom may be more severe - in worst case, the block index may become negative.

# Actual result

tedyu:
Fixes #2244


tedyu:
@egernst 
I checked out stable-1.9 branch, created block-index-update-19 from that and applied the same fix from #2245.

Please take a look.

egernst:
It looks like you may need to rebase on stable-1.9?

tedyu:
I ran the following commands:

git checkout stable-1.9
git checkout -b block-index-update-19
git push origin block-index-update-19

Let me find the correct sequence.

egernst:
That is fine. When you are in GH and creating the actual PR, the target branch should be stable-1.9.  That's the spot.

tedyu:
In favor of #2267

PR #2268, opened 2019-11-25
vc: Sandbox#getAndSetSandboxBlockIndex shouldn't alter BlockIndex in case of error

### LINKED ISSUE #2244: Sandbox#getAndSetSandboxBlockIndex shouldn't alter BlockIndex in case of error

tedyu:
Fixes #2244

This is for stable-1.8 branch


tedyu:
@egernst @grahamwhaley 
Please take a look.

PR #2273, opened 2019-11-26
cache-factory: a few bug fix

Pennyzct:
# Description of problem
For now, we will encounter nil pointer runtime panic when launching kata containers from cache factory.
```
kata-runtime[133160]: time="2019-11-25T15:57:05.799620876+08:00" level=error msg="fatal
error" arch=arm64 name=kata-runtime panic="runtime error: invalid memory address or nil pointer
dereference" pid=133160 source=runtime
```
After fixing this, we will still have missing bridge address error.
```
docker: Error response from daemon: OCI runtime create failed: failed to get available address from bridges: unknown.
```
That's because that although we've already passed bridges info to clients from cache factory server, we've still missed the setting part when creating vm.

### DIFF
--- a/cli/factory.go
+++ b/cli/factory.go
@@ -88,7 +88,7 @@ func (s *cacheServer) Quit(ctx context.Context, empty *types.Empty) (*types.Empt
 		time.Sleep(time.Second)
 		s.quit()
 	}()
-	return nil, nil
+	return &types.Empty{}, nil
 }
 
 func (s *cacheServer) Status(ctx context.Context, empty *types.Empty) (*pb.GrpcStatus, error) {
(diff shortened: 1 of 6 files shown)


amshinde:
lgtm @Pennyzct !
As @jodh-intel pointed, would be good to have some unit tests for this.

Pennyzct:
Hi guys. thanks for the advice!
@jodh-intel @amshinde 
I think that unit test may not be enough here. It couldn't detect above runtime bugs.
I may need to create one integration test, just like what you guys did for [template factory](https://github.com/kata-containers/tests/blob/master/integration/vm_factory/vm_templating_test.sh). ;).

jodh-intel:
If unit tests aren't feasible, integration tests are fine with me but maybe @bergwolf has thoughts?

bergwolf:
Yeah, we definitely need an integration test for cache factory.

/cc @teawater

Pennyzct:
Hi guys 
I got all green here~~~

Issue #2165, opened 2019-11-05
Cloud Hypervisor: Use HTTP API


jcvenegas:
Recently cloud hypervisor added support for http API based, this will be useful in the future integration for upcomming dynamic configurations like hotplug cpu or memory, query the state of the VM. 
Related to https://github.com/kata-containers/runtime/issues/2046

Issue #516, opened 2018-07-26
ci: cri-containerd test fail under centos.


jcvenegas:
The ci startet to fail on centos tests runnning cri-tools tests.

All test work

(replace this text with details of what actually happened)

All cri-tools test fail with the error:

```
rpc error: code = Unknown desc = failed to start sandbox container: failed to create containerd task: OCI runtime create failed: rpc error: code = DeadlineExceeded desc = context deadline exceeded: unknown
```

When I debug a bit more the test, I found that failed when run mutiples sandbox in parallel.
By default cri-containerd test execute 8 jobs in parallel. This casue that each VM take more than
5 seconds to start to serve the grpc api.

We need to find if there is a waty to reduce startup VM time when mutiples VM are running.

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #2275, opened 2019-11-26
clh: use http client

jcvenegas:
- Generate API client using OpenAPI tools ( similar to fc implementation does)
- Instead of build a command, use Cloud Hypervisor http API

Despite of all the noise in the PR for generated code, the main changes are in

`virtcontainers/clh.go`
- [x] [Simplify bootVM timeout check](https://github.com/kata-containers/runtime/pull/2275#discussion_r351194689)
- [x]  [use docker for openapi requeriments](https://github.com/kata-containers/runtime/pull/2275#discussion_r351185584)
- [x] Add unit test
- [ ] [Check if can handle block based volumes for could plug boot](https://github.com/kata-containers/runtime/pull/2275#discussion_r351014580)

Fixes: #2165

Depends-on: github.com/kata-containers/tests#2100


jcvenegas:
\cc @ericooper could you review this PR - please notice that  cli builder will be gone, so we may want to wait adding more unit test on current code base.

Gopkg.lock
```diff
@@ -604,10 +604,11 @@
   revision = "8dd112bcdc25174059e45e07517d9fc663123347"
 
 [[projects]]
-  digest = "1:b20c63a56900e442d5f435613fefc9392cbe8849467510fcc3869dbdad9441bb"
+  digest = "1:514d2147b3fd915870b534e3578feed1c40a82aa751a015b160785616bfa2130"
```
egernst: @jcvenegas -- just because i'm lame -- What/why did we need to update? Were we using openapi before?

virtcontainers/pkg/cloud-hypervisor/Makefile
```diff
@@ -0,0 +1,17 @@
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+all: clean-generated-code
```
egernst: Did you add this Makefile, and then do a make all in order to pull the rest?  Perhaps you can clarify what part you added versus what was created from the OpenAPI tooling?  It'll help me better understand how you carried this out / reproduce from scratch.  It *may* make sense to have that as two commits, to make it clear.  Otherwise, at least detail it out on the commit message for "clh: Generate client code"

virtcontainers/clh.go
```diff
@@ -294,13 +246,18 @@ func (clh *cloudHypervisor) startSandbox(timeout int) error {
 	if err != nil {
 		return fmt.Errorf("failed to launch cloud-hypervisor: %s, error messages from log: %s", err, strErr)
 	}
+	clh.state.PID = pid
```
egernst: @jcvenegas Can you share a snapshot of jaeger trace output when launching?  I'm curious how long it takes to start the VMM itself, and if it makes sense to do this at top of function instead of here (similar to what's in FC)?

virtcontainers/clh.go
```diff
@@ -294,13 +246,18 @@ func (clh *cloudHypervisor) startSandbox(timeout int) error {
...
 	if err := clh.waitVMM(clhTimeout); err != nil {
 		clh.Logger().WithField("error", err).Warn("cloud-hypervisor init failed")
 		clh.shutdownVirtiofsd()
 		return err
 	}
 
-	clh.state.PID = pid
+	if err := clh.bootVM(clhTimeout); err != nil {
+		return err
+	}
```
egernst: Are we ready to do this yet? What if we have devices to add that are part of a further container request?

egernst:
For at least the functions which implement the Hypervisor interface, it'd be helpful to describe exactly what is being done with respect to the configuration, the VMM and the VM itself within each function.  See comments in fc.go as a reference for what I'm hoping to read.

virtcontainers/clh.go
```diff
@@ -421,28 +378,58 @@ func (clh *cloudHypervisor) addDevice(devInfo interface{}, devType deviceType) e
...
+		devLogger.WithFields(log.Fields{
+			"path": v.UdsPath,
+			"cid":  v.ContextID,
+			"port": v.Port,
 		}).Info("Adding HybridVSock")
-		clh.cliBuilder.SetVsock(&CLIVsock{
-			cid:        uint32(v.ContextID),
-			socketPath: v.UdsPath,
-			iommu:      false,
-		})
```
egernst: can we make this a function? Easier testing, simpler case statement...

virtcontainers/clh.go
```diff
@@ -421,28 +378,58 @@ func (clh *cloudHypervisor) addDevice(devInfo interface{}, devType deviceType) e
...
+		devLogger.Infof("Adding Endpoint of type %v", v)
+		n := chclient.NetConfig{}
+		n.Mac = v.HardwareAddr()
+		n.Tap = v.NetworkPair().TapInterface.TAPIface.Name
+
+		//TODO server does not allow empty IP fail serde to get it
+		n.Ip = "0.0.0.0"
+		n.Mask = "0.0.0.0"
+		clh.vmconfig.Net = append(clh.vmconfig.Net, n)
+
```
egernst: similar to below -- can we make this a simple function?

virtcontainers/clh.go
```diff
@@ -421,28 +378,58 @@ func (clh *cloudHypervisor) addDevice(devInfo interface{}, devType deviceType) e
...
 		}).Info("Adding HybridVSock")
-		clh.cliBuilder.SetVsock(&CLIVsock{
-			cid:        uint32(v.ContextID),
-			socketPath: v.UdsPath,
-			iommu:      false,
-		})
+		clh.vmconfig.Vsock = []chclient.VsockConfig{{Cid: defaultGuestVSockCID, Sock: v.UdsPath}}
+	case types.Volume:
+
+		if clh.config.SharedFS != config.VirtioFS {
```
egernst: does it support block based volumes?

virtcontainers/clh.go
```diff
@@ -421,28 +378,58 @@ func (clh *cloudHypervisor) addDevice(devInfo interface{}, devType deviceType) e
...
+				CacheSize: int64(clh.config.VirtioFSCacheSize << 20),
+				Sock:      vfsdSockPath,
+				//TODO
+				NumQueues: 1,
+				//TODO
+				QueueSize: 512,
+				//TODO dax
+			},
+		}
+		devLogger.Debug("Adding Volume to hypervisor ", v.HostPath, ":", v.MountTag)
```
egernst: 1. A few TODO's here?
2. perhaps would be cleaner to add these to their own functions? Ie "clhAddVolume" ?

Gopkg.lock
jcvenegas: It was missing `golang.org/x/oauth2` added to the commit message

virtcontainers/clh.go
```diff
@@ -861,439 +784,165 @@ func (clh *cloudHypervisor) LaunchClh() (string, int, error) {
 //###########################################################################
 
 const (
-	cctOFF  string = "off"
-	cctFILE string = "file"
+	cctOFF  string = "Off"
```
jodh-intel: Why? Is this a CLH change since the previous version (I still say why btw though :smile:)?

virtcontainers/clh.go
```diff
@@ -861,439 +784,165 @@ func (clh *cloudHypervisor) LaunchClh() (string, int, error) {
...
-	cmdline.args = append(cmdline.args, cscFs)
-
-	fsarg := "tag=" + o.tag + ",sock=" + o.socketPath
-	if o.dax {
-		fsarg += ",dax=on"
-	} else {
-		fsarg += ",num_queues=" + strconv.FormatUint(uint64(o.queues), 10) + ",queue_size=" + strconv.FormatUint(uint64(o.queueSize), 10)
+func (c *cloudHypervisor) bootVM(timeout int) error {
+	if timeout < 0 {
+		return fmt.Errorf("Invalid timeout %ds", timeout)
```
jodh-intel: Great that we are checking this, but the code seems to be littered with this check. Could you find a way to wrap it in a function or just perform the check once so that other functions can rely on the timeout?

virtcontainers/clh.go
```diff
@@ -861,439 +784,165 @@ func (clh *cloudHypervisor) LaunchClh() (string, int, error) {
...
-		if net.iommu {
-			networks += "tap=" + tapName + ",mac=" + net.mac + ",iommu=on"
-		} else {
-			networks += "tap=" + tapName + ",mac=" + net.mac
+	if c.config.Debug {
+		bodyBuf, err := json.Marshal(c.vmconfig)
+		if err != nil {
+			return err
 		}
+		c.Logger().Debugf("%s", string(bodyBuf))
```
jodh-intel: Could you make this log call structured. Something like the following:

```go
c.Logger().WithField("body", string(bodyBuf)).Debug("VM config")
```

virtcontainers/clh.go
```diff
@@ -861,439 +784,165 @@ func (clh *cloudHypervisor) LaunchClh() (string, int, error) {
...
-	if d.kernel != nil {
-		d.kernel.Build(cmdLine)
-	}
-	if d.nets != nil {
-		d.nets.Build(cmdLine)
-	}
-	if d.cmdline != nil {
-		d.cmdline.Build(cmdLine)
+	if info.State != "Running" {
+		return fmt.Errorf("VM state is not created after create")
```
jodh-intel: "not running after create"?

virtcontainers/clh.go
```diff
@@ -421,28 +378,58 @@ func (clh *cloudHypervisor) addDevice(devInfo interface{}, devType deviceType) e
...
+				//TODO
+				QueueSize: 512,
+				//TODO dax
+			},
+		}
+		devLogger.Debug("Adding Volume to hypervisor ", v.HostPath, ":", v.MountTag)
+
 	default:
-		clh.Logger().WithField("function", "addDevice").Warnf("Add device of type %v is not supported.", v)
+		devLogger.Warnf("Add device of type %v is not supported.", v)
```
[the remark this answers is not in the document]
jcvenegas: Just refactor as it logs in different places so not add those extra characters but just in one place

virtcontainers/clh.go
jcvenegas: it is just to match the the API definition

virtcontainers/clh.go
```diff
@@ -157,98 +132,76 @@ func (clh *cloudHypervisor) createSandbox(ctx context.Context, id string, networ
...
-		backingFile: "/dev/shm",
-	})
+	clh.vmconfig.Memory.Size = int64(clh.config.MemorySize) << utils.MibToBytesShift
+	clh.vmconfig.Memory.File = "/dev/shm"
 	// Set initial amount of cpu's for the cloud hypervisor
-	clh.cliBuilder.SetCpus(&CLICpus{
-		cpus: clh.config.NumVCPUs,
-	})
+	clh.vmconfig.Cpus = chclient.CpuConfig{
+		CpuCount: int32(clh.config.NumVCPUs),
```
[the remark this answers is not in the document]
jcvenegas: yeah, this is a limitation from openAPI format defintion, I added a comment

virtcontainers/pkg/cloud-hypervisor/Makefile
```diff
@@ -0,0 +1,17 @@
...
+	java -jar openapi-generator-cli.jar generate -i cloud-hypervisor.yaml -g go -o ./client
```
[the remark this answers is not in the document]
jcvenegas: The CI wont run this but this is  a good idea, I'll add it

virtcontainers/clh.go
jcvenegas: I have not tried but is not possible to hotAdd volumes,  for K8s this the only way to go.

virtcontainers/clh.go
```diff
@@ -421,28 +378,58 @@ func (clh *cloudHypervisor) addDevice(devInfo interface{}, devType deviceType) e
...
-		})
+		devLogger.Infof("Adding Endpoint of type %v", v)
+		n := chclient.NetConfig{}
+		n.Mac = v.HardwareAddr()
+		n.Tap = v.NetworkPair().TapInterface.TAPIface.Name
+
+		//TODO server does not allow empty IP fail serde to get it
+		n.Ip = "0.0.0.0"
+		n.Mask = "0.0.0.0"
+		clh.vmconfig.Net = append(clh.vmconfig.Net, n)
```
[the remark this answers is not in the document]
jcvenegas: Fixed, depends on https://github.com/cloud-hypervisor/cloud-hypervisor/pull/480

egernst:
nit on https://github.com/kata-containers/runtime/pull/2275/commits/1dd2135447bc2d6ef1e89590195ff6de18bde331

Can we be consistent on naming (s/CH/CLH/)

virtcontainers/clh.go
```diff
@@ -877,461 +767,256 @@ func MaxClhVCPUs() uint32 {
...
-func (d *DefaultCLIBuilder) SetLogFile(logFile *CLILogFile) {
-	d.logFile = logFile
+	clh.vmconfig.Fs = []chclient.FsConfig{
+		{
+			Tag:       volume.MountTag,
+			CacheSize: int64(clh.config.VirtioFSCacheSize << 20),
+			Sock:      vfsdSockPath,
+			//TODO this is not expose to user
+			//TODO change if dax is set in config
+			NumQueues: 1,
```
egernst: Need these "todo's" documented a bit more, with corresponding GH issues.

virtcontainers/clh.go
```diff
@@ -265,13 +215,12 @@ func (clh *cloudHypervisor) createSandbox(ctx context.Context, id string, networ
...
 	}
-	clh.cliBuilder.SetAPISocket(&CLIAPISocket{
-		socketPath: apiSocketPath,
-	})
+	clh.state.apiSocket = apiSocketPath
 
 	return nil
 }
 
+// startSandbox will start the VMM for the given sandbox.
```
egernst:
```suggestion
// startSandbox will start the VMM and boot the virtual machine for the given sandbox.
```

virtcontainers/clh.go
```diff
@@ -165,98 +136,77 @@ func (clh *cloudHypervisor) createSandbox(ctx context.Context, id string, networ
...
-		backingFile: "/dev/shm",
-	})
+	clh.vmconfig.Memory.Size = int64(clh.config.MemorySize) << utils.MibToBytesShift
+	clh.vmconfig.Memory.File = "/dev/shm"
 	// Set initial amount of cpu's for the cloud hypervisor
-	clh.cliBuilder.SetCpus(&CLICpus{
-		cpus: clh.config.NumVCPUs,
-	})
+	clh.vmconfig.Cpus = chclient.CpuConfig{
+		// cast to int32, this is a openAPI limitation that not support unsigned values
```
egernst: Does a negative number of CPUs make sense here, anyway? If we want to still add this, suggestion:

```suggestion
		// cast to int32, as openAPI has a limitation that it does not support unsigned values 
```

virtcontainers/clh.go
```diff
@@ -165,98 +136,77 @@ func (clh *cloudHypervisor) createSandbox(ctx context.Context, id string, networ
...
 
-	clh.cliBuilder.SetConsole(&CLIConsole{
-		consoleType: cctOFF,
-	})
+	clh.vmconfig.Console = chclient.ConsoleConfig{
+		Mode: cctOFF,
+	}
 
 	// Move the API endpoint socket location for the
 	// by default enabled api endpoint
```
egernst: this isn't clear to me -- can you clarify this comment?

virtcontainers/clh.go
```diff
@@ -426,25 +387,14 @@ func (clh *cloudHypervisor) addDevice(devInfo interface{}, devType deviceType) e
...
-			cid:        uint32(v.ContextID),
-			socketPath: v.UdsPath,
-			iommu:      false,
-		})
+		clh.addVSock(defaultGuestVSockCID, v.UdsPath)
+	case types.Volume:
+		err = clh.addVolume(v)
 	default:
 		clh.Logger().WithField("function", "addDevice").Warnf("Add device of type %v is not supported.", v)
+		return fmt.Errorf("Not implemented support for %s", v)
```
egernst: Are we adding any docs with this PR?  I think its important to call out that we don't support block devices, specifically.

virtcontainers/clh.go
```diff
@@ -877,461 +767,256 @@ func MaxClhVCPUs() uint32 {
...
+		return openApiClientError(err)
+	}
 
-func (d *DefaultCLIBuilder) AddKernelParameters(params []Param) {
+	c.Logger().Debugf("VM state after boot: %#v", info)
 
-	if d.cmdline == nil {
-		d.cmdline = &CLICmdline{}
+	if info.State != "Running" {
+		return fmt.Errorf("VM state is not created after create")
```
egernst: This doesn't look correct based on the clause which this comment is within? Can you clarify here, and line 908?

virtcontainers/clh.go
```diff
@@ -520,11 +471,22 @@ func (clh *cloudHypervisor) terminate() (err error) {
...
+
+	if err != nil {
+		return err
+	}
+
+	if !clh_running {
+		return nil
+	}
+
+	defer clh.cleanupVM()
```
egernst: can/should this be part of the defer func above?

virtcontainers/clh.go
```diff
@@ -877,461 +767,256 @@ func MaxClhVCPUs() uint32 {
...
+			QueueSize: 512,
+		},
+	}
+	// TODO: should virtiofsd be started where for each add is done?
+	clh.Logger().Debug("Adding share volume to hypervisor: ", volume.MountTag)
+	return nil
 }
 
-func (d *DefaultCLIBuilder) GetCommandLine() (*CommandLine, error) {
+func (clh *cloudHypervisor) cleanupVM() error {
```
egernst: description of function always welcome

virtcontainers/clh.go
likebreath: Sorry for the confusion. Those are typos in the error message, while the API calls and checks are correct.

likebreath:
```suggestion
		return fmt.Errorf("VM state is not 'Running' after 'BootVM'")
```

virtcontainers/clh.go
```diff
@@ -877,461 +767,256 @@ func MaxClhVCPUs() uint32 {
...
-type CLICmdline struct {
-	params []Param
-}
 
-func (o *CLICmdline) Build(cmdline *CommandLine) {
+	c.Logger().Debugf("VM state after create: %#v", info)
 
-	cmdline.args = append(cmdline.args, cscCmdline)
+	if info.State != "Created" {
+		return fmt.Errorf("VM state is not running after create")
```
likebreath:
```suggestion
		return fmt.Errorf("VM state is not 'Created' after 'CreateVM'")
```

virtcontainers/clh.go
likebreath: I believe the comment means: "Overwrite the default value of HTTP API socket path for cloud hypervisor"

virtcontainers/clh.go
likebreath:
```suggestion
// Remove temporary folder and files used by the virtual machine
func (clh *cloudHypervisor) cleanupVM() error {
```

virtcontainers/clh.go
jcvenegas: It does not make sense but a limitation with the API today

virtcontainers/clh.go
egernst: Let's update with that clarity?

virtcontainers/clh.go
```diff
@@ -164,99 +135,78 @@ func (clh *cloudHypervisor) createSandbox(ctx context.Context, id string, networ
...
 	}
 
-	clh.cliBuilder.SetConsole(&CLIConsole{
-		consoleType: cctOFF,
-	})
+	clh.vmconfig.Console = chclient.ConsoleConfig{
+		Mode: cctOFF,
+	}
 
 	// Move the API endpoint socket location for the
```
egernst:
```suggestion
	// Overwrite the default value of HTTP API socket path for cloud hypervisor
```

virtcontainers/clh.go
```diff
@@ -164,99 +135,78 @@ func (clh *cloudHypervisor) createSandbox(ctx context.Context, id string, networ
...
 	// by default enabled api endpoint
```
egernst:
```suggestion
```

jcvenegas:
travis keeps failing seems that one case was missing for spell check.
PR: https://github.com/kata-containers/tests/pull/2143
@jodh-intel  please take a look

virtcontainers/clh.go
```diff
@@ -302,19 +284,26 @@ func (clh *cloudHypervisor) startSandbox(timeout int) error {
...
+	clh.state.PID = pid
+
 	if err := clh.waitVMM(clhTimeout); err != nil {
 		clh.Logger().WithField("error", err).WithField("output", clh.cmdOutput.String()).Warn("cloud-hypervisor init failed")
 		clh.shutdownVirtiofsd()
 		return err
 	}
 
-	clh.state.PID = pid
+	if err := clh.bootVM(ctx); err != nil {
```
sboeuf: One comment here. You chose to create the VMM and boot the VM in the same `startSandbox()` function, but you could have done it slightly differently too. The VMM can be started with only the API socket as argument (from `createSandbox()`), and later on, you can issue a `VmCreate(vmConfig)` to start the VM from `startSandbox()`.

virtcontainers/clh.go
```diff
@@ -1202,141 +1170,188 @@ func (o *CLICmdline) Build(cmdline *CommandLine) {
...
-	logFile   *CLILogFile
-	vsock     *CLIVsock
-	cmdline   *CLICmdline
-	nets      *CLINets
-}
+//****************************************
+// API calls
+//****************************************
+func (clh *cloudHypervisor) isClhRunning(timeout int) (bool, error) {
+	if timeout < 0 {
```
sboeuf: You could avoid checking for a valid timeout if you were using `uint` instead of `int`

virtcontainers/clh.go
sboeuf: @jcvenegas 1 queue is fine for now since virtio-fs is still under development regarding the multiqueue support.

virtcontainers/clh.go
```diff
@@ -1202,141 +1170,188 @@ func (o *CLICmdline) Build(cmdline *CommandLine) {
...
+
+	clh.vmconfig.Fs = []chclient.FsConfig{
+		{
+			Tag:       volume.MountTag,
+			CacheSize: int64(clh.config.VirtioFSCacheSize << 20),
+			Sock:      vfsdSockPath,
+			// need to look for the best value or provide a way the user can modifiy
+			// See: https://github.com/cloud-hypervisor/cloud-hypervisor/issues/508
+			NumQueues: 1,
+			QueueSize: 512,
```
sboeuf: Use 1024 as the default, this should slightly improve the performance.

virtcontainers/clh.go
```diff
@@ -51,8 +56,8 @@ const (
 	supportedMajorVersion = 0
 	supportedMinorVersion = 3
 	defaultClhPath        = "/usr/local/bin/cloud-hypervisor"
-	virtioFsCacheAlways   = "always"
-	maxClhVcpus           = uint32(64)
+	//virtioFsCacheAlways   = "always"
```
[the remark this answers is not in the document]
jcvenegas: Thanks for look at this, actually I wanted to ask you  in the  case of the cli we used if "always" we enabled the dax  via cloud-hypervisor --fs  dax=on, in the http api there is not an option for that, in the case of `always` what values should be setup, and for non as well what should be used?

virtcontainers/pkg/cloud-hypervisor/cloud-hypervisor.yaml
```diff
@@ -0,0 +1,406 @@
+openapi: 3.0.1
+info:
```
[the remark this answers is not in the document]
jcvenegas: we could point at some point to a branch, I prefer keep a copy of what is used, just to have an snapshot at any moment of how the yaml looks.

virtcontainers/clh.go
```diff
@@ -164,118 +167,97 @@ func (clh *cloudHypervisor) createSandbox(ctx context.Context, id string, networ
...
-	clh.cliBuilder.SetCpus(&CLICpus{
-		cpus: clh.config.NumVCPUs,
-	})
+	// Set initial memomory size of the virtual machine
+	clh.vmconfig.Memory.Size = int64(clh.config.MemorySize) << utils.MibToBytesShift
+	clh.vmconfig.Memory.File = "/dev/shm"
+	// Set initial amount of cpu's for the virtual machine
+	clh.vmconfig.Cpus = chclient.CpuConfig{
+		// cast to int32, as openAPI has a limitation that it does not support unsigned values
+		CpuCount: int32(clh.config.NumVCPUs),
```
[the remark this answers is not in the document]
jcvenegas: Yes, I say I needed to lock the hypervisor to a commit previous to that, you want we update now or after merge this PR, the functionality for the version that is lock this PR should be enough for what it cover this PR.

virtcontainers/clh.go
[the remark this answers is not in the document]
jcvenegas: I got it now, values updated.

virtcontainers/pkg/cloud-hypervisor/cloud-hypervisor.yaml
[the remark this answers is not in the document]
sboeuf: Ok that's fine!

virtcontainers/clh.go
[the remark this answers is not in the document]
sboeuf: It's really up to you, but I don't think you should worry about it right now. A follow up PR will be fine. I simply wanted to let you know that this part had changed very recently :)

virtcontainers/clh.go
```diff
@@ -1202,141 +1175,195 @@ func (o *CLICmdline) Build(cmdline *CommandLine) {
...
+	if clh.config.VirtioFSCache == virtioFsCacheAlways {
+		clh.vmconfig.Fs = []chclient.FsConfig{
+			{
+				Tag:       volume.MountTag,
+				CacheSize: int64(clh.config.VirtioFSCacheSize << 20),
+				Sock:      vfsdSockPath,
+				NumQueues: clhFsQueues,
+				QueueSize: clhFsQueueSize,
+			},
+		}
```
sboeuf: This part is fine and will work as expected :+1:

virtcontainers/clh.go
```diff
@@ -1202,141 +1175,195 @@ func (o *CLICmdline) Build(cmdline *CommandLine) {
...
+	} else {
+		clh.vmconfig.Fs = []chclient.FsConfig{
+			{
+				Tag:       volume.MountTag,
+				Sock:      vfsdSockPath,
+				NumQueues: clhFsQueues,
+				QueueSize: clhFsQueueSize,
+			},
+		}
```
sboeuf: I'm not sure about the behavior from this part. By not providing the `CacheSize`, and because the OpenAPI says that it expects an `int64`, I think the default value is going to be 0. But that will be translated into `Some(0)` for the Rust code, meaning it will still try to use DAX with the shared memory region when really what you want is no cache at all.
Please give it a try and let me know if that's the case, but TBH I think that's something we need to fix in CH.

sboeuf: @jcvenegas I've just opened https://github.com/cloud-hypervisor/cloud-hypervisor/pull/515 to fix this. It'd be nice if you could give it a shot as I think this is the right way to configure `FsConfig`.
Note that because I added `dax` default to `true` and `cache_size` default to `8G`, you shouldn't need to specify `dax` and `cache_size` when the only information you get from your user is `virtioFsCacheAlways   = "always"`.

virtcontainers/clh.go
```diff
@@ -1202,141 +1175,195 @@ func (o *CLICmdline) Build(cmdline *CommandLine) {
...
+	clh.Logger().WithFields(log.Fields{
+		"mac": mac,
+		"tap": tapPath,
+	}).Info("Adding Net")
+
+	// FIXME: This is required by CH
+	// remove after PR is merged:
+	// https://github.com/cloud-hypervisor/cloud-hypervisor/pull/480
+	ip := "0.0.0.0"
+	mask := "0.0.0.0"
```
sboeuf: @jcvenegas @likebreath I've just submitted https://github.com/cloud-hypervisor/cloud-hypervisor/pull/516 which should solve this issue. Let me know if that works for you.

virtcontainers/clh.go
egernst: I think matching FC.go is okay for now.  We can look to see if there are any measurable benefits in a follow up.

egernst:
@jcvenegas @likebreath can you PTAL at the CI failure?

jcvenegas:
For ARM I just check  and is not related with it, I restarted the job.

The CLH CI fail is  failing in different  docker run every time, this is due to https://github.com/kata-containers/tests/issues/2141 , so this is expected to fail until we find the root cause. I am trying seems that is a race condition with the agent, but I could not reproduce it recently, lets ignore clh CI until is more stable.

egernst:
Since this occurs w CLH on master already today (not a regression), let’s get this change in and focus on resolving the test failures. Thanks folks.

virtcontainers/clh.go
```diff
@@ -877,461 +864,258 @@ func MaxClhVCPUs() uint32 {
...
-//**********************************
-// The API socket
-//**********************************
-type CLIAPISocket struct {
-	socketPath string
-}
+	timeStart := time.Now()
+	cl := clh.client()
+	for {
+		ctx, cancel := context.WithTimeout(context.Background(), clhAPITimeout*time.Second)
```
tedyu: If, at the end of iteration N, the remaining time (w.r.t. timeout) is shorter than clhAPITimeout, we would still wait for clhAPITimeout in iteration N+1. This would make the total wait longer than timeout.

Is this Okay ?

virtcontainers/clh.go
tedyu: #516 has been merged.
Can the above be removed ?

PR #2278, opened 2019-11-26
Fc comment

egernst:
correct a comment in fc.go.


amshinde:
@egernst So hypervisor != VMM ?  :)

amshinde:
@egernst So hypervisor != VMM ?  :)

Issue #2279, opened 2019-11-29
virtiofsd will dead if cache mode not set in configuration


jongwu:
# Description of problem

when using virtiofs as the shareFS, if cache mode not set in configuration, virtiofsd will dead which makes container won't start.

# Expected result

default cache mode should be set to avoid that failure.

# Actual result

container won't start as the virtiofsd dead.

PR #2280, opened 2019-11-29
virtiofs: add default value for virtioFsCache type.

jongwu:
If no virtioFsCache type set in configuration file, virtiofsd will
not start well as a daemon, which make container start fail if virtio-fs
as its shared file system.

Fixed: #2279

@jodh-intel @grahamwhaley

### DIFF
--- a/pkg/katautils/config.go
+++ b/pkg/katautils/config.go
@@ -342,6 +342,14 @@ func (h hypervisor) defaultBridges() uint32 {
 	return h.DefaultBridges
 }
 
+func (h hypervisor) defaultVirtioFSCache() string {
+	if h.VirtioFSCache == "" {
+		return defaultVirtioFSCacheMode
+	}
+
+	return h.VirtioFSCache
+}
+
 func (h hypervisor) blockDeviceDriver() (string, error) {
 	supportedBlockDrivers := []string{config.VirtioSCSI, config.VirtioBlock, config.VirtioMmio, config.Nvdimm, config.VirtioBlockCCW}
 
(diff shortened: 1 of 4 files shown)


virtcontainers/pkg/oci/utils.go
```diff
@@ -642,6 +642,8 @@ func addHypervisporVirtioFsOverrides(ocispec specs.Spec, sbConfig *vc.SandboxCon
 
 	if value, ok := ocispec.Annotations[vcAnnotations.VirtioFSCache]; ok {
 		sbConfig.HypervisorConfig.VirtioFSCache = value
+	} else {
+		sbConfig.HypervisorConfig.VirtioFSCache = "none"
```
bergwolf: Please define it in `pkg/katautils/config-settings.go` and set it in `GetDefaultHypervisorConfig()`.

jongwu: Thanks, I set it in the wrong place.

grahamwhaley:
CI moaning - you need to s/Fixed/Fixes/ in your commit text :-)
```
ERROR: No "Fixes" found
```

Question - why does the daemon not start well? Is it because kata forms the commandline badly, or does the daemon not work if no mode is supplied? Just wondering if there is anything we need to discuss/push to the virtio-fs authors??

jongwu:
> CI moaning - you need to s/Fixed/Fixes/ in your commit text :-)
> 
> ```
> ERROR: No "Fixes" found
> ```
> 
> Question - why does the daemon not start well? Is it because kata forms the commandline badly, or does the daemon not work if no mode is supplied? Just wondering if there is anything we need to discuss/push to the virtio-fs authors??

virtiofsd will exit or change into a zombie process immediately when it execute, as no value provided for "--cache=".
I thinks it's kata part of job to offer the correct argument for the command.

pkg/katautils/config.go
```diff
@@ -342,6 +342,14 @@ func (h hypervisor) defaultBridges() uint32 {
 	return h.DefaultBridges
 }
 
+func (h hypervisor) defaultVirtioFSCache() string {
```
devimc: please add a unit test

jongwu: yeah, done , thanks @devimc

devimc: thanks

jongwu:
I'm not sure the ci failure is related to this commit. WDYT @devimc?

devimc:
@jongwu nothing related to this change, restarting debian CI

jongwu:
thanks @devimc

Issue #2281, opened 2019-11-29
enable virtiofs when using kernel v4.19


jongwu:
# Description of problem
now, using the way like "mount source dest" to mount virtiofs will fail for kernel v4.19

# Expected result

using "-o tag=xxx" to let it work for current kernel.

# Actual result

mount virtiofs will fail since the mount command is not correct for current kernel.

PR #2282, opened 2019-11-29
virtiofs: enable virtiofs for kernel v4.19

jongwu:
For kernel v4.19, we can't use the "mount source dest" way to mount
virtiofs, so change the source mount point to none and add tag in options.
more info refer to https://virtio-fs.gitlab.io/howto-qemu.html 

Fixed: #2281
@jodh-intel @grahamwhaley

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -60,6 +60,7 @@ var (
 	defaultKataHostSharedDir    = "/run/kata-containers/shared/sandboxes/"
 	defaultKataGuestSharedDir   = "/run/kata-containers/shared/containers/"
 	mountGuestTag               = "kataShared"
+	mountSourceVirtioFS         = "none"
 	defaultKataGuestSandboxDir  = "/run/kata-containers/sandbox/"
 	type9pFs                    = "9p"
 	typeVirtioFS                = "virtio_fs"
@@ -72,7 +73,7 @@ var (
 	kataNvdimmDevType           = "nvdimm"
 	kataVirtioFSDevType         = "virtio-fs"
 	sharedDir9pOptions          = []string{"trans=virtio,version=9p2000.L,cache=mmap", "nodev"}
-	sharedDirVirtioFSOptions    = []string{"default_permissions,allow_other,rootmode=040000,user_id=0,group_id=0", "nodev"}
+	sharedDirVirtioFSOptions    = []string{"tag=" + mountGuestTag + ",default_permissions,allow_other,rootmode=040000,user_id=0,group_id=0", "nodev"}
 	sharedDirVirtioFSDaxOptions = "dax"
 	shmDir                      = "shm"
 	kataEphemeralDevType        = "ephemeral"
@@ -884,7 +885,7 @@ func setupStorages(sandbox *Sandbox) []*grpc.Storage {
 			}
 			sharedVolume := &grpc.Storage{
 				Driver:     kataVirtioFSDevType,
-				Source:     mountGuestTag,
+				Source:     mountSourceVirtioFS,
 				MountPoint: kataGuestSharedDir(),
 				Fstype:     typeVirtioFS,
 				Options:    sharedDirVirtioFSOptions,


virtcontainers/kata_agent.go
```diff
@@ -72,7 +73,7 @@ var (
 	kataNvdimmDevType           = "nvdimm"
 	kataVirtioFSDevType         = "virtio-fs"
 	sharedDir9pOptions          = []string{"trans=virtio,version=9p2000.L,cache=mmap", "nodev"}
-	sharedDirVirtioFSOptions    = []string{"default_permissions,allow_other,rootmode=040000,user_id=0,group_id=0", "nodev"}
+	sharedDirVirtioFSOptions    = []string{"tag=kataShared,default_permissions,allow_other,rootmode=040000,user_id=0,group_id=0", "nodev"}
```
bergwolf: Please use `mountGuestTag` above.

jongwu: OK, really better. Thanks @bergwolf

jongwu:
> I guess you'll need to update CI to fetch proper virtiofsd binary otherwise ci would fail.

I think I have no authorization to update ci. How could I do that or could someone help me?

jongwu:
> Thanks for raising @jongwu.
> 
> I'm not clear why this change is _specifically_ for the 4.19 kernel. Please can you add details and specify whether the change will continue to work for older/newer kernels?

I get this info from the official document of virtio-fs (https://virtio-fs.gitlab.io/howto-qemu.html) and I have checked for it.
I highlight kernel version in commit message as current kernel version is v4.19.
IMO, mount command needs support from kernel, so it must obey the kernel rule. 
I have no idea for now what's the behavior in new kernel or the older one. I think it just depends on the implementation of virtio-fs part in kernel.

jodh-intel:
@jongwu - thanks for the link. I think it would be useful to specify those details in the commit to make it easier to understand why this change was needed.

jongwu:
done @jodh-intel . need test @devimc, thanks

jongwu:
Are there someone know the exact qemu and kernel source code in virtio-fs ci test?@jodh-intel @devimc @bergwolf . I find the the source of qemu and kernel significantly impact the test result of virtiofs when running kata. I need more test to locate the root cause of the failure.

grahamwhaley:
Hi @jongwu - afaik, all versions of components used should come from the https://github.com/kata-containers/runtime/blob/master/versions.yaml file.
Looks like the relevant sections are at:

https://github.com/kata-containers/runtime/blob/master/versions.yaml

https://github.com/kata-containers/runtime/blob/master/versions.yaml#L160-L163

Also, you should be able to verify maybe by checking the CI logs to see exactly what they built/ran.
/cc @chavafg

jongwu:
thanks @grahamwhaley

jongwu:
As virtiofs ci test choose 5.3 kernel，this commit won't pass. If kata will not use 4.19 kernel when using virtiofs, this pr doesn't need any more. @jodh-intel  @grahamwhaley  @devimc  @bergwolf

grahamwhaley:
@jongwu - good find on the kernel version...!!!
/cc @chavafg @jcvenegas - not sure if this is a CI thing we can fix, or just something we have to live with right now.

raravena80:
@jongwu any updates?

Your weekly Kata herder.

jongwu:
@raravena80 @grahamwhaley virtio-fs will be used in different way for kernel 4.19 and 5.2+. if kernel version will move to 5.4 (this version is LTS) in a short time, enable virtio-fs for 4.19 maybe unnecessary,  then this pr can close.

grahamwhaley:
It does indeed seem that [Greg expects to make 5.4 the longerm](https://people.kernel.org/gregkh/next-long-term-supported-kernel-release), and that is listed in the [kernel releases page](https://www.kernel.org/category/releases.html), but has not made it to the [kernel front page yet](https://www.kernel.org/). I guess that will happen in the next few days or at the next minor version bump.... so, yes, when that is all fully official let's bump to 5.4 I think, and this PR can close by the sound of it.
Thanks for the update @jongwu !

amshinde:
Would have been good to choose the right mount options based on the kernel, but since this is the guest kernel, there is no reliable way to find this out.
So yeah close this once we move to 5.4.

c3d:
Closing, per @amshinde comment on Jan 2 (we are now well past 5.4)

PR #2304, opened 2019-12-01
vc: Use map to represent volumeStorages

### LINKED ISSUE #2303: Use map to represent volumeStorages

Currently kataAgent#replaceOCIMountsForStorages uses nested loop to match volumeStorages with spec.Mounts.
We can remove one loop.

**How can it be improved?**

Use map to represent volumeStorages.
The match would become a lookup in the volumeStorages map.

**Additional Information**

tedyu:
This PR removes one loop in kataAgent#replaceOCIMountsForStorages

    Fixes #2303

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -999,31 +999,24 @@ func (k *kataAgent) removeIgnoredOCIMount(spec *specs.Spec, ignoredMounts []Moun
 	return nil
 }
 
-func (k *kataAgent) replaceOCIMountsForStorages(spec *specs.Spec, volumeStorages []*grpc.Storage) error {
+func (k *kataAgent) replaceOCIMountsForStorages(spec *specs.Spec, volumeStorages map[string]*grpc.Storage) error {
 	ociMounts := spec.Mounts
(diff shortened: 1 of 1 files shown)


virtcontainers/kata_agent.go
```diff
@@ -999,31 +999,24 @@ func (k *kataAgent) removeIgnoredOCIMount(spec *specs.Spec, ignoredMounts []Moun
...
-			if m.Destination != v.MountPoint {
-				continue
-			}
-
-			// Create a temporary location to mount the Storage. Mounting to the correct location
-			// will be handled by the OCI mount structure.
-			filename := fmt.Sprintf("%s-%s", uuid.Generate().String(), filepath.Base(m.Destination))
-			path := filepath.Join(kataGuestSharedDir(), filename)
+	for index, m = range ociMounts {
+		if _, ok := volumeStorages[m.Destination]; !ok {
```
bergwolf: The side effect is that you can only have one mount for each destination. While it is not a common case, docker does allow users to define multiple mounts for the same destination. So I'd prefer to keep the semantics.

PR #2308, opened 2019-12-02
vc: Detach device when unable to store sandbox device

### LINKED ISSUE #2301: Detach device when unable to store sandbox device
# Description of problem

In Container#mountSharedDirMounts, if sandbox.storeSandboxDevices() returns error, we should detach the device.

# Expected result

This addresses existing comment:
```
                               if err := c.sandbox.storeSandboxDevices(); err != nil {
                                       //TODO: roll back?
```
# Actual result

tedyu:
In Container#mountSharedDirMounts, if sandbox.storeSandboxDevices() returns error, we should detach the device.

Fixes #2301

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -495,9 +495,17 @@ func (c *Container) shareFiles(m Mount, idx int, hostSharedDir, guestSharedDir s
 // It also updates the container mount list with the HostPath info, and store
 // container mounts to the storage. This way, we will have the HostPath info
 // available when we will need to unmount those mounts.
-func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (map[string]Mount, map[string]Mount, error) {
-	sharedDirMounts := make(map[string]Mount)
-	ignoredMounts := make(map[string]Mount)
+func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (sharedDirMounts map[string]Mount, ignoredMounts map[string]Mount, err error) {
+	sharedDirMounts = make(map[string]Mount)
+	ignoredMounts = make(map[string]Mount)
+	var devicesToDetach []string
+	defer func() {
+		if err != nil {
+			for _, id := range devicesToDetach {
+				c.sandbox.devManager.DetachDevice(id, c.sandbox)
+			}
+		}
+	}()
 	for idx, m := range c.mounts {
 		// Skip mounting certain system paths from the source on the host side
 		// into the container as it does not make sense to do so.
@@ -521,9 +529,10 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
 		// instead of passing this as a shared mount.
 		if len(m.BlockDeviceID) > 0 {
 			// Attach this block device, all other devices passed in the config have been attached at this point
-			if err := c.sandbox.devManager.AttachDevice(m.BlockDeviceID, c.sandbox); err != nil {
+			if err = c.sandbox.devManager.AttachDevice(m.BlockDeviceID, c.sandbox); err != nil {
 				return nil, nil, err
 			}
+			devicesToDetach = append(devicesToDetach, m.BlockDeviceID)
 			continue
 		}
 
@@ -534,7 +543,9 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
 			continue
 		}
 
-		guestDest, ignore, err := c.shareFiles(m, idx, hostSharedDir, guestSharedDir)
+		var ignore bool
+		var guestDest string
+		guestDest, ignore, err = c.shareFiles(m, idx, hostSharedDir, guestSharedDir)
 		if err != nil {
 			return nil, nil, err
 		}


tedyu:
@jodh-intel @grahamwhaley
Mind triggering ubuntu test run ?

thanks

tedyu:
@amshinde 
Mind triggering ubuntu test run ?

Thanks

tedyu:
@jodh-intel @grahamwhaley
Mind triggering ubuntu test run ?

thanks

tedyu:
@jodh-intel @grahamwhaley @amshinde 
All tests passed

tedyu:
@grahamwhaley @amshinde
Please take a look.

virtcontainers/container.go
```diff
@@ -545,6 +545,15 @@ func (c *Container) shareFiles(m Mount, idx int, hostSharedDir, guestSharedDir s
 func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (map[string]Mount, map[string]Mount, error) {
 	sharedDirMounts := make(map[string]Mount)
 	ignoredMounts := make(map[string]Mount)
+	var err error
+	var devicesToDetach []string
+	defer func() {
+		if err != nil {
+			for _, id := range devicesToDetach {
+				c.sandbox.devManager.DetachDevice(id, c.sandbox)
```
egernst: If I understand correctly, if the mounting fails, today we return an error, and the container creation flow fails.  Shouldn't the VM be getting torn down already?

In that scenario, are we failing to undo the bind mount on the host?

@amshinde can you help here?

jodh-intel: Ping @amshinde.

amshinde: We are hotplugging block devices to the VM here. If the `storeSandboxDevices` function fails, then the container will fail, but the VM may not be brought down with other containers running. So, this seems like valid use case, although if we fail to write to file system, we may be running into much bigger issues, eventually causing the pod/VM to fail.

tedyu: @egernst 
Do you have any more comment ?

tedyu:
@egernst @amshinde 
Can this be merged ?

jodh-intel:
@tedyu - please can you update this branch and `push -f` as it is conflicted.

tedyu:
@jodh-intel 
I have updated the PR.

jodh-intel:
Thanks @tedyu.

/test

tedyu:
For the ARM-ubuntu build:
```
05:58:20 rm: cannot remove '/run/vc/vm/template': Device or resource busy
05:58:20 INFO: Remove installed kata packages
05:58:20 ERROR: packages not installed
```

virtcontainers/container.go
```diff
@@ -498,6 +498,15 @@ func (c *Container) shareFiles(m Mount, idx int, hostSharedDir, guestSharedDir s
 func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (map[string]Mount, map[string]Mount, error) {
 	sharedDirMounts := make(map[string]Mount)
 	ignoredMounts := make(map[string]Mount)
+	var err error
```
devimc: this won't work, because of `:=` creates a new variable so this variable will always be `nil`

virtcontainers/container.go
```diff
@@ -524,6 +533,7 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
 			if err := c.sandbox.devManager.AttachDevice(m.BlockDeviceID, c.sandbox); err != nil {
```
devimc: `err` is a new variable and the previous `err` that you declared is not considered.  For example: 

```go
package main

import "fmt"

func f() error {
	return fmt.Errorf("foo")
}

func main() {
	var err error
	if err := f(); err != nil {
		fmt.Printf("#1: %v\n", err)
	}
	fmt.Printf("#2 error: %v\n", err)
}
```

**output**
```
#1 foo
#2 <nil>
```

tedyu:
@devimc 
Your comment has been addressed.

devimc:
Thanks @tedyu but you should also change this line:

https://github.com/kata-containers/runtime/blob/df802cc359fce03b79a8bfe602d4a1d91034cfd4/virtcontainers/container.go#L537

and `err` variable should be declared in function declaration, for example:

```go
func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (m map[string]Mount, m2 map[string]Mount, err error) {
```

tedyu:
@devimc 
Please take another look.

virtcontainers/container.go
```diff
@@ -495,9 +495,17 @@ func (c *Container) shareFiles(m Mount, idx int, hostSharedDir, guestSharedDir s
 // It also updates the container mount list with the HostPath info, and store
 // container mounts to the storage. This way, we will have the HostPath info
 // available when we will need to unmount those mounts.
-func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (map[string]Mount, map[string]Mount, error) {
+func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (map[string]Mount, map[string]Mount, err error) {
```
devimc: this won't compile

```
./container.go:498:80: syntax error: mixed named and unnamed function parameters
```

tedyu:
@devimc 
I have updated the PR

virtcontainers/container.go
```diff
@@ -495,9 +495,17 @@ func (c *Container) shareFiles(m Mount, idx int, hostSharedDir, guestSharedDir s
 // It also updates the container mount list with the HostPath info, and store
 // container mounts to the storage. This way, we will have the HostPath info
 // available when we will need to unmount those mounts.
-func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (map[string]Mount, map[string]Mount, error) {
+func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (sharedDirMounts map[string]Mount, ignoredMounts map[string]Mount, err error) {
 	sharedDirMounts := make(map[string]Mount)
 	ignoredMounts := make(map[string]Mount)
```
devimc: ehmm , `:=` ?

tedyu:
@devimc 
All tests passed.

PR #2309, opened 2019-12-02
virtcontainers/store: make VCStoreUUIDPath rootless

### LINKED ISSUE #2133: rootless: make VCStoreUUIDPath rootless
see https://github.com/kata-containers/runtime/issues/2132

amshinde:
The uuid file shouldn't be created at `/var` if running rootless.
Modify `VMUUIDStoragePath` to get a path accessible for non-root users
if running rootless.

fixes #2133


jodh-intel:
Oh - strike that - hadn't noticed the branch. However, it would be good to add such a unit test in master (first :)

amshinde:
Yes, a test should go in master. @devimc Can you add that?
Meanwhile, since we have a release this week, would be good if this fix went it.

PR #2323, opened 2019-12-06
rootless: Disable vhost-net for rootless(backport)

### LINKED ISSUE #2321: Disable vhost-net when running as rootless

amshinde:
Since the /dev/vhost-net device is owned by root, we cannot used
vhost-net networking in rootless mode. Instead of having to
do this manually in the toml, disable vhost-net when the runtime
detects it is running rootless.

Fixes #2321

### DIFF
--- a/virtcontainers/network.go
+++ b/virtcontainers/network.go
@@ -24,6 +24,7 @@ import (
 	"github.com/vishvananda/netns"
 	"golang.org/x/sys/unix"
 
+	"github.com/kata-containers/runtime/pkg/rootless"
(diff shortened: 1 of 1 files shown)


bergwolf:
@amshinde Can we access `/dev/vhost-net` with proper user/group setup for rootless, like we did for `/dev/kvm`?

amshinde:
I am little skeptical about alllowing vhost based networking to be done by normal user due to its security implications. In any case I would like to have a discussion on this, before allowing this. I'll open a separate issue for this to have some discussion on it.

Issue #2327, opened 2019-12-06
makefile: qemu-virtiofs config is not honor if set as default


jcvenegas:
# Description of problem

Change default hypervisor to qemu-virtiofs not work if changed
in makefile

```diff
+DEFAULT_HYPERVISOR = $(HYPERVISOR_QEMU_VIRTIOFS)
```

# Expected result

It honor the configj

# Actual result

it does not work

---

(replace this text with the output of the `kata-collect-data.sh` script, after
you have reviewed its content to ensure it does not contain any private
information).

PR #2328, opened 2019-12-06
makefile: honor virtiofs config for default config

jcvenegas:
Fix config used when qemu-virtiofs is set as default.

Fixes: #2327


bergwolf:
Does it mean we make virtio-fs the default share fs implementation for kata now?

jcvenegas:
@bergwolf hi, no it is just to allow to set
`DEFAULT_HYPERVISOR=$HYPERVISOR_QEMU_VIRTIOFS`

This is just to use the correct config when the HYPERVISOR is qemu-virtiofs,  I was testing locally  and when I changed the default was not using the correct config, qemu-vanilla is is still the default  implementation.

Issue #2332, opened 2019-12-09
support config path via annotations


bergwolf:
**Which feature do you think can be improved?**

Right now we support tweaking kata specific config options via pod annotations. While it is working, it seems less convenient. We can support kata config file path via pod annotation as well. While containerd has an runtime option to specify different config file path, it is bound to runtime handler and other CRI implementations are still missing it.

PR #2333, opened 2019-12-09
shimv2: support runtime config path via annotation

bergwolf:
So that user can have per pod runtime config which is easier
to manage in some cases.

Fixes: #2332


bergwolf:
@grahamwhaley yes, let me try to fix the doc as well.

bergwolf:
@grahamwhaley fyi, https://github.com/kata-containers/documentation/pull/576

bergwolf:
@jodh-intel CI added.

Issue #2343, opened 2019-12-10
clh: update to latest master


jcvenegas:
**Which feature do you think can be improved?**

Update to latest master version

**How can it be improved?**

Update  versions.yaml and fix new API changes.

**Additional Information**

Anything else to add?

**Before raising this feature request**

Have you looked at the [limitations document](https://github.com/kata-containers/documentation/blob/master/Limitations.md)?

PR #2344, opened 2019-12-10
clh: update to latest master

jcvenegas:
Update to latest master code.

Fixes: #2343


sboeuf:
@jcvenegas please rebase on latest (latest) master from CH :)
We merged some new code today, and we added the OpenAPITools generation validation in our CI.
Two main things changed which will affect this PR:
- `CpuConfig` has been changed into `CpusConfig`
- `FsConfig` has a boolean field `dax`, which will allow you to explicitly say if you want DAX to be enabled or not.

versions.yaml
```diff
@@ -75,7 +75,7 @@ assets:
       url: "https://github.com/intel/cloud-hypervisor"
       uscan-url: >-
         https://github.com/intel/cloud-hypervisor/tags.*/v?(\d\S+)\.tar\.gz
-      version: "ca97385da5ebb798"
+      version: "ee528ae80836e9d59ff50d124ae356ed1b07002a"
```
jodh-intel: I wonder if we should treat this as a re-vendor? If so, would be good to include the changes between these commits in one of the commits in this PR.

jcvenegas: good idea, added the changes in the commit

Issue #2347, opened 2019-12-11
no timestamp offered by demsg in kata-container on arm64


jongwu:
# Description of problem

when running dmesg in kata-container on arm64, timestamp is always 0

# Expected result

correct timestamp should be offered by kata-container on arm64

# Actual result

timestamp is full of 0.

PR #2348, opened 2019-12-11
enable timestamp for dmesg log on arm64.

jongwu:
Currently, no timestamp offered by dmesg in container for arm64.
This patch set "printk.time" to 1 in kernel parameters to enable it.

Fixes: #2347

@jodh-intel @grahamwhaley  @Pennyzct  @devimc

### DIFF
--- a/virtcontainers/qemu_arm64.go
+++ b/virtcontainers/qemu_arm64.go
@@ -40,6 +40,7 @@ var kernelParams = []Param{
 	{"console", "hvc0"},
 	{"console", "hvc1"},
 	{"iommu.passthrough", "0"},
+	{"printk.time", "1"},
 }
 
 // For now, AArch64 doesn't support DAX, so we couldn't use


virtcontainers/qemu_arm64.go
```diff
@@ -40,6 +40,7 @@ var kernelParams = []Param{
 	{"console", "hvc0"},
 	{"console", "hvc1"},
 	{"iommu.passthrough", "0"},
+	{"printk.time", "1"},
```
grahamwhaley: ack. Alternatively, you could set `PRINTK_CONFIG_TIME` in the kernel config file at https://github.com/kata-containers/packaging/blob/master/kernel/configs/arm64_kata_kvm_4.19.x#L2277 ? That way we shift the decision to print time or not to the kernel builder, and not permanently enabled on the qemu command line.

/me notes, might be nice to consider moving to the [kernel config fragments setup](https://github.com/kata-containers/packaging/tree/master/kernel/configs/fragments) one day?

jongwu: OK, I will enable printk_time in kernel/config and this pr can be closed. thanks @grahamwhaley

jcvenegas:
@jongwu this is going to be closed? feel free to closet it please

virtcontainers/qemu_arm64.go
jodh-intel: Ping @jongwu

PR #2355, opened 2019-12-13
versions: update qemu to 4.2.0

### LINKED ISSUE #2354: update qemu to 5.0.0

devimc:
New features that can improve/impact in kata containers:

x86
* VMX features can be enabled/disabled via the "-cpu" flag.
* When nested virtualization is enabled with an option like
  "-cpu Haswell,+vmx", the set of VMX features will also be constrained to
  what was available on the corresponding CPU model.
* New "microvm" machine type that has virtio-mmio instead of PCI, and no ACPI
  support (so no hotplug too). The new machine type is meant as a baseline
  for performance optimizations of QEMU, firmware and guests. While inspired
  by Firecracker it is not entirely compatible with it (for example it does
  not have Firecracker's userspace IP stack and MicroVM Metadata Service).

ARM:
* We now correctly support more than 256 CPUs when using KVM
* The virt board now supports memory hotplugging, when used with a UEFI
  guest BIOS and ACPI.

s390:
* Using KVM now explicitly requires a host kernel version of at least 3.15
  (which includes the 'flic' KVM device). This had been broken since QEMU
  2.10 already.

# Metrics

s1

s2

fixes #2354

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -88,8 +88,8 @@ assets:
     qemu:
       description: "VMM that uses KVM"
       url: "https://github.com/qemu/qemu"
-      version: "4.1.1"
-      tag: "v4.1.1"
+      version: "4.2.0"
+      tag: "v4.2.0"
       # Do not include any non-full release versions
       # Break the line *without CR or space being appended*, to appease
       # yamllint, and note the deliberate ' ' at the end of the expression.
--- a/virtcontainers/qemu_amd64.go
+++ b/virtcontainers/qemu_amd64.go
@@ -131,9 +131,9 @@ func (q *qemuAmd64) cpuModel() string {
 		cpuModel += ",pmu=off"
 	}
 
-	// VMX is not migratable yet.
-	// issue: https://github.com/kata-containers/runtime/issues/1750
-	if q.vmFactory {
+	if q.nestedRun {
+		// VMX is not migratable yet.
+		// issue: https://github.com/kata-containers/runtime/issues/1750
 		virtLog.WithField("subsystem", "qemuAmd64").Warn("VMX is not migratable yet: turning it off")
 		cpuModel += ",vmx=off"
 	}


jodh-intel:
I just saw a :unicorn:!

jodh-intel:
... which probably explains the double approval ;)

egernst:
/test 

Hoping CI likes this PR more, a month later.

devimc:
@egernst seems like there is a kvm bug in some azure VMs

```
failed to set MSR 0x48b to 0x1582e00000000
```

we have two options:
* upgrade all azure VMs to D4s v3
* disable vmx if nested (a patch in kata-runtime)

cc @chavafg

chavafg:
I have verified and all VMs launched by the CI are configured to be `D4s_v3`.

devimc:
@chavafg thanks, then let me try disabling vmx

egernst:
@devimc let’s get a root cause before disabling nested.

raravena80:
@devimc nudge.

Pennyzct:
Hi @devimc 
just wondering, What's the progress here~~ 
We also want to move qemu into the new `4.2.x` on aarch64 to be in consistency with all other archs, since in case you did not notice, we've stuck into one very old version `v3.1.0-rc2` for some reasons, see https://github.com/kata-containers/runtime/blob/master/versions.yaml#L105. 😫

fidencio:
Isn't this issue related to https://github.com/qemu/qemu/commit/4a910e1f6ab4155ec8b24c49b2585cc486916985 ?

Pennyzct:
Hi @fidencio 
thanks for the explanation, so we don't move into `v4.2.x`, and just waiting for the latest stable `v5.0.x` release, right?

fidencio:
> Hi @fidencio
> thanks for the explanation, so we don't move into `v4.2.x`, and just waiting for the latest stable `v5.0.x` release, right?

If the issue is related to that, we can just wait for a stable v5.0 release.

devimc:
@fidencio  @Pennyzct fox x86 a kernel >= 5.3 is required in the azure VMs to have QEMU >= 4.2 [1]. unfortunately not all distros have that kernel version :worried:  

[1] - https://lists.nongnu.org/archive/html/qemu-devel/2020-04/msg00099.html

amshinde:
@fidencio @Pennyzct Trying to see how we can resolve this. The mentioned patch (https://github.com/qemu/qemu/commit/4a910e1f6ab4155ec8b24c49b2585cc486916985) requires 5.3 kernel. However not all distros in Azure carry the 5.3 kernel.
Again, to reiterate, this is an issue seen just on Azure and not on bare-metal.
(@devimc @chavafg Might be good to confirm this on vexxhost machines)

If we decide to go with 5.0 at this point, we may have to go with a solution such as which is not ideal:
https://github.com/kata-containers/runtime/pull/2657#discussion_r419528667

I am trying to understand the root-cause of this better:
```
In particular, it was reported that Azure VMs have RDSEED but lack
VMX_SECONDARY_EXEC_RDSEED_EXITING; attempts to enable this feature
bit result in QEMU abort.
```
So this looks like a bug in HyperV ?
I want to see if this is something that the Azure folks can fix on their side without having a kernel requirement of 5.3 version for the Virtual Machines.
@dagrh @bonzini Can you provide some input on this?

fidencio:
@amshinde, which distros are the problematic ones on Azure?

I wonder whether we could push the distros hard enough so they'd have such patch backported, makings us free to not depend on 5.3 kernel.

amshinde:
@fidencio  None of the distros are using the 5.3 version. @chavafg Can you provide info on what kernel versions are being used in the Azure distros.

> I wonder whether we could push the distros hard enough so they'd have such patch backported, makings us free to not depend on 5.3 kernel.
This kernel patch https://github.com/torvalds/linux/commit/95c5c7c77c has landed in 5.3, but I dont think this has been backported to stable kernel versions. 

@devimc Can you update this PR to use qemu 5.0 instead and run the CI again?

chavafg:
@amshinde @devimc @fidencio 
I just saw that Azure ubuntu 18.04 has: Kernel = "5.3.0-1020-azure". 
Last week it was still using 5.0 iirc. 

Fedora31 also uses 5.3.7, image that we uploaded to azure.  
@devimc you should be able to test qemu 5.0 there, right?

The other distros still have pretty old kernels, ranging from 3.10 on centos to 4.19 on debian

devimc:
@chavafg 

> @devimc you should be able to test qemu 5.0 there, right?

yes, we need a kernel >= 5.3.0 

> The other distros still have pretty old kernels, ranging from 3.10 on centos to 4.19 on debian

we have two options here:
1. build and install qemu 4.1.1 on these distros
2. azure has to fix hyperv

bonzini:
It's not a Hyper-V bug, only a slightly weird configuration; QEMU wants to work around a KVM bug that assumes that the configuration is not the weird one that Hyper-V provides.

Note that kernels that do not support customizing VMX features (4.16 or earlier) should not be affected. The bug thus only.remains from 4.17 to 5.2 inclusive.

You could also try a command line option like `-cpu  host,-vmx-rdseed-exit`.

amshinde:
Thanks @bonzini.  Good to know the range of affected kernels on Hyper-V. 
I am not clear about the last line in your comment though. What does `-vmx-rdseed-exit` do and how will that help us to overcome this issue on affected kernels?

bonzini:
> What does -vmx-rdseed-exit do and how will that help us to overcome this issue on affected kernels?

It forces QEMU not to include the flag that it thinks should be added (but Hyper-V doesn't support), thus undoing the KVM workaround and fixing Hyper-V. I haven't tested it, but it should work.

amshinde:
Thanks @bonzini for the clarification.
@devimc Can we try this out on your upgrade PR for qemu. If it works out, we can add this flag while running tests in our CI.

devimc:
@amshinde sounds good, but firstly we need this pr https://github.com/kata-containers/runtime/pull/2577

PR #2358, opened 2019-12-15
version.yaml: Update golang.newest-version to 1.13.5

teawater:
Update golang.newest-version to 1.13.5.

Fixes: #2357

Signed-off-by: Hui Zhu <teawater@antfin.com>

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -300,7 +300,7 @@ languages:
       description: |
         'newest-version' is the latest version known to work when
         building Kata
-      newest-version: "1.12.3"
+      newest-version: "1.13.5"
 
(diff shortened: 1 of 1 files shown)


teawater:
https://github.com/kata-containers/tests/pull/2175

teawater:
The test still got the test cmd argument error even if I updated the version of ginkgo.

teawater:
I reproduced the fail with ginkgo and go 1.3.5 on my part.
Reported to https://github.com/onsi/ginkgo/issues/634

And I think the update to 1.3.5 is too early for kata.  Maybe we can do it later.

PR #2363, opened 2019-12-18
FC: jailer failed when importing new flag "--config-file"

Pennyzct:
# Description of problem
When we used jailer to launch firecracker, kata container failed.
```
docker: Error response from daemon: OCI runtime create failed: Failed to 
connect to firecrackerinstance (timeout 10s): unknown.
```
Causes and Fixes are as follows:
1. new flag `-config-file` belongs to the jailed firecracker.
    so, adhering to the `end of command options` convention, we need to give `config-file` a 
prefix `--`.
2. The path of the config file(`fcConfig.json`) should be also relative to the jailed firecracker.
3. Since we do the configuration before func `fcInit` now, we also need to bring `jailer check` ahead.

### DIFF
--- a/virtcontainers/fc.go
+++ b/virtcontainers/fc.go
@@ -303,6 +303,7 @@ func (fc *firecracker) newFireClient() *client.Firecracker {
 func (fc *firecracker) vmRunning() bool {
 	resp, err := fc.client().Operations.DescribeInstance(nil)
 	if err != nil {
+		fc.Logger().WithError(err).Error("getting vm status failed")
 		return false
 	}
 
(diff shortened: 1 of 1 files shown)


virtcontainers/fc.go
```diff
@@ -303,6 +303,7 @@ func (fc *firecracker) newFireClient() *client.Firecracker {
 func (fc *firecracker) vmRunning() bool {
 	resp, err := fc.client().Operations.DescribeInstance(nil)
 	if err != nil {
+		fc.Logger().WithError(err).Error("getting vm status failed")
```
jodh-intel: Nice!

devimc:
thanks @Pennyzct , seems like CI is not happy

```
virtcontainers/fc.go:713: File is not `gofmt`-ed with `-s` (gofmt)

        if fc.config.JailerPath != "" {

                fc.jailed = true

        }
```

Pennyzct:
Hi @devimc 
so sorry... my neglect. updated. ;).

Pennyzct:
Hi guys
Maybe we should integration test for running firecracker with jailer, then the bug like this will be detected. ;).

Pennyzct:
### Update:
The config file `fcConfig.json` should be unmounted and cleaned up. ;)
/test-fc

Pennyzct:
Hi~ guys
All green. ;).
But need another reviewer, @kata-containers/runtime. ;)

PR #2368, opened 2019-12-19
clh: Increase unit test using mock testing

jcvenegas:
This add unit test in clh.go driver to achieve that two new interfaces are added to allow mock testing.

clhClient: An interface that define the methods described by the HTTP Cloud hypervisor client.
virtiofsd interface: An interface around virtiofsd.

Additionally `utils.StartCmd(cmd)` was added this is aim to use instead of `cmd.Start()`  this allow to modify it at unit testing and not run any command but just simulate is started.

Both virtiofsd and HTTP client are dependencies of cloud hypervisor that now are implemented as interfaces to have a loose depedency in them.

As a result this allow to write more extensive unit test in `clh.go`

### DIFF
--- a/virtcontainers/virtiofsd.go
+++ b/virtcontainers/virtiofsd.go
@@ -1,0 +1,239 @@
+// Copyright (c) 2019 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package virtcontainers
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"strings"
+	"syscall"
+	"time"
+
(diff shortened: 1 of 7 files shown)


virtcontainers/virtiofsd.go
```diff
@@ -0,0 +1,234 @@
+package virtcontainers
```
devimc: and the license ?

virtcontainers/virtiofsd_test.go
```diff
@@ -0,0 +1,69 @@
+package virtcontainers
```
devimc: same here

jcvenegas:
@devimc thanks headers added.
@jodh-intel squashed some commits.

PR #2376, opened 2019-12-25
kata-check: Add version consistency check

Pennyzct:
We may need multiple kata components to launch one kata container. The longest legacy one could have `kata-shim`, `kata-proxy`, `kata-runtime`, and `kata-agent.`
For users, especially for developers, we often encounter confused runtime error due to each kata component at different version.
E.g. Laterly, I've encountered following error in launching kata container with firecracker, and it's because of stale version of `kata-shim`, which is lower than `v1.9.0`.
```
docker: Error response from daemon: cannot start a stopped process: unknown.
ERRO[0002] error waiting for container: context canceled
```
**How can it be improved?**
So, here, I want to introduce version consistency check in `kata-runtime kata-check`.
If we have kata components at different version, it should issue a warning, like this:
```
$ kata-runtime kata-check
System is capable of running Kata Containers
System can currently create Kata Containers
Warning: there exists version inconsistency in kata components. kata-proxy: v1.10, kata-shim: v1.7, kata-runtime: v1.10
```
Otherwise, it should output like this:
```
$ kata-runtime kata-check
INFO[0000] VSOCK supported, configure to not use proxy
System is capable of running Kata Containers
System can currently create Kata Containers
Version consistency of Kata Containers is verified
```

**Updates:**
Following upstream comments to use `--strict/-s` option to perform version consistency check:
```
$ kata-runtime kata-check --strict
INFO[0000] VSOCK supported, configure to not use proxy
System is capable of running Kata Containers
System can currently create Kata Containers
Version consistency of Kata Containers is verified
```
```
$ kata-runtime kata-check --strict         
INFO[0000] VSOCK supported, configure to not use proxy
System is capable of running Kata Containers
System can currently create Kata Containers
ERRO[0000] there exists version inconsistency in Kata Components. kata-shim: v1.8.1, kata-runtime: v1.10.0
  arch=arm64 name=kata-runtime pid=35284 source=runtime
there exists version inconsistency in Kata Components. kata-shim: v1.8.1, kata-runtime: v1.10.0
```

### DIFF
--- a/cli/kata-check.go
+++ b/cli/kata-check.go
@@ -63,6 +63,7 @@ const (
 	moduleParamDir        = "parameters"
 	successMessageCapable = "System is capable of running " + project
 	successMessageCreate  = "System can currently create " + project
+	successMessageVersion = "Version consistency of " + project + " is verified"
 	failMessage           = "System is not capable of running " + project
 	kernelPropertyCorrect = "Kernel property value correct"
 
@@ -309,6 +310,10 @@ var kataCheckCLICommand = cli.Command{
 			Name:  "verbose, v",
 			Usage: "display the list of checks performed",
 		},
+		cli.BoolFlag{
+			Name:  "strict, s",
+			Usage: "perform strict checking",
+		},
 	},
 
 	Action: func(context *cli.Context) error {
@@ -355,6 +360,16 @@ var kataCheckCLICommand = cli.Command{
 			}
 
 			fmt.Println(successMessageCreate)
+		}
+
+		strict := context.Bool("strict")
+		if strict {
+			err = checkVersionConsistencyInComponents(runtimeConfig)
+			if err != nil {
+				return err
+			}
+
+			fmt.Println(successMessageVersion)
 		}
 
 		return nil
@@ -453,4 +468,36 @@ func genericCheckKVMExtensions(extensions map[string]kvmExtension) (map[string]i
 	}
 
 	return results, nil
+}
+
+// checkVersionConsistencyInComponents checks version consistency in Kata Components.
+func checkVersionConsistencyInComponents(config oci.RuntimeConfig) error {
+	proxyInfo := getProxyInfo(config)
+
+	shimInfo, err := getShimInfo(config)
+	if err != nil {
+		return err
+	}
+	shimVersionInfo := shimInfo.Version
+
+	runtimeVersionInfo := constructVersionInfo(version)
+
+	// kata-proxy exists
+	if proxyInfo.Type != string(vc.NoProxyType) {
+		proxyVersionInfo := proxyInfo.Version
+		if !versionEqual(proxyVersionInfo, runtimeVersionInfo) || !versionEqual(shimVersionInfo, runtimeVersionInfo) {
+			return fmt.Errorf("there exists version inconsistency in kata components. kata-proxy: v%d.%d.%d, kata-shim: v%d.%d.%d, kata-runtime: v%d.%d.%d",
+				proxyVersionInfo.Major, proxyVersionInfo.Minor, proxyVersionInfo.Patch,
+				shimVersionInfo.Major, shimVersionInfo.Minor, shimVersionInfo.Patch,
+				runtimeVersionInfo.Major, runtimeVersionInfo.Minor, runtimeVersionInfo.Patch)
+		}
+	} else {
+		if !versionEqual(shimVersionInfo, runtimeVersionInfo) {
+			return fmt.Errorf("there exists version inconsistency in kata components. kata-shim: v%d.%d.%d, kata-runtime: v%d.%d.%d",
+				shimVersionInfo.Major, shimVersionInfo.Minor, shimVersionInfo.Patch,
+				runtimeVersionInfo.Major, runtimeVersionInfo.Minor, runtimeVersionInfo.Patch)
+		}
+	}
+
+	return nil
 }
--- a/cli/kata-env.go
+++ b/cli/kata-env.go
@@ -74,11 +74,18 @@ type RuntimeInfo struct {
 	Path                string
 }
 
+type VersionInfo struct {
+	Semver string
+	Major  uint64
+	Minor  uint64
+	Patch  uint64
+	Commit string
+}
+
 // RuntimeVersionInfo stores details of the runtime version
 type RuntimeVersionInfo struct {
-	Semver string
-	Commit string
-	OCI    string
+	Version VersionInfo
+	OCI     string
 }
 
 // HypervisorInfo stores hypervisor details
@@ -100,7 +107,7 @@ type HypervisorInfo struct {
 // ProxyInfo stores proxy details
 type ProxyInfo struct {
 	Type    string
-	Version string
+	Version VersionInfo
 	Path    string
 	Debug   bool
 }
@@ -108,7 +115,7 @@ type ProxyInfo struct {
 // ShimInfo stores shim details
 type ShimInfo struct {
 	Type    string
-	Version string
+	Version VersionInfo
 	Path    string
 	Debug   bool
 }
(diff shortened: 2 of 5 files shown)


Pennyzct:
Hi @lifupan 
I finally detected the root cause of that error....
stale kata-shim...🤬🤮

cli/kata-check.go
```diff
@@ -454,3 +459,45 @@ func genericCheckKVMExtensions(extensions map[string]kvmExtension) (map[string]i
 
 	return results, nil
 }
+
+// func checkVersionConsistencyInComponents checks version consistency in kata components.
```
jodh-intel: Nit: The comment should start with the name of the function (without `func`).

cli/kata-check.go
```diff
@@ -454,3 +459,45 @@ func genericCheckKVMExtensions(extensions map[string]kvmExtension) (map[string]i
...
+func checkVersionConsistencyInComponents(c *cli.Context) error {
+	config, ok := c.App.Metadata["runtimeConfig"].(oci.RuntimeConfig)
+	if !ok {
+		return errors.New("cannot determine runtime config")
+	}
+
+	proxyInfo, _ := getProxyInfo(config)
```
jodh-intel: It might be worth removing the `error` return value for `getProxyInfo()` as it's always `nil` (same for `getNetmonInfo()`).

cli/kata-check.go
```diff
@@ -357,6 +358,10 @@ var kataCheckCLICommand = cli.Command{
 			fmt.Println(successMessageCreate)
 		}
 
+		if err := checkVersionConsistencyInComponents(context); err != nil {
```
jodh-intel: You could simplify this block to just:

```go
return checkVersionConsistencyInComponents(context)
```

cli/kata-check.go
```diff
@@ -454,3 +459,45 @@ func genericCheckKVMExtensions(extensions map[string]kvmExtension) (map[string]i
...
+
+	proxyInfo, _ := getProxyInfo(config)
+
+	shimInfo, err := getShimInfo(config)
+	if err != nil {
+		return err
+	}
+
+	var runtimeVersion VersionInfo
+	versions := strings.Split(version, ".")
```
jodh-intel: This works but it may be better to  use the already-vendored semver package (`github.com/blang/semver`).

cli/kata-check.go
```diff
@@ -454,3 +459,45 @@ func genericCheckKVMExtensions(extensions map[string]kvmExtension) (map[string]i
...
+	if len(versions) > 1 {
+		runtimeVersion.Major = versions[0]
+		runtimeVersion.Minor = versions[1]
+	} else {
+		return errors.New("getting kata-runtime version failed")
+	}
+
+	if proxyInfo.Version.Version != "" {
+		if !versionCompare(proxyInfo.Version, runtimeVersion) || !versionCompare(shimInfo.Version, runtimeVersion) {
+			fmt.Printf("Warning: there exists version inconsistency in kata components. kata-proxy: v%s.%s, kata-shim: v%s.%s, kata-runtime: v%s.%s\n", proxyInfo.Version.Major, proxyInfo.Version.Minor,
```
jodh-intel: - s/kata components/Kata components/
- We use semver consistently so I think we should display all three parts of the version here (major, minor + patch) , or just the entire version string with a single `%s`.
- Random thought: this would expand the scope of this PR a little, but I wonder if we want to add a `--strict` option so that if you run `kata-runtime kata-check --strict` and one of these checks fails, it returns `1` rather than `0`. That could be more useful for scripts, *etc* which shouldn't have to parse `stderr`. That would imply this function would then return an actual error which itself would make it easier to unit test.

cli/kata-env.go
```diff
@@ -34,6 +34,12 @@ type MetaInfo struct {
 	Version string
 }
 
+type VersionInfo struct {
+	Version string
```
jodh-intel: This is a little confusing - what *is* `Version`? I think it's the raw output of `$cmd --version` so I'd make that clearer by adding a comment or renaming the field. But again, I'd look at the `semver` package to avoid "re-inventing the wheel" - you could just store a `semver.Version` as that contains everything we'd need.

cli/kata-env_test.go
```diff
@@ -30,8 +30,9 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-const testProxyVersion = "proxy version 0.1"
-const testShimVersion = "shim version 0.1"
+var testProxyVersion = VersionInfo{Version: "proxy version 0.1", Major: "0", Minor: "1"}
```
jodh-intel: Great to see some new tests. It would be great though if you could test a larger number of scenarios:

- runtime older than proxy or shim
- runtime older than proxy and shim
- runtime newer than proxy or shim
- runtime newer than proxy and shim
- *etc*

I think the easiest way to add in a table-based set of semi-exhaustive tests would be to add in that strict option to allow you to test for `error` in `checkVersionConsistencyInComponents()`.

cli/kata-check.go
```diff
@@ -454,3 +459,45 @@ func genericCheckKVMExtensions(extensions map[string]kvmExtension) (map[string]i
...
+	versions := strings.Split(version, ".")
+	if len(versions) > 1 {
+		runtimeVersion.Major = versions[0]
+		runtimeVersion.Minor = versions[1]
+	} else {
+		return errors.New("getting kata-runtime version failed")
+	}
+
+	if proxyInfo.Version.Version != "" {
+		if !versionCompare(proxyInfo.Version, runtimeVersion) || !versionCompare(shimInfo.Version, runtimeVersion) {
```
amshinde: Add a condition comparing the shim and proxy versions here?

cli/kata-env.go
amshinde: Ya, same question, what is the `Version` here?  Shouldnt this be `Major`, `Minor` and `Patch` instead? 
Are we not comparing the "Patch" version at all?

Pennyzct:
Hi guys @jodh-intel @amshinde 
Thanks for all your detailed comments, really appreciated. ;) 😚
I'm with the @jodh-intel's suggestion to add a `--strict` option to do the binary-check, instead of just output the warning. 
And I'll use `semver` package to replace the original design. ;). 
Update ASAP...

jodh-intel:
Thanks @Pennyzct !

Pennyzct:
Hi guys 
@jodh-intel @amshinde 
I've updated my code, Plz take a review~~~~;)

Pennyzct:
Hi~ guys
FC failed as usual, and should be irrelevant with this patch.
ARM CI failed because of PR [kata-containers/tests/#2236](https://github.com/kata-containers/tests/pull/2236). 
And I need one reviewer here~~~ any one cares to take a look?? @kata-containers/runtime

devimc:
restarting ubuntu-ci

Pennyzct:
ARM CI failed on:
```
10:42:56 pkg/rootless/rootless.go:57:2: comment on exported var `IsRootless` should be of the form `IsRootless ...` (golint)
10:42:56 	//The function is declared this way for mocking in unit tests
10:42:56 	^
10:43:01 Build step 'Execute shell' marked build as failure
```
It should be related to the recent merged PR #2418. @devimc Would you like to refine this comment by yourself or I just do the refinement here? ;).

amshinde:
Rerunning ARM CI now that https://github.com/kata-containers/runtime/pull/2435 is merged.

amshinde:
ARM CI is still failing @Pennyzct

Pennyzct:
Hi~ @amshinde 
Sorry for the ARM CI failure, and I'm already raised an PR [kata-containers/test#2294](https://github.com/kata-containers/tests/pull/2294) to fix them. 
But it should be irrelevant with this PR. ;).

Pennyzct:
re-based. ;)
/test

Pennyzct:
Hi~ guys. Got all green here~~~~~. ;).

Issue #2380, opened 2020-01-06
Assign vhost-user-blk/scsi device to Kata container as ephemeral or persistent storage


dong-liuliu:
**Which feature do you think can be improved?**

Host block devices can be assigned to container as ephemeral or persistent storage. But user can only add kernel based block device to Kata container currently. 
Even userspace storage application like SPDK already has a proven vhost solution which can accelerate IO performance a lot. Seems userspace storage devices like vhost-user-blk and vhost-user-scsi still can't be hotplugged to run a kata container. 
 I would like to create a general way which can assign vhost-user-blk/scsi device to kata container through docker cmd, or later Kubernetes.

**How can it be improved?**

Since kata container is hypervisor/VM based, not like bare-metal container, all vhost-user storage solution can be applied into Kata to improve storage performance.
In the source code of Kata repos, most code part of vhost-user storage devices already exists. But code to hotplug vhost-user-blk/scsi are still incomplete.
Method to assign vhost-user-blk/scsi is not defined in OCI. So it is still not clear how to assign vhost-user-blk/scsi devices through docker to kata container.

Some patches can be added to complete vhost-user-blk/scsi related code in kata and enable their hotplug ability. Also designing a method to assign vhost-user-blk/scsi device is necessary.

**Additional Information**

Currently, SR-IOV, GPU can be passhrought with Kata, and VPP(vhost-user-net) also can be added to support Kata. (https://github.com/kata-containers/documentation#kata-use-cases)
A similar documentation also can be added there to explain how to run vhost-user-blk with kata through docker. It will describe the basic thoery and steps to run Kata together with SPDK vhost-user-blk devices.

Also, a docker storage plugin can be created to simplify the handmade SPDK vhost-user operations. It would be easy and convenient for Kata user to try and apply SPDK accelerated storage device in their container workload.

**Before raising this feature request**

I've drafted a POC patch set in kata runtime and verified the work of run Kata with SPDK vhost-user-blk deivces as ephemeral or persistent storage.

Issue #587, opened 2018-08-16
cli: Kata Containers fails to launch without apt error if neither initrd/rootfs is mentioned.


nitkon:
# Description of problem

If one does not specify initrd or rootfs path in the configuration.toml file then Kata Containers fails to launch without any apt warning/error.

# Expected result

```
# sudo docker run -ti --runtime kata-runtime busybox sh
docker: Error response from daemon: oci runtime error: /usr/share/defaults/kata-containers/configuration.toml: having neither an image nor an initrd defined in the configuration file is not supported.
```

# Actual result
```
# sudo docker run -ti --runtime kata-runtime busybox sh
docker: Error response from daemon: oci runtime error: Create new sandbox failed.
```

nitkon:
FWIW: The latest upstream code fails with apt error message. 

```
sudo docker run -ti --runtime kata-runtime busybox sh
docker: Error response from daemon: OCI runtime create failed: Missing image and initrd path: unknown.
```

PR #2390, opened 2020-01-13
Assign vhost-user-blk device to Kata container as ephemeral or persistent storage #2380

dong-liuliu:
With this patch set, kata container can run with vhost-usre-blk device provided by SPDK.
Fixes: #2380


dong-liuliu:
Here are the steps to assign SPDK vhost-user-blk device to container.
    1. set enable_hugepage=true and enable_vhost_user_store in kata toml file.
    2. run spdk_tgt and create a vhost-user-blk device
    * file vhost-blk.conf is <
    [AIO]
      AIO /home/vhost-blk-file AIO1 4096
    [VhostBlk0]
      Name vhostblk0
      Dev AIO1
    >
    *spdk_tgt -c vhost-blk.conf -S /var/run/kata-containers/vhost-user-socket/

    3. Create fake block node and uevent file
    * mkdir -p /var/run/kata-containers/vhost-user-mknod/block/
    * mkdir -p /var/run/kata-containers/vhost-user-mknod/block/241:0/
    * file /var/run/kata-containers/vhost-user-mknod/block/241\:0/uevent is: <
    MAJOR=241
    MINOR=0
    DEVNAME=vhostblk0
    DEVTYPE=vhost-user-blk
    >

    4. mknod for vhostblk0 device
    * mknod /tmp/vhostblk0 b 241 0

    5. Run docker with vhost-user-blk
    * docker run --runtime kata-runtime --device=/tmp/vhostblk0:/dev/vda -it busybox sh

virtcontainers/device/drivers/vhost_user_blk.go
```diff
@@ -35,23 +45,37 @@ func (device *VhostUserBlkDevice) Attach(devReceiver api.DeviceReceiver) (err er
...
 		return err
 	}
-	id := hex.EncodeToString(randBytes)
 
-	device.DevID = id
-	device.Type = device.DeviceType()
+	vAttrs := &config.VhostUserDeviceAttrs{
+		DevID:      utils.MakeNameID("blk", device.DeviceInfo.ID, maxDevIDSize),
+		SocketPath: device.DeviceInfo.HostPath,
+		//Type:       VhostUserBlk,
```
jodh-intel: ?

virtcontainers/device/drivers/vhost_user_blk.go
```diff
@@ -35,23 +45,37 @@ func (device *VhostUserBlkDevice) Attach(devReceiver api.DeviceReceiver) (err er
...
+	}
 
-	return devReceiver.AppendDevice(device)
+	//return devReceiver.AppendDevice(device)
```
jodh-intel: Please remove all commented out code.

virtcontainers/pkg/oci/utils.go
```diff
@@ -414,6 +414,15 @@ func addHypervisorConfigOverrides(ocispec specs.Spec, config *vc.SandboxConfig)
 		config.HypervisorConfig.DisableVhostNet = disableVhostNet
 	}
 
+	if value, ok := ocispec.Annotations[vcAnnotations.EnableVhostUserStore]; ok {
+		enableVhostUserStore, err := strconv.ParseBool(value)
+		if err != nil {
+			return fmt.Errorf("Error parsing annotation for enable_vhost_user_store: Please specify boolean value 'true|false'")
```
jodh-intel: I don't think we need the `'true|false'` part of this error.

vendor/github.com/intel/govmm/qemu/qmp.go
```diff
@@ -946,6 +946,23 @@ func (q *QMP) ExecuteBlockdevDel(ctx context.Context, blockdevID string) error {
 	return q.executeCommand(ctx, "x-blockdev-del", args, nil)
 }
 
+func (q *QMP) ExecuteChardevDel(ctx context.Context, chardevID string) error {
```
jodh-intel: Files below `vendor/` should not be modified directly as the must be pristine upstream files. See https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#re-vendor-prs.

virtcontainers/device/config/config.go
```diff
@@ -75,6 +80,12 @@ var SysDevPrefix = "/sys/dev"
 // SysIOMMUPath is static string of /sys/kernel/iommu_groups
 var SysIOMMUPath = "/sys/kernel/iommu_groups"
 
+// VhostUserSockDir is static string
```
jodh-intel: This comment (and the one below) don't really describe what the variable is for - it's obvious they are static strings, but what are the strings for? :)

virtcontainers/pkg/oci/utils.go
dong-liuliu: Yes, it is not necessary there

virtcontainers/device/config/config.go
```diff
@@ -66,6 +66,11 @@ const (
 	VirtioFS = "virtio-fs"
 )
 
+const (
```
devimc: Please add a body to all commit messages

virtcontainers/device/config/config.go
```diff
@@ -232,7 +261,9 @@ func GetHostPath(devInfo DeviceInfo) (string, error) {
 	}
 
 	format := strconv.FormatInt(devInfo.Major, 10) + ":" + strconv.FormatInt(devInfo.Minor, 10)
-	sysDevPath := filepath.Join(SysDevPrefix, pathComp, format, "uevent")
+	sysDevPath := filepath.Join(sysDevPrefix, pathComp, format, "uevent")
+
+	fmt.Errorf("sysdevpath %s", sysDevPath)
```
devimc: and this is for ...

virtcontainers/device/config/config.go
```diff
@@ -213,11 +233,20 @@ var GetHostPathFunc = GetHostPath
...
 	if devInfo.ContainerPath == "" {
 		return "", fmt.Errorf("Empty path provided for device")
 	}
 
+	sysDevPrefix := SysDevPrefix
+	devDirPrefix := "/dev"
+
+	// Filter out vhost-user storage devices by device Major numbers.
+	if vhostUserStoreEnabled && (devInfo.Major == VhostUserSCSIMajor || devInfo.Major == VhostUserBlkMajor) {
+		sysDevPrefix = VhostUserMknodDir
```
devimc: I don't follow, who is going to populate these directories? can you document it?

dong-liuliu:
Thanks for your comments @jodh-intel .
Let me add UT and fix vendor qmp's change later.

The reason why user has to manually create a bunch of files is because of the limitation of OCI spec. It only record Major/Minor number for block devices. But vhost-user devices is specified by path of its socket. So I have to use the trick to emulate a sys directory for  uevent, and use Major number 241 and 242 as the signs of vhost-user-blk and vhost-user-scsi devices.

Yes, the prepare operations doesn't seem ideal to users. So I'm going to first add a docker volume plugin to automaticly handle these steps, just referencing kata's SRIOV and VPP's cases.

virtcontainers/device/config/config.go
dong-liuliu: Thanks for you review @devimc 
Let me add a link of drafted documentation here later.

virtcontainers/device/config/config.go
dong-liuliu: Sure, let me add the details to a document

virtcontainers/device/config/config.go
```diff
@@ -75,6 +83,12 @@ var SysDevPrefix = "/sys/dev"
 // SysIOMMUPath is static string of /sys/kernel/iommu_groups
 var SysIOMMUPath = "/sys/kernel/iommu_groups"
 
+// VhostUserSockDir is static string of vhost-user socket directory
+var VhostUserSockDir = "/var/run/kata-containers/vhost-user-socket"
+
+// VhostUserSockDir is static string of emulated vhost-user block node directory
+var VhostUserMknodDir = "/var/run/kata-containers/vhost-user-mknod"
```
c3d: Curious if it's there is a way to avoid having hard-coded paths here.

Isn't that something that we might want to overload from a `configuration.toml` file later? I'm a bit new to this code, but it looks to me like `config-settings.go` has a number of these paths defined. I would put yours there as well.

sameo:
> The reason why user has to manually create a bunch of files is because of the limitation of OCI spec. It only record Major/Minor number for block devices. But vhost-user devices is specified by path of its socket. So I have to use the trick to emulate a sys directory for uevent, and use Major number 241 and 242 as the signs of vhost-user-blk and vhost-user-scsi devices.
> 
> Yes, the prepare operations doesn't seem ideal to users. So I'm going to first add a docker volume plugin to automaticly handle these steps, just referencing kata's SRIOV and VPP's cases.

A few general comments on this PR:

1. We should use a well defined scheme for where we expect vhost-user sockets to live. For example, any vhost-user-blk sockets should be under `/var/run/kata-containers/vhost-user/block/sockets/`. And I agree with @c3d here: The base prefix should be a configurable value, with the default being `/var/run/kata-containers/vhost-user`. The rest of the path (`block`, `block/sockets`, etc) should be an immutable scheme.

2. I understand the OCI spec is limiting us in terms of the kind of device paths that we can pass to the containers, and it basically has to be a device node, either a block or a character one. (Side note here: We should consider extending the OCI spec for that, as devices can be more than this. They can live behind sockets for example...). Instead of going through the pain of creating an ad-hoc device node under e.g. tmpfs, and then a fake uevent file associated with it, can we do:
    a. Use a `/var/run/kata/containers/vhost-user/block/devices/` directory per vhost-user type to create those devices.
    b. The only reason why we need to parse the uevent file is in order to get the actual complete host device path through `GetHostPath()`. Can we simply infer the socket path from the device node path itself? In your example, we would create the node as:
    ```
    mknod /var/run/kata/containers/vhost-user/block/devices/vhostblk0 b 241 0
    ```
   and from `GetHostPath`, if we get a device that lives under the vhost-user/*/devices/ path and has a reserved major number, then we can infer from its full path where the vhost-user socket is (In that case `/var/run/kata/containers/vhost-user/block/sockets/vhostblk0`. So the device node name carries all the information we need. That way, the only artifact that you have to create when setting your vhost-user target is this fake device node.

Besides those changes, I think the PR looks good overall (Except for the vendor part, but I think you fixed that in govmm already).

@dong-liuliu @c3d  @jodh-intel @devimc WDYT?

c3d:
> 1. We should use a well defined scheme for where we expect vhost-user sockets to live. For example, any vhost-user-blk sockets should be under `/var/run/kata-containers/vhost-user/block/sockets/`.

Agreed.

> And I agree with @c3d here: The base prefix should be a configurable value, with the default being `/var/run/kata-containers/vhost-user`. The rest of the path (`block`, `block/sockets`, etc) should be an immutable scheme.

Agreed.

> 2. I understand the OCI spec is limiting us in terms of the kind of device paths that we can pass to the containers, and it basically has to be a device node, either a block or a character one. (Side note here: We should consider extending the OCI spec for that, as devices can be more than this. They can live behind sockets for example...). Instead of going through the pain of creating an ad-hoc device node under e.g. tmpfs, and then a fake uevent file associated with it, can we do:
>    a. Use a `/var/run/kata/containers/vhost-user/block/devices/` directory per vhost-user type to create those devices.

How would you name vhost-user types? Would you infer that from how it's used (and then the kata code would provide the name), or is there some standard way to go from device to name?

>    b. The only reason why we need to parse the uevent file is in order to get the actual complete host device path through `GetHostPath()`. Can we simply infer the socket path from the device node path itself? In your example, we would create the node as:
>    ```
>    mknod /var/run/kata/containers/vhost-user/block/devices/vhostblk0 b 241 0
>    ```

I assume in your example, the `0` at the end of `vhostblk0` is the same as the minor number. But the part I am wondering about is the `block` and `vhostblk`.

>    
>    
>    and from `GetHostPath`, if we get a device that lives under the vhost-user/*/devices/ path and has a reserved major number, then we can infer from its full path where the vhost-user socket is (In that case `/var/run/kata/containers/vhost-user/block/sockets/vhostblk0`. So the device node name carries all the information we need. That way, the only artifact that you have to create when setting your vhost-user target is this fake device node.

That would indeed be easier.

> 
> Besides those changes, I think the PR looks good overall (Except for the vendor part, but I think you fixed that in govmm already).
> 
> @dong-liuliu @c3d @jodh-intel @devimc WDYT?

dong-liuliu:
hi, @sameo @c3d 
I just changed the vhost-user device filtering method based on @sameo 's proposal.

As jodh advised, the patch set is still lack of a full document and unit test. And I'm working on them, will resubmit the patch set with UT and documetation next week.

sameo:
> > b. The only reason why we need to parse the uevent file is in order to get the actual complete host device path through `GetHostPath()`. Can we simply infer the socket path from the device node path itself? In your example, we would create the node as:
> > ```
> > mknod /var/run/kata/containers/vhost-user/block/devices/vhostblk0 b 241 0
> > ```
> 
> I assume in your example, the `0` at the end of `vhostblk0` is the same as the minor number. But the part I am wondering about is the `block` and `vhostblk`.
> 

Let me clarify what I have in mind, assuming we use the defaut Kata created vhost-user prefix set at `var defaultVhostUserPrefix = /var/run/kata/containers/vhost-user/`:

* The Kata runtime creates the `block`, `block/devices` and `block/sockets` directories on top of `defaultVhostUserPrefix`. Same for the `network` and `filesystem` vhost-user types.

* The operator starts the SPDK device: `spdk_tgt -c vhost-blk.conf -S /var/run/kata-containers/vhost-user/block/sockets/` where `vhost-blk.conf` is:

```
[AIO]
 AIO /home/vhost-blk-file AIO1 4096
[VhostBlk0]
 Name vhostblk0
 Dev AIO1
```
which means the operator tells SPDK to create the vhost-user socker at `/var/run/kata-containers/vhost-user/block/sockets/vhostblk0`

* The operator creates the fake vhost-user block device

```
mknod /var/run/kata/containers/vhost-user/block/devices/vhostblk0 b 241 0
```

* Kata eventually gets a request to pass the `/var/run/kata/containers/vhost-user/block/devices/vhostblk0` device into the container. This device's major is 241 and this is how Kata knows that this device is meant to be used as a vhost-user-blk. We use the experimental block major range (240-254) to map each vhost-user block type to a major. See @dong-liuliu code:
```
const (
	// From OCI spec, devices are only identified by its Major:Minor number.
	// Use these linux unused device major number to identify vhost-user
	// devices.
	VhostUserBlkMajor  = 241
	VhostUserSCSIMajor = 242
)
```

Now Kata knows that it needs to use a vhost-user-blk device, which name is `vhostblk0`. From there it can infer that the socket will be at `/var/run/kata/containers/vhost-user/block/sockets/vhostblk0`.

Basically the vhost-user socket path to pass down to QEMU can be built as:
`defaultVhostUserPrefix/<map between device major and vhu type>/sockets/<name of the fake device>`

Hopefully the proposal is clearer now, and it still makes sense :-)

dong-liuliu:
> Now Kata knows that it needs to use a vhost-user-blk device, which name is `vhostblk0`. From there it can infer that the socket will be at `/var/run/kata/containers/vhost-user/block/sockets/vhostblk0`.
> 
> Basically the vhost-user socket path to pass down to QEMU can be built as:
> `defaultVhostUserPrefix/<map between device major and vhu type>/sockets/<name of the fake device>`
> 
> Hopefully the proposal is clearer now, and it still makes sense :-)

Yes, it makes sense, and really simplifies extra operations.

c3d:
> Now Kata knows that it needs to use a vhost-user-blk device, which name is `vhostblk0`. From there it can infer that the socket will be at `/var/run/kata/containers/vhost-user/block/sockets/vhostblk0`.

I see, you map by name. I thought you were mapping by minor.

> 
> Basically the vhost-user socket path to pass down to QEMU can be built as:
> `defaultVhostUserPrefix/<map between device major and vhu type>/sockets/<name of the fake device>`
> 
> Hopefully the proposal is clearer now, and it still makes sense :-)

It makes sense to me.

dong-liuliu:
A full document is added in kata-containers/documentation PR [#587](https://github.com/kata-containers/documentation/pull/587)

pkg/katautils/config.go
```diff
@@ -123,6 +123,8 @@ type hypervisor struct {
 	DisableImageNvdimm      bool     `toml:"disable_image_nvdimm"`
 	HotplugVFIOOnRootBus    bool     `toml:"hotplug_vfio_on_root_bus"`
 	DisableVhostNet         bool     `toml:"disable_vhost_net"`
+	EnableVhostUserStore    bool     `toml:"enable_vhost_user_store"`
```
amshinde: Do we need this flag? Would we want to disable this feature for not supporting vhost-user devices?
The reason I bring this up, is that we already have a bloat of supported configs and do not want to add to these unless really required.

virtcontainers/device/config/config.go
```diff
@@ -205,6 +215,10 @@ type VhostUserDeviceAttrs struct {
 	Tag       string
 	CacheSize uint32
 	Cache     string
+
+	// PCIAddr is the PCI address used to identify the slot at which the drive is attached.
+	// It is only meaningful for vhost user devices
+	PCIAddr string
```
amshinde: Wondering if we should rename the existing `MacAddress` string in this to `Addr` and use that to store the PCIAddr in case of vhost-usr-block, mac address in case of vhost-user-net and in case of (still to be handled )vhost-user-scsi devices maybe use that to store SCSI address if required.
Would like to hear what others think about this.
cc @jodh-intel

virtcontainers/persist/api/device.go
```diff
@@ -66,6 +66,10 @@ type VhostUserDeviceAttrs struct {
 
 	// MacAddress is only meaningful for vhost user net device
 	MacAddress string
+
+	// PCIAddr is the PCI address used to identify the slot at which the drive is attached.
+	// It is only meaningful for vhost user devices
+	PCIAddr string
```
amshinde: s/only meaningful for vhost user devices/ only meaningful for vhost user block devices/

dong-liuliu:
hi, @amshinde Thanks for your comments.

I has a local patch for vhost-user-scsi, but didn't submit it yet. Just think the whole work can be split into several phases  to avoid a tedious revision/review process. Like:
- Phase1 is to add vhost-user-blk to qemu.
- Phase2 is to add vhost-user-scsi to qemu. Not like blk, SCSI is a bit more complicate, it also requires some change in kata-agent repo.
- Phase3 may be to verify and enable other hypervisors.
After this PR is merged, we can start new PR for next steps. Is this acceptable?

amshinde:
@dong-liuliu I thought I had replied to your comment above, but dont see it here!
I am fine with implementing the vhost-scsi part at a later phase, just asked as you had already added some logic here to detect vhost-scsi.

> * Phase1 is to add vhost-user-blk to qemu.
> 
> * Phase2 is to add vhost-user-scsi to qemu. Not like blk, SCSI is a bit more complicate, it also requires some change in kata-agent repo.
> 
> * Phase3 may be to verify and enable other hypervisors.
>   After this PR is merged, we can start new PR for next steps. Is this acceptable?

Your plan looks good to me. Can you rebase and address some of the other comments here?

virtcontainers/device/config/config.go
```diff
@@ -66,6 +68,14 @@ const (
 	VirtioFS = "virtio-fs"
 )
 
+const (
+	// From OCI spec, devices are only identified by its Major:Minor number.
+	// Use these linux unused device major number to identify vhost-user
```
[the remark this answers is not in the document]
dong-liuliu: Ack

virtcontainers/device/manager/manager.go
```diff
@@ -119,6 +119,12 @@ func (dm *deviceManager) createDevice(devInfo config.DeviceInfo) (dev api.Device
 	}
 	if isVFIO(path) {
 		return drivers.NewVFIODevice(&devInfo), nil
+	} else if isVhostUserBlk(devInfo) {
```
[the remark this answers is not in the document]
dong-liuliu: ack

pkg/katautils/config.go
dong-liuliu: Eh... Yes, having too many configs is really a problem.
At the time, I added this config option about the enablement. I thought this feature will occupy some Linux reserved block device numbers. But we didn't know whether these numbers are already utilized for other purpose in user's circumstance. If some user is going to use vhost-user storage devices, he should be aware that there is no conflict about that by the enablement config option.
So this config is still there.

virtcontainers/persist/api/device.go
dong-liuliu: Ack

virtcontainers/device/config/config.go
dong-liuliu: sounds good to me.
And could it be revised in some later PR to remove duplicated elements if jodh also agrees with us? :)

dong-liuliu:
Sorry for the late update @amshinde 
Also added some essential unit tests, together with your comments.

pkg/katautils/config.go
amshinde: This is fine with me. It may be useful to have the reserved block numbers configurable in the future. But we can get to that later if we see the need.

amshinde:
@devimc @jodh-intel Can you take a look as well.

pkg/katautils/config.go
dong-liuliu: :)

virtcontainers/device/config/config.go
dong-liuliu: Got it, and just added a new issue #2433 as a reminder.

virtcontainers/persist/api/device.go
dong-liuliu: Sorry, missed it.

virtcontainers/device/config/config.go
```diff
@@ -66,6 +68,14 @@ const (
 	VirtioFS = "virtio-fs"
 )
 
+const (
+	// The OCI spec requires the major-minor number to be provided for a
+	// device. Use these linux unused device major numbers to identify
+	// vhost-user storage devices.
```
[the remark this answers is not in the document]
dong-liuliu: Agree with you.

pkg/katautils/config.go
```diff
@@ -124,6 +124,8 @@ type hypervisor struct {
...
+	VhostUserStorePath      string   `toml:"vhost_user_store_path"`
```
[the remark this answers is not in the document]
dong-liuliu: Ack. Yes, devimc. They should be modified.
Let me add the modification in this patch set later.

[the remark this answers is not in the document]
dong-liuliu: Hi, @devimc. Added these 2 options into *toml.in.
Aftter reading Makefile, I'm not very clear on the criteria what option should be added into makefile. And also found some options, like DEFENABLEMEMPREALLOC for enable_mem_prealloc, are not used already.  So I didn't modify Makefile. Do you have any further suggestion on this?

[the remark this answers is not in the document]
devimc: Hi @dong-liuliu , yes paths shouldn't be hardcoded in the `toml.in` file, instead use Makefile variables. This commit is a good example https://github.com/kata-containers/runtime/commit/75f75862c2cd388393d2d1d06ee76a07d9a5145b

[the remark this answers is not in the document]
dong-liuliu: Great suggestion, @devimc.
Added related modifications to Makefile/toml.in

virtcontainers/qemu.go
```diff
@@ -1550,7 +1567,12 @@ func (q *qemu) hotplugAddMemory(memDev *memoryDevice) (int, error) {
 		memDev.slot = maxSlot + 1
 	}
 
-	share, target, memoryBack := q.getMemArgs()
+	share, target, memoryBack, err := q.getMemArgs()
+	if err != nil {
+		q.Logger().WithError(err).Error("hotplug memory")
```
devimc: is this log for debugging? if not, why it was not added in the previous call to `q.getMemArgs()`

virtcontainers/device/config/config.go
```diff
@@ -275,3 +295,53 @@ func GetHostPath(devInfo DeviceInfo) (string, error) {
...
 }
+
+// getVhostUserHostPath is used to fetch host path for the vhost-user device.
+// For vhost-user block device like vhost-user-blk or vhost-user-scsi, its
+// socket should be under directory "<vhostUserStorePath>/block/sockets/";
+// its corresponding device node should be under directory
+// "<vhostUserStorePath>/block/devices/"
+func getVhostUserHostPath(devInfo DeviceInfo, vhostUserStorePath string) (string, error) {
+	vhostUserDevNodePath := filepath.Join(vhostUserStorePath, "/block/devices/")
+	vhostUserSockPath := filepath.Join(vhostUserStorePath + "/block/sockets/")
```
devimc: s/+/,

virtcontainers/device/drivers/vhost_user_blk.go
```diff
@@ -35,30 +43,75 @@ func (device *VhostUserBlkDevice) Attach(devReceiver api.DeviceReceiver) (err er
...
+
+	// Since vhost-user-blk uses "vd" prefix in Linux kernel, not "sd",
+	// sandbox block index should be updated only if sandbox default block
+	// driver is "virtio-blk"/"virtio-blk-ccw"/"virtio-mmio" which uses
+	// "vd" prefix in Linux kernel.
+	updateBlockIndex := false
+	customOptions := device.DeviceInfo.DriverOptions
+	if customOptions["block-driver"] == "virtio-blk" ||
+		customOptions["block-driver"] == "virtio-blk-ccw" ||
+		customOptions["block-driver"] == "virtio-mmio" {
```
devimc: how about saving `customOptions["block-driver"]` in a variable?

virtcontainers/device/drivers/vhost_user_blk.go
```diff
@@ -35,30 +43,75 @@ func (device *VhostUserBlkDevice) Attach(devReceiver api.DeviceReceiver) (err er
...
-	device.DevID = id
-	device.Type = device.DeviceType()
+	vAttrs := &config.VhostUserDeviceAttrs{
+		DevID:      utils.MakeNameID("blk", device.DeviceInfo.ID, maxDevIDSize),
+		SocketPath: device.DeviceInfo.HostPath,
+		Type:       config.VhostUserBlk,
+	}
 
-	return devReceiver.AppendDevice(device)
+	deviceLogger().WithField("device", device.DeviceInfo.HostPath).WithField("SocketPath", vAttrs.SocketPath).Infof("Attaching %s device", config.VhostUserBlk)
```
devimc: please, don't use `Infof` and try with `WithFields`

virtcontainers/qemu.go
dong-liuliu: yes, it's for debugging.
Removed it.

virtcontainers/device/drivers/vhost_user_blk.go
dong-liuliu: ack

virtcontainers/device/drivers/vhost_user_blk.go
dong-liuliu: Ack. Using `Info` with `WithField`.

...Actually, it referenced `attach` function in file `block.go`

devimc:
travis is failing, may be this needs a rebase

dong-liuliu:
Hi, @devimc 
It's really a strange failure there -- it reported:
`ERROR: Document vendor/github.com/containernetworking/plugins/plugins/main/windows/CONTRIBUTORS.md is not referenced`.

I extracted the related markdown checking script out, and applied the current master branch. Seems `Document vendor/github.com/containernetworking/plugins/plugins/main/windows/CONTRIBUTORS.md` is also not referenced.

Pulled master branch and rebase on it, but the error is still there.

grahamwhaley:
Hmm, vendor documents should not be getting checked in the markdown check scripts... didn't we see this somewhere else recently ?

devimc:
uhmm, yeah, this error is new, @GabyCT @chavafg @jodh-intel have you seen this before?

dong-liuliu:
Hi, @devimc 
The extra check is triggered by change in ./virtcontainers/documentation/api/1.0/api.md.
Seems some other added configurations also aren't in api.md. So I just removed that change.

devimc:
@dong-liuliu seems like you need to rebase this, I see a [commit](https://github.com/kata-containers/runtime/pull/2390/commits/64bb720478c7bc44c0799eb2691f3353706a0efe) that shouldn't be here

dong-liuliu:
@devimc rebased :)

dong-liuliu:
Hi, @devimc @amshinde @c3d @jodh-intel 
Any other review comments? :)

virtcontainers/device/config/config.go
```diff
@@ -275,3 +295,53 @@ func GetHostPath(devInfo DeviceInfo) (string, error) {
...
+		}
+
+		devMajor := unix.Major(devStat.Rdev)
+		devMinor := unix.Minor(devStat.Rdev)
+		if devMajor == majorNum && devMinor == minorNum {
+			return file.Name(), nil
+		}
+	}
+
+	return "", fmt.Errorf("Required device node doesn't exist under directory")
```
jodh-intel: This only needs to be `errors.new()` as this is only a static string. But maybe you want to add `dirname`, `majorNum` and `minorNum` to the returned error?

virtcontainers/device/drivers/vhost_user_blk.go
```diff
@@ -35,30 +43,84 @@ func (device *VhostUserBlkDevice) Attach(devReceiver api.DeviceReceiver) (err er
...
 
-	device.DevID = id
-	device.Type = device.DeviceType()
+	vAttrs := &config.VhostUserDeviceAttrs{
+		DevID:      utils.MakeNameID("blk", device.DeviceInfo.ID, maxDevIDSize),
+		SocketPath: device.DeviceInfo.HostPath,
+		Type:       config.VhostUserBlk,
+	}
+
+	deviceLogger().WithField("device", device.DeviceInfo.HostPath).WithField("SocketPath", vAttrs.SocketPath).WithField("Type", config.VhostUserBlk).Info("Attaching device")
```
jodh-intel: You could use `WithFields()`:

```go
deviceLogger().WithFields(logrus.Fields{
    "socket-path" : vAttrs.SocketPath,
    // ...
})
```

virtcontainers/qemu.go
```diff
@@ -1152,6 +1203,38 @@ func (q *qemu) hotplugBlockDevice(drive *config.BlockDrive, op operation) error
...
+	if err != nil {
+		return err
+	}
+
+	devID := "virtio-" + vAttr.DevID
+
+	if op == addDevice {
+		switch vAttr.Type {
+		case config.VhostUserBlk:
+			err = q.hotplugAddVhostUserBlkDevice(vAttr, op, devID)
```
jodh-intel: I wonder if it would be safer to change this to:

```go
return q.hotplugAddVhostUserBlkDevice(vAttr, op, devID)
```

The last line of the function can then be simply:

```go
return nil
```

virtcontainers/qemu.go
```diff
@@ -692,12 +701,16 @@ func (q *qemu) getMemArgs() (bool, string, string) {
 	} else if q.config.SharedFS == config.VirtioFS || q.config.FileBackedMemRootDir != "" {
 		target = q.qemuConfig.Memory.Path
 		memoryBack = "memory-backend-file"
+	} else if q.config.EnableVhostUserStore {
```
c3d: Why is the `enable_vhost_user_store` hugepage test disabled if virtiofs is configured?

To me it should be:

```
if q.qemuConfig.Knobs.HugePages {
...
} else {
   if q.config.sharedFS ... { ... }
   if q.config.EnableVhostUserStore {  ... }
```

Or is it because you test it elsewhere?

virtcontainers/device/config/config.go
dong-liuliu: Thanks for your suggestion and approval, @jodh-intel 

Add detail information about `dirname`, `majorNum` and `minorNum` to the returned error.
```
return "", fmt.Errorf("Required device node (%s:%s) doesn't exist under directory %s",
		majorNum, minorNum, dirname)
```

virtcontainers/device/drivers/vhost_user_blk.go
dong-liuliu: Ack. Applied~

virtcontainers/qemu.go
dong-liuliu: Ack. Applied

virtcontainers/qemu.go
dong-liuliu: Ack. :)  Applied.
It was caused unintentionally by a previous manual rebase.

jodh-intel:
Hi @dong-liuliu - please can you update this branch as it is conflicted?

dong-liuliu:
> Hi @dong-liuliu - please can you update this branch as it is conflicted?
Ack. Sorry, didn't notice the confilct.

jodh-intel:

dong-liuliu:
Thanks for your approval, @devimc 
Just found and fixed the extra test failure in which a added case should be skipped as requiring root privilege.

Please help me retrigger the extra test :)

dong-liuliu:
Added more unit tests to improve code coverage :)

dong-liuliu:
hi, @amshinde 
need your help to trigger /test again 👍

dong-liuliu:
hi, @chavafg,
Seems a test retrigger is still required.
The failures were some random errors from their error log:
one was caused by GPRC `Failed to check if grpc server is working: rpc error: code = Unavailable desc = transport is closing`; 
Or ` Trying to pull registry.fedoraproject.org/fedora-minimal:latest... 04:25:54   received unexpected HTTP status: 503 Service Temporarily Unavailable`;
And `ERROR: kata-log-parser: missing pid`

grahamwhaley:
@dong-liuliu - I've re-triggered the F31 CI.
@GabyCT - is the F31 CI stable? I see a lot of fails there - is it something we can/should currently ignore?

chavafg:
Seems that F31 started to fail when running with podman, I saw the same issue on other PRs. 
About the other failures, yes, they seem to be random, I have retrigger those jobs.

dong-liuliu:
Thanks @grahamwhaley and @chavafg,
we've gotten debian-10 passed~.

Seems F31 with Podman really failed weirdly many times.
And for U1804 with initrd also failed weirdly. I saw for its [one successful build log](http://jenkins.katacontainers.io/job/kata-containers-runtime-ubuntu-18-04-PR-initrd/2920/console) showed same content with  [this failure's log](http://jenkins.katacontainers.io/job/kata-containers-runtime-ubuntu-18-04-PR-initrd/2922/console)

dong-liuliu:
Seems CI-`F31 with podman` and CI-`U1804 with initrd` failed a lot.
Is it possible to just ignore them currently? @GabyCT

GabyCT:
@dong-liuliu , the previous failure with the podman CI, it is a bug that we currently have, however, I re-trigger the jenkins job and now it seems we have a failure but this is related with this PR
```
13:46:55      BUILD    /tmp/jenkins/workspace/kata-containers-runtime-podman-fedora-31-PR/go/src/github.com/kata-containers/runtime/kata-runtime
13:47:06 # github.com/kata-containers/runtime/virtcontainers/device/drivers
13:47:06 ../virtcontainers/device/drivers/vhost_user_blk.go:76:16: devReceiver.DecrementSandboxBlockIndex undefined (type api.DeviceReceiver has no field or method DecrementSandboxBlockIndex)
13:47:08 make: *** [Makefile:539: /tmp/jenkins/workspace/kata-containers-runtime-podman-fedora-31-PR/go/src/github.com/kata-containers/runtime/kata-runtime] 
```
Please see http://jenkins.katacontainers.io/job/kata-containers-runtime-podman-fedora-31-PR/66/console

GabyCT:
@dong-liuliu , the initrd CI is also failing, and it also seems related with this PR
```
# github.com/kata-containers/runtime/virtcontainers/device/drivers
../virtcontainers/device/drivers/vhost_user_blk.go:76:16: devReceiver.DecrementSandboxBlockIndex undefined (type api.DeviceReceiver has no field or method DecrementSandboxBlockIndex)
Makefile:539: recipe for target '/tmp/jenkins/workspace/kata-containers-runtime-ubuntu-18-04-PR-initrd/go/src/github.com/kata-containers/runtime/kata-runtime' failed
make: *** [/tmp/jenkins/workspace/kata-containers-runtime-ubuntu-18-04-PR-initrd/go/src/github.com/kata-containers/runtime/kata-runtime] Error 2
Failed at 67: "${cidir}/install_runtime.sh" "${tag}"
```

dong-liuliu:
Thank you, @GabyCT 
I just rebased the patch set on recently merged patches, and fixed the issue.

dong-liuliu:
hi, @amshinde, @devimc 
What's the next step I should do to get this patchset merged into branch? :)

PR #2399, opened 2020-01-17
unit-test: cleaning up stale files under /tmp

Pennyzct:
I've found a few unit tests which were generating stale files under `/tmp`

**- Using `defer` in TestMain.**
os.Exit will skip all deferred instructions.
So we should reconstruct TestMain to leave all setup-related code in setup(), and all cleanup-related code in shutdown().

**- Missing deleting what `ioutil.TempDir()` creates**
Normally, ` ioutil.TempDir` will create a new temporary dir under `/tmp`.

**- Missing deleting what `ioutil.TempFile()` creates**
`ioutil.TempFile` creates a new temporary file in the directory `/tmp`.
It is the caller's responsibility to remove the file when no longer needed.

### DIFF
--- a/virtcontainers/store/manager_test.go
+++ b/virtcontainers/store/manager_test.go
@@ -27,6 +27,9 @@ var storeRoot, storeRootDir = func() (string, string) {
 	dir, _ := ioutil.TempDir("", "")
 	return "file://" + dir, dir
 }()
+var testDir = ""
+var ConfigStoragePathSaved = func() string { return "" }
+var RunStoragePathSaved = func() string { return "" }
 
 func TestNewStore(t *testing.T) {
 	s, err := New(context.Background(), storeRoot)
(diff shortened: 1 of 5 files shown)


virtcontainers/store/manager_test.go
```diff
@@ -111,31 +114,36 @@ func TestManagerFindStore(t *testing.T) {
 // TestMain is the common main function used by ALL the test functions
 // for the store.
 func TestMain(m *testing.M) {
-	testDir, err := ioutil.TempDir("", "store-tmp-")
+	setup()
+	rt := m.Run()
+	shutdown()
+	os.Exit(rt)
```
grahamwhaley: OOI, in case you have never seen this - we check for no uses of `os.Exit()` in the main codebase in the CI:
https://github.com/kata-containers/runtime/blob/master/.ci/go-no-os-exit.sh

(I think that is a feature we've had for a very very long time and is little seen, so just raising here for general awareness).
but, maybe that does not apply to test code.

vijaydhanraj:
Thanks @Pennyzct . acrn test changes look fine to me.

virtcontainers/store/manager_test.go
Pennyzct: Hi~ @grahamwhaley Thanks for the notification. ;).
I'm afraid that it's not applicable for go unit test code, referring from [Go.Doc](https://golang.org/pkg/testing/#hdr-Main):
```
TestMain runs in the main goroutine and can do whatever setup and teardown 
is necessary around a call to m.Run. 
It should then call os.Exit with the result of m.Run.
```

Pennyzct:
firecracker CI failed irrelevantly. All the other are green~~~

PR #2401, opened 2020-01-17
virtcontainers/qemu: fix createSandbox to admin de morgan's law

freeHackOfJeff:


grahamwhaley:
Hi @freeHackOfJeff - thanks for your contribution!
To get the PR past our CI, you'll have to conform to some of our formatting requirements, such as adding an Issue to reference the PR against, and signing off your commit. Please have a read of https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md and do ask us if you need some help or guidance!

virtcontainers/qemu.go
```diff
@@ -509,7 +509,7 @@ func (q *qemu) createSandbox(ctx context.Context, id string, networkNS NetworkNa
 	// subsequent ones with shared=off. virtio-fs always requires shared=on for
 	// memory.
 	if q.config.SharedFS == config.VirtioFS || q.config.FileBackedMemRootDir != "" {
-		if !(q.config.BootToBeTemplate || q.config.BootFromTemplate) {
+		if (!q.config.BootToBeTemplate && !q.config.BootFromTemplate) {
```
WeiZhang555: They're same logic with different code writting. 

What is this trying to fix?

freeHackOfJeff: That's  is more readable.

grahamwhaley: I think for this relatively simple formula, there is little difference.
If the formula were more complex, then yes, maybe [De Morgan's](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) might help us simplify things.
I'm not sure it really applies here - the risk of breaking something probably outweighs the debatable improvement in readability (yes yes, I know in this case it is pretty easy to visually check the code is functionally the same....)

devimc: thanks for raising this @freeHackOfJeff , but I agree with @WeiZhang555 and @grahamwhaley
If we accept this change, someone else could apply `De Morgan's` again arguing readability and we'll end up with the same original expression, sorry

grahamwhaley:
I think the consensus here is that the change is small enough and effectively an 'equivalent', so is not needed. Thanks for your contribution @freeHackOfJeff - but at this time, we'll close this PR.

PR #2407, opened 2020-01-22
qemu: Add virtio-mem support

teawater:
This PR adds qemu virtio-mem support.
Then qemu can use virtio-mem support memory resize.

To enable this function, need the Linux and the qemu that support virtio-mem.
The kernel is at https://github.com/davidhildenbrand/linux/tree/virtio-mem-rfc-v4.
The kernel config that can work with kata is at https://gist.github.com/teawater/016194ee84748c768745a163d08b0fb9.
The qemu is at https://github.com/teawater/qemu/tree/kata-virtio-mem.   (The original source is at https://github.com/davidhildenbrand/qemu/tree/virtio-mem.  Its base version of qemu cannot work with kata.  So I merge the commit of virtio-mem to upstream qemu.)

Use command "echo 1 > /proc/sys/vm/overcommit_memory" to enable memory overcommitment of a Linux kernel.  Because qemu virtio-mem device need to allocate a lot of memory.

Set "enable_virtio_mem" of kata configuration to true.


virtcontainers/sandbox.go
```diff
@@ -1764,6 +1764,7 @@ func (s *Sandbox) calculateSandboxMemory() int64 {
 
 		if m := c.Resources.Memory; m != nil && m.Limit != nil {
 			memorySandbox += *m.Limit
+			s.Logger().WithField("container-id", c.ID).Debugf("jojo size %d", memorySandbox/1024/1024)
```
jodh-intel: Jojo? :smile: Also, can't that size be a field (`WithFields()`)?

devimc:
@teawater do you plan to document this?

virtcontainers/sandbox.go
teawater: > Jojo? 😄 Also, can't that size be a field (`WithFields()`)?

@jodh-intel Oops!  This is a temporary log when I wrote the code.  I had removed it in the new version.

teawater:
> @teawater do you plan to document this?

@devimc I will post a PR for the doc when current PR is merged.

virtcontainers/sandbox.go
jodh-intel: 哈哈 :smile:

jodh-intel:
Ping @kata-containers/runtime for reviews...

teawater:
jenkins-ci-ubuntu-1804-firecracker cannot pass because kata still not support fc new hybrid vsock.

PR #2409, opened 2020-01-22
virtcontainers: constrain docker container when sandbox_cgroup_only=true

### LINKED ISSUE #2408: cgroups: unconstrained docker container when sandbox_cgroup_only=true
# Description of problem

set `sandbox_cgroup_only=true` in configuration file and run a docker containers

```
$ docker run --cpus 3 --runtime=kata-runtime -dt busybox tail -f /dev/null
5fc14ead2036d267f
```

check container's cpu constraints in the host

```
$ sudo cat /sys/fs/cgroup/cpu/docker/kata_5fc14ead2036d267f/cpu.cfs_quota_us
-1
````

`cpu.cfs_quota_us` must be `300000` because of `cpu.cfs_period_us` is `100000`.

`quota / period = cpus`

cc @GabyCT  @jcvenegas

### PR DESCRIPTION
**Pull request by devimc**
The sandbox cgroup will be constrained if there is no container
 type annotation, otherwise kata will rely on container engine's cgroup
configuration

fixes #2408

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -122,6 +122,9 @@ type SandboxConfig struct {
 
 	DisableGuestSeccomp bool
 
+	// HasCRIContainerType specifies whether container type was set explicitly through annotations or not.
+	HasCRIContainerType bool
+
 	// Experimental features enabled
 	Experimental []exp.Feature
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: virtcontainers/sandbox.go [hunk 1]
```diff
@@ -127,7 +127,10 @@ type SandboxConfig struct {
 
 	// Cgroups specifies specific cgroup settings for the various subsystems that the container is
 	// placed into to limit the resources the container has available
-	Cgroups *configs.Cgroup `json:"cgroups"`
+	Cgroups *configs.Cgroup
+
+	// CRIContainerType specifies whether container type was set explicitly through annotations or not.
+	CRIContainerType bool
```
jodh-intel: Nit: It might be worth renaming this variable to make it slightly clearer as on first inspection is seems like it's specifying a type for a container (which it isn't ;)

devimc:
@jodh-intel change applied, please take a look

/test

jodh-intel:
Restarted Fedora CI which timed out (due to a docker repo problem possibly?)

Firecracker failed with:

```
not ok 42 ctr with non-root user has no effective capabilities
# (in test file ctr.bats, line 1366)
```

That could be a bug, but let's respin and see what happens on the 2nd attempt...

devimc:
FC CI failed again, but now different error:

```
not ok 13 ctr log max with default value
# (in test file ctr.bats, line 463)
```

nothing related to this PR

PR #2425, opened 2020-01-29
versions: bump conmon version to v2.0.5

haircommander:
in hopes the old failures were fixed, and to investigate them if not
if passes, this will fix: https://github.com/kata-containers/tests/issues/2271

Signed-off-by: Peter Hunt <pehunt@redhat.com>

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -183,7 +183,7 @@ externals:
   conmon:
     description: "An OCI container runtime monitor"
     url: "https://github.com/containers/conmon"
-    version: "v2.0.1"
+    version: "v2.0.5"
 
   crio:
     description: |


amshinde:
So I see the ubuntu tests are failing with the same reason that we saw earlier.
@haircommander Have a look at https://github.com/kata-containers/runtime/issues/2352#issuecomment-565642204

haircommander:
(starting to pin down where it failed)

chavafg:
Job failed on other (flaky?) tests, but cri-o passed OK. 

We started to see the issue with conmon on Dec 11, and conmon v2.0.5 was released on Dec 10. I think this could be the version that started to give problems to us.

haircommander:
looks like cri-o tests are going to pass. we can live with 2.0.5 for now (it has the changes needed to make cri-o tests pass again), though I don't want to lose track of the fact we're not following conmon master. I'll make an issue here after this closes :)

haircommander:
PTAL @chavafg @amshinde @devimc

PR #2429, opened 2020-01-31
rootless: implement rootless fs and support --rootless option

devimc:


devimc:
review please

virtcontainers/clh_test.go
```diff
@@ -172,7 +172,14 @@ func TestCloudHypervisorBootVM(t *testing.T) {
 }
 
 func TestCloudHypervisorCleanupVM(t *testing.T) {
-	clh := &cloudHypervisor{}
+	store, err := persist.GetDriver()
+	if err != nil {
+		t.Errorf("persist.GetDriver() unexpected error != %v", err)
```
jodh-intel: Nit: The assert syntax save you two lines of code :smile:

```go
assert.NoError(err)
```

devimc: Fixed, thanks

cli/main.go
```diff
@@ -266,6 +271,17 @@ func beforeSubcommands(c *cli.Context) error {
 		return nil
 	}
 
+	r, err := parseBoolOrAuto(c.GlobalString("rootless"))
+	if err != nil {
+		return err
+	}
+	// If nil, not set a value, let vc auto-detect rootless
```
amshinde: This comment is slightly confusing, as we are checking non-nil value below. Perhaps, this could be worded as
`If flag is true/false, assign the rootless flag. vc will not perform any auto-detection in that case. In case flag is nil or auto, vc detects if the runtime is running as rootless.`

cli/utils.go
```diff
@@ -131,3 +132,13 @@ func genericGetCPUDetails() (vendor, model string, err error) {
 
 	return vendor, model, nil
 }
+
+// from runC
+// parseBoolOrAuto returns (nil, nil) if s is empty or "auto"
+func parseBoolOrAuto(s string) (*bool, error) {
```
amshinde: I know this is a small function, but I think its better to include the runc license in the file if not included since we have taken this code varbatim. cc @grahamwhaley wdyt?

cli/main.go
devimc: it's set in `TestMain` (entrypoint) https://github.com/kata-containers/runtime/pull/2429/files#diff-e8e7e93733a02e75346a2f9d43ad6149R113

cli/utils.go
devimc: fixed, thanks

virtcontainers/acrn_test.go
```diff
@@ -208,7 +213,12 @@ func TestAcrnGetSandboxConsole(t *testing.T) {
 func TestAcrnCreateSandbox(t *testing.T) {
 	assert := assert.New(t)
 	acrnConfig := newAcrnConfig()
-	a := &Acrn{}
+	store, err := persist.GetDriver()
```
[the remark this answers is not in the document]
devimc: it's set in TestMain (entrypoint) https://github.com/kata-containers/runtime/pull/2429/files#diff-e8e7e93733a02e75346a2f9d43ad6149R113

devimc:
@amshinde changes applied, thanks

PR #2436, opened 2020-02-04
virtcontainers: clh: Set the serial to NULL instead of OFF

### LINKED ISSUE #2437: clh: Set serial to "null"
If we want to avoid any potential weird behavior regarding the boot of the guest kernel, it's better to provide an emulated serial port. Because we don't care about getting the logs from this port, we simply need to set it to "null" so that these logs get discarded.

sboeuf:
The guest Linux kernel might try to access the serial port, and in case
the serial is "off", this might cause some slowness because the port is
not emulated at all. Problem is, when the port is not emulated, the
default value when reading the I/O port will be 0, which has a special
meaning in case of the serial port. It means there is some data ready
to be read, which the kernel might try to read for some time, causing
global system slowness.

That's why it's safer to use "null" by default as this means the serial
port will be emulated but anything written to it will be redirected to
/dev/null.

Fixes #2437 

### DIFF
--- a/virtcontainers/clh.go
+++ b/virtcontainers/clh.go
@@ -265,7 +265,7 @@ func (clh *cloudHypervisor) createSandbox(ctx context.Context, id string, networ
 
 	} else {
 		clh.vmconfig.Serial = chclient.ConsoleConfig{
-			Mode: cctOFF,
+			Mode: cctNULL,
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


likebreath:
LGTM.

sboeuf:
The whole CI looks pretty unstable. I can't imagine this small change causing so much trouble. Is it expected that multiple builds would not pass?

jcvenegas:
Agree, actually this only changes clh codebase so lets, I restarted and then I say we merge it

PR #2443, opened 2020-02-07
selinux: Disable selinux

### LINKED ISSUE #2442: Disable selinux
Kata fails when selinux policy is enforced and the container manager passes those labels in the OCI spec as reported in this issue:
https://github.com/kata-containers/runtime/issues/784

We should implement support for selinux. The first basic step would be to add selinux file system mount in the guest.
Before implementing support for selinux, lets disable it by not passing the selinux labels to the container running in the guest, to unblock users running kata on a system with selinux policy enforced.

amshinde:
Till we implement support for selinux, disable selinux
by not passing selinux labels in the container spec.

Fixes #2442

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -1007,7 +1007,7 @@ func (k *kataAgent) replaceOCIMountsForStorages(spec *specs.Spec, volumeStorages
 	return nil
 }
 
-func constraintGRPCSpec(grpcSpec *grpc.Spec, passSeccomp bool) {
+func (k *kataAgent) constraintGRPCSpec(grpcSpec *grpc.Spec, passSeccomp bool) {
 	// Disable Hooks since they have been handled on the host and there is
 	// no reason to send them to the agent. It would make no sense to try
 	// to apply them on the guest.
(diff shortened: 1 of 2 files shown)


grahamwhaley:
as per @c3d comments - I'm not a fan of silently disabling a feature. It's a security feature, the user has asked for it (even if it was implicit and they did not know they had), and we silently disabled it.
At minimum we need to issue a warning somewhere (which, sadly is likely to be in a log nobody reads, because the container launch will 'work', so nobody has any reason to check the logs....).

grahamwhaley:
I guess... in the spirit of moving forwards ;-), if we:
- can make this issue an appropriate warning
- and document in the Limitations document that although selinux is **not supported**, the runtime will moderately silently disable it, whilst logging a warning
- and we work towards the longer term solution of implementing selinux support for kata....

then, ack!

virtcontainers/kata_agent.go
```diff
@@ -1019,6 +1019,12 @@ func constraintGRPCSpec(grpcSpec *grpc.Spec, passSeccomp bool) {
 		grpcSpec.Linux.Seccomp = nil
 	}
 
+	// Disable selinux
+	if grpcSpec.Process.SelinuxLabel != "" {
+		k.Logger().Warnf("Selinux label specified in config, but not supported in Kata yet, running container without selinux")
```
grahamwhaley: nitpick, for which I don't know if you might fail the CI - but, you don't need `Warnf` if you have no % formatters in the string - you might be able to just use `Warn` :-)

c3d: Looks good to me with this change, but I am not a member, so I cannot approve.

PR #2458, opened 2020-02-14
cri-o: fix netns mount point leaking from cri-o

Pennyzct:
# Description of problem
Hi~ guys
I've found tons of leaking network namespace mount point on ARM CI, which comes from cri-o.
```
root@arm-testing-2:~# mount | grep "/run/netns/cni-" | wc -l
3084
```
`cri-o v1.16.x` forgets to un-mount them when they're in no use. But it got fixed in `cri-o v1.17.0`,
see [related code here](https://github.com/cri-o/cri-o/blob/release-1.17/internal/lib/sandbox/namespaces_linux.go#L183) for confirmation. ;)

So we need to do the version update for cri-o. And Since CRI-O and Kubernetes follow the same release cycle and deprecation policy, I will also update k8s as well. ;).

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -189,7 +189,7 @@ externals:
     description: |
       OCI-based Kubernetes Container Runtime Interface implementation
     url: "https://github.com/cri-o/cri-o"
-    version: "v1.16.0"
+    version: "0eec454168e381e460b3d6de07bf50bfd9b0d082"
     meta:
       openshift: "6273bea4c9ed788aeb3d051ebf2d030060c05b6c"
       crictl: 1.0.0-beta.2
@@ -204,7 +204,7 @@ externals:
   critools:
     description: "CLI tool for Container Runtime Interface (CRI)"
     url: "https://github.com/kubernetes-sigs/cri-tools"
-    version: "1.16.1"
+    version: "1.17.0"
 
   docker:
     description: "Moby project container manager"
@@ -235,7 +235,7 @@ externals:
     uscan-url: >-
       https://github.com/kubernetes/kubernetes/tags
       .*/v?([\d\.]+)\.tar\.gz
-    version: "1.16.2-00"
+    version: "1.17.3-00"
 
   openshift:
     description: |


Pennyzct:
I've also needed to do some coding on tests repo, so adding wip and DNM flag. ;)

Pennyzct:
**Updates:**
This PR https://github.com/cri-o/cri-o/pull/3239 is critical for running cri-o v1.17.x on AArch64, and specific reason is elaborated here  https://github.com/cri-o/cri-o/issues/3233.
Although this PR has been cherry-picked to the `v.1.17.x`, now it's in the branch `release-v1.17`, not the tag `v1.17.0`.

versions.yaml
```diff
@@ -189,7 +189,7 @@ externals:
     description: |
       OCI-based Kubernetes Container Runtime Interface implementation
     url: "https://github.com/cri-o/cri-o"
-    version: "v1.16.0"
+    version: "release-1.17"
```
devimc: `release-1.17` is a branch, not a tag. afaik some branches are still under development (can accept new patches), therefore we won't have a 100% reproducible build/testing

cc @chavafg @GabyCT

chavafg: hmm, not sure what would be the best option here. In the past we have used the commit ID of the fix, In this case: [1].  
Then when new release is out, we can again update.

[1] https://github.com/cri-o/cri-o/commit/44c9932a4ab9c3836cb748eb9cd7736037d798eb

amshinde: @chavafg Using a commit id instead of a branch makes sense here.

Pennyzct: Hi @chavafg @devimc @amshinde 
I've changed to use a commit id here. In the future, it may be included in the v1.17.1. ;).

Pennyzct:
This PR is related to and dependent on https://github.com/kata-containers/tests/pull/2305. ;)

versions.yaml
devimc: thanks @Pennyzct 

> I've changed to use a commit id here. In the future, it may be included in the v1.17.1.

Could you file an issue for that?

Pennyzct: okay, raise the issue asap. ;).

Pennyzct:
### Updates:
My PR https://github.com/kata-containers/tests/pull/2305 needs to include this commit https://github.com/cri-o/cri-o/pull/3203 in cri-o `v1.17.x`.
So I update cri-o version to this more recent commit here.

versions.yaml
Pennyzct: Done. See https://github.com/kata-containers/runtime/issues/2486. ;)

jodh-intel:
Thanks @Pennyzct. This is effectively a "re-vendor" PR, so could you add details of what changed between the two versions as shown in https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#re-vendor-prs?

Pennyzct:
Hi~ @jodh-intel 
This version update on `cri-o` and `k8s` from `v1.6.x` to `v1.17.x` is really a big update. Changes consist of hundreds of commits. 
Taking `cri-o` as example:
```
$ git log --no-merges --abbrev-commit --pretty=oneline "${old_commit}..${new_commit}" | sed 's/^/    /g'
227
```
So here I import their official update docs to illustrate the changes.
CRI-O: https://github.com/cri-o/cri-o/releases/tag/v1.17.0
Kubernetes: https://kubernetes.io/docs/setup/release/notes/

Pennyzct:
Hi~ @chavafg @GabyCT 
podman CI failed irrelevantly, and the error is as follows:
```
09:44:52 ERROR: /tmp/jenkins/workspace/kata-containers-runtime-podman-fedora-31-PR/go/src/github.com/kata-containers/osbuilder/rootfs is not a directory
09:44:52 Failed at 137: sudo AGENT_INIT="${AGENT_INIT}" USE_PODMAN=true DOCKER_RUNTIME="crun" ./initrd-builder/initrd_builder.sh "$ROOTFS_DIR"
09:44:52 Failed at 26: "${cidir}/install_kata_image.sh" "${tag}"
```
It may be because of that in [installing script](https://github.com/kata-containers/tests/blob/master/.ci/install_kata_image.sh#L123), variable `ROOTFS_DIR` hasn't been passed in due to missing `-E` flag.

Issue #2460, opened 2020-02-17
Rescan PCI conflicts with shpchp / pciehp


Jimmy-Xu:
# Description of problem

My case is VFIO passthrough Nvidia GPU to kata container.

For MMIO space over 4G for PCI hotplug (for example: Nvidia Tesla P100), Hot-plugging VFIO devices requires shpchp/pciehp in guest OS.

The kata agent ([a628496](https://github.com/kata-containers/agent/commit/a6284965adab127ef8b0c597379ed898546c9f99)) in the VM will rescan the PCI bus before waiting for new devices.

Usually rescan will initialize the VFIO device before shpchp / pciehhp, which will cause the MMIO BAR space allocation to fail.

# Dependency
To support VFIO and enable loading modules, I customized the guest kernel
https://github.com/kata-containers/packaging/pull/938

# Expected result

```
$ dmesg
...
[    9.566010] shpchp 0000:00:02.0: PCI bridge to [bus 01]
[    9.566024] shpchp 0000:00:02.0:   bridge window [io  0xc000-0xcfff]
[    9.567377] shpchp 0000:00:02.0:   bridge window [mem 0xfe200000-0xfe3fffff]
[    9.568225] shpchp 0000:00:02.0:   bridge window [mem 0xfe800000-0xfe9fffff 64bit pref]
[    9.570095] PCI: No. 2 try to assign unassigned res
[    9.570098] shpchp 0000:00:02.0: resource 8 [mem 0xfe200000-0xfe3fffff] released
[    9.570099] shpchp 0000:00:02.0: PCI bridge to [bus 01]
[    9.571044] shpchp 0000:00:02.0: resource 9 [mem 0xfe800000-0xfe9fffff 64bit pref] released
[    9.571046] shpchp 0000:00:02.0: PCI bridge to [bus 01]
[    9.574381] shpchp 0000:00:02.0: BAR 9: assigned [mem 0x4400000000-0x49ffffffff 64bit pref]
[    9.574383] shpchp 0000:00:02.0: BAR 8: assigned [mem 0xc0000000-0xc0ffffff]
[    9.574386] pci 0000:01:01.0: BAR 1: assigned [mem 0x4400000000-0x47ffffffff 64bit pref]
[    9.574467] pci 0000:01:01.0: BAR 3: assigned [mem 0x4800000000-0x4801ffffff 64bit pref]
[...]

PR #2461, opened 2020-02-17
Fix rescan PCI conflicts with shpchp / pciehp

Jimmy-Xu:
This is a solution to issue #2460 .

Since the pci rescan will conflict with shpchp / pciehp, we can attach the Nvidia GPU after creating the container, and then the pci rescan will not include the Nvidia GPU. 
shpchp / pciehp will handle hot plugging of Nvidia GPU correctly.

There may be better solutions, but this is a working solution.

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -861,9 +861,38 @@ func (c *Container) create() (err error) {
 		}
 	}
 
-	// Attach devices
-	if err = c.attachDevices(); err != nil {
-		return
+	var (
+		machineType        = c.sandbox.config.HypervisorConfig.HypervisorMachineType
+		normalAttachedDevs []ContainerDevice //for q35: normally attached devices
+		delayAttachedDevs  []ContainerDevice //for q35: delay attached devices, for example, large bar space device
+	)
+	// Fix: https://github.com/kata-containers/runtime/issues/2460
+	if machineType == QemuQ35 {
+		// add Large Bar space device to delayAttachedDevs
+		for _, device := range c.devices {
+			var isLargeBarSpace bool
+			isLargeBarSpace, err = manager.IsVFIOLargeBarSpaceDevice(device.ContainerPath)
+			if err != nil {
+				return
+			}
+			if isLargeBarSpace {
+				delayAttachedDevs = append(delayAttachedDevs, device)
+			} else {
+				normalAttachedDevs = append(normalAttachedDevs, device)
+			}
+		}
+	} else {
+		normalAttachedDevs = c.devices
+	}
+
+	c.Logger().WithFields(logrus.Fields{
+		"machine_type": machineType,
+		"devices":      normalAttachedDevs,
+	}).Info("normal attach devices")
+	if len(normalAttachedDevs) > 0 {
+		if err = c.attachDevices(normalAttachedDevs); err != nil {
+			return
+		}
 	}
 
 	// Deduce additional system mount info that should be handled by the agent
@@ -875,6 +904,17 @@ func (c *Container) create() (err error) {
 		return err
 	}
 	c.process = *process
+
+	// lazy attach device after createContainer for q35
+	if machineType == QemuQ35 && len(delayAttachedDevs) > 0 {
+		c.Logger().WithFields(logrus.Fields{
+			"machine_type": machineType,
+			"devices":      delayAttachedDevs,
+		}).Info("lazy attach devices")
+		if err = c.attachDevices(delayAttachedDevs); err != nil {
+			return
+		}
+	}
 
 	if !rootless.IsRootless() && !c.sandbox.config.SandboxCgroupOnly {
 		if err = c.cgroupsCreate(); err != nil {
@@ -1359,11 +1399,15 @@ func (c *Container) removeDrive() (err error) {
 	return nil
 }
 
-func (c *Container) attachDevices() error {
+func (c *Container) attachDevices(devices []ContainerDevice) error {
 	// there's no need to do rollback when error happens,
 	// because if attachDevices fails, container creation will fail too,
 	// and rollbackFailingContainerCreation could do all the rollbacks
-	for _, dev := range c.devices {
+
+	// since devices with large bar space require delayed attachment,
+	// the devices need to be split into two lists, normalAttachedDevs and delayAttachedDevs.
+	// so c.device is not used here. See issue https://github.com/kata-containers/runtime/issues/2460.
+	for _, dev := range devices {
 		if err := c.sandbox.devManager.AttachDevice(dev.ID, c.sandbox); err != nil {
 			return err
 		}
--- a/virtcontainers/device/manager/utils.go
+++ b/virtcontainers/device/manager/utils.go
@@ -7,10 +7,16 @@
 package manager
 
 import (
+	"fmt"
+	"io/ioutil"
 	"path/filepath"
+	"strconv"
 	"strings"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/kata-containers/runtime/virtcontainers/device/config"
+	"github.com/kata-containers/runtime/virtcontainers/device/drivers"
 )
 
 const (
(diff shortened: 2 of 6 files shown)


devimc:
@Jimmy-Xu does `pci-rescan` is still needed? I added it in clear-containers because a bug in the pcie-pci-bridges (q35) that was fixed long time ago (qemu > 3.0)

Jimmy-Xu:
> @Jimmy-Xu does `pci-rescan` is still needed? I added it in clear-containers because a bug in the pcie-pci-bridges (q35) that was fixed long time ago (qemu > 3.0)

@devimc 
I tried to comment out rescanPciBus in kata-agent and this solved my problem. However, I am not sure if it will affect the hot-plugging of other types of devices, so I did not remove it.

devimc:
@Jimmy-Xu I think we can simplify the things a lot with a simple agent's option in the kernel command like, how about adding a new option (agent.no_pci_rescan) in order to don't rescan the PCI bus?

cc @bergwolf

Jimmy-Xu:
@devimc This is also a solution.

amshinde:
@Jimmy-Xu I am not in favor of adding a config option in the configuration.toml for delaying attaching devices. 
Can we instead have logic in the runtime itself to detect {devices that are hotplugged using pcie native hotplug}/{devices that have a large configuration space} and instead attach those devices in a delayed fashion. 
We typically want to have configuration options for supporting diff configurations to the user, I dont think that is the case here.

Also, @devimc pointed out about not rescanning PCI bus again. Here is the background for why the rescan was added in the first place:
https://github.com/kata-containers/agent/pull/380#issuecomment-423634195

According to that comment and the one following it, we do not need to rescan the PCI bus at all for `pc` machine type. (@devimc We can probably fix that right away for  `pc` machine). The rescan was added to support attaching devices in case of `q35` on a pci-bridge to account for the 5 sec SHPC delay so that we do not have to wait for 5 sec for a device to be detected by the agent. It is certainly not required for devices that are attached using native pcie hotplug. 
I dont think we hotplug all devices on pcie ports in case of q35, I think today it is a mix of pci-bridges and pcie ports. So until we get rid of attaching any devices on pci-bridges in case of q35, I dont think we can get rid of the pci bus rescan in case of `q35`.

lifupan:
Hi @devimc @amshinde,

In summary, we couldn't remove the rescan for q35 to avoid the 5 sec SHPC delay. 
But for those pci devices need much more io spaces shouldn't be rescaned, since the rescan
couldn't reallocate enough IO space for it, and it must depend on the Hotplug mechanism(pcie native hotplug or SHPC). Thus, we should delay those devices hotplug until the "rescan" action had done.  

I think @Jimmy-Xu 's way is a good choice: dividing those device into two classes: one was didn't need big io space and  could be rescaned, we could hotplug them before the "rescan" action, and the other class is those devices need a big io spaces and depened on the hotplug 
mechanism, do the hotplug after the "rescan" action done.

Of course, it's better to do this classification automatically, but I could't figure out an easy way to recognize whether the specific device needs a big IO  space or not.

devimc:
@lifupan 

> In summary, we couldn't remove the rescan for q35 to avoid the 5 sec SHPC delay.

does kata-agent wait 5 seconds for the device before creating the container?  otherwise the workload has to wait 5 seconds until the device is visible in the container

lifupan:
> @lifupan
> 
> > In summary, we couldn't remove the rescan for q35 to avoid the 5 sec SHPC delay.
> 
> does kata-agent wait 5 seconds for the device before creating the container? otherwise the workload has to wait 5 seconds until the device is visible in the container

Hi, @devimc  , the answer is no. That's why we shouldn't remove the "rescan" action.

Jimmy-Xu:
@amshinde 

`pci rescan` is used to ensure that certain devices are connected when the container is created, it will discover all devices connected to the VM via QMP.
`shpchp / pciehp` is required to hot-plug PCIe devices with large BAR space, with a delay of 5 seconds when the guest operating system starts.

By default, the `pci rescan` runs before `shpchp / pciehp`, which prevents proper hot-plugging of PCIe devices.

Since the `pci rescan` cannot be removed, and the Large BAR space devices require `shpchp / pciehp`. The **solution** might be to run `pci rescan` after `shpchp / pciehp` woks, or attach the large BAR space devices after creating the container.

For the **first solution**, the kata-agent needs to check when shpchp / pciehp works, but
Not all solutions require shpchp / pciehp, so kata-agent also needs to distinguish whether shpchp / pciehp is required.

For the **second solution**, the kata shim can identify large BAR space devices and delay attachment.

**In this PR**:
I identified all PCIe display VFIO device, then delay to attach them in q35.  
`enable_lazy_attach_device` is just a switch, which is off by default. After all, few devices have Large BAR space. 

**Another way**:
Remove the `enable_lazy_attach_device` parameter in config. Identify all Large BAR space devices, then delay to attach them by default in q35.  

Maybe there is a way to identify Large BAR space device
(run the following command line on the host):
```
$ lspci -nn  | grep -i nvidia
04:00.0 3D controller [0302]: NVIDIA Corporation Device [10de:15f8] (rev a1)
84:00.0 3D controller [0302]: NVIDIA Corporation Device [10de:15f8] (rev a1)

$lspci -s 04:00.0 -vv | grep Region
	Region 0: Memory at c6000000 (32-bit, non-prefetchable) [disabled] [size=16M]
	Region 1: Memory at 383800000000 (64-bit, prefetchable) [disabled] [size=16G] <<<<<<<
	Region 3: Memory at 383c00000000 (64-bit, prefetchable) [disabled] [size=32M]

$cat /sys/bus/pci/devices/0000\:04\:00.0/resource
0x00000000c6000000 0x00000000c6ffffff 0x0000000000040200
0x0000383800000000 0x0000383bffffffff 0x000000000014220c   <<<<<<<
0x0000000000000000 0x0000000000000000 0x0000000000000000
0x0000383c00000000 0x0000383c01ffffff 0x000000000014220c
0x0000000000000000 0x0000000000000000 0x0000000000000000
0x0000000000000000 0x0000000000000000 0x0000000000000000
0x0000000000000000 0x0000000000000000 0x0000000000000000
0x0000000000000000 0x0000000000000000 0x0000000000000000
0x0000000000000000 0x0000000000000000 0x0000000000000000
0x0000000000000000 0x0000000000000000 0x0000000000000000
0x0000000000000000 0x0000000000000000 0x0000000000000000
0x0000000000000000 0x0000000000000000 0x0000000000000000
0x0000000000000000 0x0000000000000000 0x0000000000000000
```

Jimmy-Xu:
@amshinde @devimc @bergwolf @lifupan:
I updated the solution: 
detect VFIO device, if it is a device with large BAR space, delay to attach it.

virtcontainers/device/manager/utils.go
```diff
@@ -35,3 +40,70 @@ func isVFIO(hostPath string) bool {
...
+		deviceFileName := filepath.Join(iommuDevicesPath, deviceFile.Name())
+		var isLarge bool
+		switch vfioDeviceType {
+		case config.VFIODeviceNormalType:
+			if isLarge, err = isLargeBarSpace(filepath.Join(deviceFileName, "resource")); err != nil {
+				return false, err
+			}
+			return isLarge, nil
+		case config.VFIODeviceMediatedType:
+			logrus.Warnf("Check sysfs resource for VFIODeviceMediatedType is not yet supported: %v", deviceFileName)
```
devimc: `WithField` please

virtcontainers/device/manager/utils.go
```diff
@@ -35,3 +40,70 @@ func isVFIO(hostPath string) bool {
...
+		default:
+			logrus.Warnf("Incorrect tokens found while check sysfs resource: %s", deviceFileName)
```
devimc: same here

virtcontainers/device/manager/utils.go
```diff
@@ -35,3 +40,70 @@ func isVFIO(hostPath string) bool {
...
+		size := end - start + 1
+		sIdx := 0
+		for i := range suffix {
+			if size/1024 < 1 {
+				break
+			}
+			size /= 1024
+			sIdx = i + 1
+		}
+		logrus.WithField("resource", resourcePath).Infof("Region %d: start=%s(%d) end=%s(%d) size:%d%s",
```
devimc: `WithFields` this can look better

virtcontainers/device/manager/utils.go
```diff
@@ -35,3 +40,70 @@ func isVFIO(hostPath string) bool {
...
+	return false, nil
+}
+
+func isLargeBarSpace(resourcePath string) (bool, error) {
+	buf, err := ioutil.ReadFile(resourcePath)
+	if err != nil {
+		return false, fmt.Errorf("failed to read sysfs resource: %v", err)
+	}
+	suffix := []string{"", "K", "M", "G", "T"}
+	for rIdx, line := range strings.Split(string(buf), "\n") {
```
devimc: is possible to include an example (how the line looks like)

virtcontainers/device/manager/utils.go
```diff
@@ -35,3 +40,70 @@ func isVFIO(hostPath string) bool {
...
+	}
+	// Pass all devices in iommu group
+	for _, deviceFile := range deviceFiles {
+		tokens := strings.Split(deviceFile.Name(), ":")
+		vfioDeviceType := config.VFIODeviceErrorType
+		if len(tokens) == 3 {
+			vfioDeviceType = config.VFIODeviceNormalType
+		} else {
+			tokens = strings.Split(deviceFile.Name(), "-")
+			if len(tokens) == 5 {
```
devimc: why 3? why 5?  please include an example

virtcontainers/device/manager/utils.go
Jimmy-Xu: Need a formatted output here:)

virtcontainers/device/manager/utils.go
Jimmy-Xu: OK

virtcontainers/device/manager/utils.go
Jimmy-Xu: OK

virtcontainers/device/manager/utils.go
Jimmy-Xu: Added.

virtcontainers/device/manager/utils.go
Jimmy-Xu: Added.

virtcontainers/device/manager/utils.go
```diff
@@ -35,3 +40,92 @@ func isVFIO(hostPath string) bool {
...
+				"device-file":        deviceFile.Name(),
+				"device-type":        vfioDeviceType,
+				"is-large-bar-space": isLarge,
+			}).Info("Detect large bar space device")
+			return isLarge, nil
+		case config.VFIODeviceMediatedType:
+			logrus.WithFields(logrus.Fields{
+				"device-file": deviceFile.Name(),
+				"device-type": vfioDeviceType,
+			}).Warn("Detect large bar space device is not yet supported for VFIODeviceMediatedType")
```
amshinde: @Jimmy-Xu Any reason why this is not supported for mediated devices?

virtcontainers/device/manager/utils.go
```diff
@@ -35,3 +40,92 @@ func isVFIO(hostPath string) bool {
...
+	}
+	// Pass all devices in iommu group
+	for _, deviceFile := range deviceFiles {
+		//For example, 0000:04:00.0
+		tokens := strings.Split(deviceFile.Name(), ":")
+		vfioDeviceType := config.VFIODeviceErrorType
+		if len(tokens) == 3 {
+			vfioDeviceType = config.VFIODeviceNormalType
+		} else {
+			//For example, 83b8f4f2-509f-382f-3c1e-e6bfe0fa1001
```
amshinde: Do mediated devices expected to have a UUID here?

virtcontainers/device/manager/utils.go
```diff
@@ -35,3 +40,92 @@ func isVFIO(hostPath string) bool {
...
+			tokens = strings.Split(deviceFile.Name(), "-")
+			if len(tokens) == 5 {
+				vfioDeviceType = config.VFIODeviceMediatedType
+			}
```
amshinde: Should be not error out or return here ?

virtcontainers/device/manager/utils.go
Jimmy-Xu: I refer here https://github.com/kata-containers/runtime/blob/master/virtcontainers/device/drivers/vfio.go#L194

virtcontainers/device/manager/utils.go
Jimmy-Xu: becase `case config.VFIODeviceMediatedType` is TODO.

virtcontainers/device/manager/utils.go
Jimmy-Xu: I have no test environment currently :(

virtcontainers/device/manager/utils.go
devimc: then can you use `getVFIODetails` ? :thinking:

virtcontainers/container.go
```diff
@@ -1359,11 +1399,11 @@ func (c *Container) removeDrive() (err error) {
 	return nil
 }
 
-func (c *Container) attachDevices() error {
+func (c *Container) attachDevices(devices []ContainerDevice) error {
 	// there's no need to do rollback when error happens,
 	// because if attachDevices fails, container creation will fail too,
 	// and rollbackFailingContainerCreation could do all the rollbacks
-	for _, dev := range c.devices {
+	for _, dev := range devices {
```
jodh-intel: Might be worth adding a comment explaining why `c.devices` isn't being used as this looks slightly odd (since there are two sets of devices).

virtcontainers/device/manager/utils.go
```diff
@@ -35,3 +40,92 @@ func isVFIO(hostPath string) bool {
 func isBlock(devInfo config.DeviceInfo) bool {
 	return devInfo.DevType == "b"
 }
+
+func IsVFIOLargeBarSpaceDevice(hostPath string) (bool, error) {
+	if !isVFIO(hostPath) {
+		return false, nil
+	}
```
jodh-intel: Nit: Can you add some blank lines between blocks to allow the code to "breathe"? It looks rather cramped like this.

virtcontainers/device/manager/utils.go
jodh-intel: ooi why is that? Is something post-processing this log message?

virtcontainers/device/manager/utils.go
```diff
@@ -35,3 +40,92 @@ func isVFIO(hostPath string) bool {
...
+	// $ cat /sys/bus/pci/devices/0000:04:00.0/resource
+	// 0x00000000c6000000 0x00000000c6ffffff 0x0000000000040200
+	// 0x0000383800000000 0x0000383bffffffff 0x000000000014220c
+	// Refer:
+	// resource format: https://github.com/torvalds/linux/blob/63623fd44972d1ed2bfb6e0fb631dfcf547fd1e7/drivers/pci/pci-sysfs.c#L145
+	// calculate size : https://github.com/pciutils/pciutils/blob/61ecc14a327de030336f1ff3fea9c7e7e55a90ca/lspci.c#L388
+	suffix := []string{"", "K", "M", "G", "T"}
+	for rIdx, line := range strings.Split(string(buf), "\n") {
+		cols := strings.Fields(line)
+		start, _ := strconv.ParseUint(cols[0][2:], 16, 64)
```
jodh-intel: You could simplify these to:

```go
start, _ := strconv.ParseUint(cols[0], 0, 64)
end, _ := strconv.ParseUint(cols[1], 0, 64)
```

virtcontainers/device/manager/utils.go
Jimmy-Xu: improved

virtcontainers/container.go
Jimmy-Xu: added.

virtcontainers/device/manager/utils.go
Jimmy-Xu: ok

virtcontainers/device/manager/utils.go
Jimmy-Xu: updated

virtcontainers/device/manager/utils.go
Jimmy-Xu: updated.

Issue #2462, opened 2020-02-17
Use "virtiofs" instead of "virtio_fs" for the virtio-fs module


fidencio:
# Description of problem

In the early stages of virtio-fs development, its module was called "virtio_fs". However, it's changed to "virtiofs" and that's the name used upstream.

Whenever we're trying to use a new enogh kernel, we'd look for virtio_fs instead of "virtiofs" causing a failure to load the module.

# Expected result

Work with the new enough kernels.

# Actual result

Doesn't work with new enough kernels.

NOTE: It's important to note that I'm not exactly familiar with the kata project with regards to when the kernel used by the project would / should be updated in order to reflect the changes I'm asking for in this bug report.

dagrh:
I think this is the right way to go. My only question for @grahamwhaley is whether we need to step the kernel version kata is using for ci firs/at the same time.

grahamwhaley:
@dagrh good point about the CIs. This is 'host kernel', yes? @chavafg - I'm going to guess we don't have control over all the CI host kernel versions...
How about we search for the new/real name first (`virtiofs`), and if that fails, then we search for the old name as a backup (`virtio_fs`) - gives us both a safety net for the CI systems, and also should allow any current or future users with older kernels to continue working?
/cc @devimc

dagrh:
No, this is guest kernel.

c3d:
Even if it's the guest kernel, I think you should, at least for a transition period, be ready for both.

c3d:
The last commit that refers to `virtio_fs` seems to be this one:
https://github.com/torvalds/linux/commit/112e72373d1f60f1e4558d0a7f0de5da39a1224d

Dated October 11, 2019. I think it's safe to proceed as if there is only `virtiofs` to consider now, given that there was a really short window where `virtio_fs` was useful, and it always required some sort of bleeding edge kernel and qemu to make it work.

fidencio:
So, I'd personally prefer to:
- Have a 1.11.0 release at some point soon;
- From 1.12.0 onwards, require 5.4 kernel.

If we really have to deal with compatibility with both virtiofs and virtio_fs, we may end up having to adapt the code in several different places with no much gain.

Does the suggestion make sense?

amshinde:
@fidencio We had been on the 4.19 longterm kernel for a while, and recently updated to LTS kernel version of 5.14.15. Related PR :https://github.com/kata-containers/runtime/pull/2422

[...]

PR #2463, opened 2020-02-17
agent: Use "virtiofs" instead of "virtio_fs"

fidencio:
virtio_fs was the name used for the module in the very early stages of
its development.

Resolves: #2462

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -61,7 +61,7 @@ var (
 	mountGuestTag               = "kataShared"
 	defaultKataGuestSandboxDir  = "/run/kata-containers/sandbox/"
 	type9pFs                    = "9p"
-	typeVirtioFS                = "virtio_fs"
+	typeVirtioFS                = "virtiofs"
 	typeVirtioFSNoCache         = "none"
 	kata9pDevType               = "9p"
 	kataMmioBlkDevType          = "mmioblk"


bergwolf:
@jodh-intel @grahamwhaley How do we test virtio-fs right now? The upstream kernel fs type has been `virtiofs` since the code is merged (See https://github.com/torvalds/linux/commit/a62a8ef9d97da23762a588592c8b8eb50a8deb6a#diff-0d441a1f46622acafe70edfb889020b1R1161). Are we testing an old version of virtio-fs guest kernel?

@fidencio Thanks for the patch! I think it is the right thing to do and should go in to kata stable branch as well, to match the expected behavior for the upstream kernel instead of a development one. BTW, I think you should reference https://github.com/torvalds/linux/commit/a62a8ef9d97da23762a588592c8b8eb50a8deb6a#diff-0d441a1f46622acafe70edfb889020b1R1161 instead because the `typeVirtioFS` is passed to mount not `modprobe`.

bergwolf:
We should fix `virtio-fs` packaging and CI too (possibly along with the virtio-fs kernel we are testing against).

chavafg:
currently we use the kernel from the viritiofs repository
https://github.com/kata-containers/runtime/blob/master/versions.yaml#L160-L163

bergwolf:
We just discussed this in today's arch committee meeting and @dagrh will check back with a new version of virtio-fs in the development tree that matches most of the upstream code plus dax support. Then kata can move forward to the new development version and make dax an optional feature so that a user with upstream kernel can also use virtio-fs w/o dax.

bergwolf:
Maybe we should just test the upstream virtio-fs by default? @kata-containers/architecture-committee

chavafg:
closing this PR as changes were included in #2604, which is already merged

Issue #2464, opened 2020-02-17
Provide a way to configure sharedDirVirtioFSOptions and sharedDirVirtioFSDaxOptions


fidencio:
**Which feature do you think can be improved?**

There's absolutely no need to use any specific option in sharedDirVirtioFSOptions and sharedDirVirtioFSDaxOptions.

**How can it be improved?**

Knowing that, the default should be an empty string with the possibility for the admin to change that in the configuration.toml file.

Currently, this piece can only be changed by changing these bits of code and re-building the project:
https://github.com/kata-containers/runtime/blob/master/virtcontainers/kata_agent.go#L74 and https://github.com/kata-containers/runtime/blob/master/virtcontainers/kata_agent.go#L75

PR #2473, opened 2020-02-19
sharedDirVirtioFS*Options changes / cleanups

fidencio:
sharedDirVirtioFSOptions has, by default, the following values: "default_permissions,allow_other,rootmode=040000,user_id=0,group_id=0", "nodev"

Although those values made sense in the earlier stages of development, the default options got baked into virtiofs.ko when it got merged.

Keeping those options as they ar would case the following issue:
```
[root@f32 runtime]# podman run --security-opt label=disable  --runtime=/usr/local/bin/kata-runtime --rm -id fedora sh
Error: rpc error: code = Internal desc = Could not mount kataShared to /run/kata-containers/shared/containers/: invalid argument: OCI runtime error
```

sharedDirVirtioFSOptions has, by default, the following value "dax". Although dax will be re-introduced soon, right now it just causes the very same issue described above:
```
[root@f32 runtime]# podman run --security-opt label=disable  --runtime=/usr/local/bin/kata-runtime --rm -id fedora sh
Error: rpc error: code = Internal desc = Could not mount kataShared to /run/kata-containers/shared/containers/: invalid argument: OCI runtime error
```

IMHO, the best way to work this around would be to **not** set the "dax" option when `virtio_fs_cache_size = 0`.

Those two fixes are related to #2464, but don't exactly go along with the conversation I've had with @devimc on #kata-dev. There we talked about:
- deprecate virtio_fs_extra_args;
- add virtio_fs_args (documenting what's already set as default);

The reasons I didn't take the suggested path are:
- virtio_fs_extra_args still makes sense, as those would be options passed to the virtiofsd daemon;
- add virtio_fs_args still makes sense, but  avoiding setting dax whenever `virtio_fs_cache = 0` also does, which makes me want to postpone the addition of virtio_fs_args till we explicitly need it.

Does everything make sense?

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -868,7 +868,10 @@ func setupStorages(sandbox *Sandbox) []*grpc.Storage {
 			// options should not contain 'dax' lest the virtio-fs daemon crashing
 			// with an invalid address reference.
 			if sandbox.config.HypervisorConfig.VirtioFSCache != typeVirtioFSNoCache {
-				sharedDirVirtioFSOptions = append(sharedDirVirtioFSOptions, sharedDirVirtioFSDaxOptions)
+				// If virtio_fs_cache_size = 0, dax should not be used.
+				if sandbox.config.HypervisorConfig.VirtioFSCacheSize != 0 {
+					sharedDirVirtioFSOptions = append(sharedDirVirtioFSOptions, sharedDirVirtioFSDaxOptions)
+				}
 			}
 			sharedVolume := &grpc.Storage{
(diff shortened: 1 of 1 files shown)


fidencio:
One more comment, this situation is quite similar to #2462 and, if merged, could be agreed on merging together with the changes proposed on #2463 and https://github.com/kata-containers/agent/pull/738

jcvenegas:
virtiofs job failed but the log is not longer available, restarting jobs

chavafg:
these changes are also included in #2604, so closing this PR

PR #2475, opened 2020-02-19
vc: do not follow symlink when umounting contanier host path

### LINKED ISSUE #2474: harden container hostpath cleanup
# Description of problem

A container's host path might be changed by the guest to point to some other places by placing a symlink there. kata runtime should not follow link when unmounting them otherwise we might end up unmounting some other mountpoints unexpectedly.

bergwolf:
So that if a guest changes it, we do not end up
propagating the error.

Fixes: #2474

### DIFF
--- a/virtcontainers/mount.go
+++ b/virtcontainers/mount.go
@@ -331,11 +331,24 @@ type Mount struct {
 	BlockDeviceID string
 }
 
+func isSymlink(path string) bool {
+	stat, err := os.Stat(path)
+	if err != nil {
+		return false
+	}
+	return stat.Mode()&os.ModeSymlink != 0
+}
+
 func bindUnmountContainerRootfs(ctx context.Context, sharedDir, sandboxID, cID string) error {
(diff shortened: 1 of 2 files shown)


devimc (approved):

lifupan (approved):

virtcontainers/mount.go
```diff
@@ -277,6 +281,10 @@ func bindMount(ctx context.Context, source, destination string, readonly bool) e
 		return fmt.Errorf("Could not create destination mount point %v: %v", destination, err)
 	}
 
+	if dest, err := filepath.EvalSymlinks(destination); err == nil && !strings.HasPrefix(dest, filepath.Clean(topDir)) {
+		return fmt.Errorf("Invalid mount destination %s must not escape", destination)
+	}
+
```
laijs: there is a "TOCTOU" here

bergwolf: I dropped it as it also breaks vm template.

Issue #2478, opened 2020-02-19
runtime fail to build if host has additional LDFLAGS


jcvenegas:
# Description of problem

Recently was added support to STRIP binaries and by default it adds LDFLAGS, if the host has flags unknown  golang linker it will fail 

Example:

```
[  149s]      BUILD    /usr/src/packages/BUILD/kata-runtime
[  149s] flag provided but not defined: -Wl,-Bsymbolic-functions
[  149s] usage: go build [-o output] [-i] [build flags] [packages]
[  149s] Run 'go help build' for details.
[  149s] make[2]: *** [Makefile:535: /usr/src/packages/BUILD/kata-runtime] Error 2
[  149s] make[2]: Leaving directory '/usr/src/packages/BUILD`
```
# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

PR #2479, opened 2020-02-19
makefile: do not use LDFLAGS for extra kata flags.

jcvenegas:
Some flags defined by the host may not be compatible with golang,
not use LDFLAGS but use our own variable.

Fixes: #2478


amshinde:
@grahamwhaley ptal

Issue #2483, opened 2020-02-20
let virtiofsd and qemu run in parallel to speed up startup time of kata


jongwu:
**Which feature do you think can be improved?**

virtiofsd

**How can it be improved?**

let virtiofsd run in parallel with qemu

**Additional Information**

tests show, it will speed up startup time of kata by dozens to one hundred milliseconds.

**Before raising this feature request**

Have you looked at the [limitations document](https://github.com/kata-containers/documentation/blob/master/Limitations.md)?
@jodh-intel @grahamwhaley @bergwolf

PR #2484, opened 2020-02-20
let virtiofsd and qemu run in parallel

jongwu:
Currently, qemu will be launched after virtiofsd start if virtiofs enabled,
which is not necessary. Both of virtiofsd and qemu only depend on unix
socket file. so we can run virtiofsd and qemu in parallel after socket
file settles down.
tests show it will speed up startup time of kata around dozens to one
hundred milliseconds.

Fixes: #2483

@jodh-intel @grahamwhaley @bergwolf @Pennyzct

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -625,13 +625,12 @@ func (q *qemu) virtiofsdArgs(fd uintptr) []string {
 	return args
 }
 
-func (q *qemu) setupVirtiofsd() (err error) {
+func (q *qemu) setupVirtiofsdSock() (fd *os.File, err error) {
 	var listener *net.UnixListener
-	var fd *os.File
 
 	sockPath, err := q.vhostFSSocketPath(q.id)
 	if err != nil {
-		return err
+		return fd, err
 	}
 
 	listener, err = net.ListenUnix("unix", &net.UnixAddr{
@@ -639,16 +638,17 @@ func (q *qemu) setupVirtiofsd() (err error) {
 		Net:  "unix",
 	})
 	if err != nil {
-		return err
+		return fd, err
 	}
 	listener.SetUnlinkOnClose(false)
 
 	fd, err = listener.File()
 	listener.Close() // no longer needed since fd is a dup
 	listener = nil
-	if err != nil {
-		return err
-	}
+	return fd, err
+}
+
+func (q *qemu) setupVirtiofsd(fd *os.File) (err error) {
 
 	const sockFd = 3 // Cmd.ExtraFiles[] fds are numbered starting from 3
 	cmd := exec.Command(q.config.VirtioFSDaemon, q.virtiofsdArgs(sockFd)...)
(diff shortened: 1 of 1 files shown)


grahamwhaley:
nitpick. Your commit has the title `subsystem: let virtiofsd and qemu run in parallel.` - you are meant to replace `subsystem` with the name of the subsystem the patch affects, in this case, probably something like `vc: qemu: let virtiofsd and qemu run in parallel.` would work.

jongwu:
@grahamwhaley  oh! I have misunderstood the subsystem. thanks.

stefanha:
I'm surprised that this change makes a measurable difference since startup was already parallelized.  cmd.Start() does not wait after fork+exec (https://golang.org/pkg/os/exec/#Cmd.Start) and therefore QEMU will already be launched while virtiofsd is starting up.

Also, does moving the code into goroutines guarantee parallelism?  I thought Go has an M:N thread model and requires explicit configuration to make goroutines always run in separate threads.

I don't object to this patch but I don't understand why it's faster and whether it will be faster on all systems.  I would suggest investigating the performance difference using tracing (e.g. perf record -a -e sched:\* -e syscalls:\*) to understand why there is a difference.  Then we'll know whether the improvement is accidental or really something that can be relied on.

jongwu:
Hi @stefanha , yeah, in most cases, without this patch, virtiofsd has nearly run in parallel with qemu. But it is not the case of the first kata runs in the system or you clean cache(including inode, page...) before run kata. Loading virtiofsd binary will consume dozens to more than a hundred of milliseconds in different systems. This patch will work well in that case. If there is cache in the system, this patch can also speed up a few milliseconds in my tests in different system.

stefanha:
> Hi @stefanha , yeah, in most cases, without this patch, virtiofsd has nearly run in parallel with qemu. But it is not the case of the first kata runs in the system or you clean cache(including inode, page...) before run kata. Loading virtiofsd binary will consume dozens to more than a hundred of milliseconds in different systems. This patch will work well in that case. If there is cache in the system, this patch can also speed up a few milliseconds in my tests in different system.

This performance improvement seems like an accident - a side-effect of how your kernel, Go implementation, etc work today - and not something that is guaranteed work in the future or on anyone else's machine.

The kata-runtime process already forked and was able to continue running before this patch.  The exec syscall, which is where the cold page cache slows things down, is not invoked from the parent kata-runtime process, so it's unclear to me why this patch should make a difference.

Also, I mentioned that goroutines are not mapped to threads 1:1, so this patch is not guaranteed to add parallelism.  It depends on the Go implementation and configuration.  If threading is necessary then it should be done explicitly.

Can you explain why the patch improves performance?

jongwu:
I agree with you @stefanha. Before this patch virtiofsd and qemu have started in different processes and run in parallelism. This patch will not  improve it again.
If I find nothing special after further tests, I will close this PR.

stefanha:
> I agree with you @stefanha. Before this patch virtiofsd and qemu have started in different processes and run in parallelism. This patch will not improve it again.
> If I find nothing special after further tests, I will close this PR.

@jongwu I hope I did not discourage you.  The startup time improvement you measured was real and would be nice to have!  Now it's a question of figuring out why and how to make it work every time.

jongwu:
> > I agree with you @stefanha. Before this patch virtiofsd and qemu have started in different processes and run in parallelism. This patch will not improve it again.
> > If I find nothing special after further tests, I will close this PR.
> 
> @jongwu I hope I did not discourage you. The startup time improvement you measured was real and would be nice to have! Now it's a question of figuring out why and how to make it work every time.

Very kind of you @stefanha , I will do further research before next step. Thanks for your kindness and help!

jongwu:
Hi @stefanha, after lots of elaborate test, maybe I can explain how this patch work.
Originally,  virtiofsd startup func use vfork() to fork a child process and then exec, which may block the parent process, so that may lead to a fake parallelism with qemu.
This patch may cancel this issue and make virtiofsd and qemu run in parallelism virtually.
This patch will work more efficient when start of kata in the first time, means with no cache available and with low speed io device.  
WDYT?

jongwu:
> Hi @stefanha, after lots of elaborate test, maybe I can explain how this patch work.
> Originally, virtiofsd startup func use vfork() to fork a child process and then exec, which may block the parent process, so that may lead to a fake parallelism with qemu.
> This patch may cancel this issue and make virtiofsd and qemu run in parallelism virtually.
> This patch will work more efficient when start of kata in the first time, means with no cache available and with low speed io device.
> WDYT?

@stefanha , just ignore this rush conclusion.  I have no decided conclusion now.

jongwu:
@stefanha ,  the delay exist in preparing config for qemu before launchCustomqemu func, I can't explain very clear why this delay exist, maybe because of gc or schedule but no relationship with the intention of this PR.
optimization for that delay relates to govmm/qemu so I will close this PR.

PR #2487, opened 2020-02-20
clh: update CLH to stable/v0.5.x

likebreath:
Use CLH branch stable/v0.5.x, and also re-generate the openAPI client
code with the new 'cloud-hypervisor.yaml'.

Fixes: #2488


virtcontainers/clh.go
```diff
@@ -120,6 +120,7 @@ var clhKernelParams = []Param{
 	{"no_timer_check", ""}, // do not check broken timer IRQ resources
 	{"noreplace-smp", ""},  // do not replace SMP instructions
 	{"agent.log_vport", fmt.Sprintf("%d", vSockLogsPort)}, // tell the agent where to send the logs
+	{"rootflags", "data=ordered,errors=remount -ro ro"},   // mount the root filesystem as readonly
```
sameo: typo: `remount -ro` should be `remount-ro`.

likebreath: @sameo Thanks for correcting me. Just being updated.

likebreath: @sameo Just separate the two commits. Any other comments are appreciated. Thanks.

virtcontainers/pkg/cloud-hypervisor/client/docs/DiskConfig.md
```diff
@@ -5,9 +5,13 @@
 Name | Type | Description | Notes
 ------------ | ------------- | ------------- | -------------
 **Path** | **string** |  | 
+**Readonly** | **bool** |  | [optional] [default to true]
```
sboeuf: I thought we agreed this shouldn't default to true.

likebreath: @sboeuf I believe the `default to true` comes from the `cloud-hypervsior.yaml` [here](https://github.com/sameo/cloud-hypervisor-1/blob/76dc26b0d1715b254680cd715273ab682ef437ed/vmm/src/api/openapi/cloud-hypervisor.yaml#L279). Note that all the code under `client/` here are auto-generated based on the yaml file.

sboeuf: Yes this needs to be fixed from CH directly.

amshinde:
@likebreath Travis is failing with `ERROR: No "Fixes" found`. Can you open an issue and address it with `Fixes :` in the commit message.

likebreath:
> @likebreath Travis is failing with `ERROR: No "Fixes" found`. Can you open an issue and address it with `Fixes :` in the commit message.

@amshinde Thanks for pointing this out. Yes, I am looking at the same failure. I am wondering do I need to put `Fixes xxx` in both commits or just the first commit? Thanks.

grahamwhaley:
You need a Fixes in at least one commit. You can have multiple different ones if you need across multiple commits (but generally we go with one-PR fixes one-Issue, generally)

virtcontainers/pkg/cloud-hypervisor/client/docs/DiskConfig.md
sameo: My bad, this is fixed now. @likebreath You probaly want to fix accordingly.

Issue #2501, opened 2020-03-02
I would like to contribute SELinux support to kata-runtime.


rhatdan:
Basically I want to launch the KVM launcher tool (qemu?) with an SELinux label, similar to what we do with libvirt.
Currently when I use kata with podman, it complains if we specify a label that kata does not support SELinux labels.  What I would like to do is have kata just use this label to apply to the KVM launcher.  Then I will work to generate a new policy type (container_kvm_t) that will allow the KVM Launcher tool to do its thing, but prevent breakout.

The problem I have is I don't know where to put the SELinux code changes into the kata source code.  The SELinux code has to be executed as close to the execution of the KVM Launcher as possible.  I could spend a lot of time trying to figure out the kata code, but I would prefer a quick pointer to where this should be added.

devimc:
@rhatdan I think you are looking for this function:

https://github.com/kata-containers/runtime/blob/6bed2a724d3944f2226c9abd03c0b8718dab4be1/virtcontainers/sandbox.go#L993

PR #2503, opened 2020-03-02
Add SELinux support for running VM Confinement

rhatdan:
We want to launch the KVM launcher tool (qemu?) with an SELinux label, similar
to what we do with libvirt.

Currently when I use kata with Podman, it complains if we specify a label that
kata does not support SELinux labels. What I would like to do is have kata just
use this label to apply to the KVM launcher. Then I will work to generate a new
policy type (container_kvm_t) that will allow the KVM Launcher tool to do its
thing, but prevent breakout.

Fixes: https://github.com/kata-containers/runtime/issues/2501

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

### DIFF
--- a/hack/selinux_tag.sh
+++ b/hack/selinux_tag.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+#
+# Copyright 2020 Red Hat Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+pkg-config libselinux 2> /dev/null && echo selinux
--- a/virtcontainers/hypervisor.go
+++ b/virtcontainers/hypervisor.go
@@ -406,6 +406,9 @@ type HypervisorConfig struct {
 	// VMid is the id of the VM that create the hypervisor if the VM is created by the factory.
 	// VMid is "" if the hypervisor is not created by the factory.
 	VMid string
+
+	// SELinux label for the VM
+	SELinuxProcessLabel string
 }
 
 // vcpu mapping from vcpu number to thread number
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -1025,9 +1025,10 @@ func (k *kataAgent) constraintGRPCSpec(grpcSpec *grpc.Spec, passSeccomp bool) {
 		grpcSpec.Linux.Seccomp = nil
 	}
 
-	// Disable selinux
+	// Disable SELinux inside of the virtual machine, the label will apply
+	// to the KVM process
 	if grpcSpec.Process.SelinuxLabel != "" {
-		k.Logger().Warn("Selinux label specified in config, but not supported in Kata yet, running container without selinux")
+		k.Logger().Info("SELinux label from config will be applied to the hypervisor process, not the VM workload")
 		grpcSpec.Process.SelinuxLabel = ""
 	}
 
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -25,6 +25,7 @@ import (
 	"unsafe"
 
 	govmmQemu "github.com/intel/govmm/qemu"
+	"github.com/opencontainers/selinux/go-selinux/label"
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -793,6 +794,15 @@ func (q *qemu) startSandbox(timeout int) error {
 			}
 		}
 	}()
+
+	// This needs to be done as late as possible, just before launching
+	// virtiofsd are executed by kata-runtime after this call, run with
+	// the SELinux label. If these processes require privileged, we do
+	// notwant to run them under confinement.
+	if err := label.SetProcessLabel(q.config.SELinuxProcessLabel); err != nil {
+		return err
+	}
+	defer label.SetProcessLabel("")
 
 	if q.config.SharedFS == config.VirtioFS {
 		err = q.setupVirtiofsd()
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -498,6 +498,18 @@
   revision = "a1b50f621a48ad13f8f696a162f684a241307db0"
 
 [[projects]]
+  digest = "1:fe18b198e372b89e9d1868a2ac24741360ed3e44367e3c96665bf2ba3d78c8d3"
+  name = "github.com/opencontainers/selinux"
+  packages = [
+    "go-selinux",
+    "go-selinux/label",
+    "pkg/pwalk",
+  ]
+  pruneopts = "NUT"
+  revision = "069a5a99e1e5e46c9be197aa5ca6388b041aa4c7"
+  version = "v1.4.0"
+
+[[projects]]
   digest = "1:7da29c22bcc5c2ffb308324377dc00b5084650348c2799e573ed226d8cc9faf0"
   name = "github.com/opentracing/opentracing-go"
   packages = [
@@ -798,6 +810,7 @@
     "github.com/opencontainers/runc/libcontainer/system",
     "github.com/opencontainers/runc/libcontainer/utils",
     "github.com/opencontainers/runtime-spec/specs-go",
+    "github.com/opencontainers/selinux/go-selinux/label",
     "github.com/opentracing/opentracing-go",
     "github.com/opentracing/opentracing-go/log",
     "github.com/pkg/errors",
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -98,3 +98,7 @@
 [[constraint]]
   name = "github.com/blang/semver"
   version = "3.6.1"
+
+[[constraint]]
+  name = "github.com/opencontainers/selinux"
+  version = "1.3.3"
--- a/Makefile
+++ b/Makefile
@@ -464,8 +464,11 @@ QUIET_GENERATE = $(Q:@=@echo    '     GENERATE '$@;)
 QUIET_INST     = $(Q:@=@echo    '     INSTALL  '$@;)
 QUIET_TEST     = $(Q:@=@echo    '     TEST     '$@;)
 
+SELINUXTAG := $(shell ./hack/selinux_tag.sh)
+BUILDTAGS := --tags "$(SELINUXTAG)"
+
 # go build common flags
-BUILDFLAGS := -buildmode=pie
+BUILDFLAGS := -buildmode=pie ${BUILDTAGS}
 
 # whether stipping the binary
 ifeq ($(STRIP),yes)
--- a/vendor/github.com/opencontainers/selinux/LICENSE
+++ b/vendor/github.com/opencontainers/selinux/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
+      communication on electronic mailing lists, source code control systems,
+      and issue tracking systems that are managed by, or on behalf of, the
+      Licensor for the purpose of discussing and improving the Work, but
+      excluding communication that is conspicuously marked or otherwise
+      designated in writing by the copyright owner as "Not a Contribution."
+
+      "Contributor" shall mean Licensor and any individual or Legal Entity
+      on behalf of whom a Contribution has been received by Licensor and
+      subsequently incorporated within the Work.
+
+   2. Grant of Copyright License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      copyright license to reproduce, prepare Derivative Works of,
+      publicly display, publicly perform, sublicense, and distribute the
+      Work and such Derivative Works in Source or Object form.
+
+   3. Grant of Patent License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      (except as stated in this section) patent license to make, have made,
+      use, offer to sell, sell, import, and otherwise transfer the Work,
+      where such license applies only to those patent claims licensable
+      by such Contributor that are necessarily infringed by their
+      Contribution(s) alone or by combination of their Contribution(s)
+      with the Work to which such Contribution(s) was submitted. If You
+      institute patent litigation against any entity (including a
+      cross-claim or counterclaim in a lawsuit) alleging that the Work
+      or a Contribution incorporated within the Work constitutes direct
+      or contributory patent infringement, then any patent licenses
+      granted to You under this License for that Work shall terminate
+      as of the date such litigation is filed.
+
+   4. Redistribution. You may reproduce and distribute copies of the
+      Work or Derivative Works thereof in any medium, with or without
+      modifications, and in Source or Object form, provided that You
+      meet the following conditions:
+
+      (a) You must give any other recipients of the Work or
+          Derivative Works a copy of this License; and
+
+      (b) You must cause any modified files to carry prominent notices
+          stating that You changed the files; and
+
+      (c) You must retain, in the Source form of any Derivative Works
+          that You distribute, all copyright, patent, trademark, and
+          attribution notices from the Source form of the Work,
+          excluding those notices that do not pertain to any part of
+          the Derivative Works; and
+
+      (d) If the Work includes a "NOTICE" text file as part of its
+          distribution, then any Derivative Works that You distribute must
+          include a readable copy of the attribution notices contained
+          within such NOTICE file, excluding those notices that do not
+          pertain to any part of the Derivative Works, in at least one
+          of the following places: within a NOTICE text file distributed
+          as part of the Derivative Works; within the Source form or
+          documentation, if provided along with the Derivative Works; or,
+          within a display generated by the Derivative Works, if and
+          wherever such third-party notices normally appear. The contents
+          of the NOTICE file are for informational purposes only and
+          do not modify the License. You may add Your own attribution
+          notices within Derivative Works that You distribute, alongside
+          or as an addendum to the NOTICE text from the Work, provided
+          that such additional attribution notices cannot be construed
+          as modifying the License.
+
+      You may add Your own copyright statement to Your modifications and
+      may provide additional or different license terms and conditions
+      for use, reproduction, or distribution of Your modifications, or
+      for any such Derivative Works as a whole, provided Your use,
+      reproduction, and distribution of the Work otherwise complies with
+      the conditions stated in this License.
+
+   5. Submission of Contributions. Unless You explicitly state otherwise,
+      any Contribution intentionally submitted for inclusion in the Work
+      by You to the Licensor shall be under the terms and conditions of
+      this License, without any additional terms or conditions.
+      Notwithstanding the above, nothing herein shall supersede or modify
+      the terms of any separate license agreement you may have executed
+      with Licensor regarding such Contributions.
+
+   6. Trademarks. This License does not grant permission to use the trade
+      names, trademarks, service marks, or product names of the Licensor,
+      except as required for reasonable and customary use in describing the
+      origin of the Work and reproducing the content of the NOTICE file.
+
+   7. Disclaimer of Warranty. Unless required by applicable law or
+      agreed to in writing, Licensor provides the Work (and each
+      Contributor provides its Contributions) on an "AS IS" BASIS,
+      WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+      implied, including, without limitation, any warranties or conditions
+      of TITLE, NON-INFRINGEMENT, MERCHANTABILITY, or FITNESS FOR A
+      PARTICULAR PURPOSE. You are solely responsible for determining the
+      appropriateness of using or redistributing the Work and assume any
+      risks associated with Your exercise of permissions under this License.
+
+   8. Limitation of Liability. In no event and under no legal theory,
+      whether in tort (including negligence), contract, or otherwise,
+      unless required by applicable law (such as deliberate and grossly
+      negligent acts) or agreed to in writing, shall any Contributor be
+      liable to You for damages, including any direct, indirect, special,
+      incidental, or consequential damages of any character arising as a
+      result of this License or out of the use or inability to use the
+      Work (including but not limited to damages for loss of goodwill,
+      work stoppage, computer failure or malfunction, or any and all
+      other commercial damages or losses), even if such Contributor
+      has been advised of the possibility of such damages.
+
+   9. Accepting Warranty or Additional Liability. While redistributing
+      the Work or Derivative Works thereof, You may choose to offer,
+      and charge a fee for, acceptance of support, warranty, indemnity,
+      or other liability obligations and/or rights consistent with this
+      License. However, in accepting such obligations, You may act only
+      on Your own behalf and on Your sole responsibility, not on behalf
+      of any other Contributor, and only if You agree to indemnify,
+      defend, and hold each Contributor harmless for any liability
+      incurred by, or claims asserted against, such Contributor by reason
+      of your accepting any such warranty or additional liability.
+
+   END OF TERMS AND CONDITIONS
+
+   APPENDIX: How to apply the Apache License to your work.
+
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "{}"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright {yyyy} {name of copyright owner}
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
--- a/vendor/github.com/opencontainers/selinux/go-selinux/label/label.go
+++ b/vendor/github.com/opencontainers/selinux/go-selinux/label/label.go
@@ -1,0 +1,77 @@
+package label
+
+import (
+	"github.com/opencontainers/selinux/go-selinux"
+)
+
+// Deprecated: use selinux.ROFileLabel
+var ROMountLabel = selinux.ROFileLabel
+
+// SetProcessLabel takes a process label and tells the kernel to assign the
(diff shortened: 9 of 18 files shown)


rhatdan:
@devimc PTAL
This is totally untested, but wanted to make sure I am going in the right direction.

rhatdan:
I am not sure about Cloud Hyper-visor or acrn?
Theoretically SELinux should work with firecracker and qemu.

virtcontainers/acrn.go
```diff
@@ -17,6 +17,7 @@ import (
 	"time"
 	"unsafe"
 
+	"github.com/opencontainers/selinux/go-selinux/label"
```
devimc: please run  `dep ensure` and verdor this package in a different commit

virtcontainers/acrn.go
```diff
@@ -444,6 +445,11 @@ func (a *Acrn) startSandbox(timeoutSecs int) error {
 		}
 	}()
 
+	if err := label.SetProcessLabel(a.config.ProcessLabel); err != nil {
```
devimc: I wonder if ACRN needs selinux...
@vijaydhanraj  please take a look,

virtcontainers/kata_agent.go
```diff
@@ -1015,12 +1015,6 @@ func (k *kataAgent) constraintGRPCSpec(grpcSpec *grpc.Spec, passSeccomp bool) {
 		grpcSpec.Linux.Seccomp = nil
 	}
 
-	// Disable selinux
```
amshinde: @rhatdan Is this tested? Without this, the kata-agent applies the SelinuxLabel passed in the grpcSpec inside the guest, and since we dont have selinux suuport inside the guest, this was failing earlier.

virtcontainers/kata_agent.go
```diff
@@ -1305,6 +1299,8 @@ func (k *kataAgent) createContainer(sandbox *Sandbox, c *Container) (p *Process,
 		return nil, err
 	}
 
+	c.sandbox.config.HypervisorConfig.ProcessLabel = grpcSpec.Process.SelinuxLabel
```
devimc: this will not work, `k.createContainer` is called after `startSandbox`

virtcontainers/qemu.go
```diff
@@ -779,6 +780,11 @@ func (q *qemu) startSandbox(timeout int) error {
 		}
 	}
 
+	if err := label.SetProcessLabel(q.config.ProcessLabel); err != nil {
```
devimc: I'm curious, why not do this in `startVM()` ?

virtcontainers/kata_agent.go
devimc: @amshinde what kind of `selinux support` is missing?

amshinde: @devimc Without this we were seeing this failure earlier with Kata `selinux label is specified in config, but selinux is disabled or not supported`, which essentially comes from https://github.com/opencontainers/runc/blob/4769cdf607b0d1cbd73e5407d9655da7511fe193/libcontainer/configs/validate/validator.go#L103

The cause of the failure is this : https://github.com/opencontainers/selinux/blob/b485497fbee5ba8d2d3e8f3bfea0d5f9f2d40e86/go-selinux/selinux_linux.go#L195

Its a failure in detecting an selinuxfs mount point.  
I think we would need to first enable selinux configs in kata guest kernel.

grahamwhaley:
> I am not sure about Cloud Hyper-visor or acrn?
> Theoretically SELinux should work with firecracker and qemu.

/cc @sboeuf @sameo for any Cloud Hypervisor selinux input and thoughts.

virtcontainers/acrn.go
vijaydhanraj: Hi @devimc, I don't think selinux applies to acrn. Isn't this a linux kernel feature? or I am missing something? 
If the guest wants to enable or disable, we can pass it through the guest kernel parameters. But other than that not sure how it would apply for acrn hypervisor.

rhatdan: Ok so arcn is not an executable that runs the VM?

virtcontainers/kata_agent.go
rhatdan: Yes we should not pass the selinux label into the guest kernel.  For now, we only care about the process running the VM not used inside of the VM.

virtcontainers/qemu.go
rhatdan: This needs to be done as late as possible, since all processes that are executed by kata-runtime after this call, will now run them with the SELinux label.  If those are fairly privileged processes, we do not want to run them under confinement or we end up giving a lot more privs to the KVMLauncher, then is necessary.

sboeuf:
>/cc @sboeuf @sameo for any Cloud Hypervisor selinux input and thoughts.

/cc @rbradford

virtcontainers/acrn.go
vijaydhanraj: acrn-dm is the hypervisor device model that runs in the SOS (Service OS) and is responsible for creating guest UOS (user OS) VMs.

virtcontainers/kata_agent.go
rhatdan: Where should I call this?  I need to get the label out of the OCI spec before the VM is launched.

virtcontainers/kata_agent.go
rhatdan: Ok updated patch to remove SELinux label from oci spec before running container inside of VM.

virtcontainers/kata_agent.go
rhatdan: Figured it out now setting the ProcessLabel inside of startSandbox.

rhatdan:
Ok I am trying to test, but can not get kata to work right now on Fedora 32, I have switched to cgroup V1 in case this is a problem.

```
# make install prefix=/usr
# podman run --security-opt label=type:container_lvm_t --runtime=/usr/bin/kata-runtime --rm -id fedora sh
/usr/share/defaults/kata-containers/configuration-qemu.toml: having both an image and an initrd defined in the configuration file is not supported
ERRO[0001] Error removing container e06ba5cc663cc8de0f355d7c2a417ffcc39dd2eeb8082a3e9e5768d129ac2f4d from runtime after creation failed 
Error: /usr/share/defaults/kata-containers/configuration-qemu.toml: having both an image and an initrd defined in the configuration file is not supported: OCI runtime error
```
Comment out the initrd
```
# podman run --security-opt label=type:container_lvm_t --runtime=/usr/bin/kata-runtime --rm -id fedora sh
Error: Failed to check if grpc server is working: rpc error: code = Unavailable desc = transport is closing: OCI runtime error
```
Comment out the image, and use initrd
```
# podman run --security-opt label=type:container_lvm_t --runtime=/usr/bin/kata-runtime --rm -id fedora sh
Error: rpc error: code = Internal desc = Could not run process: container_linux.go:346: starting container process caused "process_linux.go:297: applying cgroup configuration for process caused \"mountpoint for cgroup not found\"": OCI runtime error
```

devimc:
@rhatdan for cgroups v2 you need to enable `sandbox_cgroup_only` in the configuration file

virtcontainers/kata_agent.go
```diff
@@ -761,6 +761,10 @@ func (k *kataAgent) getDNS(sandbox *Sandbox) ([]string, error) {
 func (k *kataAgent) startSandbox(sandbox *Sandbox) error {
 	span, _ := k.trace("startSandbox")
 	defer span.Finish()
+	ociSpec := sandbox.GetPatchedOCISpec()
```
devimc: I think this may not work. `k.startSandbox` is called after `h.startSandbox`. I still think that this should be in `startVM()`

https://github.com/kata-containers/runtime/blob/83eef430bd9eaa3d313bb3c1e73bdcbe4fbf23cc/virtcontainers/sandbox.go#L1016-L1050

virtcontainers/kata_agent.go
amshinde: Thanks @rhatdan 
We do need to confine the container process inside the VM using selinux eventually, but this can be done later.
What I see is that we need two selinux labels, one for the hypervisor and one for the container process itself.

In case of runc, the selinux label passed in the OCI spec is applied to the container process itself. With this PR, we have a different behaviour as this label is applied to the hypervisor process instead, we can document this for now.
But to have expected behaviour, it looks like we need to have support for an additional selinux label for vm based runtimes in the spec.

sboeuf:
@grahamwhaley @rhatdan 
>any Cloud Hypervisor selinux input and thoughts.

Well the same way it should work for QEMU and Firecracker, I don't see any issue to make this work with Cloud-Hypervisor.
It's not clear to me whose responsibility it is to apply SELinux policies, but Cloud-Hypervisor is a simple process running multiple threads for its virtio devices, so I don't think there's any blocker there.
I would like to understand the expectation better as I'd like to make sure nothing's missing in Cloud-Hypervisor to support SELinux.

virtcontainers/kata_agent.go
rhatdan: Using SELinux inside of the VM requires the VM to load SELinux policy.  We could potentially have a stripped down policy for containers within a VM, or another option would be to use the policy from the host system within the VM.  Anyways, yes we would need to create a new selinux type for the VM, or just use the types for the sandbox to apply to the VM and the types for each container to apply within the VM.  Bottom line we are a long way from this kind of support at this point.

jodh-intel: I'm still slightly unclear on this. If we're applying the generic SELinux policy to the hypervisor process (qemu, fc, etc), is that policy the same as that used for `runc`? What if we need to add support for a new hypervisor that needs additional SELinux privs beyond what a `runc` process would need?

- Would those privs be added to the generic policy (which would effectively give *too many* privs to non-VM containers like `runc`)?
- Would we need to create a new policy for all hypervisors at that point in time (so that *all* the VM-based containers had the extra privs, but non-VM containers didn't?)
- Would we need to create a new policy for that specific hypervisor at that point in time (so that *only* that hypervisor got the extra privs?

I think we should create some documentation on what type of SELinux support is being provided and what isn't. The question is where best to put it :smile: I suspect the [`docs`](https://github.com/kata-containers/documentation) repo might be the best place with a link to that from this repos top-level `README.md`.

rhatdan: > I'm still slightly unclear on this. If we're applying the generic SELinux policy to the hypervisor process (qemu, fc, etc), is that policy the same as that used for `runc`? What if we need to add support for a new hypervisor that needs additional SELinux privs beyond what a `runc` process would need?
It will be different then what runc provides.  Runc needs ability to work with the host kernel and certain permissions, but in general can not work with /dev/kvm for instance.  When running a Virt Launching tool like qemu or Kata we will have to give it very different access.  But the bottom line from a security point of view is to guarantee separation between containers and the host.  We want to make sure that Container A can not interfere with Container B, and that Container A can not attack the host os in places like /var, /home, /root, /etc ...

The overlap between the policies will be that both can read/write/execute container_file_t content, and they will be separated by MCS Separation.  We need to make sure also that a runc/crun container/pod can not attack a KVM container/pod or vice versa.  

But bottom line, they both do very different things.
> 
>     * Would those privs be added to the generic policy (which would effectively give _too many_ privs to non-VM containers like `runc`)?
> 
Yes, I don't want runc containers to have the added allow rules of a KVM launcher, Similarly I would like to remove rules from a KVM Launcher that I might need to allow to runc/crun containers/pods.
>     * Would we need to create a new policy for all hypervisors at that point in time (so that _all_ the VM-based containers had the extra privs, but non-VM containers didn't?)
Yes
> 
>     * Would we need to create a new policy for that specific hypervisor at that point in time (so that _only_ that hypervisor got the extra privs?
> 
No, I will attempt to group all the hypervisors in the same policy.  Only issue would be if a particular hypervisor needed a grossly different and powerful permission to run.
> 
> I think we should create some documentation on what type of SELinux support is being provided and what isn't. The question is where best to put it smile I suspect the [`docs`](https://github.com/kata-containers/documentation) repo might be the best place with a link to that from this repos top-level `README.md`.

We could build out the github.com/containers/selinux README. which is where this would be best described.

virtcontainers/qemu.go
rhatdan: Done

virtcontainers/hypervisor.go
```diff
@@ -399,6 +399,9 @@ type HypervisorConfig struct {
 	// VMid is the id of the VM that create the hypervisor if the VM is created by the factory.
 	// VMid is "" if the hypervisor is not created by the factory.
 	VMid string
+
+	// SELinux label for the VM
```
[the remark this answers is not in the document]
rhatdan: Fixed

virtcontainers/kata_agent.go
```diff
@@ -1015,9 +1015,11 @@ func (k *kataAgent) constraintGRPCSpec(grpcSpec *grpc.Spec, passSeccomp bool) {
 		grpcSpec.Linux.Seccomp = nil
 	}
 
-	// Disable selinux
+	// Disable SELinux inside of the virtual machine, the label will apply
+	// to the KVM process
 	if grpcSpec.Process.SelinuxLabel != "" {
-		k.Logger().Warn("Selinux label specified in config, but not supported in Kata yet, running container without selinux")
+		k.Logger().Info("SELinux label specified in config, will be applied to the virtual machine")
+		k.Logger().Info("No SELinux label be applied inside the virtual machine")
```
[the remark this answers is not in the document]
rhatdan: Fixed

virtcontainers/hypervisor.go
```diff
@@ -399,6 +399,9 @@ type HypervisorConfig struct {
...
+	ProcessLabel string
```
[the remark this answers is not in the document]
rhatdan: Currently only one LSM is supported at a time, and AppArmor is not label based.

We originally called this ProcessLabel, so that a different labeling LSM like Smack could be used and would use the same infrastructure.  Over time this has not proven to be important, and slowly SELinux has snuck it's way into the tool.  This is why the commands are label rather then just calling it SELinux directly.

fidencio:
I've run a bunch of tests on this series, together with @rhatdan.

It does work for Fedora as expected. For instance, running:
```

File: hack/selinux_tag.sh [hunk 10]
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
fidencio: CI is breaking due to missing license here:
```
ERROR: Required license identifier ('SPDX-License-Identifier: Apache-2.0') missing from following files:
hack/selinux_tag.sh
The command ".ci/static-checks.sh" failed and exited with 1 during .
```

fidencio:
@rhatdan, @devimc already fixed my `Signed-off-by: ...` (sorry for the trouble).

Seems that (hopefully) the only missing piece is the CI error related to missing the lincese in the newly added `hack/selinux_tag.sh` file.

rhatdan:
@sboeuf @devimc @grahamwhaley @jodh-intel @amshinde PTAL

fidencio:
\test

jodh-intel:
Thanks @fidencio (and welcome! ;)! You actually need to use a forward slash to trigger the tests (see https://github.com/kata-containers/community/wiki/Controlling-the-CI)

/test

fidencio:
I've just tested the latest version of the patch and it does work as expected.

Right now, distros which want to have SELinux support enabled, will have to depend on `libselinux-devel` in order to have the support automatically enabled (this is more like a note for the future me, working on adding this to Fedora).

Without `libselinux-devel` installed, we get the processes running using the following label:
```
[root@rawhide fidencio]# ps -eZ | grep -E 'qemu|virtiofsd'
unconfined_u:system_r:container_runtime_t:s0 23673 ? 00:00:00 virtiofsd
unconfined_u:system_r:container_runtime_t:s0 23681 ? 00:00:05 qemu-system-x86
unconfined_u:system_r:container_runtime_t:s0 23684 ? 00:00:00 virtiofsd
```

With `libselinux-devel` installed, we get the processes running using the new container_kvm_t label:
```
[root@rawhide fidencio]# ps -eZ | grep -E 'qemu|virtiofsd'
system_u:system_r:container_kvm_t:s0:c573,c878 26016 ? 00:00:00 virtiofsd
system_u:system_r:container_kvm_t:s0:c573,c878 26024 ? 00:00:05 qemu-system-x86
system_u:system_r:container_kvm_t:s0:c573,c878 26027 ? 00:00:00 virtiofsd
```

The command-line used to start podman is:
```
[root@rawhide fidencio]# podman --runtime=/usr/local/bin/kata-runtime run --security-opt label=type:container_kvm_t -it fedora /bin/bash
```

So, +1 from me from the functional perspective.

There's one nit, though, that I'd like to have fixed before having this merged:
`Signed-off-by: fidencio@redhat.com <fidencio@redhat.com> ` is wrong and it's my own fault. Would be nice if whoever merges this patch, or even Dan Walsh, ends up fixing that, changing the current content to `Signed-off-by: Fabiano Fidêncio <fidencio@redhat.com>`

virtcontainers/kata_agent.go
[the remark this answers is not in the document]
jodh-intel: I'm still seeing two log calls here? Have you pushed your latest branch?

hack/selinux_tag.sh
```diff
@@ -0,0 +1,8 @@
+#
+# Copyright 2020 Red Hat Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+##!/usr/bin/env bash
```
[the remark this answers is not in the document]
rhatdan: Oops, added the copyright and extra # 
Fixed

hack/selinux_tag.sh
```diff
@@ -0,0 +1,8 @@
...
+if pkg-config libselinux 2> /dev/null ; then
```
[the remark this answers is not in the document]
rhatdan: Fixed

jodh-intel:
Adding dnm label to allow the incorrect commit sign off to be amended...

fidencio:
@jodh-intel, just to clearly understand the situation, **all** the CI jobs must pass before having this PR merged, is it? If not, is there some easy way to get the CI status with the current master so we could at least compare the failures?

I'll start to dig into the remaining failures during the day and will keep updating the status in this very same comment here.

For now, I can see that this is not related at all to this PR:
- **jenkins-ci-rhel-7:** This is not related to the patch set. I've been in contact with @GabyCT about this one, more details here (https://github.com/kata-containers/ci/issues/261#issuecomment-606848214)

jodh-intel:
@fidencio - yes, that's right. Merging is disabled until we get a good CI run.

You can see a summary of all the CI jobs here:

- http://jenkins.katacontainers.io/

I think we may have disabled some of the CI runs as we don't have the capacity to run every combination (if you can help out with that please let us know ;)

Note that if you have suitable (development or test) systems, you can run the CI locally:

- https://github.com/kata-containers/tests#run-the-tests

/cc @chavafg.

fidencio:
@jodh-intel,
I'd say that based on the failures we have, none of those are related to this PR.

@rhatdan,
As I cannot force-push to this branch, could you fix my `Signed-off-by: ...` in your commit? It should be `Signed-off-by: Fabiano Fidêncio <fidencio@redhat.com>`.

Once the Signed-off-by is fixed, I'd say we can go ahead and have this one merged.

amshinde:
Thanks @fidencio.  I am going to restart the CI for the failing jobs.
@chavafg @GabyCT, Noting the failures seen here before I restart. We need to take a look to see if these are known flakiness and see if we can avoid these.

For centos-q35,  I am seeing this: 
```
Create cni directories /opt/cni/bin and /etc/cni/net.d
Create bundles in /tmp/jenkins/workspace/kata-containers-runtime-centos-7-q-35-PR/go/src/github.com/kata-containers/runtime/virtcontainers/utils/supportfiles/bundles
Using default tag: latest
latest: Pulling from library/busybox
0669b0daf1fb: Pulling fs layer
Performing Post build task...
Match found for :.* : True
```
Looks like, the failure is caused due to a network issue, perhaps we should fix this by checking if error caused is due to network, and a couple of retries in out test script, so that we do not have to restart entire job.

For Power-8-ubuntu job, this is the failure seen in the ginkgo tests:
```
state container 
  with workload [true], timeWait 5
  /home/jenkins/workspace/kata-containers-runtime-Power8-ubuntu-18-04-PR-initrd/go/src/github.com/kata-containers/tests/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:43
Running command '/usr/local/bin/kata-runtime [kata-runtime --log=/tmp/bundle347234458/log run --bundle=/tmp/bundle347234458 --console= --pid-file=/tmp/bundle347234458/pid --detach P6NutYZJpe3tNooi94BK]'
[kata-runtime --log=/tmp/bundle347234458/log run --bundle=/tmp/bundle347234458 --console= --pid-file=/tmp/bundle347234458/pid --detach P6NutYZJpe3tNooi94BK]
Timeout: 120 seconds
Exit Code: 1
Stdout: 
Stderr: rpc error: code = Unknown desc = rootfs (/run/kata-containers/shared/containers/P6NutYZJpe3tNooi94BK/rootfs) does not exist

Running command '/usr/local/bin/kata-runtime [kata-runtime list --quiet]'
[kata-runtime list --quiet]
Timeout: 120 seconds
Exit Code: 0
Stdout: 
Stderr: 
```

The test cannot find the bundle, again lets look at if this is failure in downloading the bundle.

For Rhel-7:
```
--> Finished Dependency Resolution
Error: Package: moreutils-0.49-2.el7.x86_64 (epel)
           Requires: perl(IPC::Run)
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
Failed at 22: sudo -E yum install -y moreutils
Performing Post build task...
Match found for :.* : True
```
@GabyCT Is this related to https://github.com/kata-containers/ci/issues/261?

Finally for ARM:
```
/workspace/kata-containers-runtime-ARM-18.04-PR/go/src/github.com/golangci/golangci-lint ~/workspace/kata-containers-runtime-ARM-18.04-PR/go/src/github.com/kata-containers/runtime
12:03:55 INFO: Using github.com/golangci/golangci-lint version v1.16.0
12:03:56 Switched to a new branch 'v1.16.0'
12:03:56 INFO: Building github.com/golangci/golangci-lint
12:03:56 go build -o golangci-lint ./cmd/golangci-lint
12:04:05 go: github.com/go-critic/go-critic@v0.0.0-20181204210945-ee9bf5809ead: invalid pseudo-version: does not match version-control timestamp (2019-02-10T22:04:43Z)
12:04:05 Makefile:8: recipe for target 'build' failed
12:04:05 make: *** [build] Error 1
```
@Pennyzct Is this a known issue as well?

fidencio:
> Thanks @fidencio. I am going to restart the CI for the failing jobs.
> @chavafg @GabyCT, Noting the failures seen here before I restart. We need to take a look to see if these are known flakiness and see if we can avoid these.

@amshinde, seems you missed my comment here https://github.com/kata-containers/runtime/pull/2503#issuecomment-607110831

> 
> For centos-q35, I am seeing this:
> 
> ```
> Create cni directories /opt/cni/bin and /etc/cni/net.d
> Create bundles in /tmp/jenkins/workspace/kata-containers-runtime-centos-7-q-35-PR/go/src/github.com/kata-containers/runtime/virtcontainers/utils/supportfiles/bundles
> Using default tag: latest
> latest: Pulling from library/busybox
> 0669b0daf1fb: Pulling fs layer
> Build step 'Execute shell' marked build as failure
> Performing Post build task...
> Match found for :.* : True
> ```
> 
> Looks like, the failure is caused due to a network issue, perhaps we should fix this by checking if error caused is due to network, and a couple of retries in out test script, so that we do not have to restart entire job.

Agreed!

> 
> For Power-8-ubuntu job, this is the failure seen in the ginkgo tests:
> 
> ```
> state container 
>   with workload [true], timeWait 5
>   /home/jenkins/workspace/kata-containers-runtime-Power8-ubuntu-18-04-PR-initrd/go/src/github.com/kata-containers/tests/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:43
> Running command '/usr/local/bin/kata-runtime [kata-runtime --log=/tmp/bundle347234458/log run --bundle=/tmp/bundle347234458 --console= --pid-file=/tmp/bundle347234458/pid --detach P6NutYZJpe3tNooi94BK]'
> command failed error 'exit status 1'
> [kata-runtime --log=/tmp/bundle347234458/log run --bundle=/tmp/bundle347234458 --console= --pid-file=/tmp/bundle347234458/pid --detach P6NutYZJpe3tNooi94BK]
> Timeout: 120 seconds
> Exit Code: 1
> Stdout: 
> Stderr: rpc error: code = Unknown desc = rootfs (/run/kata-containers/shared/containers/P6NutYZJpe3tNooi94BK/rootfs) does not exist
> 
> Running command '/usr/local/bin/kata-runtime [kata-runtime list --quiet]'
> [kata-runtime list --quiet]
> Timeout: 120 seconds
> Exit Code: 0
> Stdout: 
> Stderr: 
> ```
> 
> The test cannot find the bundle, again lets look at if this is failure in downloading the bundle.

It's related to: https://github.com/kata-containers/ci/issues/263

> 
> For Rhel-7:
> 
> ```
> --> Finished Dependency Resolution
> Error: Package: moreutils-0.49-2.el7.x86_64 (epel)
>            Requires: perl(IPC::Run)
>  You could try using --skip-broken to work around the problem
>  You could try running: rpm -Va --nofiles --nodigest
> Failed at 22: sudo -E yum install -y moreutils
> Build step 'Execute shell' marked build as failure
> Performing Post build task...
> Match found for :.* : True
> ```
> 
> @GabyCT Is this related to [kata-containers/ci#261](https://github.com/kata-containers/ci/issues/261)?

Yes, it's related to https://github.com/kata-containers/ci/issues/261
And I'm working together with @GabyCT in order to get that fixed.

> 
> Finally for ARM:
> 
> ```
> /workspace/kata-containers-runtime-ARM-18.04-PR/go/src/github.com/golangci/golangci-lint ~/workspace/kata-containers-runtime-ARM-18.04-PR/go/src/github.com/kata-containers/runtime
> 12:03:55 INFO: Using github.com/golangci/golangci-lint version v1.16.0
> 12:03:56 Switched to a new branch 'v1.16.0'
> 12:03:56 INFO: Building github.com/golangci/golangci-lint
> 12:03:56 go build -o golangci-lint ./cmd/golangci-lint
> 12:04:05 go: github.com/go-critic/go-critic@v0.0.0-20181204210945-ee9bf5809ead: invalid pseudo-version: does not match version-control timestamp (2019-02-10T22:04:43Z)
> 12:04:05 Makefile:8: recipe for target 'build' failed
> 12:04:05 make: *** [build] Error 1
> ```
> 
> @Pennyzct Is this a known issue as well?

It's related to https://github.com/kata-containers/tests/issues/2434

rhatdan:
@fidencio Repushed with fixed signature.

jodh-intel:
Thanks @rhatdan.

/test

jodh-intel:
Restarted Ubuntu CI (timeout) and Fedora CI (tracing test anomoly I think - known issue).

fidencio:
@jodh-intel, seems the only two issues we have are the ones already reported here[0] and here[1].
For some reason, travis-ci seems stuck since Yesterday.

Are we good to go and have this one merged?

0: https://github.com/kata-containers/ci/issues/263
1: https://github.com/kata-containers/tests/issues/2434

chavafg:
travis already passed, but not reporting back, I think it is safe to merge.
https://travis-ci.org/github/kata-containers/runtime/builds/670142930?utm_medium=notification&utm_source=github_status

amshinde:
@fidencio @rhatdan Thanks for the work done on this PR. Merged :)

jodh-intel:
\o/ - thanks again @rhatdan !

andreabolognani:
@rhatdan This commit causes container startup to fail on my Fedora 31 machine with SELinux in enforcing mode:

```
$ docker run --rm -it alpine true
docker: Error response from daemon: OCI runtime create failed: failed to launch qemu: exit status 1, error messages from qemu log: qemu-system-x86_64: cannot create PID file: Cannot open pid file: No such file or directory: unknown.
```

The AVCs are:

```
Apr 16 15:56:56 kata audit[23788]: AVC avc:  denied  { read write } for  pid=23788 comm="qemu-system-x86" path="/dev/vhost-net" dev="devtmpfs" ino=24766 scontext=system_u:system_r:container_t:s0:c99,c363 tcontext=system_u:object_r:vhost_device_t:s0 tclass=chr_file permissive=0
Apr 16 15:56:56 kata audit[23788]: AVC avc:  denied  { read write } for  pid=23788 comm="qemu-system-x86" path="/dev/net/tun" dev="devtmpfs" ino=15375 scontext=system_u:system_r:container_t:s0:c99,c363 tcontext=system_u:object_r:tun_tap_device_t:s0 tclass=chr_file permissive=0
```

If I revert the commit, or flip SELinux to permissive, everything works; alternatively, I can pass `--security-opt label=disable` as was necessary before https://github.com/kata-containers/runtime/pull/2443 got merged.

I think what's needed to make it work properly is a SELinux policy update, but I'm not sure whether the corresponding changes have been merged upstream - they certainly haven't made it to Fedora 31.

I'm wondering if it's a good idea to have this in 11.1 at all: the out-of-the-box experience is not any worse than it was before https://github.com/kata-containers/runtime/pull/2443, but after that PR things had gotten better and now we're back to the previous status quo. Once the SELinux policy updates have trickled down to distributions we definitely want this, but perhaps it's >11.1 material and should be pulled out of this release?

fidencio:
> Once the SELinux policy updates have trickled down to distributions we definitely want this, but perhaps it's >11.1 material and should be pulled out of this release?

Personally I don't agree with the suggestion as 1.11.0 is not Fedora 31 material.
In any case, I do believe this is something that woud fit better in the specific distro bugzilla.

andreabolognani:
> Personally I don't agree with the suggestion as 1.11.0 is not Fedora 31 material.
> In any case, I do believe this is something that woud fit better in the specific distro bugzilla.

Whether or the Fedora project will update Kata to 1.11 in Fedora 31 (probably not) is a different question than whether the Kata projects cares about Kata 1.11 working on Fedora 31.

I'm not sure what the exact rules when it comes to target platforms are, and one could certainly make the case that running Kata containers without putting SELinux into permissive mode or passing `--security-opt label=disable` was not working in 1.10 either, so there's no regression.

I just though it was a good idea to bring up the subject, since I've run into this issue myself and there's a chance other people will as well :)

rhatdan:
container-selinux should be available on Fedora now to handle this, we don't have a version of podman that can handle this yet.  I think for F31 we should disable the SELinux support for now.

Users can do 
podman run --security-opt label=type:container_kvm_t ...
And this will work also, for those still using docker.
docker run --security-opt label=type:container_kvm_t ...

rhatdan:
Should we add a configuration option to kata to ignore the label?

fidencio:
@rhatdan, can we make that if building **without** passing `+selinux` we use the old behaviour?

andreabolognani:
> @rhatdan, can we make that if building **without** passing `+selinux` we use the old behaviour?

If I understand correctly the way the build system works, there is currently no way to configure whether or not SELinux labels are used: the `selinux` build tag is used internally by https://github.com/opencontainers/selinux to know whether libselinux is available on the host, and I guess there will be some stub functions in that library that will be compiled in if it's not available?

So you can't just do the equivalent of `./configure --disable-selinux` on Fedora 31 and `./configure --enable-selinux` on Fedora 32; even if that was the case, that would work for packagers but for people building from source you'd want the correct value to be detected automatically, and that's another can of worms.

Maybe I'm just worrying too much: as I mentioned earlier, even with 1.10 you had to make SELinux permissive or disable labeling for Kata to work, so this is not any worse, and in fact it works better on Fedora 32 and other targets that have a new enough SELinux policy installed. Perhaps the best way to handle this is to do nothing at all :)

andreabolognani:
> So you can't just do the equivalent of `./configure --disable-selinux` on Fedora 31 and `./configure --enable-selinux` on Fedora 32; even if that was the case, that would work for packagers but for people building from source you'd want the correct value to be detected automatically, and that's another can of worms.

I've addressed the first part with https://github.com/kata-containers/runtime/pull/2624.

This will not give Fedora 31 users the perfect out-of-the-box experience, but it will at least provide a way for them to pick a different behavior than the default, and it will also help packagers ensure that SELinux support is either enabled or disabled based on what's more appropriate for the target.

Issue #2495, opened 2020-02-24
clh: Add support CPU hotplug


jcvenegas:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

**Before raising this feature request**

Have you looked at the [limitations document](https://github.com/kata-containers/documentation/blob/master/Limitations.md)?

PR #2509, opened 2020-03-04
clh: enable CPU hotplug

likebreath:
With the HTTP API 'vm.resize()', the CPU hotplug with CLH is much simpler
comparing with QEMU. This is because we don't need to distinguish adding from
removing CPUs.

Fixes: #2495

Note: hotplug with CLH requires a patch to the guest kernel (as noted from [CLH's docs](https://github.com/cloud-hypervisor/cloud-hypervisor/blob/master/docs/hotplug.md#kernel-support)), which is missing in kata's guest kernel for now. We will need to introduce [this patch](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/patch/drivers/acpi/Makefile?id=ac36d37e943635fc072e9d4f47e40a48fbcdb3f0) to our kata packaging first, before we can merge this PR.

### DIFF
--- a/virtcontainers/clh.go
+++ b/virtcontainers/clh.go
@@ -76,6 +76,7 @@ type clhClient interface {
 	// No lint: golint suggest to rename to VMInfoGet.
 	VmInfoGet(ctx context.Context) (chclient.VmInfo, *http.Response, error) //nolint:golint
 	BootVM(ctx context.Context) (*http.Response, error)
+	VmResizePut(ctx context.Context, vmResize chclient.VmResize) (*http.Response, error)
 }
 
 type CloudHypervisorVersion struct {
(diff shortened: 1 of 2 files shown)


jcvenegas:
Looks good :)  seems that you'll need to update the mock testing  in travis.

I think the CPU tests are skiped for clh you can raise a PR remove the skip for the ones you want to be added and use 
`Depends-on: github.com/kata-containers/tests#PR_NUMBER`  in the commit message so CI can test before even merge. 
Same for your PR for kernel changes.

virtcontainers/clh.go
```diff
@@ -391,8 +392,42 @@ func (clh *cloudHypervisor) resizeMemory(reqMemMB uint32, memoryBlockSizeMB uint
...
+	cl := clh.client()
+
+	// Retrieve the number of current vCPUs via HTTP API
+	ctx, cancel := context.WithTimeout(context.Background(), clhAPITimeout*time.Second)
+	info, _, err := cl.VmInfoGet(ctx)
+	if err != nil {
+		clh.Logger().WithField("function", "resizeVCPUs").WithError(openAPIClientError(err)).Info("[clh] VmInfoGet failed")
+		return 0, 0, openAPIClientError(err)
+	}
+	cancel()
```
devimc: I think you should move this after `ctx, cancel := context.WithTimeout(context.Background(), clhAPITimeout*time.Second)` as `defer cancel()`

virtcontainers/clh.go
```diff
@@ -391,8 +392,42 @@ func (clh *cloudHypervisor) resizeMemory(reqMemMB uint32, memoryBlockSizeMB uint
...
+	currentVCPUs = uint32(info.Config.Cpus.BootVcpus)
+	newVCPUs = currentVCPUs
+
+	// Sanity check
+	if reqVCPUs == 0 {
+		clh.Logger().WithField("function", "resizeVCPUs").Debugf("Cannot resize vCPU to 0")
+		return currentVCPUs, newVCPUs, fmt.Errorf("Cannot resize vCPU to 0")
+	}
+	if reqVCPUs > clh.config.DefaultMaxVCPUs || reqVCPUs > uint32(info.Config.Cpus.MaxVcpus) {
+		clh.Logger().WithField("function", "resizeVCPUs").Debugf("Cannot resize vCPU to %d: exceeding the maximum amount of vCPUs (%d or %d)",
```
devimc: please use `WithFields`, not "%d"

likebreath: @devimc Thanks for the comments. Just adjusted as you advised. One question I have was: what is the right way to break the long line into multiple lines in golang? The line I changed now is even longer, and is ready hard to read. :<

devimc: this is a good example:

https://github.com/kata-containers/runtime/blob/fde6447c16721dc5fa3139872bc791ec08afea4e/virtcontainers/device/manager/utils.go#L67-L72

virtcontainers/clh.go
likebreath: @devimc What I want to achieve with this `cancel()` here is to reset the timer after the first HTTP API call (`VmInfoGet`), and use a new timer and `cancel()` call for the second HTTP API (`VmResizePut`). I was assuming with a `defer cancel()` here, the timer will be imposed for the  rest of the `resizeVCPUs()` function (not the first HTTP API call `VmResizePut`). 

Basically I want two separate timer for two HTTP API calls. Please let me know if my understanding is incorrect or there is better/nicer way to do it. Thanks a lot.

virtcontainers/clh.go
likebreath: Much easier to read now. Thanks a lot, @devimc.

virtcontainers/clh.go
devimc: ok, but please add a comment  :smile:

PR #2515, opened 2020-03-06
Support persistent memory volumes

### LINKED ISSUE #2262: support PMEM inside Kata Containers when running under Kubernetes
When using [PMEM-CSI](https://github.com/intel/pmem-csi) to manage PMEM storage, individual apps are going to have volumes created for them by Kubernetes with that storage driver and then want to use them like RAM, i.e. `mmap` a file or the entire volume. Depending on the application, `MAP_SYNC` and thus additional persistency guarantees will be needed.

Filesystem volumes currently get passed into kata-containers with 9p or virtio-fs. Neither of them supports `MAP_SYNC`. While virtio-fs supports mmap, performance is likely to be lower than native access because not all pages can be mapped at once (see below).

Block volumes are passed in as SCSI disk devices which is even worse (no mmap).

**Describe the solution you'd like**

A way to pass in the volume such that all of it can me mapped into an application's address space with `MAP_SYNC` semantic. Once that mapping exists, applications should be able to read and write bytes with native performance (= as if they weren't running under Kata Containers).

At this point, the most promising approach for achieving this seems to be to detect such special volumes and map them to QEMU objects and nvdimm device (https://github.com/qemu/qemu/blob/master/docs/nvdimm.txt). How to activate this special behavior is to be decided.

**Describe alternatives you've considered**

virtio-fs was considered, but doesn't meet all objectives because:
- It does not map all pages at once. Instead, it maintains a cache of mapped pages which is considerably smaller ("a few GB") than the available PMEM. This should lead to lower performance and unpredictable latency spikes.
- Because a page might not be currently mapped when written to, it does not meet MAP_SYNC requirements.

**Before raising this feature request**

This was discussed on freenode IRC, `#kata-dev`, on 2019-11-25:

(11:42:07 AM) pohly: stefanha: hello. I am trying to understand how (and how well) virtio-fs supports mmap. Background: I work on PMEM-CSI, a driver which enables the use of PMEM in Kubernetes. Ultimately the goal is that an application can do mmap(MAP_SYNC) and then do byte read/writes directly to the the underlying hardware. That works without kata-containers involved. I now looked at kata-containers 1.9.1 with the kata-qemu-virtiofs. I can see that this passes the dax-capable filesystem (XFS, in case that this matters) into the qemu instance with virtiofs. A test program can do mmap(MAP_SYNC) on a file.
(11:43:13 AM) pohly: But... it can also do that with 9p as file system and with the container root filesystem served by virtio-fs although that filesystem on the host does not support dax (hosted by plain SSD).
(11:45:00 AM) pohly: I was under the (perhaps mistaken) impression that virtio-fs would somehow support mmap. I though I had read that somewhere. Is that really true?
(11:46:42 AM) pohly: I checked the /proc/<pid>/maps for the /opt/kata/bin/qemu-virtiofs-system-x86_64 process that runs the pod. It doesn't have any entry for the file that currently is mapped inside the container.
(11:51:52 AM) brtknr: pohly: following this discussion
(11:57:44 AM) davidgiluk: pohly: is the mount mounted with DAX?
(11:58:25 AM) pohly: Yes: kataShared on /data type virtio_fs (rw,nodev,relatime,user_id=0,group_id=0,default_permissions,allow_other,dax)
(11:58:35 AM) pohly: That is inside qemu.
(11:59:16 AM) pohly: And also outside of it: /dev/mapper/ndbus0region0fsdax-e7660acd0fd86e6aea32589af51903654f6a4e41 on /var/lib/kubelet/pods/6576fed5-5488-4ee4-a6a2-578c5519ae9c/volumes/kubernetes.io~csi/my-csi-volume/mount type xfs (rw,relatime,attr2,dax,inode64,noquota)
(12:00:53 PM) stefanha: pohly: virtio-fs isn't intended for pmem.  QEMU won't use MAP_SYNC.
(12:01:10 PM) stefanha: pohly: If you need MAP_SYNC semantics then QEMU's nvdimm device can do that.
(12:01:37 PM) stefanha: pohly: MAP_SYNC support could be added to virtio-fs but today it doesn't do that.
(12:01:48 PM) pohly: stefanha: if virtio-fs doesn't support MAP_SYNC, shouldn't it then reject the mmap call?
(12:03:02 PM) stefanha: pohly: Probably.  Inside the guest the virtio-fs and FUSE code isn't doing anything that violates MAP_SYNC,
(12:03:19 PM) stefanha: but the problem is that the host side doesn't necessarily honor those semantics.
(12:03:48 PM) pohly: But plain mmap works?
(12:03:57 PM) stefanha: pohly: Yep, plain mmap is supported.
(12:04:13 PM) pohly: Should I then see a /proc/*/maps entry for the file? I don't have that.
(12:04:50 PM) pohly: Or am I checking the wrong process? I looked at qemu-virtiofs-system-x86_64, because that is where the code runs.
(12:05:39 PM) stefanha: pohly: There isn't necessarily a 1:1 mmap relationship between guest application mmaps and host qemu-virtiofs-system-x86_64 mmaps.
(12:05:54 PM) stefanha: pohly: What are you trying to confirm by looking at qemu-virtiofs-system-x86_64 mmaps?
(12:06:26 PM) pohly: Looking more closely I do see one entry that has at least the right size: 7f2f1bffe000-7f2f1bfff000 ---p 00000000 00:00 0
(12:06:40 PM) pohly: But it doesn't have a file name associated with it. Should it have that?
(12:07:06 PM) pohly: I am trying to verify that a file on the host has indeed been mapped into the address space of the process running inside qemu.
(12:07:55 PM) pohly: If that isn't the case, then how does mmap support work?
(12:07:56 PM) stefanha: pohly: The lack of filename could be due to file descriptor passing
(12:08:10 PM) stefanha: The file is opened by virtiofsd and passed to QEMU.  Maybe that's why no name is reported.
(12:08:15 PM) stefanha: But that's just a guess.
(12:08:30 PM) pohly: That might be it. Let me remove the mapping inside qemu...
(12:08:31 PM) davidgiluk: the name normally does show up
(12:08:53 PM) davidgiluk: pohly: Have you accessed the mmap'd area, or just done the mmap?
(12:10:20 PM) pohly: Just the mmap. So it's waiting for a page fault before doing anything on the host side? I can add that.
(12:11:12 PM) stefanha: Yes, that sounds likely.
(12:11:52 PM) stefanha: pohly: But again, if your goal is to get pmem semantics then virtio-fs in its current state doesn't guarantee that.
(12:12:15 PM) davidgiluk: pohly: Yes, I think so - remember for virtiofs we only have a fixed sized cache window, so we can't guarantee to mmap the whole region
(12:12:19 PM) stefanha: pohly: QEMU has -device nvdimm and -device virtio-pmem-pci for that.
(12:13:36 PM) pohly: Using those for a mounted filesystem in kata-containers isn't going to be easy.
(12:14:22 PM) pohly: virtio-fs looked much more promising ;-}
(12:15:00 PM) davidgiluk: stefanha: What stops us passing the MAP_SYNC all the way through?
(12:17:06 PM) pohly: davidgiluk: even if you do, "fixed size cache window" sounds like another big roadblock. PMEM comes in higher capacity than DRAM, that's partly why it is appealing for some workloads.
(12:18:10 PM) pohly: MAP_SYNC isn't even needed for all workloads. In fact, most apps currently don't depend on it.
(12:18:33 PM) pohly: So virtio-fs may already be a good step forward and sufficient.
(12:19:09 PM) pohly: OTOH, if it needs to set up and tear down mappings on the host side often, then that may affect performance.
(12:20:07 PM) pohly: memcached uses PMEM as DRAM replacement and stores its data there. Predictable access times for that data probably is important.
(12:20:41 PM) davidgiluk: pohly: Right; if you've got a single PMEM device to pass through then as stefan says using the -device stuff is the right way; if you're trying to pass through files that on the host are mountedon a filesystem that's backed by pmem, then virtiofs might be interesting
(12:22:06 PM) pohly: davidgiluk: we are trying the former. PMEM-CSI basically splits up a single PMEM device and hands out portions of it to individual apps. We cannot assume that only a single app uses that device; that would be rather limiting.
(12:22:26 PM) pohly: Ahem, I meant "we are trying the latter"...
(12:23:23 PM) davidgiluk: pohly: But does the PMEM-CSI portions look like individual block devices that you then put a filesystem on, and is that filesystem built in the host or the guest?
(12:26:10 PM) pohly: davidgiluk: it is a block device. But applications in Kubernetes typically will ask for a filesystem, so PMEM-CSI formats and mounts that device.
(12:26:26 PM) sameo left the room (quit: Ping timeout: 276 seconds).
(12:26:32 PM) pohly: And then Kubernetes passes the directory name of the mounted FS to the runtime.
(12:27:01 PM) pohly: I heard that kata-containers sometimes does tricks like then passing the device into qemu and mounting again inside.
(12:27:38 PM) pohly: That's a bit dirty, because there are two Linux kernels which both might write to the same block device.
(12:27:40 PM) davidgiluk: pohly: OK, if it's a device+filesystem just for that container then it does feel like passing that block device into the container is right rather than passing the filesystem through virtiofs
(12:28:27 PM) pohly: davidgiluk: yes, that would be the better alternative, except for the "is already mounted" part.
(12:29:49 PM) pohly: Also, does it have to be some actual device? Currently the block devices are either LVM logical volumes or PMEM namespaces (/dev/pmem*).
(12:30:16 PM) pohly: We can't use PCI device pass-through - it's not even on the PCI bus.
(12:30:32 PM) pohly: Nor do we want to pass in the entire NVDIMM.
(12:39:19 PM) davidgiluk: pohly: Does one of the chunks of a PMEM-CSI look like a pmem device? i.e. would it make sense to pass that in using -device nvdimm or virtio-pmem-pci ?
(12:40:58 PM) pohly: davidgiluk: I need to check what those options expect, but for the LVM case the answer is probably "no" - it's just a logical volume.
(12:43:14 PM) pohly: Hmm, according to https://github.com/qemu/qemu/blob/master/docs/nvdimm.txt the "mem-path" can be an ordinary file. So we could just point that at the block device.
(12:45:06 PM) pohly: But how would kata-containers even recognize that it needs to do this special handling? All it gets is a path to a mounted filesystem or a loop device (block mode, which also works in Kubernetes).
(12:46:03 PM) pohly: This sounds doable, but I fear that it will be rather hacky and I have no idea how many different components need to be adapted to make this work.
(12:47:41 PM) pohly: May I copy this discussion into an issue in https://github.com/kata-containers/runtime/issues? Is that the right tracker for "add PMEM support to kata-containers"?
(12:51:49 PM) davidgiluk: pohly: Yeh probably best to make an issue; I'm also not sure the best way to wire it through - but if it looks like a block device, and that block device is intended just for this container, then treat it as a block device and let the guest handle it
(12:52:58 PM) gwhaley: pohly: include 'devimc' on that Issue, if not already - he'll have a good idea I think of what knitting would be required.
(12:53:27 PM) gwhaley: yes, the hard bit is how to annotate that volume/mount/device to ensure it ends up mapped via the correct route. It may be that 'annotations' are the route.
(12:53:40 PM) gwhaley: oh, amshinde might have good input as well
(12:55:03 PM) gwhaley: so, historically we've always noted that nvdimm/dax could be used to pass items in (kata uses it for iirc the kernel image, or is it the rootfs....) - but, I don't believe there is a defined mechanism to set that all up via the orchestrators and runtime, and I don't think I've ever seen anybody actually using an nvdimm/dax mount/map for themselves ... yet....
(12:59:52 PM) pohly: gwhaley: /opt/kata/share/kata-containers/kata-containers-image_clearlinux_1.9.1_agent_d4bbd8007f.img is passed via "-object" + "-device nvdimm".
(01:00:47 PM) pohly: Looks like the rootfs. There's also "root=/dev/pmem0p1".
(01:00:50 PM) gwhaley: pohly: right, the rootfs for the VM (I can never remember if it is the rootfs or the kernel we do it with ;-) )... so, we use it, we know it works.... now it would be how do we enable 'users' to do it...
(01:03:11 PM) pohly: davidgiluk: to get closure on this: when actually writing into the memory mapped region via virtio-fs, I do see map entries on the host side, including the file name.
(01:03:41 PM) pohly: davidgiluk: how large is this "fixed size cache window"?
(01:04:51 PM) davidgiluk: pohly: It's configurable via an option, normally a few GB
(01:05:08 PM) gwhaley: https://github.com/kata-containers/runtime/blob/master/cli/config/configuration-qemu-virtiofs.toml.in#L118-L131 :-)
(01:05:24 PM) ***davidgiluk disappears for a 2 hours
(01:05:31 PM) ***gwhaley goes for lunch...
(01:06:43 PM) pohly: So a lot less than the hundreds of GB that people may have as PMEM. MIght be worth testing how that affects performance. Thanks!

CC @devimc @GabyCT

devimc:
A persistent memory volume MUST meet the following conditions:
* A loop device must be mounted in the directory passed as volume
* The loop device must have a backing file
* The backing file must have the PFN signature at offset 4k [1][2]

The backing file is used as backend file for a NVDIMM device in the guest

fixes #2262

[1] - https://github.com/kata-containers/osbuilder/blob/master/image-builder
/nsdax.gpl.c
[2] - https://github.com/torvalds/linux/blob/master/drivers/nvdimm/pfn.h

cc @pohly

### DIFF
--- a/virtcontainers/device/config/config.go
+++ b/virtcontainers/device/config/config.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 
 	"github.com/go-ini/ini"
 	"golang.org/x/sys/unix"
@@ -90,6 +91,8 @@ var SysBusPciDevicesPath = "/sys/bus/pci/devices"
 
 // SysBusPciSlotsPath is static string of /sys/bus/pci/slots
 var SysBusPciSlotsPath = "/sys/bus/pci/slots"
+
+var getSysDevPath = getSysDevPathImpl
 
 // DeviceInfo is an embedded type that contains device data common to all types of devices.
 type DeviceInfo struct {
@@ -257,29 +260,14 @@ func GetHostPath(devInfo DeviceInfo, vhostUserStoreEnabled bool, vhostUserStoreP
 		return "", fmt.Errorf("Empty path provided for device")
 	}
 
-	var pathComp string
-
-	switch devInfo.DevType {
-	case "c", "u":
-		pathComp = "char"
-	case "b":
-		pathComp = "block"
-	default:
-		// Unsupported device types. Return nil error to ignore devices
-		// that cannot be handled currently.
-		return "", nil
-	}
-
 	// Filter out vhost-user storage devices by device Major numbers.
 	if vhostUserStoreEnabled && devInfo.DevType == "b" &&
 		(devInfo.Major == VhostUserSCSIMajor || devInfo.Major == VhostUserBlkMajor) {
 		return getVhostUserHostPath(devInfo, vhostUserStorePath)
 	}
 
-	format := strconv.FormatInt(devInfo.Major, 10) + ":" + strconv.FormatInt(devInfo.Minor, 10)
-	sysDevPath := filepath.Join(SysDevPrefix, pathComp, format, "uevent")
-
-	if _, err := os.Stat(sysDevPath); err != nil {
+	ueventPath := filepath.Join(getSysDevPath(devInfo), "uevent")
+	if _, err := os.Stat(ueventPath); err != nil {
 		// Some devices(eg. /dev/fuse, /dev/cuse) do not always implement sysfs interface under /sys/dev
 		// These devices are passed by default by docker.
 		//
@@ -293,7 +281,7 @@ func GetHostPath(devInfo DeviceInfo, vhostUserStoreEnabled bool, vhostUserStoreP
 		return "", err
 	}
 
-	content, err := ini.Load(sysDevPath)
+	content, err := ini.Load(ueventPath)
 	if err != nil {
 		return "", err
 	}
@@ -304,6 +292,35 @@ func GetHostPath(devInfo DeviceInfo, vhostUserStoreEnabled bool, vhostUserStoreP
 	}
 
 	return filepath.Join("/dev", devName.String()), nil
+}
+
+// getBackingFile is used to fetch the backing file for the device.
+func getBackingFile(devInfo DeviceInfo) (string, error) {
+	backingFilePath := filepath.Join(getSysDevPath(devInfo), "loop", "backing_file")
+	data, err := ioutil.ReadFile(backingFilePath)
+	if err != nil {
+		return "", err
+	}
+
+	return strings.TrimSpace(string(data)), nil
+}
+
+func getSysDevPathImpl(devInfo DeviceInfo) string {
+	var pathComp string
+
+	switch devInfo.DevType {
+	case "c", "u":
+		pathComp = "char"
+	case "b":
+		pathComp = "block"
+	default:
+		// Unsupported device types. Return nil error to ignore devices
+		// that cannot be handled currently.
+		return ""
+	}
+
+	format := strconv.FormatInt(devInfo.Major, 10) + ":" + strconv.FormatInt(devInfo.Minor, 10)
+	return filepath.Join(SysDevPrefix, pathComp, format)
 }
 
 // getVhostUserHostPath is used to fetch host path for the vhost-user device.
--- a/virtcontainers/device/config/config.go
+++ b/virtcontainers/device/config/config.go
@@ -113,6 +113,10 @@ type DeviceInfo struct {
 	Major int64
 	Minor int64
 
+	// Pmem enabled persistent memory. Use HostPath as backing file
+	// for a nvdimm device in the guest.
+	Pmem bool
+
 	// FileMode permission bits for the device.
 	FileMode os.FileMode
 
@@ -169,6 +173,10 @@ type BlockDrive struct {
 
 	// ReadOnly sets the device file readonly
 	ReadOnly bool
+
+	// Pmem enables persistent memory. Use File as backing file
+	// for a nvdimm device in the guest
+	Pmem bool
 }
 
 // VFIODeviceType indicates VFIO device type
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -412,11 +412,11 @@
   revision = "2f1d1f20f75d5404f53b9edf6b53ed5505508675"
 
 [[projects]]
-  digest = "1:0044fb81f517f480ca3c33675a3af6b4ada77a6faf699a302bc2388c98cacba9"
+  digest = "1:fafdb4aa5b6207f51ec7557818d5f7a534ed44ea4fb31c6f2e8abb01d1627b74"
   name = "github.com/intel/govmm"
   packages = ["qemu"]
   pruneopts = "NUT"
-  revision = "3700c55dd766d37e17af354fb9975dc801619d62"
+  revision = "e969afbec52cf687bbe97b76654c664128cdb04b"
 
 [[projects]]
   digest = "1:d6e9b99fe0150d4c26d81612676e8d59ad045642e4cbc8646e494b50d4f245ef"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -48,7 +48,7 @@
 
 [[constraint]]
   name = "github.com/intel/govmm"
-  revision = "3700c55dd766d37e17af354fb9975dc801619d62"
+  revision = "e969afbec52cf687bbe97b76654c664128cdb04b"
 
 [[constraint]]
   name = "github.com/kata-containers/agent"
--- a/vendor/github.com/intel/govmm/qemu/qmp.go
+++ b/vendor/github.com/intel/govmm/qemu/qmp.go
@@ -1428,8 +1428,9 @@ func (q *QMP) ExecHotplugMemory(ctx context.Context, qomtype, id, mempath string
 // a NVDIMM driver with the device_add command.
 // id is the id of the device to add.  It must be a valid QMP identifier.
 // mempath is the path of the device to add, e.g., /dev/rdb0.  size is
-// the data size of the device.
-func (q *QMP) ExecuteNVDIMMDeviceAdd(ctx context.Context, id, mempath string, size int64) error {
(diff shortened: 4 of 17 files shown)


devimc:
@jodh-intel @amshinde @pohly @bergwolf @lifupan  please take a look

virtcontainers/device/config/config.go
```diff
@@ -278,3 +275,36 @@ func GetHostPath(devInfo DeviceInfo) (string, error) {
 
 	return filepath.Join("/dev", devName.String()), nil
 }
+
+// getBackingFile is used to fetch the backing file for the device.
+func getBackingFile(devInfo DeviceInfo) (string, error) {
+	backingFilePath := filepath.Join(getSysDevPath(devInfo), "loop", "backing_file")
+	if _, err := os.Stat(backingFilePath); err != nil {
```
jodh-intel: You could remove the `Stat` by checking the error from `ReadFile()` using `os.IsNotExist(err)`.

devimc:
@jodh-intel change applied, thanks

devimc:
can we merge this?

cc @amshinde @pohly

pohly:
This all looks reasonable to me, with the caveat of course that I don't know the code base.

For user-facing documentation we can use the PMEM-CSI project, because that is what users will have to start with to use the new feature. I can also add end-to-end testing there once there is release of Kata Containers with this feature.

Are you doing canary builds, i.e. once this is merged, can I directly install a pre-release of Kata Containers which supports PMEM?

devimc:
@pohly sure, you can download the static tarball form here https://github.com/kata-containers/runtime/releases/

pohly:
> sure, you can download the static tarball form here https://github.com/kata-containers/runtime/releases/

Can you ping me when there's an alpha with this feature included?

devimc:
@pohly 
> Can you ping me when there's an alpha with this feature included?

sure thing :+1:

PR #2517, opened 2020-03-09
vc: Remove container share dir when stopping

### LINKED ISSUE #2516: vc: rootfsDest and shareDir should be removed when stopping the container
# Description of problem
`rootfsDest` and `shareDir` are created when creating the container but remains after the container is removed. This may cause large amount of inode usage when there are many pods having containers that are restarted frequently.

darfux:
Remove the rootfs bind dest and finally remove the created share directory when stopping the container.

Fixes #2516

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -1108,6 +1108,11 @@ func (c *Container) stop(force bool) error {
 		return err
 	}
 
+	shareDir := filepath.Join(kataHostSharedDir(), c.sandbox.id, c.id)
+	if err := syscall.Rmdir(shareDir); err != nil {
+		c.Logger().WithError(err).WithField("share-dir", shareDir).Warn("Could not remove container share dir")
+	}
+
 	// container was killed by force, container MUST change its state
 	// as soon as possible just in case one of below operations fail leaving
 	// the containers in a bad state.
--- a/virtcontainers/mount.go
+++ b/virtcontainers/mount.go
@@ -354,6 +354,10 @@ func bindUnmountContainerRootfs(ctx context.Context, sharedDir, sandboxID, cID s
 		logrus.Warnf("%s: %s", err, rootfsDest)
 		return nil
 	}
+	if err := syscall.Rmdir(rootfsDest); err != nil {
+		logrus.WithError(err).WithField("rootfs-dir", rootfsDest).Warn("Could not remove container rootfs dir")
+	}
+
 	return err
 }
 
--- a/virtcontainers/mount_test.go
+++ b/virtcontainers/mount_test.go
@@ -304,3 +304,29 @@ func TestBindUnmountContainerRootfsENOENTNotError(t *testing.T) {
 	err := bindUnmountContainerRootfs(context.Background(), testMnt, sID, cID)
 	assert.NoError(err)
 }
+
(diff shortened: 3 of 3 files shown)


grahamwhaley:
Nice find.... I thought we had a CI check that ensured Kata (vc) cleaned up all of its paths once we had deleted a pod/container - soooo, @GabyCT can you check that, and figure out why we were not catching this? And @darfux - maybe we can add some tests (unit tests in vc, and ideally some CI tests as well) to make sure :
- this is working
- we never break it in the future? ;-)

sboeuf:
@grahamwhaley the PR looks good to me. It'd be better to have some feedback from @devimc @jcvenegas @amshinde for more accurate review.
About the question why our CI didn't catch this, I think that's because we do the proper cleanup at the pod level anyway. But I agree this PR makes sense.

virtcontainers/container.go
```diff
@@ -1108,6 +1108,11 @@ func (c *Container) stop(force bool) error {
 		return err
 	}
 
+	shareDir := filepath.Join(kataHostSharedDir(), c.sandbox.id, c.id)
+	if err := syscall.Rmdir(shareDir); err != nil {
+		c.Logger().WithError(err).Warnf("Could not remove container share dir %v", shareDir)
```
devimc: please use `WithField`, not `%v` and  s/Warnf/Warn/

virtcontainers/mount.go
```diff
@@ -354,6 +354,10 @@ func bindUnmountContainerRootfs(ctx context.Context, sharedDir, sandboxID, cID s
 		logrus.Warnf("%s: %s", err, rootfsDest)
 		return nil
 	}
+	if err := syscall.Rmdir(rootfsDest); err != nil {
+		logrus.WithError(err).Warnf("Could not remove container rootfs dir")
```
devimc: `WithField("rootfs-dir", rootfsDest)`   and   s/Warnf/Warn/

darfux:
Thanks @grahamwhaley @sboeuf @devimc . Updated log & added UT for `bindUnmountContainerRootfs` :)  And maybe the share dir case should be done in the tests repo?

darfux:
**ubuntu-18-04-virtiofs**  failed due to `unexpected /opt/kata/bin/qemu-virtiofs-system-x86_64`
**ubuntu-18-04** , **ubuntu-18-04-initrd** and **fedora-31** failed due to `rpc error: code = Unavailable desc = transport is closing`
**podman** failed due to `ERROR: /tmp/jenkins/workspace/kata-containers-runtime-podman-fedora-31-PR/go/src/github.com/kata-containers/osbuilder/rootfs is not a directory`

Could someone give me a hand to restart them 👀

grahamwhaley:
Sure. I've re-scheduled all 5 that failed. That's quite a lot of fails, so let's check if they happen again, or if you were just very unlucky!

darfux:
Thanks @grahamwhaley ! 4 of 5 cases are happy now but virtiofs still fails unluckily because of `curl: (7) Failed to connect to 10.107.143.200 port 8080: Connection timed out` 🤣

grahamwhaley:
nudged virtiofs CI one more time....

PR #2523, opened 2020-03-10
logging: reduce default log level to Warn from Info

grahamwhaley:
Our default logging level was Info, which is rather noisy. Reduce it to Warn, so we still see any critical
items, but don't get swamped with Info messages.

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -57,8 +57,8 @@ var kataLog *logrus.Entry
 
 // originalLoggerLevel is the default log level. It is used to revert the
 // current log level back to its original value if debug output is not
-// required.
-var originalLoggerLevel logrus.Level
+// required. We set the default to 'Warn' for the runtime.
+var originalLoggerLevel = logrus.WarnLevel
 
 var debug = false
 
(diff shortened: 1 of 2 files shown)


grahamwhaley:
@gnawux - do you have any active users with whom you can check this change makes sense, and will not disrupt any existing infra etc.
/cc @awprice - although I believe you are not collecting/filtering kata logs right now.

Personally, I see this as a moderately 'breaking' change, but only if somebody is currently relying on the info messages - which, maybe is only us for our CIs.... /cc @chavafg

grahamwhaley:
/test-ubuntu
As I wrote this on a machine with no dev tools installed yet ;-)....

devimc:
thanks @grahamwhaley , I like this change, but:

`./main.go:181:22: logrus.WarnLevel is not a type`   :thinking:

grahamwhaley:
> thanks @grahamwhaley , I like this change, but:
> 
> `./main.go:181:22: logrus.WarnLevel is not a type` 🤔

;-) - right, I'll pull that down onto a machine where I can actually build it this time - I thought I'd copied that from similar existing code, but the `logrus` was probably a different type or class... np, I'll go fix...

Issue #2546, opened 2020-03-21
Failed to load Nvidia driver for GeForce GT 1030 in Kata container


Jimmy-Xu:
@amshinde @Jimmy-Xu 

Thank you for your kindly documentation for using Nvidia GPU with Kata-containers.
Also, I'm sorry to keep you inquiring about the completed PR.
I followed the guide at the documentation, and I'm successfully able to pass one of the GPU to the guest VM through vfio-pci.
It show up with lspci on container.
```
00:08.0 VGA compatible controller [0300]: NVIDIA Corporation GP108 [GeForce GT 1030] [10de:1d01] (rev a1) (prog-if 00 [VGA controller])
	Subsystem: NVIDIA Corporation GP108 [GeForce GT 1030] [10de:11c7]
	Physical Slot: 8
	Flags: bus master, fast devsel, latency 0, IRQ 10
	Memory at c0000000 (32-bit, non-prefetchable) [size=16M]
	Memory at 1ac0000000 (64-bit, prefetchable) [size=256M]
	Memory at 1ad0000000 (64-bit, prefetchable) [size=32M]
	I/O ports at 1000 [size=128]
	Capabilities: [60] Power Management version 3
	Capabilities: [68] MSI: Enable- Count=1/1 Maskable- 64bit+
	Capabilities: [78] Express Legacy Endpoint, MSI 00
	Kernel driver in use: nvidia
```
But, I have another problem.
When I check Nvidia device status with nvidia-smi, it is failed like as below.
And, the nvidia gpu information has an errors.
```
# nvidia-smi 
Unable to determine the device handle for GPU 0000:00:08.0: Unknown Error

# cat /proc/driver/nvidia/gpus/0000\:00\:08.0/information 
Model: 		 Unknown
IRQ:   		 35
GPU UUID: 	 GPU-????????-????-????-????-????????????
Video BIOS: 	 ??.??.??.??.??
Bus Type: 	 PCIe
DMA Size: 	 36 bits
DMA Mask: 	 0xfffffffff
Bus Location: 	 0000:00:08.0
Device Minor: 	 0
Blacklisted:	 No
```

Is there anything I missed that I should do?

_Originally posted by @wansuyoo in https://github.com/kata-containers/runtime/pull/2410#issuecomment-600445263_

PR #2547, opened 2020-03-21
add kvm=off for qemu

Jimmy-Xu:
Solve the problem of loading drivers for some nvidia graphics cards.

Fixes: #2546

> Just for test

### DIFF
--- a/virtcontainers/qemu_amd64.go
+++ b/virtcontainers/qemu_amd64.go
@@ -129,6 +129,7 @@ func (q *qemuAmd64) bridges(number uint32) {
 
 func (q *qemuAmd64) cpuModel() string {
 	cpuModel := defaultCPUModel
+	cpuModel += ",kvm=off"
 	if q.nestedRun {
 		cpuModel += ",pmu=off"
 	}


wenlxie:
Can we use the configuration to disable this but not make it as default? This will impact the performance?

chestack:
came into the same problem, this patch worked for me.

virtcontainers/qemu_amd64.go
```diff
@@ -129,6 +129,7 @@ func (q *qemuAmd64) bridges(number uint32) {
 
 func (q *qemuAmd64) cpuModel() string {
 	cpuModel := defaultCPUModel
+	cpuModel += ",kvm=off"
```
jodh-intel, line 132: If this is really the solution for Nvidia GPUs, couldn't we just document to add `kvm=off` to `cpu_features=` in `configuration.toml`?

Aside: If the number of GPU-specific settings in `configuration.toml` grows, we might want to think about adding an actual GPU config option which if set automatically sets these "ancillary" config options to make life easier for GPU users.

devimc: :+1:  please don't hard code this , instead you can document it as a GPU-specific setting for some specific GPUs

fidencio:
As this PR is marked as `do-not-merge`, we're assuming it won't make for the final 1.13.0 release, thus we're closing it.
In case you feel this has been wrongly closed, please, feel free to re-open it.

PR #2549, opened 2020-03-23
shimv2: ignore runc options

bergwolf:
create requests might bring in runc options that makes no sense
to kata. Just ignore it.

### DIFF
--- a/containerd-shim-v2/create.go
+++ b/containerd-shim-v2/create.go
@@ -155,14 +155,14 @@ func loadRuntimeConfig(s *service, r *taskAPI.CreateTaskRequest, anno map[string
 	configPath := oci.GetSandboxConfigPath(anno)
 	if configPath == "" && r.Options != nil {
 		v, err := typeurl.UnmarshalAny(r.Options)
-		if err != nil {
-			return nil, err
-		}
-		option, ok := v.(*crioption.Options)
-		// cri default runtime handler will pass a linux runc options,
-		// and we'll ignore it.
-		if ok {
-			configPath = option.ConfigPath
+		// Ignore error as cri daemon might send v1 options that make no sense to us
+		if err == nil {
+			option, ok := v.(*crioption.Options)
+			// cri default runtime handler will pass a linux runc options,
+			// and we'll ignore it.
+			if ok {
+				configPath = option.ConfigPath
+			}
 		}
 	}
 


containerd-shim-v2/create.go
```diff
@@ -155,14 +155,14 @@ func loadRuntimeConfig(s *service, r *taskAPI.CreateTaskRequest, anno map[string
...
 		v, err := typeurl.UnmarshalAny(r.Options)
-		if err != nil {
-			return nil, err
-		}
-		option, ok := v.(*crioption.Options)
-		// cri default runtime handler will pass a linux runc options,
-		// and we'll ignore it.
-		if ok {
-			configPath = option.ConfigPath
+		// Ignore error as cri daemon might send v1 options that make no sense to us
```
jodh-intel: - I think I undestand the intent of this PR, but `crioption` is an alias for the CRI v1 options (`github.com/containerd/cri-containerd/pkg/api/runtimeoptions/v1`) isn't it? Which means I'm confused by this comment :)

- Is it worth logging the error though?

bergwolf: It depends on which containerd client is calling.  For `ctr`, it just sends the runc v1 options. For `cri`, it sends cri v1 options.  Since it is a normal case not any failure for `ctr` case, I don't think we need to log it every time.

lifupan: Hi @bergwolf 

This issue was bought by https://github.com/containerd/containerd/commit/7f5d900769ad02eb339ff2b8ed191af13d663464, 
is it better to complain and fix the issue in contained by ignoring the runc options if ctr wasn't passed the "runc-binary" flag, since the ctr command is a general containerd utils, and not particularly to runc? If we just ignore the option parsed error in kata, then if the users do pass an option and failed to parse here, we'll give the users a wrong illusion
that he's options is valid and come into effect.

bergwolf: @lifupan hmm, makes sense. Let's push a containerd side change. The specific commit is only in v1.3.x release so it's easier to just fix it there. Also we do try to decode runc options by including https://github.com/kata-containers/runtime/blob/511cad9df21286f95e08f73ae9571e51580249bf/containerd-shim-v2/create.go#L24

Is there a real need for the runc types?

bergwolf: @lifupan Thanks for the pointer. I've sent a PR to containerd to fix it there. https://github.com/containerd/containerd/pull/4122

For this PR, let keep it around until containerd merges that fix.

lifupan: > @lifupan hmm, makes sense. Let's push a containerd side change. The specific commit is only in v1.3.x release so it's easier to just fix it there. Also we do try to decode runc options by including
> 
> https://github.com/kata-containers/runtime/blob/511cad9df21286f95e08f73ae9571e51580249bf/containerd-shim-v2/create.go#L24
> 
> Is there a real need for the runc types?

Yeah, I remember that's added by  keeping compatible with old containerd which hadn't supported runtime classes which would pass a shim v1 runc options to kata. For now, I think we can remove it safely.

bergwolf:
Close this as the containerd fix has been merged.

Issue #2551, opened 2020-03-23
clh: Report warning when requested hot-plug CPUs exceeds maxVCPU allowed


likebreath:
**Which feature do you think can be improved?**

The behavior of CPU hotplug in `clh` is not consistent with qemu when the requested vCPUs exceeds the maximum number of vCPUs allowed. This inconsistency also failed some integration tests related to `CPU update`.

**How can it be improved?**
To follow Kata's behavior with `qemu`, we should only report warning instead of report errors.

PR #2552, opened 2020-03-23
clh: Report warning when requested vCPUs exceeds maxVCPU allowed

likebreath:
To have a consistent behavior of CPU hotplug with qemu in clh, we should
only report warning instead of errors when the requested vCPUs exceeds
the maximum number of vCPUs allowed.

Fixes: #2551

### DIFF
--- a/virtcontainers/clh.go
+++ b/virtcontainers/clh.go
@@ -514,15 +514,14 @@ func (clh *cloudHypervisor) resizeVCPUs(reqVCPUs uint32) (currentVCPUs uint32, n
 		clh.Logger().WithField("function", "resizeVCPUs").Debugf("Cannot resize vCPU to 0")
 		return currentVCPUs, newVCPUs, fmt.Errorf("Cannot resize vCPU to 0")
 	}
-	if reqVCPUs > clh.config.DefaultMaxVCPUs || reqVCPUs > uint32(info.Config.Cpus.MaxVcpus) {
+	if reqVCPUs > uint32(info.Config.Cpus.MaxVcpus) {
 		clh.Logger().WithFields(log.Fields{
-			"function":              "resizeVCPUs",
-			"reqVCPUs":              reqVCPUs,
-			"configDefaultMaxVCPUs": clh.config.DefaultMaxVCPUs,
-			"clhMaxVCPUs":           info.Config.Cpus.MaxVcpus,
-		}).Debug("exceeding the maxVCPUs")
-		return currentVCPUs, newVCPUs, fmt.Errorf("Cannot resize vCPU to %d: exceeding the maximum amount of vCPUs (%d or %d)",
-			reqVCPUs, clh.config.DefaultMaxVCPUs, info.Config.Cpus.MaxVcpus)
+			"function":    "resizeVCPUs",
+			"reqVCPUs":    reqVCPUs,
+			"clhMaxVCPUs": info.Config.Cpus.MaxVcpus,
+		}).Warn("exceeding the 'clhMaxVCPUs' (resizing to 'clhMaxVCPUs')")
+
+		reqVCPUs = uint32(info.Config.Cpus.MaxVcpus)
 	}
 
 	// Resize (hot-plug) vCPUs via HTTP API


virtcontainers/clh.go
```diff
@@ -455,9 +455,13 @@ func (clh *cloudHypervisor) resizeVCPUs(reqVCPUs uint32) (currentVCPUs uint32, n
 			"reqVCPUs":              reqVCPUs,
 			"configDefaultMaxVCPUs": clh.config.DefaultMaxVCPUs,
 			"clhMaxVCPUs":           info.Config.Cpus.MaxVcpus,
-		}).Debug("exceeding the maxVCPUs")
-		return currentVCPUs, newVCPUs, fmt.Errorf("Cannot resize vCPU to %d: exceeding the maximum amount of vCPUs (%d or %d)",
-			reqVCPUs, clh.config.DefaultMaxVCPUs, info.Config.Cpus.MaxVcpus)
+		}).Warnf("exceeding the maxVCPUs (resizing to maxVCPUs)")
+
+		if clh.config.DefaultMaxVCPUs < uint32(info.Config.Cpus.MaxVcpus) {
+			reqVCPUs = clh.config.DefaultMaxVCPUs
```
jcvenegas: I wonder when this case would be useful,  at the VM creation the VM got the max amount of cpu based in the host, so unless the host user want to limit the cpus the max vcpus via config at runtime, I wonder if we just want to check for the max amount we provided at the VM creation. @devimc @egernst   and suggestion here?

virtcontainers/clh.go
```diff
@@ -455,9 +455,13 @@ func (clh *cloudHypervisor) resizeVCPUs(reqVCPUs uint32) (currentVCPUs uint32, n
 			"reqVCPUs":              reqVCPUs,
 			"configDefaultMaxVCPUs": clh.config.DefaultMaxVCPUs,
 			"clhMaxVCPUs":           info.Config.Cpus.MaxVcpus,
-		}).Debug("exceeding the maxVCPUs")
-		return currentVCPUs, newVCPUs, fmt.Errorf("Cannot resize vCPU to %d: exceeding the maximum amount of vCPUs (%d or %d)",
-			reqVCPUs, clh.config.DefaultMaxVCPUs, info.Config.Cpus.MaxVcpus)
+		}).Warnf("exceeding the maxVCPUs (resizing to maxVCPUs)")
```
jodh-intel: - Nit: This doesn't need to be `Warnf()` as there are no formats in the string, so you could make it a `Warn()`.
- `maxVCPUs` isn't a field that is being logged, so the user won't know what value is being used. I think this message probably wants to mention `clhMaxVCPUs` instead?

virtcontainers/clh.go
jodh-intel: If this is still useful to have, could you create a new function for this logic? Something like the following (although I don't like this function name particularly ;) ...

```go
func handleResizeVCPUs(vmConfig VmConfig, config HypervisorConfig) (newVCPUs uint32, err error)
    :

newVCPUs, err := handleResizeVCPUs(info.Config, clh.config)
```

That way, the logic is clearer plus you can then add a table-based test to `clh_test.go` for `handleResizeVCPUs()` to assert the expected behaviour for various values.

likebreath: @jodh-intel Thanks for the suggestion. It is nice to have related unit tests in place. I will add it if we decide to keep this checks. :)

@jcvenegas Good point. The questions break down to:
If we don't want to check both `clh.config.DefaultMaxVCPUs` (from the running CLH instance) and `info.Config.Cpus.MaxVcpus` (from Kata's config file), which one we want to use?
1. The reason to pick `clh.config.DefaultMaxVCPUs` is that this is the hard-limit enforced by the running VM/clh.
2. The reason to pick `info.Config.Cpus.MaxVcpus` is that this is Kata's interface for users to config the limit of vCPUs.

For common usage cases, the above two reasons do not conflict with each other, as `clh.config.DefaultMaxVCPUs` is assigned based on `info.Config.Cpus.MaxVcpus` at VM creation. The only corner case (as @jcvenegas mentioned above) is that users can change the kata configuration file after VM creation with a different/larger value of `clh.config.DefaultMaxVCPUs`. If this is a valid usage case, it seems that we will need to check on both (to respect Kata's configuration and clh's hard-limit). Any thoughts?

jcvenegas: Well IMO post VM creation  config modification is not something we honor,  not even documented, so for simplification I would go just checking the hypervisor.

jcvenegas: @sameo @bergwolf @sboeuf  any preference here?

amshinde: @likebreath The change in the config should not affect running containers as the config should be stored separately for a pod. If you are seeing otherwise, then this should be filed as a bug.

likebreath: @amshinde With your input, @jcvenegas and I decided to check the `info.Config.Cpus.MaxVcpus` only for now. Also the patch has been updated accordingly.

jcvenegas:
Travis not reported the result, but was sucessful merging now

PR #2556, opened 2020-03-25
clh: Enable memory hotplug

jcvenegas:
Enable memory hotplug for cloud hypervisor.  Use clh openAPI to start VM with max amount of memory (based in host memory). 

- Implement memory hotplug
- Memory resize only can be done to increase memory

Depends-on: github.com/kata-containers/tests#2413

### DIFF
--- a/virtcontainers/clh.go
+++ b/virtcontainers/clh.go
@@ -214,7 +214,8 @@ func (clh *cloudHypervisor) createSandbox(ctx context.Context, id string, networ
 	clh.Logger().WithField("function", "createSandbox").WithError(err).Info("Sandbox not found creating ")
 
 	// Set initial memomory size of the virtual machine
-	clh.vmconfig.Memory.Size = int64(clh.config.MemorySize) << utils.MibToBytesShift
+	// Convert to int64 openApiClient only support int64
+	clh.vmconfig.Memory.Size = int64((utils.MemUnit(clh.config.MemorySize) * utils.MiB).ToBytes())
 	clh.vmconfig.Memory.File = "/dev/shm"
 	// Set initial amount of cpu's for the virtual machine
 	clh.vmconfig.Cpus = chclient.CpusConfig{
--- a/virtcontainers/clh.go
+++ b/virtcontainers/clh.go
@@ -432,14 +432,11 @@ func (clh *cloudHypervisor) resizeVCPUs(reqVCPUs uint32) (currentVCPUs uint32, n
 	cl := clh.client()
 
 	// Retrieve the number of current vCPUs via HTTP API
-	ctx, cancel := context.WithTimeout(context.Background(), clhAPITimeout*time.Second)
-	info, _, err := cl.VmInfoGet(ctx)
+	info, err := clh.vmInfo()
 	if err != nil {
-		clh.Logger().WithField("function", "resizeVCPUs").WithError(openAPIClientError(err)).Info("[clh] VmInfoGet failed")
+		clh.Logger().WithField("function", "resizeVCPUs").WithError(err).Info("[clh] vmInfo failed")
 		return 0, 0, openAPIClientError(err)
 	}
-	// Reset the timer after the first HTTP API call
-	cancel()
 
 	currentVCPUs = uint32(info.Config.Cpus.BootVcpus)
 	newVCPUs = currentVCPUs
@@ -461,7 +458,7 @@ func (clh *cloudHypervisor) resizeVCPUs(reqVCPUs uint32) (currentVCPUs uint32, n
 	}
 
 	// Resize (hot-plug) vCPUs via HTTP API
-	ctx, cancel = context.WithTimeout(context.Background(), clhAPITimeout*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), clhAPITimeout*time.Second)
 	defer cancel()
 	if _, err = cl.VmResizePut(ctx, chclient.VmResize{DesiredVcpus: int32(reqVCPUs)}); err != nil {
 		return currentVCPUs, newVCPUs, errors.Wrap(err, "[clh] VmResizePut failed")
@@ -960,10 +957,10 @@ func (clh *cloudHypervisor) bootVM(ctx context.Context) error {
 		return openAPIClientError(err)
 	}
 
-	info, _, err := cl.VmInfoGet(ctx)
+	info, err := clh.vmInfo()
 
 	if err != nil {
-		return openAPIClientError(err)
+		return err
 	}
 
 	clh.Logger().Debugf("VM state after create: %#v", info)
@@ -979,10 +976,10 @@ func (clh *cloudHypervisor) bootVM(ctx context.Context) error {
 		return openAPIClientError(err)
 	}
 
-	info, _, err = cl.VmInfoGet(ctx)
+	info, err = clh.vmInfo()
 
 	if err != nil {
-		return openAPIClientError(err)
+		return err
 	}
 
 	clh.Logger().Debugf("VM state after boot: %#v", info)
@@ -1119,4 +1116,18 @@ func (clh *cloudHypervisor) cleanupVM(force bool) error {
 	clh.reset()
 
 	return nil
+}
+
+// vmInfo ask to hypervisor for current VM status
+func (clh *cloudHypervisor) vmInfo() (chclient.VmInfo, error) {
+	cl := clh.client()
+	ctx, cancelInfo := context.WithTimeout(context.Background(), clhAPITimeout*time.Second)
+	defer cancelInfo()
+
+	info, _, err := cl.VmInfoGet(ctx)
+	if err != nil {
+		clh.Logger().WithError(openAPIClientError(err)).Warn("VmInfoGet failed")
+	}
+	return info, openAPIClientError(err)
+
 }
--- a/virtcontainers/clh.go
+++ b/virtcontainers/clh.go
@@ -217,6 +217,13 @@ func (clh *cloudHypervisor) createSandbox(ctx context.Context, id string, networ
 	// Convert to int64 openApiClient only support int64
 	clh.vmconfig.Memory.Size = int64((utils.MemUnit(clh.config.MemorySize) * utils.MiB).ToBytes())
 	clh.vmconfig.Memory.File = "/dev/shm"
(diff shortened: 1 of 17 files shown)


virtcontainers/pkg/cloud-hypervisor/client/api/openapi.yaml
```diff
@@ -309,24 +330,26 @@ components:
...
+          cache_size: 4
           dax: true
           tag: tag
         - sock: sock
-          num_queues: 9
-          queue_size: 3
-          cache_size: 2
+          num_queues: 3
+          queue_size: 2
+          cache_size: 4
```
sboeuf: I don't understand all these numbers assigned to `num_queues` and `queue_size` :-/
Is this coming from the automated file generation?

virtcontainers/clh.go
```diff
@@ -217,6 +217,13 @@ func (clh *cloudHypervisor) createSandbox(ctx context.Context, id string, networ
 	// Convert to int64 openApiClient only support int64
 	clh.vmconfig.Memory.Size = int64((utils.MemUnit(clh.config.MemorySize) * utils.MiB).ToBytes())
 	clh.vmconfig.Memory.File = "/dev/shm"
+	hostMemKb, err := getHostMemorySizeKb(procMemInfo)
+	if err != nil {
+		return nil
+	}
+
+	// OpenAPI only suuports int64 values
```
sboeuf: typo: s/suuports/supports/

virtcontainers/clh.go
```diff
@@ -424,22 +432,84 @@ func (clh *cloudHypervisor) hypervisorConfig() HypervisorConfig {
...
+		clh.Logger().WithFields(log.Fields{"request": newMem, "aligned_request": alignedRequest}).Debug("aligning VM memory request")
+		newMem = alignedRequest
+	}
+
+	// Check if memory is the same as requested, a second check is done
+	// to consider the memory request now that is updated to be memory aligned
+	if currentMem == newMem {
+		clh.Logger().WithFields(log.Fields{"current_memory": currentMem, "new_memory": newMem}).Debug("VM already has requested memory(after alignment)")
+		return 0, memoryDevice{}, nil
+	}
```
jodh-intel: Like the VCPU logic PR, I think this block could be put into a separate function which can then be unit tested. This function would then become very simple:

- Call the new function to work out the resize value (which can be easily unit tested).
- Pass that value to CLH to perform the resize (which cannot be easily unit tested).

virtcontainers/clh.go
```diff
@@ -424,22 +432,84 @@ func (clh *cloudHypervisor) hypervisorConfig() HypervisorConfig {
...
+
+	blockSize := utils.MemUnit(memoryBlockSizeMB) * utils.MiB
+	hotplugSize := (newMem - currentMem).AlignMem(blockSize)
+
+	// Update memory request to increase memory aligned block
+	alignedRequest := currentMem + hotplugSize
+	if newMem != alignedRequest {
+		clh.Logger().WithFields(log.Fields{"request": newMem, "aligned_request": alignedRequest}).Debug("aligning VM memory request")
```
jodh-intel: Nit: (Here and elsewhere) We tend to use dashes for field names, so `aligned-request`.

virtcontainers/clh.go
```diff
@@ -424,22 +432,84 @@ func (clh *cloudHypervisor) hypervisorConfig() HypervisorConfig {
...
+		clh.Logger().WithError(err).Warn("vmInfo failed")
+		return 0, memoryDevice{}, openAPIClientError(err)
+	}
+
+	newMem = utils.MemUnit(info.Config.Memory.Size) * utils.Byte
+	addedMemory := newMem - currentMem
+
+	if hotplugSize != addedMemory {
+		errors.Errorf("Added memory different than expected %d != %d", newMem, addedMemory)
+	}
```
jodh-intel: Hmm, so we have to make _two_ calls? If `VmResizePut()` could in future return the actual value set, that would save an HTTP request I think?

/cc @sboeuf.

sboeuf: No there's no need for that because if the call didn't return any error, that means the expected size is now the current one. So basically I don't think there's a need for calling into `VmInfo` to verify the size.

jodh-intel: Thanks @sboeuf - so you can zap that bit @jcvenegas ;)

virtcontainers/pkg/cloud-hypervisor/client/api/openapi.yaml
jcvenegas: Yes, it is comming from the automatic file generation seems that it adds examples to that yaml. 

https://github.com/kata-containers/runtime/pull/2556/files#diff-4dc59e987e364456f339849060f41612L160

sboeuf: Ok!

virtcontainers/clh.go
```diff
@@ -425,8 +432,62 @@ func (clh *cloudHypervisor) hypervisorConfig() HypervisorConfig {
 }
 
 func (clh *cloudHypervisor) resizeMemory(reqMemMB uint32, memoryBlockSizeMB uint32, probe bool) (uint32, memoryDevice, error) {
-	clh.Logger().WithField("function", "resizeMemory").Warn("not supported")
-	return 0, memoryDevice{}, nil
+	memDev := memoryDevice{probe: probe}
+	if reqMemMB == 0 {
+		return 0, memDev, errors.New("Can not resize memory to 0")
+	}
```
likebreath: nit: it should be covered by the following check on `if currentMem > newMem `.

virtcontainers/clh.go
```diff
@@ -425,8 +432,62 @@ func (clh *cloudHypervisor) hypervisorConfig() HypervisorConfig {
 }
 
 func (clh *cloudHypervisor) resizeMemory(reqMemMB uint32, memoryBlockSizeMB uint32, probe bool) (uint32, memoryDevice, error) {
-	clh.Logger().WithField("function", "resizeMemory").Warn("not supported")
-	return 0, memoryDevice{}, nil
+	memDev := memoryDevice{probe: probe}
```
likebreath: In general, I feel we should always return `memoryDevice{}` in this function. If I read correctly the`resizeMemory()` from `virtcontainers/qemu.go`, they will need to return a non-empty `memoryDevice` only when `hotplugMemory` is required (e.g. removing/adding memory device when `virtio-mem` is not used). I am assuming we don't need to (manually) remove/add memory device with `clh` for memory hotplug.

virtcontainers/clh.go
```diff
@@ -425,8 +432,62 @@ func (clh *cloudHypervisor) hypervisorConfig() HypervisorConfig {
...
+		return 0, memDev, err
+	}
+
+	currentMem := utils.MemUnit(info.Config.Memory.Size) * utils.Byte
+	newMem := utils.MemUnit(reqMemMB) * utils.MiB
+
+	// Early check to verify if boot memory is the same as requested
+	if currentMem == newMem {
+		clh.Logger().WithField("memory", reqMemMB).Debugf("VM already has requested memory")
+		return 0, memDev, nil
```
likebreath: It should return `currentMem` if I read the `qemu.go` correctly.

virtcontainers/clh.go
```diff
@@ -425,8 +432,62 @@ func (clh *cloudHypervisor) hypervisorConfig() HypervisorConfig {
...
+	}
+
+	if currentMem > newMem {
+		clh.Logger().Warn("Remove memory is not supported, nothing to do")
+		return 0, memDev, nil
```
likebreath: Same thought as the return with `if currentMem == newMem`, if we just want to log the warning and nothing else, should we return with `currentMem`?

virtcontainers/clh.go
```diff
@@ -425,8 +432,62 @@ func (clh *cloudHypervisor) hypervisorConfig() HypervisorConfig {
...
+	}
+
+	blockSize := utils.MemUnit(memoryBlockSizeMB) * utils.MiB
```
likebreath: Two questions:
1. Do we need to respect the `memoryBlockSizeMB`? It sounds only related when `memory device` hotplug is involved, and `qemu.go` does not check on it when `virtio-mem` is enabled.
2. Do we want to make sure we respect the constraint described in CLH hotplug [doc](https://github.com/cloud-hypervisor/cloud-hypervisor/blob/master/docs/hotplug.md): `Due to guest OS limitations is is necessary to ensure that amount of memory added (between currently assigned RAM and that which is desired) is a multiple of 128MiB.`

jcvenegas: > Two questions:
> 
>     1. Do we need to respect the `memoryBlockSizeMB`? It sounds only related when `memory device` hotplug is involved, and `qemu.go` does not check on it when `virtio-mem` is enabled.

Yes the way the hotplug is without virtio-mem, so we need to verify we request to add memory as a multiple of 128MiB.

> 
>     2. Do we want to make sure we respect the constraint described in CLH hotplug [doc](https://github.com/cloud-hypervisor/cloud-hypervisor/blob/master/docs/hotplug.md): `Due to guest OS limitations is is necessary to ensure that amount of memory added (between currently assigned RAM and that which is desired) is a multiple of 128MiB.`


Yes, the block size in x86_64 is set to 128Mb and the function `func (m MemUnit) AlignMem(blockSize MemUnit) MemUnit {` will make sure to make it a multiple of it

virtcontainers/clh.go
jcvenegas: Agree :)

virtcontainers/clh.go
jcvenegas: I decided to add this corner case because resize to 0 is meaning of something went really wrong, in the case when memory is just reduced  only a warning is set and no operation is performed but not errors are returned.

virtcontainers/clh.go
jcvenegas: Good catch, yeah it is not needed I added a validation to check if probe is true and return an error.

virtcontainers/clh.go
jcvenegas: Agree, fixed.

likebreath:
As a reminder: the CI failure on clh is a known one, which is fixed by PR #2552.
@jcvenegas

Issue #2560, opened 2020-03-25
config: Add scsi_mod.scan=none for virtio-scsi


justin-he:
**Which feature do you think can be improved?**

the default scan mode of scsi is sync. Change it to async
can reduce the guest boot time:

**How can it be improved?**

Change the kernel boot parameter if virtio-scsi && no user default scsi parameter

**Additional Information**

No

**Before raising this feature request**

Have you looked at the [limitations document](https://github.com/kata-containers/documentation/blob/master/Limitations.md)? 
Yes

PR #2561, opened 2020-03-25
config: Add scsi_mod.scan=none for virtio-scsi

justin-he:
As per [1], the default scan mode of scsi is sync. Change it to none
can reduce the guest boot time:
=Before this patch=
[    0.090796]  vda: vda1
[    0.092119] scsi host0: Virtio SCSI HBA
[    0.148851] tun: Universal TUN/TAP device driver, 1.6
[    0.149023] rtc-pl031 9010000.pl031: rtc core: registered pl031 as rtc0

=After this patch=
[    0.077012]  vda: vda1
[    0.078353] scsi host0: Virtio SCSI HBA
[    0.080755] tun: Universal TUN/TAP device driver, 1.6
[    0.081884] rtc-pl031 9010000.pl031: rtc core: registered pl031 as rtc0
It reduces about 54ms on arm64 for virtio-scsi.

This patch changes the default kernel parameter:
1. If user specifies the scan mode, use that
2. If user doesn't specify it, and the block device is virtio-scsi, use
   "async" by default

[1] https://lwn.net/Articles/201898/

Fixes: #2560
Signed-off-by: Jia He <justin.he@arm.com

### DIFF
--- a/pkg/katautils/config.go
+++ b/pkg/katautils/config.go
@@ -977,7 +977,6 @@ func SetKernelParams(runtimeConfig *oci.RuntimeConfig) error {
 	if runtimeConfig.HypervisorConfig.Debug {
 		strParams := vc.SerializeParams(defaultKernelParams, "=")
 		formatted := strings.Join(strParams, " ")
-
 		kataUtilsLogger.WithField("default-kernel-parameters", formatted).Debug()
 	}
 
@@ -989,7 +988,18 @@ func SetKernelParams(runtimeConfig *oci.RuntimeConfig) error {
 
 	// first, add default values
 	for _, p := range defaultKernelParams {
-		if err := (runtimeConfig).AddKernelParam(p); err != nil {
+		if err := runtimeConfig.AddKernelParam(p); err != nil {
+			return err
+		}
+	}
+
+	// set the scsi scan mode to none for virtio-scsi
+	if runtimeConfig.HypervisorConfig.BlockDeviceDriver == config.VirtioSCSI {
+		p := vc.Param{
+			Key:   "scsi_mod.scan",
+			Value: "none",
+		}
+		if err := runtimeConfig.AddKernelParam(p); err != nil {
 			return err
 		}
 	}
@@ -1004,7 +1014,7 @@ func SetKernelParams(runtimeConfig *oci.RuntimeConfig) error {
 		params := vc.KataAgentKernelParams(agentConfig)
 
 		for _, p := range params {
-			if err := (runtimeConfig).AddKernelParam(p); err != nil {
+			if err := runtimeConfig.AddKernelParam(p); err != nil {
 				return err
 			}
 		}
@@ -1012,7 +1022,7 @@ func SetKernelParams(runtimeConfig *oci.RuntimeConfig) error {
 
 	// now re-add the user-specified values so that they take priority.
 	for _, p := range userKernelParams {
-		if err := (runtimeConfig).AddKernelParam(p); err != nil {
+		if err := runtimeConfig.AddKernelParam(p); err != nil {
 			return err
 		}
 	}
--- a/pkg/katautils/create_test.go
+++ b/pkg/katautils/create_test.go
@@ -208,6 +208,10 @@ func TestSetKernelParams(t *testing.T) {
 	err := SetKernelParams(&config)
 	assert.NoError(err)
 
+	config.HypervisorConfig.BlockDeviceDriver = "virtio-scsi"
+	err = SetKernelParams(&config)
+	assert.NoError(err)
+
 	if needSystemd(config.HypervisorConfig) {
 		assert.NotEmpty(config.HypervisorConfig.KernelParams)
 	}


pkg/katautils/config.go
```diff
@@ -994,6 +993,19 @@ func SetKernelParams(runtimeConfig *oci.RuntimeConfig) error {
 		}
 	}
 
+	// set the scsi scan mode to async for virtio-scsi
+	if runtimeConfig.HypervisorConfig.BlockDeviceDriver == config.VirtioSCSI {
+		if !strings.Contains(formatted, "scsi_mod.scan") {
```
jodh-intel: This looks odd - shouldn't the value be tested? What if `scsi_mod.scan=none` or `scsi_mod.scan=sync`?

- https://github.com/torvalds/linux/blob/master/Documentation/scsi/scsi-parameters.txt#L95..L98

devimc: also, kata-agent already scans the SCSI busses [1], `scsi_mod.scan=none` should work, right?

[1] https://github.com/kata-containers/agent/blob/649d44117a862cdcc72dfef6f3ade3bdec6e3fc4/device.go#L322

amshinde: Yes, I think we can get away with passing  `scsi_mod.scan=none`, since all scsi devices are hotplugged in our case and the agent scans the scsi bus after the scsi device is hotplugged.
@justin-he Give that a shot and see how it goes.

justin-he: > This looks odd - shouldn't the value be tested? What if `scsi_mod.scan=none` or `scsi_mod.scan=sync`?
> 
> * https://github.com/torvalds/linux/blob/master/Documentation/scsi/scsi-parameters.txt#L95..L98

My original purpose is:
1. If the user specify any parameter for scsi_mod.scan, using that specified one without checking.
2. If the user doesn't specify, using "scsi_mod.scan=async"
image

justin-he: > Yes, I think we can get away with passing `scsi_mod.scan=none`, since all scsi devices are hotplugged in our case and the agent scans the scsi bus after the scsi device is hotplugged.
> @justin-he Give that a shot and see how it goes.

Ok, I will let you know the result once I test it. suspect that "none" will be better than "async"

justin-he: scsi_mod.scan=sync
[    0.113828] [    T1] scsi host0: Virtio SCSI HBA
[    0.134006] [    T1] tun: Universal TUN/TAP device driver, 1.6
scsi_mod.scan=none
[    0.105891] [    T1] scsi host0: Virtio SCSI HBA
[    0.107868] [    T1] tun: Universal TUN/TAP device driver, 1.6

pkg/katautils/config.go
```diff
@@ -994,6 +993,19 @@ func SetKernelParams(runtimeConfig *oci.RuntimeConfig) error {
 		}
 	}
 
+	// set the scsi scan mode to none for virtio-scsi
+	if runtimeConfig.HypervisorConfig.BlockDeviceDriver == config.VirtioSCSI {
+		if !strings.Contains(formatted, "scsi_mod.scan") {
+			p := vc.Param{
+				Key:   "scsi_mod.scan",
+				Value: "none",
```
devimc, line 1001: thanks, can you update the unit test `TestSetKernelParams`?

amshinde: Also, can you change the title of this PR as well.

justin-he: ok

grahamwhaley:
I've nudged the failed CIs for rebuilds - let's keep an eye on them.

pkg/katautils/config.go
```diff
@@ -994,6 +993,19 @@ func SetKernelParams(runtimeConfig *oci.RuntimeConfig) error {
 		}
 	}
 
+	// set the scsi scan mode to none for virtio-scsi
+	if runtimeConfig.HypervisorConfig.BlockDeviceDriver == config.VirtioSCSI {
+		if !strings.Contains(formatted, "scsi_mod.scan") {
```
devimc, line 998: @justin-he after reading the code, I think this conditional is not needed, since `kernel_params` overwrites default kernel parameters.

https://github.com/kata-containers/runtime/blob/af24829c2ae78c3b811c4cff6736ddaba500d37c/cli/config/configuration-qemu.toml.in#L24-L30

devimc: so, in theory this means you can set `scsi_mod.scan=none` by default and the user can change this behaviour

justin-he: Ok I change remove the condition

pkg/katautils/config.go
```diff
@@ -973,11 +973,10 @@ func updateRuntimeConfigShim(configPath string, tomlConf tomlConfig, config *oci
 // configuration file) to the defaults so that the former take priority.
 func SetKernelParams(runtimeConfig *oci.RuntimeConfig) error {
 	defaultKernelParams := GetKernelParamsFunc(needSystemd(runtimeConfig.HypervisorConfig), runtimeConfig.Trace)
+	strParams := vc.SerializeParams(defaultKernelParams, "=")
+	formatted := strings.Join(strParams, " ")
 
 	if runtimeConfig.HypervisorConfig.Debug {
-		strParams := vc.SerializeParams(defaultKernelParams, "=")
-		formatted := strings.Join(strParams, " ")
-
```
devimc, line 980: I think you need to revert this

pkg/katautils/config.go
```diff
@@ -994,6 +993,17 @@ func SetKernelParams(runtimeConfig *oci.RuntimeConfig) error {
 		}
 	}
 
+	// set the scsi scan mode to none for virtio-scsi
+	if runtimeConfig.HypervisorConfig.BlockDeviceDriver == config.VirtioSCSI {
+		p := vc.Param{
+			Key:   "scsi_mod.scan",
+			Value: "none",
+		}
+		if err := (runtimeConfig).AddKernelParam(p); err != nil {
```
devimc, line 1002: s/(runtimeConfig)/runtimeConfig

justin-he: ok， I will also s/(runtimeConfig)/runtimeConfig globally in pkg/katautils/config.go

pkg/katautils/config.go
justin-he: yes, indeed

PR #2578, opened 2020-04-02
Kata-agent: fix client is nil

keloyang:
Fix https://github.com/kata-containers/runtime/issues/2579


grahamwhaley:
OOI @keloyang - how did you discover/generate this case?

keloyang:
I don't has this situation occured,because all sendReq are called serially now. but I think the code is not correct. If someday kata-runtime can call sendReq parallelly, the code is a hidden danger.

PR #2583, opened 2020-04-03
qemu: Don't crash if virtiofsd path is non existent

fidencio:
Instead, report an error and exit gracefully, as shown below:
```
dahmer fidencio # podman run -ti --runtime=/usr/bin/kata-runtime fedora sh
Error: stat /usr/libexec/virtiofsd: no such file or directory: OCI runtime command not found error
```

Fixes: #2582

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -638,6 +638,10 @@ func (q *qemu) setupVirtiofsd() (err error) {
 	var listener *net.UnixListener
 	var fd *os.File
 
+	if _, err = os.Stat(q.config.VirtioFSDaemon); os.IsNotExist(err) {
+		return fmt.Errorf("virtiofsd path (%s) does not exist", q.config.VirtioFSDaemon)
+	}
+
 	sockPath, err := q.vhostFSSocketPath(q.id)
 	if err != nil {
 		return err


virtcontainers/qemu.go
```diff
@@ -638,6 +638,11 @@ func (q *qemu) setupVirtiofsd() (err error) {
 	var listener *net.UnixListener
 	var fd *os.File
 
+	_, err = os.Stat(q.config.VirtioFSDaemon)
+	if err != nil {
+		return err
```
devimc, line 643: might be worth making this error a little bit more verbose .. ?

```go
return fmt.Error("No such virtio fs daemon: %v", err)
```

virtcontainers/qemu.go
```diff
@@ -638,6 +638,11 @@ func (q *qemu) setupVirtiofsd() (err error) {
 	var listener *net.UnixListener
 	var fd *os.File
 
+	_, err = os.Stat(q.config.VirtioFSDaemon)
+	if err != nil {
```
devimc, line 642: `os.Stat` can fail for different reasons, I think here you want to check if the file exist, isn't it?

```go
if os.IsNotExist(err) {
```

https://golang.org/pkg/os/#IsNotExist

virtcontainers/qemu.go
fidencio: In the new version, I went for the same message used when the hypervisor is not found.
Thanks for the suggestion!

virtcontainers/qemu.go
fidencio: Fixed!

fidencio:
The two CI failures are well known and not related to this patch:
- kata-containers/ci#263
- kata-containers/tests#2434

Issue #2584, opened 2020-04-03
qemu_ppc64le: Expose that hypervisor supports fs sharing


nitkon:
# Description of problem

https://github.com/kata-containers/ci/issues/263

PR #2585, opened 2020-04-03
qemu_ppc64le: EXpose fs support explicitly

nitkon:
Since fs sharing is not assumed as supported by default, expose
explicitly that the qemu_ppc64le supports it.

Fixes: #2584

### DIFF
--- a/virtcontainers/qemu_ppc64le.go
+++ b/virtcontainers/qemu_ppc64le.go
@@ -95,6 +95,7 @@ func (q *qemuPPC64le) capabilities() types.Capabilities {
 	}
 
 	caps.SetMultiQueueSupport()
+	caps.SetFsSharingSupport()
 
 	return caps
 }


virtcontainers/qemu_ppc64le.go
```diff
@@ -95,6 +95,7 @@ func (q *qemuPPC64le) capabilities() types.Capabilities {
 	}
 
 	caps.SetMultiQueueSupport()
+	caps.SetFsSharingSupport()
```
jodh-intel, line 98: Presumably this is for parity with:

- https://github.com/kata-containers/runtime/blob/705713b4f9bc4d1e754871d5ef1ab5e99ea71aff/virtcontainers/qemu_amd64.go#L122

... but I just found:

- https://github.com/kata-containers/runtime/blob/705713b4f9bc4d1e754871d5ef1ab5e99ea71aff/virtcontainers/qemu_amd64.go#L122

... so I think the amd64 call (and this PR) may be redundant?

/cc @jcvenegas.

nitkon: @jodh-intel: Interesting.. but the failing functional tests passes now ...
```
Ran 8 of 8 Specs in 18.976 seconds
SUCCESS! -- 8 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
```

jcvenegas: @jodh-intel  that is a good point, I based in your comment that would make sense, I think from CI healthy we want merge this first, unless @nitkon want to check a bit more and fix as part of this PR. I just open #2586. @Pennyzct is this affecting ARM CI ?

jodh-intel: Thanks @jcvenegas.

Issue #2594, opened 2020-04-07
clh: Fails doing apt-get -y upgrade


jcvenegas:
# Description of problem

CI started to fail with on docker package manager test for ubuntu
```
00:24:01.226 [docker exec Bspe1VfwE8WH9afZ8W0KJLrPg35uM1 apt-get -y upgrade]
00:24:01.226 Timeout: 900 seconds
00:24:01.226 Exit Code: 100
00:24:01.226 Stdout: Reading package lists...
00:24:01.226 Building dependency tree...
00:24:01.226 Reading state information...
00:24:01.226 Calculating upgrade...
00:24:01.226 The following packages will be upgraded:
00:24:01.226   libgnutls30
00:24:01.226 1 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
00:24:01.226 Need to get 1124 kB of archives.
00:24:01.226 After this operation, 0 B of additional disk space will be used.
00:24:01.226 Get:1 http://security.debian.org/debian-security buster/updates/main amd64 libgnutls30 amd64 3.6.7-4+deb10u3 [1124 kB]
00:24:01.226 Fetched 1124 kB in 0s (12.6 MB/s)
00:24:01.226 dpkg: error: unable to lock triggers database lock: Operation not supported
00:24:01.226 
00:24:01.226 Stderr: debconf: delaying package configuration, since apt-utils is not installed
00:24:01.226 E: Sub-process /usr/bin/dpkg returned an error code (2)
00:24:01.226 
00:24:01.227 Running command '/usr/bin/docker [docker exec Bspe1VfwE8WH9afZ8W0KJLrPg35uM1 apt-get -y upgrade]'
00:24:03.625 command failed error 'exit status 100'
00:24:03.625 [docker exec Bspe1VfwE8WH9afZ8W0KJLrPg35uM1 apt-get -y upgrade]
00:24:03.625 Timeout: 900 seconds
00:24:03.625 Exit Code: 100
00:24:03.625 Stdout: Reading package lists...
00:24:03.625 Building dependency tree...
00:24:03.625 Reading state information...
00:24:03.625 Calculating upgrade...
00:24:03.625 The following packages will be upgraded:
00:24:03.625   libgnutls30
00:24:03.625 1 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
00:24:03.625 Need to get 1124 kB of archives.
00:24:03.625 After this operation, 0 B of additional disk space will be used.
[...]

PR #2595, opened 2020-04-07
clh: virtiofs: Add no_posix_lock option

jcvenegas:
Add option in qemu driver , this will allow lock operations.

Fixes: #2594

### DIFF
--- a/virtcontainers/virtiofsd.go
+++ b/virtcontainers/virtiofsd.go
@@ -126,9 +126,11 @@ func (v *virtiofsd) args() ([]string, error) {
 
 	args := []string{
 		"-f",
-		"-o", "vhost_user_socket=" + v.socketPath,
+		"-o", "cache=" + v.cache,
+		"-o", "no_posix_lock",
 		"-o", "source=" + v.sourcePath,
-		"-o", "cache=" + v.cache}
+		"-o", "vhost_user_socket=" + v.socketPath,
+	}
 
 	if len(v.extraArgs) != 0 {
 		args = append(args, v.extraArgs...)


c3d:
This is a bit similar to #2580, although the virtiofs option is different. I believe that my comment over there, https://github.com/kata-containers/runtime/issues/2580#issuecomment-610487663, applies here as well, although the security risk associated to locking is lower.

In short, it would be better to have a "semantic" option associated to specific volumes and indicating that we _want_ to install things there. Changing options globally for all volumes seems risky, since ultimately this impacts the host.

jcvenegas:
@c3d thanks for the pointer to the issue. 
@c3d @amshinde @grahamwhaley 
For now this PR was to match what virtiofsd qemu implementation has see:
https://github.com/kata-containers/runtime/blob/master/virtcontainers/qemu.go#L625

We have two options here, 
 - just hang in this change and remove `no_posix_lock` from qemu implementation. In addition skip test  affected by this.
- Merge this and keep working in a side solution for a "sematic" option to allow users decide to enable extra options like this. 

Both options work for me, specially because I would like to unblock cloud-hypervisor CI.
\cc @kata-containers/architecture-committee

dagrh:
I believe one problem is that I don't think we have a setlkw (i.e. blocking) lock implementation - so if you declare that you support posix locks you get most but not a full implementation.   The difficulty is that a blocking lock is ... blocking; so you hang up one of the threads of the daemon waiting for the lock and there's no good kernel interface for it.  We could spawn threads just to hold the blocking locks but they could get eaten up quickly.
Really the world needs an asynchronous lock kernel interface, so you can say you'd like a lock but neither need to block nor poll for it.

amshinde:
@jcvenegas @dagrh @c3d
I am curious how this change fixes the package update issue. We discussed in another thread about a running virtiofsd with the capability to change extended attributes.

jcvenegas:
@amshinde to be honest I dont know, I figured out by checking the difference with qemu driver an clh.

Probably @dagrh could extend bit more if on the option, the effects is has, functionally and from security perspective. And finally if we should go with this change or remove the flag from qemu until a semantic option is implemented.

amshinde:
@jcvenegas Ok Lets wait for some input here.
Following git blame, similar change to qemu was done by @chavafg. Maybe he has some input here.

@jcvenegas To be clear, this fixes the apt install issue and not dnf/yum. I believe for the latter, we still need virtiofsd to be called with support for changing extended attributes. Please confirm.

jcvenegas:
@amshinde  I confirm this only fix apt, for dnf I still see the container hangs.

jcvenegas:
@kata-containers/architecture-committee  ping, please feel free to merge this PR as agreed this morning in arch meeting.

fidencio:
As discussed during the meeting Today, this option is already present for QEMU backend, let's have it merged.

PR #2606, opened 2020-04-13
virtcontainers: apply constrains to the sandbox cgroup

### LINKED ISSUE #2605: vfio device passthrough fails with “Operation not permitted”
# Description of problem

Fedora 32 beta, Kernel 5.6.3-300. Using kata-runtime from Fedora repo (1.11.0-alpha1) and encountered #2542, so forked the [upstream package spec](https://src.fedoraproject.org/rpms/kata-runtime/tree/f32) and applied the patch for that fix. This problem was encountered on that as well as changing the package spec to pull directly from commit af24829c2ae78c3b811c4cff6736ddaba500d37c. Both encounter the problem described.

# Expected result

`sudo podman run -it --rm --cap-add=ALL --runtime=kata-runtime --device /dev/vfio/72 fedora`
Container starts and has device attached.

# Actual result

`sudo podman run -it --rm --cap-add=ALL --runtime=kata-runtime --device /dev/vfio/72 fedora`
> Error: QMP command failed: vfio 0000:03:10.4: failed to open /dev/vfio/72: Operation not permitted: OCI runtime permission denied error

-----

This device being passed-through is an Intel 82599 Virtual Function from a X520, however I also encounter the same error with other devices bound to vfio-pci. I can successfully pass this device through to a VM using qemu-kvm via virt-manager, so IOMMU, etc., are configured correctly.

I am not using a custom kernel for the VM, as it doesn't seem like it should be necessary as this is not a large BAR device. I'd also expect if missing drivers were the only problem I'd still be able to start the container.

Starting a container using kata-runtime without attempting vfio passthrough is successful.

Various troubleshooting changes that had no effect:

- Adding `-v /dev:/dev` to `podman run`
- Disabling SELinux temporarily with `setenforce 0` (unsurprising as nothing in the audit logs indicates an SELinux problem)
- Setting `hotplug_vfio_on_root_bus = true`, `machine_type = "pc"`
- Setting `hotplug_vfio_on_root_bus = true`, `pcie_root_port = 1` (`machine_type = "q35"`, the default); fails (I think expectedly) with
   > Error: QMP command failed: Bus 'pcie.0' does not support hotplugging: OCI runtime error
- Running with `--privileged`, this fails instead with the following. Per logs, though, this seems to occur before attaching the vfio device, as there is no "Start hot-plug VFIO device" entry as before.
   > Error: QMP command failed: Could not open '/dev/loop0': Operation not permitted: OCI runtime permission denied error

devimc:
kata relies on the cgroup created and configured by container engine,
but sometimes the sandbox cgroup is not configured (podman) or the
cgroup implementation is very restrictive (cgroups v2) that the
workload may not have access to any device.
Use the initial (sandbox) cgroup configuration to have a functional
cgroup where the hypervisor can be started, the container engine (e.g k8s)
it's free to modify it later.

fixes #2605

### DIFF
--- a/virtcontainers/pkg/cgroups/manager.go
+++ b/virtcontainers/pkg/cgroups/manager.go
@@ -23,7 +23,6 @@ import (
 	"github.com/opencontainers/runc/libcontainer/specconv"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/sirupsen/logrus"
-	"golang.org/x/sys/unix"
 )
 
 type Config struct {
@@ -74,21 +73,7 @@ func UseSystemdCgroup() bool {
 
 // returns the list of devices that a hypervisor may need
 func hypervisorDevices() []specs.LinuxDeviceCgroup {
-	wildcard := int64(-1)
-	devicemapperMajor := int64(253)
-
 	devices := []specs.LinuxDeviceCgroup{}
-
-	devices = append(devices,
-		// hypervisor needs access to all devicemapper devices,
-		// since they can be hotplugged in the VM.
-		specs.LinuxDeviceCgroup{
-			Allow:  true,
-			Type:   "b",
-			Major:  &devicemapperMajor,
-			Minor:  &wildcard,
-			Access: "rwm",
-		})
 
 	// Processes running in a device-cgroup are constrained, they have acccess
 	// only to the devices listed in the devices.list file.
@@ -97,33 +82,16 @@ func hypervisorDevices() []specs.LinuxDeviceCgroup {
 	hypervisorDevices := []string{
 		"/dev/kvm",       // To run virtual machines
 		"/dev/vhost-net", // To create virtqueues
+		"/dev/vfio/vfio", // To access VFIO devices
 	}
 
 	for _, device := range hypervisorDevices {
-		var st unix.Stat_t
-		linuxDevice := specs.LinuxDeviceCgroup{
-			Allow:  true,
-			Access: "rwm",
-		}
-
-		if err := unix.Stat(device, &st); err != nil {
-			cgroupsLogger.WithError(err).WithField("device", device).Warn("Could not get device information")
+		ldevice, err := DeviceToLinuxDevice(device)
+		if err != nil {
+			cgroupsLogger.WithError(err).Warnf("Could not get device information")
 			continue
 		}
-
-		switch st.Mode & unix.S_IFMT {
-		case unix.S_IFCHR:
-			linuxDevice.Type = "c"
-		case unix.S_IFBLK:
-			linuxDevice.Type = "b"
-		}
-
-		major := int64(unix.Major(st.Rdev))
-		minor := int64(unix.Minor(st.Rdev))
-		linuxDevice.Major = &major
-		linuxDevice.Minor = &minor
-
-		devices = append(devices, linuxDevice)
+		devices = append(devices, ldevice)
 	}
 
 	return devices
@@ -134,8 +102,7 @@ func New(config *Config) (*Manager, error) {
 	var err error
 	useSystemdCgroup := UseSystemdCgroup()
 
-	devices := []specs.LinuxDeviceCgroup{}
-	copy(devices, config.Resources.Devices)
+	devices := config.Resources.Devices
 	devices = append(devices, hypervisorDevices()...)
 	// Do not modify original devices
 	config.Resources.Devices = devices
--- a/virtcontainers/pkg/cgroups/manager.go
+++ b/virtcontainers/pkg/cgroups/manager.go
@@ -319,3 +319,41 @@ func (m *Manager) Destroy() error {
 	defer m.Unlock()
 	return m.mgr.Destroy()
 }
+
+// AddDevice adds a device to the device cgroup
+func (m *Manager) AddDevice(device string) error {
+	cgroups, err := m.GetCgroups()
+	if err != nil {
+		return err
+	}
+
+	ld, err := DeviceToCgroupDevice(device)
+	if err != nil {
+		return err
+	}
+
+	m.Lock()
+	cgroups.Devices = append(cgroups.Devices, ld)
+	m.Unlock()
+
+	return m.Apply()
+}
+
+// RemoceDevice removed a device from the device cgroup
+func (m *Manager) RemoveDevice(device string) error {
+	cgroups, err := m.GetCgroups()
+	if err != nil {
+		return err
+	}
+
+	m.Lock()
+	for i, d := range cgroups.Devices {
+		if d.Path == device {
+			cgroups.Devices = append(cgroups.Devices[:i], cgroups.Devices[i+1:]...)
+			m.Unlock()
+			return m.Apply()
+		}
+	}
+	m.Unlock()
+	return fmt.Errorf("device %v not found in the cgroup", device)
+}
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -123,9 +123,6 @@ type SandboxConfig struct {
 
 	DisableGuestSeccomp bool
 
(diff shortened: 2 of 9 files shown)


fidencio:
@amorenoz, would you mind to give this one a try, as you've already done some work with vfio + kata + qemu and may have a setup handy?

jodh-intel:
Restarted the following failing CI's:

- OpenSUSE CI failure:
  ```sh
  not ok 23 privileged ctr device add
  # (in test file ctr.bats, line 890)
  ```
- vsock CI failure:
  ```sh
  not ok 1 test netmon
  # (in test file integration/netmon/netmon_test.bats, line 76)
  #   `[ "$before_interfaces" -eq "$final_interfaces" ]' failed
  # 
  # 4d0b0151cde4814759c897b288128567d756c9a122baefb636729c3487806ebc
  # e912b902a8b4ae67e8187bd1f09e8c8bf354e7b985bc20a18f825ea767cca5bb
  # containerA
  # containerA
  # test
  # 
  ```

- containerd CI failure: timed out.

fidencio:
Today I've talked to @amorenoz and he was not able to test it due to some other issues on his system. Please, do not get blocked by us to merge this one.

We'll ensure a test will be done and the situation is fixed before the final 1.11.0 release.

virtcontainers/sandbox.go
```diff
@@ -641,12 +638,14 @@ func (s *Sandbox) createCgroupManager() error {
...
-		if !s.config.HasCRIContainerType {
-			resources = *spec.Linux.Resources
-		}
+		// kata relies on the cgroup created and configured by container engine,
+		// but sometimes the sandbox cgroup is not configured (podman) or the
+		// cgroup implementation is very restrictive (cgroups v2) that the
+		// workload may not have access to any device.
+		// Use the initial (sandbox) cgroup configuration to have a functional
+		// cgroup where the hypervisor can be started, the container engine (e.g k8s)
+		// it's free to modify it later.
```
fidencio, line 647: nitpick: it's -> is (also in the commit message)

devimc:
@fidencio change applied, thanks

devimc:
ARM CIs failing, nothing related to this PR

cc @Pennyzct @jongwu 

```
15:16:04 debconf: unable to initialize frontend: Dialog
15:16:04 debconf: (TERM is not set, so the dialog frontend is not usable.)
15:16:04 debconf: falling back to frontend: Readline
15:16:04 Configuring tzdata
15:16:04 ------------------
15:16:04 
15:16:04 Please select the geographic area in which you live. Subsequent configuration
15:16:04 questions will narrow this down by presenting a list of cities, representing
15:16:04 the time zones in which they are located.
15:16:04 
15:16:04   1. Africa      4. Australia  7. Atlantic  10. Pacific  13. Etc
15:16:04   2. America     5. Arctic     8. Europe    11. SystemV
15:16:04   3. Antarctica  6. Asia       9. Indian    12. US
15:16:04 Geographic area: Build timed out (after 10 minutes). Marking the build as aborted.
15:26:04 Build was aborted
15:26:04 Performing Post build task...
15:26:04 Match found for :.* : True
15:26:04 Logical operation result is TRUE
15:26:04 Running script  : #!/bin/bash
````

**Comment by Pennyzct** [comment]
@devimc thanks for the heads-up. 
I've been trying to fix it on https://github.com/kata-containers/osbuilder/pull/449. Let debian/ubuntu install packages in non-interactive mode. ;)

**Comment by devimc** [comment]
thanks @Pennyzct for fixing this :+1:

**Comment by jodh-intel** [comment]
Looks like something borked the docker compose tests:

```
bats docker-compose.bats
21:06:29 1..1
21:06:29 # bats warning: Executed 0 instead of expected 1 tests
21:06:29 make: *** [Makefile:216: docker-compose] Error 1
```

**Comment by devimc** [comment]
ready to be merged :smile:

**Comment by fidencio** [comment]
merged, thanks!

File: virtcontainers/sandbox.go [hunk 2]
```diff
@@ -641,11 +641,22 @@ func (s *Sandbox) createCgroupManager() error {
...
-		// container engine *only* if actual container was
-		// marked *explicitly* as sandbox through annotations.
-		if !s.config.HasCRIContainerType {
-			resources = *spec.Linux.Resources
+		// Kata relies on the cgroup parent created and configured by the container
+		// engine, but sometimes the sandbox cgroup is not configured and the container
+		// may have access to all the resources, hence the runtime must constrain the
+		// sandbox and update the list of devices with the devices hotplugged in the
+		// hypervisor.
+		resources = *spec.Linux.Resources
```
amshinde, line 646: I know this PR has been already merged. But I would like to get some background about this change, in what circumstances are you seeing that the sandbox cgroup is not configured? 
So regardless if this is the sandbox or container type, we now apply constraints on resources?

virtcontainers/pkg/cgroups/manager.go
```diff
@@ -74,56 +73,25 @@ func UseSystemdCgroup() bool {
...
 func hypervisorDevices() []specs.LinuxDeviceCgroup {
-	wildcard := int64(-1)
-	devicemapperMajor := int64(253)
-
 	devices := []specs.LinuxDeviceCgroup{}
 
-	devices = append(devices,
-		// hypervisor needs access to all devicemapper devices,
-		// since they can be hotplugged in the VM.
-		specs.LinuxDeviceCgroup{
```
amshinde, line 85: Another question I had was the reason for removing the devicemapper devices here?
What if the container was using devicemapper storage, will it be access the device node?

virtcontainers/sandbox.go
devimc: > in what circumstances are you seeing that the sandbox cgroup is not configured?

docker and podman

> So regardless if this is the sandbox or container type, we now apply constraints on resources?

only when `sandbox_cgroup_only=true`

virtcontainers/pkg/cgroups/manager.go
devimc: > @devimc Another question I had was the reason for removing the devicemapper devices here?

device cgroup is updated before adding the device to the sandbox, so access to all the devicemapper devices is no longer needed

> What if the container was using devicemapper storage, will it be access the device node?

yes, when the devices node is added to the sandbox

amorenoz:
Although being very late (sorry about that), I and can confirm this PR does fix the issue for a Intel X710 VF

devimc:
thanks @amorenoz

egernst:
@devimc -- I think  https://github.com/kata-containers/runtime/commit/389b374e1de925c7a9e137da7c59586c2671772f will overconstrain the sandbox in the case of k8s.  You are constraining the sandbox more than kubelet, since AFAIU you aren't taking into account PodOverhead?

devimc:
@egernst you're right, but the problem is not easy to fix.

The following condition works for docker and k8s, because a docker container doesn't have the `CRIContainerType` annotation, so the constrain is applied, but a podman container *has that annotation* and the constrain is not applied, the podman container runs without constrains (bug!)

```go
if !s.config.HasCRIContainerType {
	resources = *spec.Linux.Resources
```

I guess the solution is to check whether the podoverhead annotation was set (does it exist?) and don't apply *memory* and *cpu* constrains. Any thoughts about how to do this?

amshinde:
@devimc @egernst I havent delved into the details here, but if podman is the outlier here, perhaps we should try to make podman behave as docker/k8s.

egernst:
There isn't an annotation -- it isn't a 'type' or annotation that is passed to the CRI.

egernst:
I would think it makes most sense to update the VC cgroup-manager package to allow for a way to differentiate in Apply() on whether it should apply all resources, or just the ones that make sense in the Kata+K8S case (ie, cpuset, and maybe devices?)

egernst:
Further, I don't think we should be doing it at this point anyway; probably when the container is being added?

Issue #2439, opened 2020-02-04
clh: keep the stdout/stderr open for CLH until its process exits


likebreath:
**Which feature do you think can be improved?**

The `kata-runtime` now captures `cloud-hypervisor` (CLH) output (stdout/stderr) to a local buffer `cloudHypervisor::cmdOutput` when launching CLH. The `kata-runtime` will exit when CLH launched successfully, and will leave CLH's output closed (not redirected) afterwards.

This issue can cause unexpected behaviors from CLH, e.g. causing CLH (child thread) panic while CLH tries to log warnings (to stderr) and stderr is not open (https://github.com/cloud-hypervisor/cloud-hypervisor/pull/703).

**How can it be improved?**

One potential solution is, instead of redirecting to a local buffer that lives with `kata-runtime`, CLH's output can be redirected to a file, e.g. `/dev/null` for production, or a log file for debugging.

**Additional Information**

Please share your thoughts and advises.

grahamwhaley:
I'm prone to say you might want to capture the output, even in a production deployment. The end users might want to import those into their logging stack for post-analysis etc.
I don't know the wrinkles of how/where is going to be best to re-direct/capture/push that data in the kata runtime stack - but, I reckon @jodh-intel will have an idea (basically, whatever route we do today for qemu etc.).

amshinde:
Based on the debug settings for the hypervisor in our config.toml we can either have this redirected to a log file or /dev/null when debug is not enabled.

PR #2648, opened 2020-04-27
clh: Improve logging for cloud hypervisor

jcvenegas:
Send virtiofsd logs to syslog in the same way that qemu implementation
does. This requires not to wait for messages from virtiofsd stdout. This
takes the qemu implementation approach. Give the socket fd to virtiofsd.

Use systemd-cat to collect hypervisor output. The `systemd-cat` program
will open a journal fd and call `cat(1)` to redirect all the output to
the fd. This requires an extra binary to read from hypervisor stdout
(that has combined stdin, stderr and serial terminal). But because it is
cat the overhead is minimal and only is started on Kata debug mode.

Fixes: #2439

### DIFF
--- a/virtcontainers/clh.go
+++ b/virtcontainers/clh.go
@@ -6,10 +6,11 @@
 package virtcontainers
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
@@ -59,9 +60,7 @@ const (
 	clhStopSandboxTimeout = 3
 	clhSocket             = "clh.sock"
 	clhAPISocket          = "clh-api.sock"
-	clhLogFile            = "clh.log"
 	virtioFsSocket        = "virtiofsd.sock"
-	clhSerial             = "serial-tty.log"
 	supportedMajorVersion = 0
 	supportedMinorVersion = 5
 	defaultClhPath        = "/usr/local/bin/cloud-hypervisor"
@@ -120,7 +119,6 @@ type cloudHypervisor struct {
 	APIClient clhClient
 	version   CloudHypervisorVersion
 	vmconfig  chclient.VmConfig
-	cmdOutput bytes.Buffer
 	virtiofsd Virtiofsd
 	store     persistapi.PersistDriver
 }
@@ -286,13 +284,8 @@ func (clh *cloudHypervisor) createSandbox(ctx context.Context, id string, networ
 
 	// set the serial console to the cloud hypervisor
 	if clh.config.Debug {
-		serialPath, err := clh.serialPath(clh.id)
-		if err != nil {
-			return err
-		}
 		clh.vmconfig.Serial = chclient.ConsoleConfig{
-			Mode: cctFILE,
-			File: serialPath,
+			Mode: cctTTY,
 		}
 
 	} else {
@@ -368,17 +361,12 @@ func (clh *cloudHypervisor) startSandbox(timeout int) error {
 	var strErr string
 	strErr, pid, err := clh.LaunchClh()
 	if err != nil {
-		return fmt.Errorf("failed to launch cloud-hypervisor: %s, error messages from log: %s", err, strErr)
-	}
-	clh.state.PID = pid
-
-	if err := clh.waitVMM(clhTimeout); err != nil {
-		clh.Logger().WithField("error", err).WithField("output", clh.cmdOutput.String()).Warn("cloud-hypervisor init failed")
 		if shutdownErr := clh.virtiofsd.Stop(); shutdownErr != nil {
 			clh.Logger().WithField("error", shutdownErr).Warn("error shutting down Virtiofsd")
 		}
-		return err
+		return fmt.Errorf("failed to launch cloud-hypervisor: %q, hypervisor output:\n%s", err, strErr)
 	}
+	clh.state.PID = pid
 
 	if err := clh.bootVM(ctx); err != nil {
 		return err
@@ -771,16 +759,8 @@ func (clh *cloudHypervisor) vsockSocketPath(id string) (string, error) {
 	return utils.BuildSocketPath(clh.store.RunVMStoragePath(), id, clhSocket)
 }
 
-func (clh *cloudHypervisor) serialPath(id string) (string, error) {
-	return utils.BuildSocketPath(clh.store.RunVMStoragePath(), id, clhSerial)
-}
-
 func (clh *cloudHypervisor) apiSocketPath(id string) (string, error) {
 	return utils.BuildSocketPath(clh.store.RunVMStoragePath(), id, clhAPISocket)
-}
-
-func (clh *cloudHypervisor) logFilePath(id string) (string, error) {
-	return utils.BuildSocketPath(clh.store.RunVMStoragePath(), id, clhLogFile)
 }
 
 func (clh *cloudHypervisor) waitVMM(timeout uint) error {
@@ -873,8 +853,6 @@ func (clh *cloudHypervisor) getAvailableVersion() error {
 
 func (clh *cloudHypervisor) LaunchClh() (string, int, error) {
 
-	errStr := ""
-
 	clhPath, err := clh.clhPath()
 	if err != nil {
 		return "", -1, err
@@ -882,36 +860,71 @@ func (clh *cloudHypervisor) LaunchClh() (string, int, error) {
 
 	args := []string{cscAPIsocket, clh.state.apiSocket}
 	if clh.config.Debug {
-
-		logfile, err := clh.logFilePath(clh.id)
-		if err != nil {
-			return "", -1, err
-		}
-		args = append(args, cscLogFile)
-		args = append(args, logfile)
+		// Cloud hypervisor log levels
+		// 'v' occurrences increase the level
+		//0 =>  Error
+		//1 =>  Warn
+		//2 =>  Info
+		//3 =>  Debug
+		//4+ => Trace
+		// Use Info, the CI runs with debug enabled
+		// a high level of logging increases the boot time
+		// and in a nested environment this could increase
+		// the chances to fail because agent is not
+		// ready on time.
+		args = append(args, "-vv")
 	}
 
 	clh.Logger().WithField("path", clhPath).Info()
 	clh.Logger().WithField("args", strings.Join(args, " ")).Info()
 
-	cmd := exec.Command(clhPath, args...)
-	cmd.Stdout = &clh.cmdOutput
-	cmd.Stderr = &clh.cmdOutput
+	cmdHypervisor := exec.Command(clhPath, args...)
+	var hypervisorOutput io.ReadCloser
+	if clh.config.Debug {
+		cmdHypervisor.Env = os.Environ()
+		cmdHypervisor.Env = append(cmdHypervisor.Env, "RUST_BACKTRACE=full")
+		// Get  StdoutPipe only for debug, without debug golang will redirect to /dev/null
+		hypervisorOutput, err = cmdHypervisor.StdoutPipe()
+		if err != nil {
+			return "", -1, err
+		}
+	}
+
+	cmdHypervisor.Stderr = cmdHypervisor.Stdout
+
+	err = utils.StartCmd(cmdHypervisor)
+	if err != nil {
+		return "", -1, err
+	}
+
+	if err := clh.waitVMM(clhTimeout); err != nil {
+		clh.Logger().WithField("error", err).Warn("cloud-hypervisor init failed")
+		var output string
+
+		if hypervisorOutput != nil {
+			b, errRead := ioutil.ReadAll(hypervisorOutput)
+			if errRead != nil {
+				output = "failed to read hypervisor output to get error information"
+			} else {
+				output = string(b)
+			}
+		} else {
+			output = "Please enable hypervisor logging to get stdout information"
+		}
+
+		return output, -1, err
+	}
 
 	if clh.config.Debug {
-		cmd.Env = os.Environ()
-		cmd.Env = append(cmd.Env, "RUST_BACKTRACE=full")
-	}
-
-	if err := utils.StartCmd(cmd); err != nil {
-		fmt.Println("Error starting cloudHypervisor", err)
-		if cmd.Process != nil {
-			cmd.Process.Kill()
+		cmdLogger := utils.NewProgramLogger("kata-hypervisor")
+		clh.Logger().Debugf("Starting process logger(%s) for hypervisor", cmdLogger)
+		if err := cmdLogger.StartLogger(hypervisorOutput); err != nil {
+			// Not critical to run a container, but output wont be logged
+			clh.Logger().Warnf("Failed start process logger(%s) %s", cmdLogger, err)
 		}
-		return errStr, 0, err
 	}
 
-	return errStr, cmd.Process.Pid, nil
+	return "", cmdHypervisor.Process.Pid, nil
 }
 
 //###########################################################################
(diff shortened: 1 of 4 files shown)


jcvenegas:
Example,deliberately add wrong option value to hypervisor args
```
docker run -ti --rm --runtime kata-runtime fedora bash
docker: Error response from daemon: OCI runtime create failed: failed to launch cloud-hypervisor: 'Failed to connect to API (timeout 10s): error: Get http://localhost/api/v1/vmm.ping: dial unix /run/vc/vm/1cea95b732c21488e0bbe792bd3107ce9fcbd4a0498ecd1d05650d89cf8fbda9/clh-api.sock: connect: no such file or directory reason: ', hypervisor output:
cloud-hypervisor v0.6.0-287-g3eaeba4
The Cloud Hypervisor Authors
Launch a cloud-hypervisor VMM.

USAGE:
    cloud-hypervisor [FLAGS] [OPTIONS]

FLAGS:
    -h, --help       Prints help information
    -v               Sets the level of debugging output
    -V, --version    Prints version information
...

```

Now running correctly, but going to a guest going to panic because of the wrong kernel version:

```
docker run -ti --rm --runtime kata-runtime fedora bash
docker: Error response from daemon: OCI runtime create failed: Failed to check if grpc server is working: context deadline exceeded: unknown.
```

```
journalctl -t kata-hypervisor -f --output-fields=MESSAGE -o cat
...
[    2.180980]  ? printk+0x3e/0x46
[    2.183955]  mount_block_root+0x27e/0x331
[    2.187697]  prepare_namespace+0x15b/0x18a
[    2.190992]  ? rest_init+0xa0/0xa0
[    2.193404]  kernel_init+0x5/0x100
[    2.196081]  ret_from_fork+0x35/0x40
[    2.199523] Kernel Offset: disabled
[    2.202846] Rebooting in 1 seconds..
VMM thread failed VmReboot(DeviceManager(CreateVirtioFs(VhostUserCreateMaster(VhostUserProtocol(SocketConnect(Os { code: 111, kind: ConnectionRefused, message: "Connection refused" }))))))

```

Filter cloud-hypervisor logs:
```
journalctl -t kata-hypervisor -f --output-fields=MESSAGE -o cat | grep 'cloud-hypervisor:'
cloud-hypervisor: 11.392470421s: DEBUG:vm-virtio/src/vsock/device.rs:161 -- vsock: epoll_handler::process_tx()
cloud-hypervisor: 11.392545025s: DEBUG:vm-virtio/src/vsock/unix/muxer.rs:201 -- vsock: muxer.send[rxq.len=0]: [3, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 4, 0, 0, 126, 3, 0, 64, 47, 0, 0, 0, 1, 0, 5, 0, 0, 0, 0, 0, 0, 0, 4, 0, 225, 3, 0, 0]
cloud-hypervisor: 11.392719935s: DEBUG:vm-virtio/src/vsock/device.rs:105 -- vsock: raising IRQ
cloud-hypervisor: 11.39280674s: DEBUG:vm-virtio/src/vsock/device.rs:337 -- vsock: backend event
cloud-hypervisor: 11.392832341s: DEBUG:vm-virtio/src/vsock/unix/muxer.rs:285 -- vsock: muxer received kick
cloud-hypervisor: 11.392862943s: DEBUG:vm-virtio/src/vsock/unix/muxer.rs:357 -- vsock: muxer processing event: fd=123, evset=EPOLLIN
cloud-hypervisor: 11.392905445s: DEBUG:vm-virtio/src/vsock/device.rs:161 -- vsock: epoll_handler::process_tx()
cloud-hypervisor: 11.392946648s: DEBUG:vm-virtio/src/vsock/device.rs:119 -- vsock: epoll_handler::process_rx()
cloud-hypervisor: 11.393017052s: DEBUG:vm-virtio/src/vsock/unix/muxer.rs:178 -- vsock muxer: RX pkt: [2, 0, 0, 0, 0, 0, 0, 0, 3, 0, 0, 0, 0, 0, 0, 0, 126, 3, 0, 64, 0, 4, 0, 0, 17, 0, 0, 0, 1, 0, 5, 0, 0, 0, 0, 0, 0, 0, 1, 0, 136, 1, 0, 0]
cloud-hypervisor: 11.393156059s: DEBUG:vm-virtio/src/vsock/device.rs:105 -- vsock: raising IRQ
```

virtcontainers/clh.go
```diff
@@ -877,45 +856,61 @@ func (clh *cloudHypervisor) getAvailableVersion() error {
...
+	cmdHypervisor := exec.Command(clhPath, args...)
 
 	if clh.config.Debug {
-		cmd.Env = os.Environ()
-		cmd.Env = append(cmd.Env, "RUST_BACKTRACE=full")
+		cmdHypervisor.Env = os.Environ()
+		cmdHypervisor.Env = append(cmdHypervisor.Env, "RUST_BACKTRACE=full")
+	}
+
+	hypervisoOutput, err := cmdHypervisor.StdoutPipe()
```
likebreath, line 879: Nit: `hypervisoOutput` -> `hypervisorOutput`

virtcontainers/clh.go
```diff
@@ -877,45 +856,61 @@ func (clh *cloudHypervisor) getAvailableVersion() error {
...
+	if err != nil {
+		return "", -1, err
 	}
+	cmdHypervisor.Stderr = cmdHypervisor.Stdout
```
likebreath, line 883: Also a question: is it safe to open a pipe with `stdoutPipe()` but not read from it (when debug is off)?

It looks like we are ignoring the stdout/stderr of CLH anyway when debug is off. Will it be the same if we open the pipe and redirect `cmdHypervisor.Stdour/Stderr` to it only when debug is on ( and do noting to the stderr/out when debug is off)?

virtcontainers/virtiofsd_test.go
```diff
@@ -64,6 +65,7 @@ func TestVirtiofsdStart(t *testing.T) {
 				},
 			}
 			var ctx context.Context
+			err = os.MkdirAll(filepath.Dir(v.socketPath), testDirMode)
```
jodh-intel, line 68: As shown on the PR, passing invalid args would be a good test that this is working. Could you find a way to create a unit test for that to guarantee debugging always works (distressing if it doesn't - when you need it, you need it ;)

virtcontainers/clh.go
```diff
@@ -877,45 +856,61 @@ func (clh *cloudHypervisor) getAvailableVersion() error {
...
 	args := []string{cscAPIsocket, clh.state.apiSocket}
 	if clh.config.Debug {
-
-		logfile, err := clh.logFilePath(clh.id)
-		if err != nil {
-			return "", -1, err
-		}
-		args = append(args, cscLogFile)
-		args = append(args, logfile)
+		args = append(args, "-vvv")
```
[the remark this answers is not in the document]
sboeuf: There's no such thing as `--verbose-max` for now (maybe that's something we can add in the future). The way it works, each `v` gives you more logs, and over 3 `v`s, you get the `Trace` type of logs.
This is not expected to change, but we didn't reach 1.0.0 yet, so for the time being, we'll have to adapt Kata is anything changes in Cloud-Hypervisor (which of course we're trying to avoid as much as possible :))

virtcontainers/virtiofsd_test.go
jcvenegas: Actually, related with this specific case, I needed to create the directory before if not the start function will fail because the runtime will fail open the socket for virtiofsd, in the use case the directory already exists, I wonder but I wonder if the directory does not exist the function should create before open the socket, what do you think?

virtcontainers/clh.go
jcvenegas: It is a good question, I think the system has a buffer where it dumps data, in the case when is ignore I imagine new prints to stdout, stderr will get a try again error, not sure how rust handle this. @sboeuf any suggestion here?

virtcontainers/utils/utils.go
```diff
@@ -275,3 +276,23 @@ const (
 	MiB          = KiB << 10
 	GiB          = MiB << 10
 )
+
+// Binary to use to log program output
+const LoggerBinryName = "systemd-cat"
```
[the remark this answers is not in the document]
jcvenegas: It will print the output in the format of cloud hypervisor stdout, I am afraid is not structured logged. 
```
journalctl -t kata-hypervisor -f
Apr 28 21:11:15 jcvenega kata-hypervisor[102976]: cloud-hypervisor: 10.342620707s: DEBUG:vm-virtio/src/vsock/device.rs:161 -- vsock: epoll_handler::process_tx()
Apr 28 21:11:15 jcvenega kata-hypervisor[102976]: cloud-hypervisor: 10.342638908s: DEBUG:vm-virtio/src/vsock/device.rs:119 -- vsock: epoll_handler::process_rx()
Apr 28 21:11:15 jcvenega kata-hypervisor[102976]: cloud-hypervisor: 10.34267121s: DEBUG:vm-virtio/src/vsock/unix/muxer.rs:178 -- vsock muxer: RX pkt: [2, 0, 0, 0, 0, 0, 0, 0, 3, 0, 0, 0, 0, 0, 0, 0, 51, 4, 0, 64, 0, 4, 0, 0, 17, 0, 0, 0, 1, 0, 5, 0, 0, 0, 0, 0, 0, 0, 1, 0, 231, 1, 0, 0]
Apr 28 21:11:15 jcvenega kata-hypervisor[102976]: cloud-hypervisor: 10.342782016s: DEBUG:vm-virtio/src/vsock/device.rs:105 -- vsock: raising IRQ
```

virtcontainers/virtiofsd.go
```diff
@@ -125,11 +150,23 @@ func (v *virtiofsd) args() ([]string, error) {
...
 		"-f",
+		// cache mode for virtiofsd
 		"-o", "cache=" + v.cache,
+		// disable posix locking in daemon: bunch of basic posix locks properties are broken
+		// apt-get update is broken if enabled
 		"-o", "no_posix_lock",
+		// shared directory tree
 		"-o", "source=" + v.sourcePath,
-		"-o", "vhost_user_socket=" + v.socketPath,
+		// fd number of vhost-user socket
```
amshinde, line 164: why vhost-user, isnt FdSocketNumber  just a unix socket?

virtcontainers/virtiofsd.go
```diff
@@ -125,11 +150,23 @@ func (v *virtiofsd) args() ([]string, error) {
...
+		// foreground operation
 		"-f",
+		// cache mode for virtiofsd
 		"-o", "cache=" + v.cache,
+		// disable posix locking in daemon: bunch of basic posix locks properties are broken
+		// apt-get update is broken if enabled
 		"-o", "no_posix_lock",
+		// shared directory tree
 		"-o", "source=" + v.sourcePath,
-		"-o", "vhost_user_socket=" + v.socketPath,
```
amshinde, line 132: Not looked into what `vhost_user_socket` was used for, but why is this removed here ?

virtcontainers/utils/utils.go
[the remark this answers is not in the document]
jcvenegas: @jodh-intel @sboeuf I wonder if it worth I take a look to structured log from kata-agent in rust( I think now it supports it, rigth?) and I try to use it in cloud-hypervisor

virtcontainers/clh.go
```diff
@@ -877,45 +856,61 @@ func (clh *cloudHypervisor) getAvailableVersion() error {
...
 	}
+	cmdHypervisor.Stderr = cmdHypervisor.Stdout
 
-	if err := utils.StartCmd(cmd); err != nil {
-		fmt.Println("Error starting cloudHypervisor", err)
-		if cmd.Process != nil {
-			cmd.Process.Kill()
+	err = utils.StartCmd(cmdHypervisor)
+	if err != nil {
+		return "Failed to start hypervisor", 0, errors.Errorf("Failed to start hypervisor: %s", err)
```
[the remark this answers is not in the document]
jcvenegas: Thanks, agree with `-1`, I removed the message from `output`

virtcontainers/clh.go
```diff
@@ -372,17 +364,12 @@ func (clh *cloudHypervisor) startSandbox(timeout int) error {
 	var strErr string
 	strErr, pid, err := clh.LaunchClh()
 	if err != nil {
-		return fmt.Errorf("failed to launch cloud-hypervisor: %s, error messages from log: %s", err, strErr)
-	}
-	clh.state.PID = pid
-
-	if err := clh.waitVMM(clhTimeout); err != nil {
```
amshinde, line 379: `waitVMM` is no longer required here?

virtcontainers/utils/utils.go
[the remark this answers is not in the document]
jodh-intel: Yes, the new rust agent already uses structured logging. If clh wants to, they could just grab:

- https://github.com/kata-containers/kata-containers/tree/master/src/agent/logging
  (see also https://github.com/kata-containers/kata-containers/issues/46).

virtcontainers/virtiofsd_test.go
jodh-intel: Sorry - I'm not commenting on this `mkdir` specifically, I'm just saying that if it's possible, it would be good to add a test to ensure debug output is actually captured from clh/virtiofs. Using some invalid params is one way, but there may be better ones ;)

virtcontainers/virtiofsd.go
jcvenegas: we have two options to provide the socket path and virtiofsd will create the socket or provide the file descriptor. Because no more stdout check is done by the runtime we require that the hypervisor has access to the socket when it starts. Using the `FD` let us do that without keeping checking until the socket appears.

virtcontainers/virtiofsd.go
jcvenegas: Yes it is both, it is unix socket for vhost user connection, I wanted to keep the same as virtiofsd help `virtiofsd --help`

virtcontainers/clh.go
jcvenegas: It was moved to [launchClh](https://github.com/kata-containers/runtime/pull/2648/files#diff-f1c6a53125ae38d7dbb5756ebea5771fR898) if fails to wait for the hypervisor, it read the output of the command and return it as part of the error output.

likebreath:
Seems to be a timeout for a k8s test (because of lots of logging)?
```
09:36:32 not ok 1 Check PID namespaces
09:36:32 # (in test file k8s-pid-ns.bats, line 25)
09:36:32 #   `kubectl wait --for=condition=Ready pod "$pod_name"' failed
09:36:32 # INFO: k8s configured to use runtimeclass
09:36:32 # pod/busybox created
09:36:32 # error: timed out waiting for the condition on pods/busybox
09:36:32 # pod "busybox" deleted
09:36:32 Failed at 90: bats "${K8S_TEST_ENTRY}"
```

jcvenegas:
@likebreath thanks for spot that, yeah that is what I am looking I printed both kata-runtime and hypervisor logs to see if just timeout something  else

likebreath:
All green. Ready to be merged? 😄

devimc:
ehmm.. clh CI is not happy :sweat_smile:

jcvenegas:
I have tried to re-run to confirm was OK, I feel tha PR could bring some inestability to the CI, I need more time to confirm if the CI itself is not stable or this PR could make it worst.

jcvenegas:
Ready to be merged found that what causes the instability is a missing kernel patch. I added a depends-on to test the PR and seems to be OK. 

Update: 
- I reduced the logging of the guest if needed we can still add it manually.
- Changed to get hypervisor stdout only on debug mode also to get initial setup, to avoid leave the stdout open but not be readed in that case golang will redirect to `/dev/null`

grahamwhaley:
well, we have enough acks and @jcvenegas is happy with it... I'm going to press the big green button.... I think all feedback has been addressed.....

PR #2652, opened 2020-04-28
Support microvm machine type 

merwick:


pkg/katautils/config.go
```diff
@@ -595,6 +596,13 @@ func newQemuHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 	kernelParams := h.kernelParams()
 	machineType := h.machineType()
 
+	// The "microvm" machine type doesn't support NVDIMM so override the
+	// config setting to explicitly disable it (i.e. don't require the
+	// user to add 'disable_image_nvdimm = true' in the .toml file).
+	if machineType == govmmQemu.MachineTypeMicrovm {
+		h.DisableImageNvdimm = true
```
devimc, line 603: I think this is fine, but could you add a log?

merwick: Sure. I'll make the changes below (adding a check for machineType so that a message is only printed when it's overridden).

```
--- a/pkg/katautils/config.go
+++ b/pkg/katautils/config.go
@@ -599,8 +599,10 @@ func newQemuHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
        // The "microvm" machine type doesn't support NVDIMM so override the
        // config setting to explicitly disable it (i.e. don't require the
        // user to add 'disable_image_nvdimm = true' in the .toml file).
-       if machineType == govmmQemu.MachineTypeMicrovm {
+       if machineType == govmmQemu.MachineTypeMicrovm &&
+               h.DisableImageNvdimm == false {
                h.DisableImageNvdimm = true
+               kataUtilsLogger.Info("Setting 'disable_image_nvdimm = true' as microvm does not support NVDIMM")
        }
 
        blockDriver, err := h.blockDeviceDriver()

```

merwick:
[ I have moved text from Description to this comment as the jenkins-ubuntu-18-04-vfio test seemed to choke on in when parsing ghprbPullLongDescription ]

I have vendored in the intel/govmm#121 changes (I'm not sure I should do that but I needed them as a dependency for my kata-runtime changes).

With these changes, I can specify 'machine_type = "microvm"' in /etc/kata-containers/configuration.toml and see a 170ms (13%) improvement in startup time compared with 'machine_type = "pc"' (using PVH also). Running kata metrics I see a 106ms (8%) improvement in startup time and a 6% reduction in memory footprint.

Since the only change needed to switch to using microvm, is changing the 'machine_type' key in configuration.toml, I chose not do deliver a new .toml file as it'd only result in future overhead maintaining an extra file.

jodh-intel:
The metrics CI is failing but we're aware of this issue (unrelated to this PR).

I've restarted the vsocks CI which hit a network timeout.

PR #2653, opened 2020-04-28
backport 1.11: fix podman vfio

devimc:


fidencio:
lgtm as well!

@amshinde, you raised some questions in the original PR, are those solved? shall we go ahead with this one as is or do you wanna see some changes there that should also be backported here?

Issue #1917, opened 2019-07-29
Sandbox creating is slower when use_vsock enabled


BetaXOi:
# Description of problem

Sandbox createing is slower when use vsock than virtio-serial.

Set 'enable_tracing = true' and create 10 kata containers with and without 'use_vsock=true', record the time of 'docker run' and opentracing span.
`for i in `seq 10`; do time docker run --rm --runtime kata-runtime-test -tid busybox sh; sleep 1; done`

# Expected result

The time of creating shoud be similar.

# Actual result

The time of using vsock is slower 2 second almost than virtio-serial. 
![image](https://user-images.githubusercontent.com/15086907/62033618-eeff9100-b21e-11e9-9eee-e64ef15ca047.png)

![image](https://user-images.githubusercontent.com/15086907/62032371-16a12a00-b21c-11e9-9833-6e439909c935.png)

BetaXOi:
The reason is 'connect' be invoked befor vsock is ready, and the timeout is 2 second in vsock kernel module.
`#define VSOCK_DEFAULT_CONNECT_TIMEOUT (2 * HZ)`

[...]

PR #2656, opened 2020-04-30
vsock: initiate a quick handshake to avoid VSOCK_DEFAULT_CONNECT_TIMEOUT

Pennyzct:
We will suffer default two seconds timeout, when kata-runtime initiates a connection to agent server,and right now vsock transport in guest part isn't ready.
Here, taking advices from Stefan Hajnoczi, we will do a quick handshake to ensure host initiating connection until vsock device is completely ready.
The quick handshake follows below steps:
1. kata-runtime listens on a vsock port
2. `agent.vsock_host_port=PORT` is added to the kernel command-line options
3. kata-agent parses the port number and connects to the host
This commit here covers the 1-2 parts in kata-runtime.

Fixes: #1917

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -36,6 +36,7 @@ import (
 	"github.com/kata-containers/runtime/virtcontainers/pkg/uuid"
 	"github.com/kata-containers/runtime/virtcontainers/types"
 	"github.com/kata-containers/runtime/virtcontainers/utils"
+	"github.com/mdlayher/vsock"
 )
 
 // romFile is the file name of the ROM that can be used for virtio-pci devices.
@@ -99,6 +100,9 @@ type qemu struct {
 	stopped bool
 
 	store persistapi.PersistDriver
+
+	vsockHostPort uint32
+	vsockConnect  chan bool
 }
 
 const (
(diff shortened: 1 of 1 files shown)


Pennyzct:
Hi guys
I haven't added unit tests here, so adding `wip` flag here.
I'm not sure whether giving the time out `2s` here is correct. 😰

Pennyzct:
Hi @stefanha
Plz feel free to review, hope I'm following your thinking here. ;)

virtcontainers/qemu.go
```diff
@@ -609,6 +626,35 @@ func (q *qemu) createSandbox(ctx context.Context, id string, networkNS NetworkNa
 	return nil
 }
 
+func (q *qemu) setVSockHostPort() error {
+	socket, err := unix.Socket(unix.AF_VSOCK, unix.SOCK_STREAM, 0)
+	if err != nil {
+		return err
+	}
+	defer syscall.Close(socket)
```
stefanha, line 634: There is a race condition here: we fetch the port number and then close the file descriptor.  Now another host process can reuse the same port number before we get a chance to call vsock.Listen().  When this happens our vsock.Listen() call will fail.

Please keep the file descriptor open and use it instead of calling vsock.Listen() later on.

virtcontainers/qemu.go
```diff
@@ -867,6 +937,17 @@ func (q *qemu) bootFromTemplate() error {
 	return q.waitMigration()
 }
 
+func (q *qemu) waitVSockReady() error {
+	select {
+	case <-q.vsockConnect:
+		// receive a connection from agent, that is,
+		// vsock device in guest is ready.
+		return nil
+	case <-time.After(vsockReadyTimeOut):
```
stefanha, line 946: Is it necessary to introduce a new timeout?  This timeout includes guest kernel boot and kata-agent startup, so 2 seconds might be a little short on a heavily loaded host.

I think a timeout isn't needed here since the container runtime fails container startup after some time anyway.

virtcontainers/qemu.go
Pennyzct: okay, thanks. I'll save the server socket somewhere for later use.

PR #2657, opened 2020-04-30
versions: update QEMU to 5.0.0

### LINKED ISSUE #2354: update qemu to 5.0.0

devimc:
New features that can improve/impact in kata containers:

x86

    VMX features can be enabled/disabled via the "-cpu" flag.
    When nested virtualization is enabled with an option like
    "-cpu Haswell,+vmx", the set of VMX features will also be constrained to
    what was available on the corresponding CPU model.
    New "microvm" machine type that has virtio-mmio instead of PCI, and no ACPI
    support (so no hotplug too). The new machine type is meant as a baseline
    for performance optimizations of QEMU, firmware and guests. While inspired
    by Firecracker it is not entirely compatible with it (for example it does
    not have Firecracker's userspace IP stack and MicroVM Metadata Service).
    Reduce memory footprint when booting uncompressed kernels.

ARM:

    We now correctly support more than 256 CPUs when using KVM
    The virt board now supports memory hotplugging, when used with a UEFI
    guest BIOS and ACPI.
    virtio-iommu is now supported with machvirt.
    The Cortex-M7 CPU is now supported

s390:

    Using KVM now explicitly requires a host kernel version of at least 3.15
    (which includes the 'flic' KVM device). This had been broken since QEMU
    2.10 already

Depends-on: github.com/kata-containers/tests#2496

fixes #2354

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -88,8 +88,8 @@ assets:
     qemu:
       description: "VMM that uses KVM"
       url: "https://github.com/qemu/qemu"
-      version: "4.1.1"
-      tag: "v4.1.1"
+      version: "5.0.0"
+      tag: "v5.0.0"
       # Do not include any non-full release versions
       # Break the line *without CR or space being appended*, to appease
       # yamllint, and note the deliberate ' ' at the end of the expression.
--- a/vendor/github.com/intel/govmm/qemu/qmp.go
+++ b/vendor/github.com/intel/govmm/qemu/qmp.go
@@ -993,7 +993,7 @@ func (q *QMP) ExecuteNetdevAddByFds(ctx context.Context, netdevType, netdevID st
 	}
 	if len(vhostFdNames) > 0 {
 		vhostFdNameStr := strings.Join(vhostFdNames, ":")
-		args["vhost"] = "on"
+		args["vhost"] = true
 		args["vhostfds"] = vhostFdNameStr
 	}
 


versions.yaml
```diff
@@ -88,8 +88,8 @@ assets:
     qemu:
       description: "VMM that uses KVM"
       url: "https://github.com/qemu/qemu"
-      version: "4.1.1"
-      tag: "v4.1.1"
+      version: "5.0.0"
```
jcvenegas, line 91: Do you expect that  both qemu versions will be added in obs, kata-deploy, snap?

devimc: nop, only 5.0.  4.1.1 will be used in the CI when the kernel version < 5.3.0

jcvenegas: Ok, as heads up, today we have automated testing for kata-deploy and obs packages in azure, those may be affected.

devimc: how they can be affected? you mean failures? or just a change in the qemu's version?

amshinde: @devimc If we go this route, it would also mean we would need to package both 4.1 and 5.0 qemu, as someone using Kata on Azure cloud may not be able to do so with the 5.0 qemu packages.

amshinde:
@devimc Lets hold this till we get some feedback on this from Azure folks. This reminds me I have to reach out to them.

versions.yaml
devimc: @amshinde I think we shouldn't deal with this, since this is bug between QEMU and hyperv, users have two options:
1. update Linux kernel to >=5.3.0
2. build and install QEMU 4.1.1 (if they don't like the first option)

jcvenegas: @devimc those would be affected in case we don't provide qemu 4.X as default for kata-deploy, or OBS. 
- Start the  CI for OBS/Kata-deploy in azure => CI fails because QEMU 4.0 is not used. 

>    update Linux kernel to >=5.3.0
>    build and install QEMU 4.1.1 (if they don't like the first option)

In this case, we are the users, in case we want to continue updating to 5.x, it needs to choose one of those options and implement it and none of those are really user friendly in my opinion :P

versions.yaml
```diff
@@ -104,6 +104,15 @@ assets:
           branch: "master"
           tag: "v3.1.0-rc2"
           commit: "47c1cc30e440860aa695358f7c2dd0b9d7b53d16"
+      hyperv:
+        description: |
+          Due to a missing KVM feature in hyperV is not possible to run
```
jodh-intel, line 109: OOI, is there a URL we can add here for a link to further details on the status of this missing feature in HyperV?

jcvenegas:
I just remember that qemu packaging would need an update of deps , I would be nice if you could open a PR in packaging with a depends-on and fix in parallel any build issue to avoid any future problem at release if this PR is merged.

bergwolf:
This would be a good candidate to go with the `2.0-dev` branch and we have decided to feature freeze for kata 1.x releases. Please re-submit it to https://github.com/kata-containers/kata-containers Thanks!

If it is strongly desired to include this in the 1.x kata releases like v1.12, please backport it from the 2.0-dev branch after it is merged there.

PR #2659, opened 2020-04-30
versions: Move to cloud-hypervisor v0.7.0

likebreath:
Cloud-hypervisor [v0.7.0](https://github.com/cloud-hypervisor/cloud-hypervisor/releases/tag/v0.7.0) was released earlier today. We definitely want to stay updated with it. Also, it is nicer to use the v0.7.0 tag in our versions.yaml for our next Kata release.

This patch also updated the use of `VsockConfig` to `Vmconfig` for booting the hypervisor. It reflects CLH's [changes](https://github.com/cloud-hypervisor/cloud-hypervisor/commit/10348f73e45fcfbd70f818e9dec2dc6d952febfb) on Vsock HTTP API and support.

Fixes: kata-containers#2658

Signed-off-by: Bo Chen <chen.bo@intel.com>

PR #2663, opened 2020-05-06
clh: backport: Move to v0.7.0

jcvenegas:
Changes:
2d251652 clh: vsock: Supply the right VsockConfig to Vmconfig
fcc9e93b versions: Move to cloud-hypervisor v0.7.0
34be9e0c clh: memory: remove pmem size argument
9798e8a5 versions: Move to latest cloud-hypervisor

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -75,7 +75,7 @@ assets:
       url: "https://github.com/cloud-hypervisor/cloud-hypervisor"
       uscan-url: >-
         https://github.com/cloud-hypervisor/cloud-hypervisor/tags.*/v?(\d\S+)\.tar\.gz
-      version: "a8ec8f3326628d34021ccae0f259a9509ff1e6da"
+      version: "f5debc4bc001fb14dad0aee28fef102e6c263565"
 
     firecracker:
       description: "Firecracker micro-VMM"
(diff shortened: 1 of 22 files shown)


jodh-intel:
Hi @jcvenegas - fwics, this is related to https://github.com/kata-containers/runtime/pull/2659. I can see this is a "nice to have" change, but is it actually necessary to backport it?

jcvenegas:
@jodh-intel this is a very few changes and the stable branch gets the bennefit that if there are some fixes in cloud-hypervisor we only need to bump to 0.7.1

grahamwhaley:
> I am always wondering is increasing code coverage required for each PR? I assume No, as we can't guarantee that for every changes, right?

Code Coverage does not have to *increase* for every change - but, ideally, it will not **decrease** for any change :-)  Which implies, if you add code, you may have to add tests to maintain the % coverage...

likebreath:
> > I am always wondering is increasing code coverage required for each PR? I assume No, as we can't guarantee that for every changes, right?
> 
> Code Coverage does not have to _increase_ for every change - but, ideally, it will not **decrease** for any change :-) Which implies, if you add code, you may have to add tests to maintain the % coverage...

Right. That makes sense. Thanks, @grahamwhaley.

Issue #2674, opened 2020-05-08
kata-env: Version check is broken for stable releases


jcvenegas:
# Description of problem

The PR  https://github.com/kata-containers/runtime/pull/2376 introduced a new way to check kata component versions, this was introduced after 1.10 release, it did not fail for a long time because 1.11 releases included a valid case that is 1.11.0-<alpha|rc>. Now that we are about to move to kata 1.11.0 kata-env was failing silently.  But internally was going to panic (not sure if panic is redirected somewhere else).

# Expected result

Kata-env not fail for versions sthat has no extra values.

PR #2675, opened 2020-05-08
utils: Fix case version check for stable releases

jcvenegas:
For stable versions the format used `x.y.z`.

kata-env was failing trying to make a new release from 1.11.0-rc to
1.11.0

This fix kata-env for releases 1.11+ where this regression was
introduced.

Fixes: #2674

### DIFF
--- a/cli/utils.go
+++ b/cli/utils.go
@@ -161,7 +161,13 @@ func constructVersionInfo(version string) VersionInfo {
 		return unknownVersionInfo
 	}
 
-	pres := strings.Split(sv.Pre[0].VersionStr, "-")
+	var pres string
+	if len(sv.Pre) > 0 {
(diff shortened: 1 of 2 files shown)


cli/utils.go
```diff
@@ -161,7 +161,13 @@ func constructVersionInfo(version string) VersionInfo {
 		return unknownVersionInfo
 	}
 
-	pres := strings.Split(sv.Pre[0].VersionStr, "-")
+	var pres string
+	if len(sv.Pre) > 0 {
+		presSplit := strings.Split(sv.Pre[0].VersionStr, "-")
+		if len(presSplit) > 2 {
```
devimc, line 167: `> 2`? I think you mean '>=2' or `> 1`

Issue #2683, opened 2020-05-13
File descriptor not closed properly in setupVirtiofsd


c3d:
# Description of problem

In `func (q *qemu) setupVirtiofsd() (err error)`, the `fd` variable is initialized by calling `listener.File()`, but not closed on the following error path.

# Expected result

File is closed

# Actual result

File is leaked.

PR #2684, opened 2020-05-13
qemu: Don't leak file descriptors in case of error

c3d:
If we take one of the error paths from setupVirtiofsd() after
opening the fd variable, the fd.Close() function is not called.

Fixes: #2683

Signed-off-by: Christophe de Dinechin <dinechin@redhat.com>

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -663,6 +663,7 @@ func (q *qemu) setupVirtiofsd() (err error) {
 	if err != nil {
 		return err
 	}
+	defer fd.Close()
 
 	const sockFd = 3 // Cmd.ExtraFiles[] fds are numbered starting from 3
 	cmd := exec.Command(q.config.VirtioFSDaemon, q.virtiofsdArgs(sockFd)...)


virtcontainers/qemu.go
```diff
@@ -663,6 +663,7 @@ func (q *qemu) setupVirtiofsd() (err error) {
 	if err != nil {
 		return err
 	}
+	defer fd.Close()
```
fidencio, line 666: Shouldn't we remove the `fd.Close()` call a few lines below?

fidencio:
@jcvenegas was faster than me! ;-)
Adding the DNM label till the `fd.Close()` comment is addressed.

virtcontainers/qemu.go
grahamwhaley: OOI, I had wondered if we needed to drop the later `fd.Close()` as well - but, hmm, as it is it means we close the `fd` after its last use. If we drop that, it will stay open for longer when not needed. On the flip side, we can only leave it there if calling `fd.Close()` on an already closed `fd` is safe - as it will still be called by the `defer`...

c3d: The intent of that `fd.Close()` is clearly to close the `fd` before entering the loop, so I left it there intentionally, after checking that closing twice was OK.

c3d:
> I think you need to remove line
> 
> https://github.com/kata-containers/runtime/blob/7b269ff7aa2d62fe12593ff7040798e6c9bd5d65/virtcontainers/qemu.go#L680

I disagree. First, it's safe to `fd.Close()` twice, see other comment above. Second, the existing `fd.Close()` is clearly intended to not leave a useless `fd` open while you run the loop following it.

virtcontainers/qemu.go
fidencio: So, just to confirm, the final action on that is to remove the fd.Close() line mentioned, considering "**Close will return an error if it has already been called**". is it?

(Christophe's Emphasis ;-))

fidencio: Ah, I see @c3d's point after a private conversation.

fd.Close() will error out with "already closed", but won't crash, so that may be fine as we don't check / report that error.

I'm fine with this as is now. @jcvenegas, do you agree? If so, feel free to remove the "do-not-merge" label and merge it.

jcvenegas: Hey thanks for the clarifications, I agree that will not be harmful in the case if it is called two times.  I think is just a matter of different styles but the same result. let's merge it :)

PR #2703, opened 2020-05-28
Add vIOMMU support to qemu q35

### LINKED ISSUE #2694: Add vIOMMU support
If the workload wants to use userspace drivers to drive a device, the use of vIOMMU is generally required.

I would like to propose adding vIOMMU support to Kata. We could add new configuration option (with a default value in the configuration.toml and an Annotation to override it) that requests a viommu. If set, kata-runtime should:

- create the appropriate device in the hypervisor (if supported)
- Add some default parameters to the guest kernel's commandline

What do you think?

amorenoz:
Add vIOMMU support to q35 machines

- Use vIOMMU support added to govmm ([link](https://github.com/intel/govmm/commit/7efaf0b1cde3883a3f38e656c2223f38b9086469))
- Add a configuration.toml switch and a Pod annotation
- If enabled:
  - Add "kernel_irqchip=split" to machine options.
  - Add iommu device: "--device intel-iommu,intremap=on,device-iotlb=on,caching-mode=on"
  - Add iommu params to kernel command line

Fixes #2694 
Signed-off-by: Adrian Moreno <amorenoz@redhat.com>

### DIFF
--- a/virtcontainers/pkg/annotations/annotations.go
+++ b/virtcontainers/pkg/annotations/annotations.go
@@ -151,6 +151,9 @@ const (
 	// HugePages is a sandbox annotation to specify if the memory should be pre-allocated from huge pages
 	HugePages = kataAnnotHypervisorPrefix + "enable_hugepages"
 
+	// Iommu is a sandbox annotation to specify if the VM should have a vIOMMU device
+	IOMMU = kataAnnotHypervisorPrefix + "enable_iommu"
+
 	// FileBackedMemRootDir is a sandbox annotation to soecify file based memory backend root directory
 	FileBackedMemRootDir = kataAnnotHypervisorPrefix + "file_mem_backend"
 
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -409,6 +409,13 @@ func (q *qemu) buildDevices(initrdPath string) ([]govmmQemu.Device, *govmmQemu.I
 		}
 	}
 
+	if q.config.IOMMU {
+		devices, err = q.arch.appendIOMMU(devices)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+
 	var ioThread *govmmQemu.IOThread
 	if q.config.BlockDeviceDriver == config.VirtioSCSI {
 		return q.arch.appendSCSIController(devices, q.config.EnableIOThreads)
--- a/virtcontainers/qemu_amd64.go
+++ b/virtcontainers/qemu_amd64.go
@@ -49,7 +49,6 @@ var kernelParams = []Param{
 	{"reboot", "k"},
 	{"console", "hvc0"},
 	{"console", "hvc1"},
-	{"iommu", "off"},
 	{"cryptomgr.notests", ""},
 	{"net.ifnames", "0"},
 	{"pci", "lastbus=0"},
@@ -94,12 +93,31 @@ func newQemuArch(config HypervisorConfig) qemuArch {
 		factory = true
 	}
 
+	var qemuMachines = supportedQemuMachines
+	if config.IOMMU {
+		var q35QemuIOMMUOptions = "accel=kvm,kernel_irqchip=split"
+
+		kernelParams = append(kernelParams,
+			Param{"intel_iommu", "on"})
+		kernelParams = append(kernelParams,
+			Param{"iommu", "pt"})
+
+		for _, m := range qemuMachines {
+			if m.Type == QemuQ35 {
+				m.Options = q35QemuIOMMUOptions
+			}
+		}
+	} else {
+		kernelParams = append(kernelParams,
+			Param{"iommu", "off"})
+	}
+
 	q := &qemuAmd64{
 		qemuArchBase: qemuArchBase{
 			machineType:           machineType,
 			memoryOffset:          config.MemOffset,
 			qemuPaths:             qemuPaths,
-			supportedQemuMachines: supportedQemuMachines,
+			supportedQemuMachines: qemuMachines,
 			kernelParamsNonDebug:  kernelParamsNonDebug,
 			kernelParamsDebug:     kernelParamsDebug,
 			kernelParams:          kernelParams,
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -412,11 +412,11 @@
   revision = "2f1d1f20f75d5404f53b9edf6b53ed5505508675"
 
 [[projects]]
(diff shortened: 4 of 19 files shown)


virtcontainers/qemu.go
```diff
@@ -409,6 +409,10 @@ func (q *qemu) buildDevices(initrdPath string) ([]govmmQemu.Device, *govmmQemu.I
 		}
 	}
 
+	if q.config.IOMMU {
+		devices = q.arch.appendIOMMU(devices)
```
c3d, line 413: I'm curious what the semantics is for the returned value. Is `appendIOMMU` supposed to prune the list or add more devices? Is that the reason for returning a device list? All the code examples I looked at returned their input.

virtcontainers/qemu_amd64.go
```diff
@@ -93,6 +70,42 @@ func newQemuArch(config HypervisorConfig) qemuArch {
 		factory = true
 	}
 
+	var q35QemuMachineOptions string
+
+	if config.IOMMU {
+		kernelParams = append(kernelParams,
```
c3d, line 76: I would add a little bit of logging here so that a log captures which options were sent to the kernel, etc. Probably `Info` level.

virtcontainers/qemu_amd64.go
```diff
@@ -171,7 +184,7 @@ func (q *qemuAmd64) appendIOMMU(devices []govmmQemu.Device) []govmmQemu.Device {
 	iommu := govmmQemu.IommuDev{
 		Intremap:    true,
 		DeviceIotlb: true,
-		Caching:     true,
+		CachingMode: true,
```
c3d, line 187: What are the use cases for `caching-mode=off`? Do we want to warn in that case?

virtcontainers/pkg/annotations/annotations.go
```diff
@@ -151,6 +151,9 @@ const (
 	// HugePages is a sandbox annotation to specify if the memory should be pre-allocated from huge pages
 	HugePages = kataAnnotHypervisorPrefix + "enable_hugepages"
 
+	// Iommu is a sandbox annotation to specify if the VM should have a vIOMMU device
+	Iommu = kataAnnotHypervisorPrefix + "enable_iommu"
```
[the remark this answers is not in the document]
amorenoz: When you write a bit of code every day that's the result :)
I will homogenize.

virtcontainers/qemu_amd64.go
```diff
@@ -93,6 +70,42 @@ func newQemuArch(config HypervisorConfig) qemuArch {
 		factory = true
 	}
 
+	var q35QemuMachineOptions string
+
+	if config.IOMMU {
```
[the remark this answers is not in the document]
amorenoz: Fair enough, though I'd be introducing another vendor:
https://godoc.org/github.com/jaypipes/ghw
Is that OK?

virtcontainers/qemu.go
amorenoz: Yes. It's the same semantics as [golang's slice appending](https://golang.org/pkg/builtin/#append)

virtcontainers/qemu_amd64.go
amorenoz: Adding a comment here will only log amd64 kernel params.
On every execution we will get a Info level log with all qemu params (including kernel params) in:
https://github.com/intel/govmm/blob/7efaf0b1cde3883a3f38e656c2223f38b9086469/qemu/qemu.go#L2591

Is that enough in your opinion or should we log here as well just the kernel params?

virtcontainers/qemu_amd64.go
amorenoz: AFAIK caching-mode=on is only needed for device passthrough. So a non-configurable "true" value in kata makes sense.
I don't know if we should log at the govmm level though, it is supposed to be a generic library

bergwolf:
We have decided to feature freeze for kata 1.x releases. Please re-submit it to https://github.com/kata-containers/kata-containers Thanks!

If it is strongly desired to include this in the 1.x kata releases like v1.12, please backport it from the 2.0-dev branch after it is merged there.

virtcontainers/qemu_amd64.go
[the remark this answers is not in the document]
jodh-intel: Does it actually error on AMD or just get silently ignored I wonder? We do already have code to check the CPU vendor ID, but you'd need to move it to virtcontainers (or a package under it):

- https://github.com/kata-containers/runtime/blob/master/cli/kata-check_amd64.go#L204..L218

bergwolf:
As we've discussed in PTG, we can have this in 1.x release first. But we need to have a dedicated developer to commit to porting it to kata 2.0-dev branch as well. For this PR, IIUC @fidencio will be the one to port the commits. Am I understanding it correctly?

fidencio:
> As we've discussed in PTG, we can have this in 1.x release first. But we need to have a dedicated developer to commit to porting it to kata 2.0-dev branch as well. For this PR, IIUC @fidencio will be the one to port the commits. Am I understanding it correctly?

Exactly. And I'd like to go through a review here as well.

bergwolf:
>. Exactly. And I'd like to go through a review here as well.

Sure. Thanks for confirming.

virtcontainers/qemu_amd64.go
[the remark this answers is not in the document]
amorenoz: I've consulted the qemu gurus and have been told that intel-iommu emulation should work fine on AMD machines. Maybe @bpradipt can help verifying that?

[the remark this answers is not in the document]
bpradipt: Sure @amorenoz . I need to look for an AMD system to try :-)

virtcontainers/qemu_amd64.go
bpradipt: Yes with `intel_iommu`, `caching-mode=on` is must otherwise device assignment will fail

amorenoz:
Thanks for the direction @fidencio. I will update.

virtcontainers/qemu_amd64.go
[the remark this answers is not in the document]
amorenoz: When we verify this, it could be documented in #2718

dgibson:
Looks good to me, as far as I can tell from my so far limited understanding of Go and Kata.

jodh-intel:
@amorenoz - please tal at the Travis logs - there are compile errors for this PR.

fidencio:
@devimc, is the issue with vfio suite known?

Issue #2712, opened 2020-05-29
host shared directory should be readonly


bergwolf:
We need to make sure containers cannot modify host path unless it is explicitly shared to it. Right now we expose an additional top level shared directory to the guest and allow it to be modified. This is less ideal and can be enhanced by following method:
1. create two directories for each sandbox:
  -. /run/kata-containers/shared/sandboxes/$sbx_id/mounts/, a directory to hold all host/guest shared mounts
  -. /run/kata-containers/shared/sandboxes/$sbx_id/shared/, a host/guest shared directory (9pfs/virtiofs source dir)
2. /run/kata-containers/shared/sandboxes/$sbx_id/mounts/ is bind mounted readonly to /run/kata-containers/shared/sandboxes/$sbx_id/shared/, so guest cannot modify it
3. host-guest shared files/directories are mounted one-level under /run/kata-containers/shared/sandboxes/$sbx_id/mounts/ and thus present to guest at one level under /run/kata-containers/shared/sandboxes/$sbx_id/shared/

PR #2713, opened 2020-05-29
vc: make host shared path readonly

bergwolf:
We need to make sure containers cannot modify host path unless it is explicitly shared to it. Right now we expose an additional top level shared directory to the guest and allow it to be modified. This is less ideal and can be enhanced by following method:
1. create two directories for each sandbox:
  -. /run/kata-containers/shared/sandboxes/$sbx_id/mounts/, a directory to hold all host/guest shared mounts
  -. /run/kata-containers/shared/sandboxes/$sbx_id/shared/, a host/guest shared directory (9pfs/virtiofs source dir)
2. /run/kata-containers/shared/sandboxes/$sbx_id/mounts/ is bind mounted readonly to /run/kata-containers/shared/sandboxes/$sbx_id/shared/, so guest cannot modify it
3. host-guest shared files/directories are mounted one-level under /run/kata-containers/shared/sandboxes/$sbx_id/mounts/ and thus present to guest at one level under /run/kata-containers/shared/sandboxes/$sbx_id/shared/

Fixes: #2712

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -140,6 +140,26 @@ var kataHostSharedDir = func() string {
 	return defaultKataHostSharedDir
 }
 
+// Shared path handling:
+// 1. create two directories for each sandbox:
+// -. /run/kata-containers/shared/sandboxes/$sbx_id/mounts/, a directory to hold all host/guest shared mounts
+// -. /run/kata-containers/shared/sandboxes/$sbx_id/shared/, a host/guest shared directory (9pfs/virtiofs source dir)
+//
+// 2. /run/kata-containers/shared/sandboxes/$sbx_id/mounts/ is bind mounted readonly to /run/kata-containers/shared/sandboxes/$sbx_id/shared/, so guest cannot modify it
+//
+// 3. host-guest shared files/directories are mounted one-level under /run/kata-containers/shared/sandboxes/$sbx_id/mounts/ and thus present to guest at one level under /run/kata-containers/shared/sandboxes/$sbx_id/shared/
+func getSharePath(id string) string {
+	return filepath.Join(kataHostSharedDir(), id, "shared")
+}
+
+func getMountPath(id string) string {
+	return filepath.Join(kataHostSharedDir(), id, "mounts")
+}
+
+func getSandboxPath(id string) string {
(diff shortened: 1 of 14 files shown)


amshinde:
@bergwolf CI seems to be failing. Can you take a look?

jodh-intel:
Yeah, it's slightly ironic :smile::

```
21:38:04 # time="2020-05-31 20:38:01.436721357Z" level=error msg="Container creation error: rpc error: code = Internal desc = Could not create destination mount point: /run/kata-containers/shared/containers/2aa56755a0a1442e543722b07b8c51e88e880856296cc0ee3b39bb9043e1c5a3: mkdir /run/kata-containers/shared/containers/2aa56755a0a1442e543722b07b8c51e88e880856296cc0ee3b39bb9043e1c5a3: read-only file system\n" file="oci/runtime_oci.go:210"
```

bergwolf:
Thanks @jodh-intel ! I wasn't expecting dm based rootfs to still use the shared fs path but it turns out that we do reply on shared fs path even if there is no file system to share.

bergwolf:
The two failed ci are unrelated and tracked at https://github.com/kata-containers/tests/issues/2594 and https://github.com/kata-containers/tests/issues/2592

GabyCT:
@jcvenegas and @likebreath there is an issue with the ci of clh-k8s could you please take a look?

likebreath:
@GabyCT I believe the clh-k8s-crio is still work-in-progress, and we should be able to ignore its failing for now. @jcvenegas Please correct me if I am mistaken on it.

virtcontainers/kata_agent.go
```diff
@@ -1302,6 +1344,13 @@ func (k *kataAgent) buildContainerRootfs(sandbox *Sandbox, c *Container, rootPat
 			rootfs.Options = []string{"nouuid"}
 		}
 
+		// Ensure container mount destination exists
+		// TODO: remove dependency on shared fs path. shared fs is just one kind of storage sources.
+		// we should not always use shared fs path for all kinds of storage. Stead, all storage
+		// should be bind mounted to a tmpfs path for containers to use.
```
liubogithub, line 1350: typo? "Instead"

PR #2725, opened 2020-06-03
virtcontainers: drop deferred func for GetAndSetSandboxBlockIndex()

### LINKED ISSUE #2726: the error return from GetAndSetSandboxBlockIndex() should be checked
# Description of problem

In Acrn#appendImage(), we should check the error return value from GetAndSetSandboxBlockIndex().

# Expected result

Error from GetAndSetSandboxBlockIndex() should result in Acrn#appendImage() returning error.

# Actual result

tedyu:

In Acrn#appendImage(), we should check the error return value from GetAndSetSandboxBlockIndex().

This PR also names the error return for getAndSetSandboxBlockIndex so that the deferred func checks the right variable.

Fixes #2726

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -1606,7 +1606,6 @@ const maxBlockIndex = 65535
 // the BlockIndexMap and marks it as used. This index is used to maintain the
 // index at which a block device is assigned to a container in the sandbox.
 func (s *Sandbox) getAndSetSandboxBlockIndex() (int, error) {
-	var err error
 	currentIndex := -1
 	for i := 0; i < maxBlockIndex; i++ {
 		if _, ok := s.state.BlockIndexMap[i]; !ok {
@@ -1618,11 +1617,6 @@ func (s *Sandbox) getAndSetSandboxBlockIndex() (int, error) {
 		return -1, errors.New("no available block index")
 	}
 	s.state.BlockIndexMap[currentIndex] = struct{}{}
-	defer func() {
-		if err != nil {
-			delete(s.state.BlockIndexMap, currentIndex)
-		}
-	}()
 
 	return currentIndex, nil
 }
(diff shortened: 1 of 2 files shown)


tedyu:
@fidencio @jodh-intel 
Please review.

virtcontainers/sandbox.go
```diff
@@ -1605,8 +1605,7 @@ const maxBlockIndex = 65535
 // getAndSetSandboxBlockIndex retrieves an unused sandbox block index from
 // the BlockIndexMap and marks it as used. This index is used to maintain the
 // index at which a block device is assigned to a container in the sandbox.
-func (s *Sandbox) getAndSetSandboxBlockIndex() (int, error) {
-	var err error
+func (s *Sandbox) getAndSetSandboxBlockIndex() (idx int, err error) {
```
jodh-intel, line 1608: - This change looks odd as `idx` is never set. Maybe `currentIndex` should become `idx` in the function body?
- Also, `err` is never set so I don't see how that `defer` ever gets called?

fidencio: @jodh-intel, but wasn't that the case even prior the changes?

What I'm trying to say is, I'm failing to see how the changes are making it worse than it was before.

tedyu: I noticed the second point as well.
Please confirm that the deferred func should be removed.

idx represents the first return value.
It is filled on the return statement.
If the deferred func is removed, this change is not needed.

jodh-intel: The change to `acrn.go` is fine, but I don't see that the change to this function really helps tbh. In fact as stated, I think it's confusing.

I'd either undo the change to this function, or improve on the change by identfying why this function originally called `defer` by tracing through the git history (maybe it was a copy/paste issue?) and either fixing the logic, or removing the `defer`.

tedyu: I checked commit e9a46580b10a05f9da1d48c5c5490fd981aa318c.
There is no error variable assignment which the deferred func can operate on (before or after that change).

Let me drop the deferred func.

tedyu:
@jodh-intel 
Please take another look.

jodh-intel:
This should be ported to 2.0 (https://github.com/kata-containers/kata-containers/tree/2.0-dev)

Issue #2738, opened 2020-06-03
ppc64le: Configuration.toml file generated from source fails on Fedora due to incorrect Qemu binary path


bpradipt:
# Description of problem
Using the configuration.toml file generated from kata-runtime sources fails on Fedora due to missing qemu binary. 
```
$ sudo podman run -it --runtime=/usr/local/bin/kata-runtime fedora:latest sh
Error: /usr/share/defaults/kata-containers/configuration-qemu.toml: file /usr/bin/qemu-system-ppc64le does not exist: OCI runtime error
```
The default qemu binary on Fedora is `/usr/bin/qemu-system-ppc64` . On Ubuntu `/usr/bin/qemu-system-ppc64le is a soft link to /usr/bin/qemu-system-ppc64`

# Expected result
Container should get spawned

# Actual result

file ` /usr/bin/qemu-system-ppc64le` does not exist

PR #2739, opened 2020-06-03
qemu: Fix Qemu binary path for Power across distros

bpradipt:
The default ppc64le Qemu binary path was specific for Ubuntu.
This patch fixes the default binary path for both Fedora and Ubuntu

Fixes: #2738

Signed-off-by: bpradipt@in.ibm.com


tedyu:
lgtm

Issue #2748, opened 2020-06-05
 virtiofsd: cache=always by default cause sync issue with host


jcvenegas:
**Which feature do you think can be improved?**

Today for virtiofsd kata sets by default `cache=always`. This option is useful for performance but if the shared files are modified from the host changes are not updated in the guest as virtiofsd uses cached value all time.

We should move this value to `cache=none` to keep functionality and allow users to modify is want better performance results. 

Related to https://github.com/kata-containers/tests/issues/2572

@sboeuf @dagrh @c3d @fidencio

PR #2749, opened 2020-06-05
virtiofsd: Use cache auto

jcvenegas:
Today for virtiofsd kata sets by default `cache=always`. This option is
useful for performance but if the shared files are modified from the
host changes are not updated in the guest as virtiofsd uses cached value
all time.

This patch changes to  `cache=auto` to fix consistency issues. The option
can still be set to always if it is wanted by the user.

Fixes: #2748

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,7 +176,7 @@ DEFSHAREDFS_QEMU_VIRTIOFS := virtio-fs
 DEFVIRTIOFSDAEMON := $(VIRTIOFSDBINDIR)/virtiofsd
 # Default DAX mapping cache size in MiB
 DEFVIRTIOFSCACHESIZE := 1024
-DEFVIRTIOFSCACHE ?= always
+DEFVIRTIOFSCACHE ?= auto
 # Format example:
 #   [\"-o\", \"arg1=xxx,arg2\", \"-o\", \"hello world\", \"--arg3=yyy\"]
 #


fidencio:
@jcvenegas, wouldn't be better to have this as `auto` by default? I'm suggesting this because `auto` seems to offer a reasonable trade-off.

jcvenegas:
@fidencio agree! fixed :)

Makefile
```diff
@@ -176,7 +176,7 @@ DEFSHAREDFS_QEMU_VIRTIOFS := virtio-fs
 DEFVIRTIOFSDAEMON := $(VIRTIOFSDBINDIR)/virtiofsd
 # Default DAX mapping cache size in MiB
 DEFVIRTIOFSCACHESIZE := 1024
-DEFVIRTIOFSCACHE ?= always
+DEFVIRTIOFSCACHE ?= none
```
amshinde, line 179: Does this have to be "auto" or "none"? Your commit message mentions "auto"

jcvenegas: good catch, my amend just changed the commit but missed to add the change.

amshinde: Thanks how does cache=auto work @jcvenegas ?

jcvenegas: @amshinde  from config docs:
#  - auto
#    Metadata and pathname lookup cache expires after a configured amount of
#    time (default is 1 second). Data is cached while the file is open (close
#    to open consistency).

tedyu:
lgtm

PR #2756, opened 2020-06-09
Check the correct error variable for sandbox creation

tedyu:
In newSandbox(), the deferred func checks err variable which is different from the one declared within the useOldStore(ctx) case.

This PR makes the deferred func check the correct error variable.

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -505,7 +505,7 @@ func createSandbox(ctx context.Context, sandboxConfig SandboxConfig, factory Fac
 	return s, nil
 }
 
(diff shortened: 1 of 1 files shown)


fidencio:
@tedyu, there are errors in the travis CI because of:
- You're missing a commit message (just the Signed-off-by is not allowed);
- You must file an issue and have the `Fixed: #nnn` as part of the commit message;

You can check the errors here: https://travis-ci.org/github/kata-containers/runtime/builds/696648757

tedyu:
@fidencio 
I have modified the commit message.

tedyu:
For 2.0-dev branch, can I use the same issue #2759 in the commit message ?

fidencio:
> For 2.0-dev branch, can I use the same issue #2759 in the commit message ?

Yes, you can use the very same one. Thanks!

PR #2767, opened 2020-06-11
qemu: Fix kernel_irqchip=split option for IOMMU enabled sandbox

dgibson:
When an x86 sandbox has a vIOMMU (needed for VFIO), it needs the
'kernel_irqchip=split' option or it can't start.  fdcd1f3a2 attempts to set
that, but ends up just writing it to a temporary (looks like Go for range
loops pass by value).

Fixes: fdcd1f3a2 "qemu: enable iommu on q35"
Signed-off-by: David Gibson <david@gibson.dropbear.id.au>

### DIFF
--- a/virtcontainers/qemu_amd64.go
+++ b/virtcontainers/qemu_amd64.go
@@ -102,9 +102,9 @@ func newQemuArch(config HypervisorConfig) qemuArch {
 		kernelParams = append(kernelParams,
 			Param{"iommu", "pt"})
 
-		for _, m := range qemuMachines {
+		for i, m := range qemuMachines {
 			if m.Type == QemuQ35 {
-				m.Options = q35QemuIOMMUOptions
+				qemuMachines[i].Options = q35QemuIOMMUOptions
 			}
 		}
 	} else {
--- a/virtcontainers/qemu_amd64_test.go
+++ b/virtcontainers/qemu_amd64_test.go
@@ -257,3 +257,19 @@ func TestQemuAmd64Microvm(t *testing.T) {
 
 	assert.False(amd64.supportGuestMemoryHotplug())
 }
+
+func TestQemuAmd64Iommu(t *testing.T) {
+	assert := assert.New(t)
+
+	config := qemuConfig(QemuQ35)
+	config.IOMMU = true
+	qemu := newQemuArch(config)
+
+	p := qemu.kernelParameters(false)
+	assert.Contains(p, Param{"intel_iommu", "on"})
+
+	m, err := qemu.machine()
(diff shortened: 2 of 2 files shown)


fidencio:
@dgibson, travis is failing because of:
```
ERROR: No "Fixes" found
```

You can either open a new issue or have this one tied to #2694

dgibson:
Oh, that's a little annoying.  Here the Fixes: tag gives a bug number, rather than a previous commit like in the kernel and qemu.

dgibson:
Well, without this fix you can't actually turn on the enable_iommu option added by the previous changes for #2694, so let's attach it to this.  I updated my branch.

bpradipt:
The rhel8-q35 job failed with this error
```
Errors during downloading metadata for repository 'fedora-modular':
  - Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=fedora-modular-30&arch=x86_64 [Could not resolve host: mirrors.fedoraproject.org]
Error: Failed to download metadata for repo 'fedora-modular': Cannot prepare internal mirrorlist: Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=fedora-modular-30&arch=x86_64 [Could not resolve host: mirrors.fedoraproject.org]
```

amshinde:
@dgibson lgtm, but can we add unit test that covers this?

dgibson:
@amshinde well, **I** can't, at this time, because I can't get "make check" to work even on the master tree.

That said, I'm also not sure exactly what you want to check.  Just starting a VM with the enable_iommu=on option?  Was that not in the initial IOMMU support patches, as it should have been?

I don't think explicitly checking for the presence of kernel_irqchip=split makes a good test case.  We don't care about the kernel options except insofar as they allow the VM to run with IOMMU.

amshinde:
@dgibson You can test that the correct option are added when config.IOMMU is set. I believe this PR fixes the bug, so lets make sure we test that. 
We have similar tests here: https://github.com/kata-containers/runtime/blob/master/virtcontainers/qemu_arm64_test.go#L132

Let me take a look at your `make check` failures. Meanwhile you can always run a test as `sudo go test -v -run TestQemuAmd64Microvm

dgibson:
@amshinde, it was a very frustrating day and a half, but I do finally have make check and make test working properly on the runtime repo.

My latest push adds a test.  It's mostly as an exercise in writing a new testcase for me - I still don't think it's particularly useful (low chance of catching a bug, high chance of requiring maintenance to accommodate future changes).

Issue #2790, opened 2020-06-23
Truncate the kata-runtime log file to avoid OOM


flyflypeng:
# Description of problem

If you use containerd shim v1 to create a kata container, containerd-shim will pass a --log parameter to kata-runtime create command, the log file(we call it kata-runtime.log following) specified by --log parameter will be opened by `beforeSubcommands()` function and set file as the kataLog object output file.

Normally, if you call kata-runtime exec command to do some operation in the container, the kata-runtime exec process will write the log messages to the log file we described previously. However, if huge times of kata-runtime exec command are called, the file size of kata-runtime.log will increase rapidly and consume the memory footprint in the /run dir. 

Moreover, if you put your containerd and containerd-shim processes in a separate limited memory cgroup, which may cause oom occurs and containerd process is killed. 

# Expected result

No matter how much time kata-runtime exec commands are executed, kata.log file will not be too large and never triggers oom.

# Actual result

If I put dockerd/containerd in a memory cgroup which limit max usage memory is 4GB, and I do the about thousands of times docker exec <container-id> echo "hello" operations, which triggers 
the oom event in the memory cgroup.

flyflypeng:
I think who use this log file who should process this log file(maybe should truncate it), so I recommend to do this modification in the containerd instead of kata-runtime.

I will close this issue.

PR #2791, opened 2020-06-23
truncate the kata-runtime log file to avoid OOM

flyflypeng:
reason: If huge times of kata-runtime exec command are called, kata-runtime log file
size will increase rapidly and consume the memory footprint in the run dir, which may
trigger the oom event. So fix this problem by truncate the kata-runtime log file before
kata-runtime subcommand is executed.

Fixes: #2790

Signed-off-by: flyflypeng <woshijpfgg@gmail.com>

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -296,6 +296,14 @@ func beforeSubcommands(c *cli.Context) error {
 		ignoreConfigLogs = true
 	} else {
 		if path := c.GlobalString("log"); path != "" {
+			// avoid log.json file too large consume the memory footprint in the tmpfs,
+			// truncate the log.json file every time before kata subcommand is executed
+			if path != "/dev/null" && katautils.FileExists(path) {
+				if err := os.Truncate(path, 0); err != nil {
+					return err
+				}
+			}
+
 			f, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY|os.O_APPEND|os.O_SYNC, 0640)
 			if err != nil {
 				return err


cli/main.go
```diff
@@ -296,6 +296,14 @@ func beforeSubcommands(c *cli.Context) error {
 		ignoreConfigLogs = true
 	} else {
 		if path := c.GlobalString("log"); path != "" {
+			// avoid log.json file too large consume the memory footprint in the tmpfs,
+			// truncate the log.json file every time before kata subcommand is executed
+			if path != "/dev/null" && katautils.FileExists(path) {
```
devimc, line 301: please don't do this, instead check if path is a regular file and if it exists

cli/main.go
```diff
@@ -296,6 +296,14 @@ func beforeSubcommands(c *cli.Context) error {
...
+				if err := os.Truncate(path, 0); err != nil {
```
devimc, line 302: I'd prefer to have a limit (2 MB maybe) instead of truncate it when a subcommand with --log is called, wdyt?

flyflypeng: I think have a file limit is not appropriate. If the log file size exceeds the limit, just return the log file exceed the limit error is not a good way.

I find the log file specified with --log is used only in the `getLastRuntimeError` function in the containerd-shim, this function just opens the log file and returns the last error level log message in the log file. when a new kata subcommand is executed, the log message written before is not userless in the log file, so why not truncate it.

cli/main.go
flyflypeng: 😀 could you give me some advice? thanks.

cli/main.go
bergwolf: Does containerd ever set `--log` argument? I cannot find the code that sets it.

bergwolf: When `--log` is specified, we should honor `--debug` to set kata's log level as well. Then we mirror the behavior of runc and would not write many log entries there when debug is not enabled.

flyflypeng: I agree with you. kata-runtime will not produce too much log at the most time. And I think who use this log file who should process this log file(maybe should truncate it), so I recommend to do this modification in the containerd instead of kata-runtime.

I will close this issue and PR.

PR #2794, opened 2020-06-25
shimv2 : Remove workaround for sharedPidNs

### LINKED ISSUE #2788: [question] does kata support shareProcessNamespace?
# Description of problem
In k8s, there is "shareProcessNamespace: true" to Share Process Namespace between Containers in a Pod.
Does kata support it or not?
It seems kata doesn't support it per below open issues:
https://github.com/kata-containers/runtime/issues/493
and
https://github.com/kata-containers/runtime/issues/2633
But there is one closed issue seems kata support it:
https://github.com/kata-containers/tests/issues/336

Could you share me whether kata support it or not?
Thanks.
In my test, after switch to kata as runtime, shareProcessNamespace doesn't work, so it seems kata doesn't support it.
If this feature is not supported, could you help add it to the limitations documentation explicitly? Thanks.
https://github.com/kata-containers/documentation/blob/master/Limitations.md#architectural-limitations

# Expected result
n/a

# Actual result
n/a

amshinde:
Removing code that existed as a workaround for a bug in
how shared process namespaces were handled in the agent.
That has been long fixed in the agent.
With this, sharedPidNs will now work with shimv2.

Fixes #2788

### DIFF
--- a/containerd-shim-v2/create.go
+++ b/containerd-shim-v2/create.go
@@ -132,14 +132,6 @@ func loadSpec(r *taskAPI.CreateTaskRequest) (*specs.Spec, string, error) {
 		return nil, "", err
 	}
 
-	// Todo:
-	// Since there is a bug in kata for sharedPidNs, here to
-	// remove the pidns to disable the sharePidNs temporarily,
-	// once kata fixed this issue, we can remove this line.
-	// For the bug, please see:
-	// https://github.com/kata-containers/runtime/issues/930
-	removeNamespace(&ociSpec, specs.PIDNamespace)
-
 	return &ociSpec, bundlePath, nil
 }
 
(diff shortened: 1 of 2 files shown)


amshinde:
@lifupan @bergwolf PTAL

lifupan:
It's LGTM to this modification, but I think there is an issue with sharedPidNs in the guest, and the init process  shared by the containers wouldn't reap the zombie child process which would cause resources leak.

amshinde:
> there is an issue with sharedPidNs in the guest, and the init process shared by the containers wouldn't reap the zombie child process which would cause resources leak.

Is there as easy way to reproduce this? Do you know what needs to be done to fix this.  I can take a look at it.
It would be best if we could address that in a separate issue, can you open an issue for that.
Meanwhile, I think we should merge this patch.
cc @bergwolf

amshinde:
ping @kata-containers/runtime

Issue #2803, opened 2020-06-27
clh: remove the use of deprecated '--memory file=' parameter


likebreath:
# Description of problem

Along with the recent release of cloud-hypervisor v0.8.0, the use of backing file, a.k.a "--memory file=" option, is deprecated and its alternative is using "shared" and "hugepages" controls (https://github.com/cloud-hypervisor/cloud-hypervisor/pull/1056). This is why we got the following "error" message from kata-runtime log:

```
Jun 26 17:22:04 chenb-desk0 kata-hypervisor[5346]: cloud-hypervisor: 12.258979ms: ERROR:vmm/src/config.rs:1263 -- Use of backing file ('--memory file=') is deprecated. Use the 'shared' and 'hugepages' controls.                                                          

```

PR #2804, opened 2020-06-27
clh: Remove the use of deprecated '--memory file=' parameter

likebreath:
Along with the release of cloud-hypervisor v0.8.0, this option has been
deprecated. clh now enforces to use the alternative controls,
e.g. "shared" and "hugepages", which can infer the backing file
paths. Also, we don't use "hugepages" in kata, so we are fine now as the
"shared" control  is already enabled.

Fixes: #2803


likebreath:
> I wonder if we should set "shared" explicitly though to protect against any upstream change in chl default values?

The default value for "shared" and "hugepages" are both `false`. We already explicitly set "shared=true" in clh.go code path as you suggested. Also, we will always make sure the use of clh http api in kata are updated accordingly when we upgrade the CLH version in Kata. So we should be good here.

PR #2823, opened 2020-07-10
virtcontainers: vhost-user-blk/scsi are block device nodes

sameo:
When checking if a device is an emulated vhost-user-blk or
vhost-user-scsi one, we should not only check for their major number but
also their device node type. They must be block devices.

Fixes: #2822

Signed-off-by: Samuel Ortiz <sameo@linux.intel.com>


sameo:
PR ported to 2.0: https://github.com/kata-containers/kata-containers/pull/402

Issue #2864, opened 2020-07-28
clh: Potential regression test failure `VmInfoGet failed`


likebreath:
# Description of problem

As observed from the two PRs (https://github.com/kata-containers/runtime/pull/2840 and https://github.com/kata-containers/runtime/pull/2833), we see the clh-docker CI job is failing on `run hot plug block devices`. Also, @egernst reported a similar failure of `VmInfoGet failed` after hotplug memory to kata+clh.

As @jcvenegas and I confirmed the failure is not related to the changes from the PRs, we believe this is a regression failure introduced recently. The last CI job was passing few days ago on 07/25 [here](http://jenkins.katacontainers.io/job/kata-containers-runtime-ubuntu-1804-PR-cloud-hypeprvisor-docker/141/.

I am opening a dummy PR to verify whether it is actually a regression test failure that was escaped from previous checks/CIs.

likebreath:
Interesting/unexpected experiments results from running clh-docker CI jobs on different PRs:
PR #2865 (dummy PR for debugging this issue) is passing:  http://jenkins.katacontainers.io/job/kata-containers-runtime-ubuntu-1804-PR-cloud-hypeprvisor-docker/150/
PR #2833 (support block device unplug) now is passing (while it was failing earlier today): http://jenkins.katacontainers.io/job/kata-containers-runtime-ubuntu-1804-PR-cloud-hypeprvisor-docker/153/
PR #2840 (Update qemu-virtiofs to 5.0) is still failing at the same docker test: http://jenkins.katacontainers.io/job/kata-containers-runtime-ubuntu-1804-PR-cloud-hypeprvisor-docker/152/

sboeuf:
@likebreath thanks for keeping track of these issues!
As always, I would suggest that we try to reproduce the problem by manually running a Cloud-Hypervisor instance. Unless it's not reproducible because of some weird race condition, this would simplify the debug of these issues.

jcvenegas:
I still not have a cloud-hypervisor only case, but I now have created a script with to docker to isolate a bit more.
```sh
#!/bin/bash
set -x
set -e

loops=9
loop_list=()

create_loop_devices() {
	for i in $(seq 1 ${loops}); do
		loop_name="loop${i}"
		dd if=/dev/zero of=/tmp/${loop_name} count=1 bs=50M
		bash -c printf "g\nn\n\n\n\nw\n" | sudo fdisk /tmp/${loop_name}
[...]

PR #2840, opened 2020-07-20
versions: Update qemu-virtiofs to 5.0 and cloud-hypervisor 0.9

jcvenegas:
Update to qemu 5.0.x with support for virtiofs + dax.

Depends-on: github.com/kata-containers/packaging#1097
Depends-on: github.com/kata-containers/tests#2737

Fixes: #2848

Signed-off-by: Jose Carlos Venegas Munoz <jose.carlos.venegas.munoz@intel.com>

PR #2865, opened 2020-07-29
DNM: clh: debug on issue #2864

likebreath:
It is a potential regression test failure observed as `VmInfoGet failed`
after block device hotplug and memory hotplug.

Fixes: #2864

### DIFF
--- a/virtcontainers/clh.go
+++ b/virtcontainers/clh.go
@@ -3,6 +3,8 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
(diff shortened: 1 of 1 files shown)


likebreath:
The CI seems to be fine for the first run. Let me run it again and see whether the issue is sort of random.

/tet-clh-docker

likebreath:
As we don't see regressions in the master branch, and also the similar error (w/ much higher reproducible rate) has been fixed in the PR https://github.com/kata-containers/runtime/pull/2840, I am closing the test PR now.

Issue #2866, opened 2020-07-29
QEMU instance does not exit when VM shuts itself down.


merwick:
# Description of problem

Stopping a Kata VM (e.g. using /proc/sysrq-trigger) from within the VM doesn't cause the QEMU process to exit.

# Expected result

echo b >  /proc/sysrq-trigger, Kata VM shuts down and QEMU process exits

# Actual result

QEMU instance keeps running but VM is now unresponsive having shut down.

---

merwick:
The Kata architecture does not support rebooting VMs (the lifecycle being start/exec/kill) and if a VM is killed (e.g. using sysrq-trigger), the VM does not exit fully and other layers do not notice the state change.

The QEMU instance needs to be told to exit and not reboot which can be done by passing the --no-reboot command line option when creating the VM.

I am working on a fix for this, please assign issue to me.

merwick:
@devimc This covers Kata 1.12 but I'm not sure how to deal with Kata 2.0.  I ported the changes to the 2.0-dev branch [1] and they pass Travis CI [2] - does it need a separate PR?
[1] https://github.com/merwick/kata-containers/commit/c15ef219e521a1f23de8bef57b273c9a632c0dab
[2] https://travis-ci.org/github/merwick/kata-containers/builds/713350564

devimc:
@merwick yes please, forwardport this patch, 

NOTE: 2.0-dev uses `go modules` (not `dep`)

devimc:
>  does it need a separate PR?

yes, please

merwick:
In https://github.com/merwick/kata-containers/commit/5010e3a368406f4a650def0cf787d9a6b422c9f1 I made the changes to go.mod, got.sum & vendor/modules.txt  (I'm not sure if I should have deleted the old govmm lines (167-168) in go.sum). I'll submit the PR and people can comment there.

PR #2867, opened 2020-07-29
qemu: Set govmmQemu NoReboot Knob

merwick:
The Kata architecture does not support rebooting VMs (the lifecycle being start/exec/kill) and if a VM is killed (e.g. using sysrq-trigger), the VM does not exit fully and other layers do not notice the state change.

The QEMU instance needs to be told to exit and not reboot which can be done by passing the --no-reboot command line option when creating the VM.

Fix for  #2866

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -420,11 +420,11 @@
   revision = "2f1d1f20f75d5404f53b9edf6b53ed5505508675"
 
 [[projects]]
-  digest = "1:ded834b4e437433e3f75a95041ca4e0013e1acd358218c833289986566108f67"
+  digest = "1:e800a48df8e5709baa6b78062d47d43829cddbc54871e049e8bbeb1aaf50fe40"
   name = "github.com/intel/govmm"
   packages = ["qemu"]
   pruneopts = "NUT"
-  revision = "af9e34b91ae99aff4b4ac126141259d34d98e9f2"
+  revision = "6c3315ba8a4262df4300b735b2c53ce3b15d21dd"
 
 [[projects]]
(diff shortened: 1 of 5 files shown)


merwick:
Marking as WIP - TestQemuKnobs needs more work

merwick:
@devimc @amshinde Thanks for taking a look at the first version.  Since you approved it, I have added another commit that added TestQemuKnobs with tests for the pre-existing code (fixing the initial logic) and then the fix for #2866 just adds a line testing the NoReboot var introduced.

devimc:
thanks @merwick  lgtm

@jongwu does ARM support `--no-reboot` ?

merwick:
The--no-reboot option is a universal one. From QEMU source https://git.qemu.org/?p=qemu.git;a=blob;f=qemu-options.hx;h=708583b4ce0b627bf5f74f9da6cde41d8b98fe56;hb=5772f2b1fc5d00e7e04e01fa28e9081d6550440a#l3844

```
DEF("no-reboot", 0, QEMU_OPTION_no_reboot, \
    "-no-reboot      exit instead of rebooting\n", QEMU_ARCH_ALL)
```
and it's present on ARM

```
# /usr/bin/qemu-system-aarch64 --version
QEMU emulator version 2.11.1(qemu-2.11.1-4.el7)
Copyright (c) 2003-2017 Fabrice Bellard and the QEMU Project developers
# /usr/bin/qemu-system-aarch64 --machine virt --no-reboot
VNC server running on ::1:5900
^C
# 
```

devimc:
thanks @merwick

Issue #1217, opened 2019-02-07
ARM CI is broken


sameo:
# Description of problem

Static checks fail with:

```
02:29:38 cli/kata-check_arm64.go:129:9:warning: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary) (golint)
02:29:40 cli/kata-check_test.go:141:113:warning: unused variable or constant undeclared name: TestDataa (varcheck)
02:29:40 virtcontainers/qemu_arm64_test.go:43:23:warning: unused variable or constant undeclared name: virt (varcheck)
02:29:40 virtcontainers/qemu_arm64_test.go:50:11:warning: unused variable or constant undeclared name: defaultMemSlots (varcheck)
02:29:40 virtcontainers/qemu_arm64_test.go:54:40:warning: unused variable or constant too few arguments in call to arm64.memoryTopology (varcheck)
02:29:40 virtcontainers/qemu_arm64_test.go:29:23:warning: unused variable or constant undeclared name: virt (varcheck)
02:29:41 cli/kata-check_test.go:141:113:warning: unused struct field undeclared name: TestDataa (structcheck)
02:29:41 virtcontainers/qemu_arm64_test.go:43:23:warning: unused struct field undeclared name: virt (structcheck)
02:29:41 virtcontainers/qemu_arm64_test.go:50:11:warning: unused struct field undeclared name: defaultMemSlots (structcheck)
02:29:41 virtcontainers/qemu_arm64_test.go:54:40:warning: unused struct field too few arguments in call to arm64.memoryTopology (structcheck)
02:29:41 virtcontainers/qemu_arm64_test.go:29:23:warning: unused struct field undeclared name: virt (structcheck)
02:29:41 cli/kata-check_test.go:141:113:warning: undeclared name: TestDataa (staticcheck)
02:29:42 virtcontainers/qemu_arm64_test.go:43:23:warning: undeclared name: virt (unused)

```

PR #2880, opened 2020-08-06
versions: cloud-hypervisor 0.9.0

jcvenegas:
Highlights for cloud-hypervisor version 0.9.0 include:

virtiofs updates to new dax implementation based in qemu 5.0
Fixed random issues caused due to seccomp filters

io_uring Based Block Device Support

If the io_uring feature is enabled and the host kernel supports it then io_uring will be used for block devices. This results a very significant performance improvement.
Block and Network Device Statistics

Statistics for activity of the virtio network and block devices is now exposed through a new vm.counters HTTP API entry point. These take the form of simple counters which can be used to observe the activity of the VM.
HTTP API Responses

The HTTP API for adding devices now responds with the name that was assigned to the device as well the PCI BDF.
CPU Topology

A topology parameter has been added to --cpus which allows the configuration of the guest CPU topology allowing the user to specify the numbers of sockets, packages per socket, cores per package and threads per core.
Release Build Optimization

Our release build is now built with LTO (Link Time Optimization) which results in a ~20% reduction in the binary size.
Hypervisor Abstraction

A new abstraction has been introduced, in the form of a hypervisor crate so as to enable the support of additional hypervisors beyond KVM.
Snapshot/Restore Improvements

Multiple improvements have been made to the VM snapshot/restore support that was added in the last release. This includes persisting more vCPU state and in particular preserving the guest paravirtualized clock in order to avoid vCPU hangs inside the guest when running with multiple vCPUs.
Virtio Memory Ballooning Support

A virtio-balloon device has been added, controlled through the resize control, which allows the reclamation of host memory by resizing a memory balloon inside the guest.
Enhancements to ARM64 Support

The ARM64 support introduced in the last release has been further enhanced with support for using PCI for exposing devices into the guest as well as multiple bug fixes. It also now supports using an initramfs when booting.
Intel SGX Support

The guest can now use Intel SGX if the host supports it. Details can be found in the dedicated SGX documentation.
Seccomp Sandbox Improvements

The most frequently used virtio devices are now isolated with their own seccomp filters. It is also now possible to pass --seccomp=log which result in the logging of requests that would have otherwise been denied to further aid development.
Notable Bug Fixes

    Our virtio-vsock implementation has been resynced with the implementation from Firecracker and includes multiple bug fixes.
    CPU hotplug has been fixed so that it is now possible to add, remove, and re-add vCPUs (#1338)
    A workaround is now in place for when KVM reports MSRs available MSRs that are in fact unreadable preventing snapshot/restore from working correctly (#1543).
    virtio-mmio based devices are now more widely tested (#275).
    Multiple issues have been fixed with virtio device configuration (#1217)
    Console input was wrongly consumed by both virtio-console and the serial. (#1521)

Fixes: #2864


jcvenegas:
Seems that clh is not happy to be updated with out a new version of virtiofsd, and virtiofsd PR is not happy if clh is not updated so closing this in favor of https://github.com/kata-containers/runtime/pull/2840 that includes both upgrades.

Issue #2886, opened 2020-08-17
Add version cli option to shimv2


jodh-intel:
All components should support a `--version` option to allow clear identification of the version of the component being used.

See: https://github.com/kata-containers/kata-containers/issues/307

PR #2887, opened 2020-08-17
shimv2: Add a "--version" cli option

jodh-intel:
All components should support a `--version` option to allow clear identification of the version of the component being used.

Note that the build changes are required to allow the shim binary to access the golang code generated by the build (such as the `version` variable).

This is a backport of https://github.com/kata-containers/kata-containers/pull/308.

Fixes: #2886.

### DIFF
--- a/cli/containerd-shim-kata-v2/main.go
+++ b/cli/containerd-shim-kata-v2/main.go
@@ -6,9 +6,14 @@
 package main
 
 import (
+	"fmt"
+	"os"
+
 	"github.com/containerd/containerd/runtime/v2/shim"
 	"github.com/kata-containers/runtime/containerd-shim-v2"
 )
+
+const shimID = "io.containerd.kata.v2"
 
 func shimConfig(config *shim.Config) {
 	config.NoReaper = true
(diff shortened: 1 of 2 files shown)


cli/containerd-shim-kata-v2/main.go
```diff
@@ -6,15 +6,25 @@
...
+const shim_id = "io.containerd.kata.v2"
+
 func shimConfig(config *shim.Config) {
 	config.NoReaper = true
 	config.NoSubreaper = true
 }
 
 func main() {
-	shim.Run("io.containerd.kata.v2", containerdshim.New, shimConfig)
+	if len(os.Args) == 2 && os.Args[1] == "--version" {
```
devimc, line 24: I wonder if this should be part of `parseFlags()` in `shim.go`, since `--version` is not displayed in the usage message

```
./containerd-shim-kata-v2 --help
Usage of ./containerd-shim-kata-v2:
  -address string
    	grpc address back to main containerd
  -bundle string
    	path to the bundle if not workdir
  -debug
    	enable debug output in logs
  -id string
    	id of the task
  -namespace string
    	namespace that owns the shim
  -publish-binary string
    	path to publish binary (used for publishing events) (default "containerd")
  -socket string
    	abstract socket path to serve
```
```

jodh-intel: Hi @devimc - that's a good question. However, you can't call `flag.Parse()` twice: the first call wins meaning only `--verbose` would be recognised and the shim wouldn't work. Also, this is a backport from 2.0 so I'd like to keep it as similar as possible.

Issue #2678, opened 2020-05-11
PcieDevice detection fails


amorenoz:
# Description of problem

A PCIe device should be pluggable into a pcie-root-port using the following combination of options:
```
hotplug_vfio_on_root_bus = true
pcie_root_port = 2
```
However, doing so with devices connected to root-ports in the host fails with:
```
Bus 'pcie.0' does not support hotplugging: OCI runtime error
```
The reason is that [isPCIeDevice](https://github.com/kata-containers/runtime/blob/fde6447c16721dc5fa3139872bc791ec08afea4e/virtcontainers/device/drivers/utils.go#L51) looks at /sys/bus/pci/slots to determine whether the device is PCIe based on the slot reported speed.
This mechanism is not generic enough and fails to recognize devices that are connected to root-ports such as:

```
$ lspci -t
[...]
 +-[0000:64]-+-00.0-[65-66]--+-00.0  Intel Corporation Ethernet Controller X710 for 10GbE SFP+                                                                                                                                                  
 |           |               \-00.1  Intel Corporation Ethernet Controller X710 for 10GbE SFP+
[...]
$ lspci -s 000:64:00.0 
64:00.0 PCI bridge: Intel Corporation Sky Lake-E PCI Express Root Port A (rev 04)
$ cat  /sys/bus/pci/slots/2/{address,max_bus_speed}
0000:64:00
Unknown
```

# Expected result

Hotplug into pcie-root-port should work

# Actual result

Devices are not plugged into pcie-root-ports. In q35 this means kata tries to plug them into the root bus, which does not support hot-plugging and the following error message is shown:
However, doing so with devices connected to root-ports in the host fails with:
```
Bus 'pcie.0' does not support hotplugging: OCI runtime error
```

bpradipt:
@amorenoz @devimc as a fallback mechanism can we check the value  for under /sys/bus/pci/devices/xxx/max_link_speed to be >= 5.0 GT/s? This seems to be always populated.

If it's fine I can work on a PR

amorenoz:
@bpradipt If we do that, we will have to detect if the device is SR-IOV and probe the PF:
```
$ lspci
65:00.0 Ethernet controller: Intel Corporation Ethernet Controller X710 for 10GbE SFP+ (rev 02)
65:00.1 Ethernet controller: Intel Corporation Ethernet Controller X710 for 10GbE SFP+ (rev 02)
[...]

PR #2889, opened 2020-08-19
Correct isPCIeDevice logic

dgibson:
Currently, isPCIeDevice() attempts to determine if a (host) device is
PCI-Express capable by looking up its link speed via the PCI slots
information in sysfs.  This is a) complicated and b) wrong.  PCI-e devices
don't have to have slots information, so this frequently fails.

Instead determine if devices are PCI-e by checking for the presence of
PCIe extended configuration space by looking at the size of the "config"
file in sysfs.

Fixes: #2678

Signed-off-by: David Gibson <david@gibson.dropbear.id.au>

### DIFF
--- a/virtcontainers/device/drivers/utils.go
+++ b/virtcontainers/device/drivers/utils.go
@@ -9,6 +9,7 @@ package drivers
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"strings"
 
@@ -22,6 +23,8 @@ const (
 
 	PCIDomain   = "0000"
 	PCIeKeyword = "PCIe"
+
+	PCIConfigSpaceSize = 256
 )
 
 type PCISysFsType string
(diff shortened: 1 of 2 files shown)


dgibson:
Corrected commit message format error.

virtcontainers/device/drivers/utils.go
```diff
@@ -52,23 +53,17 @@ func isPCIeDevice(bdf string) bool {
...
-		}
+		deviceLogger().WithField("dev-bdf", bdf).WithField("error", err).Warning("Couldn't stat() configuration space file")
+		return false //Who knows?
 	}
-	deviceLogger().WithField("dev-bdf", bdf).Debug("can not find slot for bdf of pci device")
-	return false
+
+	// Plain PCI devices hav 256 bytes of configuration space,
+	// PCI-Express devices have 4096 bytes
+	return fi.Size() > 256
```
devimc, line 66: please don't hard code this value, instead declare a `const`

dgibson: Done.

dgibson:
@devimc, I've addressed your most recent comments.  Can we get this fix merged?

virtcontainers/device/drivers/utils.go
```diff
@@ -52,23 +55,17 @@ func isPCIeDevice(bdf string) bool {
...
-	b := strings.Split(bdf, ".")[0]
-	for _, slot := range slots {
-		address := getPCISlotProperty(slot.Name(), PCISysFsSlotsAddress)
-		if b == address {
-			maxBusSpeed := getPCISlotProperty(slot.Name(), PCISysFsSlotsMaxBusSpeed)
-			if strings.Contains(maxBusSpeed, PCIeKeyword) {
-				return true
-			}
-		}
+		deviceLogger().WithField("dev-bdf", bdf).WithField("error", err).Warning("Couldn't stat() configuration space file")
```
fidencio, line 62: I think you can use `WithError(err)`, instead of `WithField("error", err)`, but that's a nitpick.

fidencio (commented):
@dgibson, thanks for the contribution!

I've started CI and made one comment inline, which is just a nitpick.

For now, let's wait till CI finishes its job before addressing that (or not). :-)

virtcontainers/device/drivers/utils.go
```diff
@@ -52,23 +55,17 @@ func isPCIeDevice(bdf string) bool {
...
+		return false //Who knows?
 	}
-	deviceLogger().WithField("dev-bdf", bdf).Debug("can not find slot for bdf of pci device")
-	return false
+
+	// Plain PCI devices hav 256 bytes of configuration space,
+	// PCI-Express devices have 4096 bytes
+	return fi.Size() > PCIConfigSpaceSize
```
jodh-intel, line 68: I was going to say it might be safer to check if the size is exactly 4096. But presumably if another PCI standard is released that would be a superset of PCI-e (as PCI-e is to PCI) so I guess this is "safe" as-is :smile:

jodh-intel (approved):
Thanks  for raising @dgibson !

Please could you consider raising an equivalent PR for the 2.0 branch here too: https://github.com/kata-containers/kata-containers/tree/2.0-dev/src/runtime

fidencio (approved):
LGTM.

Feel free to adapt the code with the nitpick I mentioned or just to ignore that, as it's nothing else than a nitpick.

virtcontainers/device/drivers/utils.go
dgibson: Right, and my guess would be that that hypothetical extension will probably behave closer to PCI express than classic PCI.

amshinde:
@dgibson Thanks for the fix. Can you please backport this fix to 1.10 and 1.11 branches as well along with 2.x repo.

dgibson:
> @dgibson Thanks for the fix. Can you please backport this fix to 1.10 and 1.11 branches as well along with 2.x repo.

I've sent a backport for 1.11 here: https://github.com/kata-containers/runtime/pull/2898
Backporting for 1.10 doesn't make sense, the isPCIeDevice() function doesn't exist at all there.

I'll look at 2.x later.

Issue #2899, opened 2020-08-25
clh: Disable the `seccomp` option temporarily 


likebreath:
# Description of problem

We noticed the `seccomp` option in cloud-hypervisor can introduce periodic failures in Kata's CI since the debug work on `VmInfoGet` failure (issue https://github.com/kata-containers/runtime/issues/2864). We continued to see instability of some CLH CI jobs, espcially the `clh-docker` CI [here](http://jenkins.katacontainers.io/job/kata-containers-tests-ubuntu-1804-PR-cloud-hypeprvisor-docker/), and we believe the instability is likely to be related to the `seccomp` option in CLH. 

To understand and fix the recent instability of the CLH CI, I propose to disable the `seccomp` option of CLH in Kata for now. In this way, we can separate `seccomp` related failures from other moving pieces (e.g. packaging clh as real static binary, and new changes from tests/runtime, etc). On the other hand, we can start a separate issue/PR to debug on the `seccomp` issue in Kata, which we can collect a list of missing `syscalls` that are required to run Kata workload with CLH (issue #2901 ).

# Expected result

Constantly stable CI on CLH.

# Actual result

Periodic CI failures that can be resolved by restarting the same CI job. Examples from the `clh docker CI` are:

Job [191](http://jenkins.katacontainers.io/job/kata-containers-tests-ubuntu-1804-PR-cloud-hypeprvisor-docker/191/) vs [190](http://jenkins.katacontainers.io/job/kata-containers-tests-ubuntu-1804-PR-cloud-hypeprvisor-docker/190/)/[189](http://jenkins.katacontainers.io/job/kata-containers-tests-ubuntu-1804-PR-cloud-hypeprvisor-docker/189/)

Job [184](http://jenkins.katacontainers.io/job/kata-containers-tests-ubuntu-1804-PR-cloud-hypeprvisor-docker/184/) vs [183](http://jenkins.katacontainers.io/job/kata-containers-tests-ubuntu-1804-PR-cloud-hypeprvisor-docker/183/)

PR #2900, opened 2020-08-25
clh: Disable the 'seccomp' option temporarily

likebreath:
We kept observing instabilities from CLH CI jobs periodically. To
separate the random failures caused by `seccomp` from other failures,
this patch disables the 'seccomp' option from clh in kata for now. We
will bring this option back after completing the 'seccomp' filter lists
based on Kata's CI workload.

Fixes: #2899

### DIFF
--- a/virtcontainers/clh.go
+++ b/virtcontainers/clh.go
@@ -955,6 +955,12 @@ func (clh *cloudHypervisor) LaunchClh() (string, int, error) {
 		args = append(args, "-vv")
 	}
 
+	// Disable the 'seccomp' option in clh for now.
+	// In this way, we can separate the periodic failures caused
+	// by incomplete `seccomp` filters from other failures.
+	// We will bring it back after completing the `seccomp` filter.
+	args = append(args, "--seccomp", "false")
+
 	clh.Logger().WithField("path", clhPath).Info()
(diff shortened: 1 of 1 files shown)


virtcontainers/clh.go
```diff
@@ -955,6 +955,12 @@ func (clh *cloudHypervisor) LaunchClh() (string, int, error) {
 		args = append(args, "-vv")
 	}
 
+	// Disable the 'seccomp' option in clh for now.
+	// In this way, we can separate the periodic failures caused
+	// by incomplete `seccomp` filters from other failures.
+	// We will bring it back after completing the `seccomp` filter.
```
devimc, line 961: would you mind filing an issue and mentioning it here ?

likebreath: The corresponding issue is here: https://github.com/kata-containers/runtime/issues/2899.

I will open another one to track the work on collecting `seccomp` filter lists based on kata's CI workload.

likebreath: @devimc The other issue added here: https://github.com/kata-containers/runtime/issues/2901. :)

jodh-intel: Thanks @likebreath. Please can you add the issue URLs into the code comments so we don't forget :)

Issue #2914, opened 2020-08-31
Don' call bindUnmountContainerRootfs if container's rootfs use devicemapper device


keloyang:
buildContainerRootfs don't call bindMountContainerRootfs  if container's rootfs use devicemapper device in https://github.com/kata-containers/runtime/blob/master/virtcontainers/kata_agent.go#L1300, so bindUnmountContainerRootfs should not be called  if container's rootfs use devicemapper device in https://github.com/kata-containers/runtime/blob/master/virtcontainers/container.go#L1123

PR #2915, opened 2020-08-31
runtime: Don' call bindUnmountContainerRootfs if container's rootfs u…

keloyang:
buildContainerRootfs don't call bindMountContainerRootfs if container's rootfs use devicemapper
device in https://github.com/kata-containers/runtime/blob/master/virtcontainers/kata_agent.go#L1300,
so bindUnmountContainerRootfs should not be called if container's rootfs use devicemapper device
in https://github.com/kata-containers/runtime/blob/master/virtcontainers/container.go#L1123

Fixes: #2914

### DIFF
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -850,7 +850,7 @@ func (c *Container) rollbackFailingContainerCreation() {
 	if err := c.unmountHostMounts(); err != nil {
 		c.Logger().WithError(err).Error("rollback failed unmountHostMounts()")
 	}
-	if err := bindUnmountContainerRootfs(c.ctx, getMountPath(c.sandbox.id), c.id); err != nil {
+	if err := bindUnmountContainerRootfs(c.ctx, getMountPath(c.sandbox.id), c); err != nil {
 		c.Logger().WithError(err).Error("rollback failed bindUnmountContainerRootfs()")
 	}
 }
@@ -1120,7 +1120,7 @@ func (c *Container) stop(force bool) error {
 		return err
 	}
 
-	if err := bindUnmountContainerRootfs(c.ctx, getMountPath(c.sandbox.id), c.id); err != nil && !force {
+	if err := bindUnmountContainerRootfs(c.ctx, getMountPath(c.sandbox.id), c); err != nil && !force {
 		return err
 	}
 
--- a/virtcontainers/mount.go
+++ b/virtcontainers/mount.go
@@ -315,13 +315,17 @@ func isSymlink(path string) bool {
 	return stat.Mode()&os.ModeSymlink != 0
 }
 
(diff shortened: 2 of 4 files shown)


virtcontainers/container.go
```diff
@@ -546,6 +546,7 @@ func (c *Container) mountSharedDirMounts(hostSharedDir, guestSharedDir string) (
 
 		var ignore bool
 		var guestDest string
+
```
fidencio, line 549: nitpick: extra empty line

fidencio (approved):
lgtm!

fidencio:
@bergwolf, do we have the same condition on the rust agent? If so, this PR must be forward-ported to the kata-containers/kata-containers branch.

For sure it also deserves a backport to the stable-1.11 branch as well.

egernst:
Thanks for the change. I’m wondering if we should be augmenting our unit tests to exercise this. Any interest in helping with this @keloyang ?

keloyang:
I'm happy to do this and I have change for this pr, PTAL, thanks. @egernst @fidencio

jodh-intel (approved):
Thanks @keloyang.

Would you be able to forward-port this to Kata 2.0 too (https://github.com/kata-containers/kata-containers/tree/2.0-dev/src/runtime) ?

lgtm

keloyang:
ping @egernst @jodh-intel @fidencio

cmaf:
@keloyang this is failing on 2 required checks ("jenkins-ci-ubuntu-18-04" and "jenkins-metrics-ubuntu-18-04") I'm seeing a 404 error instead of the results, so I'm retesting.

/test

virtcontainers/mount.go
```diff
@@ -351,7 +355,7 @@ func bindUnmountAllRootfs(ctx context.Context, sharedDir string, sandbox *Sandbo
 		if c.state.Fstype == "" {
```
amshinde, line 355: This check here should prevent `bindUnmountContainerRootfs` from being called. If it were being called, the `bindUnmountContainerRootfs` call would have returned an error which would have been propogated up the stack.
Are you seeing otherwise?

keloyang: @amshinde thanks for your review. I think we still need this change. **bindUnmountContainerRootfs** is called in several places ，there is no check in **https://github.com/kata-containers/runtime/blob/master/virtcontainers/container.go#L1123**. 
in addition，call **bindUnmountContainerRootfs** don't always return  error, see **https://github.com/kata-containers/runtime/blob/master/virtcontainers/mount.go#L329**.

amshinde: Ok fair enough.

amshinde (approved):

jodh-intel:
forward port PR: https://github.com/kata-containers/kata-containers/pull/940

PR #2916, opened 2020-08-31
[backport | master] sandbox: Disconnect from agent after VM shutdown

### LINKED ISSUE #2719: /usr/bin/containerd-shim-kata-v2 processes persist and consume heavy CPU after corresponding containers exit
# Description of problem

Ran a test that created 40 CPU soaker containers on a 32/64 thread system w/192 GB RAM.

Using the `clusterbuster` tool from https://github.com/RobertKrawitz/OpenShift4-tools

```clusterbuster -b 5 -p 5 -P soaker -T pod -Y -e --container-resource-request=cpu=10m -v -N 1 -d 40 -r 1 -t 30 --report --cleanup -Q --kata```

which I ran a number of times, I observed that the node in question retained hundreds of containerd-shim-kata-v2 processes that were consuming a large amount of CPU.  Accompanying this were wide variations in the CPU utilization reported by the containers, wide variation in the work (simple loop iterations) accomplished by the containers, and amount of time required for containers to start from first to last.

Without use of Kata, it took about 1.1 seconds between the first and last container to start, achieved 3986% CPU utilization (within .5% of max), and achieved about 613M loop iterations/sec.  With Kata, the first run was very close to that (3.9 seconds for the containers to start, 3984% CPU utilization, and 603M loop iterations/second.  Running this in a loop 5 times, all of these numbers got worse.  By the final iteration, I observed 506% CPU utilization, 45 seconds between first and last pod (in some cases I saw much more), and about 69M loop iterations/second.

# Expected result

Performance to remain consistent with multiple runs, no accumulation of processes.

# Actual result
See attached: 
[kata-log.txt](https://github.com/kata-containers/runtime/files/4713127/kata-log.txt)

evanfoster:
When a one-shot pod dies in CRI-O, the shimv2 process isn't killed until the pod is actually deleted, even though the VM is shut down. In this case, the shim appears to busyloop when attempting to talk to the (now dead) agent via VSOCK. To address this, we disconnect from the agent after the VM is shut down.

This is especially catastrophic for one-shot pods that may persist for hours or days, but it also applies to any shimv2 pod where Kata is configured to use VSOCK for communication.

Backport of https://github.com/kata-containers/kata-containers/pull/556 to kata-containers/runtime master branch.

Fixes #2719

Signed-off-by: Evan Foster <efoster@adobe.com>

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -1603,6 +1603,11 @@ func (s *Sandbox) Stop(force bool) error {
 		return err
 	}
 
+	// Stop communicating with the agent.
+	if err := s.agent.disconnect(); err != nil && !force {
+		return err
+	}
+
 	return nil
 }
 


fidencio (approved):
lgtm!

jodh-intel (approved):
Thanks @evanfoster.

lgtm

Issue #2921, opened 2020-09-01
 /var/lib/vc/sbs/${sid} dir residual


keloyang:
we can reporduce like the following.
1.  **kubectl create -f untrusted.yaml**   create a untrusted pod which is bound to the node of centos1 in k8s
2.  **sleep 10** sleep 10 seconds
3.  **delete the pod**
4.  **ssh centos1 ls**  login into centos with ssh ,and list the dir for /var/lib/vc/sbs/ 

execute the commands above everytime, the directory of  /var/lib/vc/sbs/  will have a new directory which's name is the ${sandbox-id}, e.g. d3e0482b22b9e25cd3268608b12ab8c1eb666960c4fa9a6a72a3e4d0b1606551
```
[root@centos0 manifest]# kubectl create -f untrusted.yaml; sleep 10; kubectl delete -f untrusted.yaml;ssh centos1 ls /var/lib/vc/sbs/ 
pod/untrusted created
pod "untrusted" deleted
d3e0482b22b9e25cd3268608b12ab8c1eb666960c4fa9a6a72a3e4d0b1606551
[root@centos0 manifest]# kubectl create -f untrusted.yaml; sleep 10; kubectl delete -f untrusted.yaml;ssh centos1 ls /var/lib/vc/sbs/
pod/untrusted created
pod "untrusted" deleted
d3e0482b22b9e25cd3268608b12ab8c1eb666960c4fa9a6a72a3e4d0b1606551
eda6fa3179b2320655b89a5b538362a491e4ec0c0fa75b2eb7f930b1d9bf2b49
[root@centos0 manifest]# kubectl create -f untrusted.yaml; sleep 10; kubectl delete -f untrusted.yaml;ssh centos1 ls /var/lib/vc/sbs/
pod/untrusted created
pod "untrusted" deleted
27c1334f7e5a32fb115173dd9d2c9bad7d8826f63bdc0a97aa0798b8b123da3c
d3e0482b22b9e25cd3268608b12ab8c1eb666960c4fa9a6a72a3e4d0b1606551
eda6fa3179b2320655b89a5b538362a491e4ec0c0fa75b2eb7f930b1d9bf2b49

[root@centos0 manifest]# cat untrusted.yaml 
apiVersion: v1
kind: Pod
metadata:
  name: untrusted
  namespace: default
  annotations:
    io.kubernetes.cri.untrusted-workload: "true"
  labels:
    env: test
spec:
  containers:
  - name: untrusted
    image: busybox
[...]

PR #2922, opened 2020-09-01
runtime: Fix /var/lib/vc/sbs/${sid} dir residual

keloyang:
Fixes: #2921

runtime call fetchSandbox-->loadSandboxConfigFromOldStore-->store.NewVCSandboxStore--> ... f.initialize -->  os.MkdirAll(f.path, DirMode) 
and at last create /var/lib/vc/sbs/${sid}, but don't delete it before delete sandbox.

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -728,14 +728,14 @@ func fetchSandbox(ctx context.Context, sandboxID string) (sandbox *Sandbox, err 
 
 	var config SandboxConfig
 
-	// Try to load sandbox config from old store at first.
-	c, ctx, err := loadSandboxConfigFromOldStore(ctx, sandboxID)
+	// Try to load sandbox config from new store at first.
+	c, err := loadSandboxConfig(sandboxID)
 	if err != nil {
-		virtLog.Warningf("failed to get sandbox config from old store: %v", err)
-		// If we failed to load sandbox config from old store, try again with new store.
-		c, err = loadSandboxConfig(sandboxID)
+		virtLog.Warningf("failed to get sandbox config from new store: %v", err)
+		// If we failed to load sandbox config from new store, try again with old store.
+		c, ctx, err = loadSandboxConfigFromOldStore(ctx, sandboxID)
 		if err != nil {
-			virtLog.Warningf("failed to get sandbox config from new store: %v", err)
+			virtLog.Warningf("failed to get sandbox config from old store: %v", err)
 			return nil, err
 		}
 	}
@@ -837,7 +837,11 @@ func (s *Sandbox) Delete() error {
 	}
 
 	s.agent.cleanup(s)
-
+	if useOldStore(s.ctx) && s.store != nil {
+		if err := s.store.Delete(); err != nil {
+			s.Logger().WithError(err).Error("store delete failed")
+		}
+	}
 	return s.newStore.Destroy(s.id)
 }
 
--- a/virtcontainers/sandbox_test.go
+++ b/virtcontainers/sandbox_test.go
@@ -13,6 +13,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"strings"
 	"sync"
 	"syscall"
 	"testing"
@@ -24,6 +25,7 @@ import (
 	exp "github.com/kata-containers/runtime/virtcontainers/experimental"
 	"github.com/kata-containers/runtime/virtcontainers/persist/fs"
 	"github.com/kata-containers/runtime/virtcontainers/pkg/annotations"
+	"github.com/kata-containers/runtime/virtcontainers/store"
 	"github.com/kata-containers/runtime/virtcontainers/types"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/stretchr/testify/assert"
@@ -1526,4 +1528,41 @@ func TestGetSandboxCpuSet(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestSandboxStoreClean(t *testing.T) {
+	ctx := context.Background()
+	contID := "SandboxStore"
+	contConfig := newTestContainerConfigNoop(contID)
+	hConfig := newHypervisorConfig(nil, nil)
+	assert := assert.New(t)
+
+	// create a sandbox
+	p, err := testCreateSandbox(t, testSandboxID, MockHypervisor, hConfig, NoopAgentType, NetworkConfig{}, []ContainerConfig{contConfig}, nil)
+	assert.NoError(err)
+	defer cleanUp()
+
+	l := len(p.GetAllContainers())
+	assert.Equal(l, 1)
+
+	// persist to disk
+	err = p.storeSandbox()
+	assert.NoError(err)
+
+	loadSandboxConfigFromOldStore(ctx, p.ID())
+
+	runtimeSidPath := store.SandboxConfigurationRoot(p.ID())
+	runtimeSidPath = strings.TrimPrefix(runtimeSidPath, "file://")
+
+	p.store, err = store.NewVCSandboxStore(ctx, testSandboxID)
+	assert.Nil(err)
+
+	p.ctx = context.WithValue(ctx, oldstoreKey, true)
+	err = p.Delete()
+	assert.NoError(err)
+
+	// expect runtimeSidPath not exist, if exist, it means this case failed.
+	_, err = os.Stat(runtimeSidPath)
+	assert.Error(err)
+	assert.True(os.IsNotExist(err))
 }


fidencio (commented):
@keloyang, I'm not sure if I understand this fix.
Would you mind to elaborate why it's needed? Please, also add the details to the commit message.

keloyang:
@fidencio I have updated the issue #2921, PTAL, thanks.

jodh-intel:
@keloyang - Thanks for raising! A couple of things:

- Is there any way we can add a unit test here to ensure the correct behaviour?
- Could you also port this to Kata 2.0 (https://github.com/kata-containers/kata-containers/tree/2.0-dev/src/runtime)?

@bergwolf, @sameo - could you tal at this change please?

fidencio:
Let me give it a  try on my environment here, thanks for the update @keloyang.

amshinde (approved):

amshinde:
@keloyang Thanks for the fix!  LGTM. Could you please add add a unit/integration test verifying this behaviour?

keloyang:
@amshinde @jodh-intel  unit test case added, ptal, thanks.

fidencio (commented):
I'd like to ask @amshinde, @jodh-intel, to not merge this one yet!

I want to try to reproduce this one and will most likely find some time this evening to do so.
There are a few things puzzling me here, which I'm commenting inline.

virtcontainers/sandbox.go
```diff
@@ -837,7 +837,10 @@ func (s *Sandbox) Delete() error {
 	}
 
 	s.agent.cleanup(s)
-
+	vcStore, err := store.NewVCSandboxStore(s.ctx, s.id)
```
fidencio, line 840: Do we really have to create a new vcStore or should we just use s.store?

virtcontainers/sandbox.go
```diff
@@ -837,7 +837,10 @@ func (s *Sandbox) Delete() error {
...
+	if err == nil {
+		vcStore.Delete()
```
fidencio, line 842: We definitely should check for errors here, at least to log them.

virtcontainers/sandbox_test.go
```diff
@@ -1527,3 +1528,35 @@ func TestGetSandboxCpuSet(t *testing.T) {
...
+	loadSandboxConfigFromOldStore(context.Background(), p.ID())
+
+	runtimSidPath := store.SandboxConfigurationRoot(p.ID())
+	runtimSidPath = strings.TrimPrefix(runtimSidPath, "file://")
+
+	err = p.Delete()
+	assert.NoError(err)
+
+	// expect runtimSidPath not exist, if exist, it means this case failed.
+	if _, err := os.Stat(runtimSidPath); err == nil || os.IsExist(err) {
```
jodh-intel, line 1559: I think it would be clearer to write this as separate tests:

```go
err := os.Stat(runtimSidPath)
assert.Error(err)
assert.True(os.IsNotExist(err))
```

jodh-intel (commented):
Thanks @keloyang - just one comment.

jodh-intel:
@fidencio - agreed - added `dnm` for now.

keloyang:
> Do we really have to create a new vcStore or should we just use s.store?

runtime set s.store only when  useOldStore(ctx) is true in https://github.com/kata-containers/runtime/blob/master/virtcontainers/sandbox.go#L573, this need loadSandboxConfigFromOldStore return nil, but it return error if old store don't exist. this means everytime when fetchSandbox is called ,it will create a vcStore, but s.store is nil. so we can't use s.store  simplely. 
loadSandboxConfigFromOldStore is for backford,and old store has been deprecated, so it's better to use loadSandboxConfig before loadSandboxConfigFromOldStore, so I have new change for this, ptal.

virtcontainers/sandbox.go
```diff
@@ -728,14 +728,14 @@ func fetchSandbox(ctx context.Context, sandboxID string) (sandbox *Sandbox, err
 
 	var config SandboxConfig
 
-	// Try to load sandbox config from old store at first.
-	c, ctx, err := loadSandboxConfigFromOldStore(ctx, sandboxID)
+	// Try to load sandbox config from new store at first.
```
fidencio, line 731: @keloyang, what was the reason of inverting the logic here?
I think we still should try to load from the old store at first.

virtcontainers/sandbox.go
```diff
@@ -837,7 +837,11 @@ func (s *Sandbox) Delete() error {
 	}
 
 	s.agent.cleanup(s)
-
+	if  useOldStore(s.ctx) && s.store != nil {
```
fidencio, line 840: @keloyang what's the reason of only doing this for an old store?

fidencio (commented):
@keloyang,

I've made some comments inline, but I must admit I'm failing to reproduce the issues.

What's exactly the version of the runtime you're using that you can reproduce it?

virtcontainers/sandbox.go
keloyang: The root cause of the residue is loadSandboxConfigFromOldStore is called but old store file is not exist,  and if load sandbox from new store successfull, there is no need to call loadSandboxConfigFromOldStore.

virtcontainers/sandbox.go
keloyang: useOldStore  return true only loadSandboxConfigFromOldStore return successfully,  so I add useOldStore here, but it's not necessary.

keloyang:
@fidencio  please show me how do you try to reproduce,  I use the master branch, and it's easy to reproduce.
```
[root@centos1 ~]# ls  /var/lib/vc/sbs/                                               
[root@centos1 ~]# docker run --rm -ti --runtime untrusted-runtime 018c9d7b792b echo; ls  /var/lib/vc/sbs/

15760f7e07c98e4f4ed9d7aba4076de14475dd71a61be9919b24037c810b37aa
[root@centos1 ~]# docker run --rm -ti --runtime untrusted-runtime 018c9d7b792b echo; ls  /var/lib/vc/sbs/

15760f7e07c98e4f4ed9d7aba4076de14475dd71a61be9919b24037c810b37aa  cd86011f6c0b779a2e3f40b49fa06463757ac1413ba70007e14462b7d0782e24
[root@centos1 ~]# docker run --rm -ti --runtime untrusted-runtime 018c9d7b792b echo; ls  /var/lib/vc/sbs/

15760f7e07c98e4f4ed9d7aba4076de14475dd71a61be9919b24037c810b37aa  93fb2abfbe151c8f12e3b2b7f24321479b2425af80acb93f2f4c2263fb679621  cd86011f6c0b779a2e3f40b49fa06463757ac1413ba70007e14462b7d0782e24
```

virtcontainers/sandbox.go
fidencio: @amshinde, @egernst,

Let me ask some more experienced people for some help.

I think we still should try to load the old store first.

In that very same code path we call `createSandbox()`, which calls `newSandbox()` and there depending on whether we use the old store or not we'd call `store.NewCVSandboxStore()`.

My question here is, why we only call that for one case?

keloyang:
ping @jodh-intel @amshinde @fidencio

jodh-intel:
@keloyang - please check the CI's at the end of your PR. Travis is failing on this one too:

```
virtcontainers/sandbox.go:840: File is not `gofmt`-ed with `-s` (gofmt)
```

virtcontainers/sandbox.go
fidencio: @jodh-intel, maybe you can help me with this one? ^^

fidencio:
> @fidencio please show me how do you try to reproduce, I use the master branch, and it's easy to reproduce.
> 
> ```
> [root@centos1 ~]# ls  /var/lib/vc/sbs/                                               
> [root@centos1 ~]# docker run --rm -ti --runtime untrusted-runtime 018c9d7b792b echo; ls  /var/lib/vc/sbs/
> 
> 15760f7e07c98e4f4ed9d7aba4076de14475dd71a61be9919b24037c810b37aa
> [root@centos1 ~]# docker run --rm -ti --runtime untrusted-runtime 018c9d7b792b echo; ls  /var/lib/vc/sbs/
> 
> 15760f7e07c98e4f4ed9d7aba4076de14475dd71a61be9919b24037c810b37aa  cd86011f6c0b779a2e3f40b49fa06463757ac1413ba70007e14462b7d0782e24
> [root@centos1 ~]# docker run --rm -ti --runtime untrusted-runtime 018c9d7b792b echo; ls  /var/lib/vc/sbs/
> 
> 15760f7e07c98e4f4ed9d7aba4076de14475dd71a61be9919b24037c810b37aa  93fb2abfbe151c8f12e3b2b7f24321479b2425af80acb93f2f4c2263fb679621  cd86011f6c0b779a2e3f40b49fa06463757ac1413ba70007e14462b7d0782e24
> ```

I was able to reproduce the issue and I raised a comment on a part that I don't have a clear understanding. Let's wait till the other developers jump into that.

keloyang:
> virtcontainers/sandbox.go

updated and sorry for not gofmt, thanks @jodh-intel

jodh-intel (approved):
Thanks @keloyang.

lgtm

jodh-intel:
@keloyang - I've restarted the failing metrics CI to see if this PR now passes. Please can you port this to 2.0 (https://github.com/kata-containers/kata-containers), or add a link to the 2.0 PR here if already done?

amshinde:
I had tested a previous version of this PR, will like to take another look before this is merged.

virtcontainers/sandbox.go
fidencio: @amshinde, @jodh-intel, @egernst,
ping about this question.

amshinde: @fidencio This patch makes sense to me. We have kept the old store code around for backward compatibility. But by checking for the old store and trying to initialize one and checking for failure while using the old store, we were inadvertently creating the directory structure under `/var/lib/vc/sbs/{sandbox_id}` and leaving that around after the sandbox is deleted. 
It makes sense to me to check for the new store first instead. I tested this patch and works as expected.

amshinde:
lgtm @fidencio. I think this one is ready to be merged.

keloyang:
I think there is no need to do porting for 2.0-dev, because there is no code to call loadSandboxConfigFromOldStore in fetchSandbox, see https://github.com/kata-containers/kata-containers/blob/2.0-dev/src/runtime/virtcontainers/sandbox.go#L650.
ping @jodh-intel @amshinde @fidencio

jodh-intel:
@keloyang - Regarding, https://github.com/kata-containers/runtime/pull/2922#issuecomment-707636127, I think you are correct, so I'm removing the `needs-forward-port` label...

Issue #2918, opened 2020-09-01
lazy Umount 


keloyang:
https://github.com/kata-containers/runtime/blob/master/virtcontainers/factory/template/template.go#L89 use a zero flag to Unmount, and I found that runtime uses lazy umount everywhere else,  do we need to modify it ? If necessary, I can modify it .

PR #2923, opened 2020-09-01
factory: Use lazy unmount

keloyang:
Fixes: #2918

### DIFF
--- a/virtcontainers/factory/template/template.go
+++ b/virtcontainers/factory/template/template.go
@@ -16,6 +16,7 @@ import (
 	pb "github.com/kata-containers/runtime/protocols/cache"
 	vc "github.com/kata-containers/runtime/virtcontainers"
 	"github.com/kata-containers/runtime/virtcontainers/factory/base"
+	"github.com/sirupsen/logrus"
 )
 
 type template struct {
@@ -24,6 +25,7 @@ type template struct {
 }
 
 var templateWaitForAgent = 2 * time.Second
+var templateLog = logrus.WithField("source", "virtcontainers/factory")
 
 // Fetch finds and returns a pre-built template factory.
 // TODO: save template metadata and fetch from storage.
@@ -86,8 +88,13 @@ func (t *template) GetVMStatus() []*pb.GrpcVMStatus {
 }
 
 func (t *template) close() {
-	syscall.Unmount(t.statePath, 0)
-	os.RemoveAll(t.statePath)
+	if err := syscall.Unmount(t.statePath, syscall.MNT_DETACH); err != nil {
+		t.Logger().WithError(err).Errorf("failed to unmount %s", t.statePath)
+	}
+
+	if err := os.RemoveAll(t.statePath); err != nil {
+		t.Logger().WithError(err).Errorf("failed to remove %s", t.statePath)
+	}
 }
 
 func (t *template) prepareTemplateFiles() error {
@@ -170,4 +177,17 @@ func (t *template) checkTemplateVM() error {
 
 	_, err = os.Stat(t.statePath + "/state")
 	return err
+}
+
+// Logger returns a logrus logger appropriate for logging template messages
+func (t *template) Logger() *logrus.Entry {
+	return templateLog.WithFields(logrus.Fields{
+		"subsystem": "template",
+	})
+}
+
+// SetLogger sets the logger for the factory template.
+func SetLogger(ctx context.Context, logger logrus.FieldLogger) {
+	fields := templateLog.Data
+	templateLog = logger.WithFields(fields)
 }
--- a/cli/main.go
+++ b/cli/main.go
@@ -22,6 +22,7 @@ import (
 	vc "github.com/kata-containers/runtime/virtcontainers"
 	exp "github.com/kata-containers/runtime/virtcontainers/experimental"
 	vf "github.com/kata-containers/runtime/virtcontainers/factory"
+	tl "github.com/kata-containers/runtime/virtcontainers/factory/template"
 	"github.com/kata-containers/runtime/virtcontainers/pkg/oci"
 	"github.com/kata-containers/runtime/virtcontainers/pkg/rootless"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
@@ -243,6 +244,9 @@ func setExternalLoggers(ctx context.Context, logger *logrus.Entry) {
 
 	// Set vm factory logger.
 	vf.SetLogger(ctx, logger)
+
+	// Set vm factory template logger.
+	tl.SetLogger(ctx, logger)
 
 	// Set the OCI package logger.
 	oci.SetLogger(ctx, logger)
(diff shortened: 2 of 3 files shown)


cmaf:
Hi @keloyang, could you update the commit with more details?

/test-ubuntu

fidencio (commented):
@keloyang, I'm not convinced about this one.
Would you mind to elaborate the need of a lazy unmount here?

keloyang:
@fidencio thanks for you review.  I have not found a  obvious problem with it, but I think it is better to use lazy umount, because it can reduce the probability of problems.

```
[root@centos0 test]# mkdir a b
[root@centos0 test]# mount -B a b
[root@centos0 test]# cd b
[root@centos0 b]# umount ../b
umount: /home/workspace/test/b: target is busy.
        (In some cases useful info about processes that use
         the device is found by lsof(8) or fuser(1))
[root@centos0 b]# umount ../b -l     # use lazy umount
[root@centos0 b]# 
```

devimc:
thanks @keloyang , could you elaborate a little bit more about what issue are you facing? do we really need this in the factory?

cc @bergwolf @lifupan

keloyang:
@devimc thanks for your review. also ping @bergwolf and @lifupan ptal.

we can have the following case,
1. start kata container with factory feature, this need kata-runtime config to enable factory and use initrd as base image.
2. start a kata container.
3. cd /root; cd /run/vc/vm/template dir, this will make /run/vc/vm/template to be in used. 
4. destroy vm template with kata-runtime factory destroy , and check the template mountpoint.
we can see  the template mountpoints will add everytime we repeat the above steps .
```
[root@centos1 template]# mount |grep template
[root@centos1 template]# docker run -ti --rm  --runtime untrusted-runtime --net none busybox echo

[root@centos1 template]# cd /root; cd /run/vc/vm/template/                
[root@centos1 template]# /kata/bin/kata-runtime factory destroy 
vm factory destroyed
[root@centos1 template]# mount |grep template
tmpfs on /run/vc/vm/template type tmpfs (rw,nosuid,nodev,relatime,seclabel,size=2105344k)
[root@centos1 template]# docker run -ti --rm  --runtime untrusted-runtime --net none busybox echo

[root@centos1 template]# cd /root; cd /run/vc/vm/template/      
[root@centos1 template]# /kata/bin/kata-runtime factory destroy 
vm factory destroyed
[root@centos1 template]# mount |grep template
tmpfs on /run/vc/vm/template type tmpfs (rw,nosuid,nodev,relatime,seclabel,size=2105344k)
tmpfs on /run/vc/vm/template type tmpfs (rw,nosuid,nodev,relatime,seclabel,size=2105344k)
```

jodh-intel:
Thanks @keloyang. This does indeed look like a bug. I'd prefer we also make `template.close()` return an `error` as that can be passed back to the caller (or atleast logged somehow) in `template.prepareTemplateFiles()`.

@keloyang - Once @bergwolf / @lifupan have commented on this, would you also be able to raise a PR on Kata 2.0 (https://github.com/kata-containers/kata-containers/tree/2.0-dev/src/runtime)? Adding forward port label so we don't forget to do the port...

lifupan:
Since the close() had ignored the error of umount, I think it's safe to do a lazy umount here.

devimc:
thanks @keloyang - Could you update the commit message with the case that you mention?

> we can have the following case, .........

keloyang:
@devimc I have updated the commit message. thanks for your review.

jodh-intel:
> Since the close() had ignored the error of umount, I think it's safe to do a lazy umount here.

I agree, but strictly `close()` shouldn't have ignored the error so we could have found this problem quicker. But if we now use lazy umount we don't expect any errors, hence should really return `error` from `close()` "just in case" to avoid any future hidden bugs.

jodh-intel:
added dnm label as there is still some debate around this.

keloyang:
> I agree, but strictly close() shouldn't have ignored the error so we could have found this problem quicker. But if we now use lazy umount we don't expect any errors, hence should really return error from close() "just in case" to avoid any future hidden bugs.

@jodh-intel thanks for your review and thinking deeply
I agree with you, we shouldn't  ignored the error. but I think we still can use lazy umount. If we keep to use syscall.Unmount(*, 0),we can find the error and then we will looking for solutions. How to fix ? we can print err info, if we  find the error info later, we can  do a manual cleaning, alternatively we use lazy unmount, as far as my concerned,I like lazy unmount better, so I have add some log to close()， ptal.

virtcontainers/factory/template/template.go
```diff
@@ -171,3 +178,10 @@ func (t *template) checkTemplateVM() error {
 	_, err = os.Stat(t.statePath + "/state")
 	return err
 }
+
+// Logger returns a logrus logger appropriate for logging template messages
+func (t *template) Logger() *logrus.Entry {
```
jodh-intel, line 183: Thanks for adding this. But can you "stitch" this into the existing logging facility by creating a `SetLogger()` function here and then calling it in:

- https://github.com/kata-containers/runtime/blob/master/virtcontainers/api.go#L49

Then, Having this logger function just return the global logger in this module that will now contain all the log fields. This is required for consistency and for https://github.com/kata-containers/tests/tree/master/cmd/log-parser#logfile-requirements.

jodh-intel (commented):
Thanks for updating @keloyang!

Just one more tweak I think.

virtcontainers/factory/template/template.go
keloyang: I have add SetLogger, ptal. thanks @jodh-intel

virtcontainers/factory/template/template.go
```diff
@@ -171,3 +178,19 @@ func (t *template) checkTemplateVM() error {
...
+	return templateLog.WithFields(logrus.Fields{
+		"subsystem": "template",
+	})
+}
+
+// SetLogger sets the logger for the factory template.
+func SetLogger(ctx context.Context, logger logrus.FieldLogger) {
+	fields := logrus.Fields{
+		"source": "virtcontainers",
```
jodh-intel, line 192: Nit: Shouldn't this be:

```go
"source": "virtcontainers/factory/template",
```

jodh-intel (approved):
Thanks @keloyang.

virtcontainers/factory/template/template.go
keloyang: I'm sorry, I  didn't understand it correctly. I have change it. @jodh-intel

jodh-intel:
Thanks @keloyang.

keloyang:
ping @lifupan @jodh-intel @fidencio

fidencio:
@keloyang, we're aware of the PR and we'll move forward with this as soon as possible.
Right now, our metrics CI is failing and that's a blocker for pushing new PRs. Of course, we could bypass and do a git push directly instead of relying on GitHub "merge", but right now I don't see any issue with a priority that's high enough to do so.

So, all-in-all, this PR is on our radar and we'll get to this (and all the others) as soon as we unblock the queue.

jodh-intel:
forward port PR: https://github.com/kata-containers/kata-containers/pull/939

PR #2972, opened 2020-09-18
cpuset: fixes to address VM sizing and constraining based on cpuset.mems

egernst:
*   CPUSet cgroup allows for pinning the memory associated with a cpuset to
    a given numa node. Similar to cpuset.cpus, we should take cpuset.mems
    into account for the sandbox-cgroup that Kata creates.
*  If quota is not being enforced on any containers but CPUsets are
    specified, take the number of CPU sets into account when sizing the
    virtual machine.

### DIFF
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -1863,7 +1863,10 @@ func (s *Sandbox) updateResources() error {
 		return fmt.Errorf("sandbox config is nil")
 	}
 
-	sandboxVCPUs := s.calculateSandboxCPUs()
+	sandboxVCPUs, err := s.calculateSandboxCPUs()
+	if err != nil {
+		return err
+	}
 	// Add default vcpus for sandbox
 	sandboxVCPUs += s.hypervisor.hypervisorConfig().NumVCPUs
 
@@ -1923,8 +1926,9 @@ func (s *Sandbox) calculateSandboxMemory() int64 {
 	return memorySandbox
 }
 
-func (s *Sandbox) calculateSandboxCPUs() uint32 {
+func (s *Sandbox) calculateSandboxCPUs() (uint32, error) {
 	mCPU := uint32(0)
+	cpusetCount := int(0)
 
 	for _, c := range s.config.Containers {
 		// Do not hot add again non-running containers resources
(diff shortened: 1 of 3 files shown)


fidencio:
@egernst, please take a look:
```
INFO: Running 'go test' as current user on package 'github.com/kata-containers/runtime/virtcontainers' with flags '-v -race -timeout 30s'
# github.com/kata-containers/runtime/virtcontainers [github.com/kata-containers/runtime/virtcontainers.test]
virtcontainers/sandbox_test.go:139:8: assignment mismatch: 1 variable but sandbox.calculateSandboxCPUs returns 2 values
FAIL	github.com/kata-containers/runtime/virtcontainers [build failed]
FAIL
make: *** [Makefile:680: go-test] Error 2
```

And a general comment, we're now mixing cpu and memory sets / getters. Wouldn't be better to have one function for each one of those, as we already do in the sandbox level?

fidencio:
> And a general comment, we're now mixing cpu and memory sets / getters. Wouldn't be better to have one function for each one of those, as we already do in the sandbox level?

Hmm. No, thinking a little bit more about this, no. It's specfically related to cpuset. So, please, ignore my previous suggestion.

virtcontainers/sandbox.go
```diff
@@ -1938,9 +1942,23 @@ func (s *Sandbox) calculateSandboxCPUs() uint32 {
 				mCPU += utils.CalculateMilliCPUs(*cpu.Quota, *cpu.Period)
 			}
 
+			//todo we should handle error
+			set, err := cpuset.Parse(cpu.Cpus)
```
crosbymichael, line 1946: should you just have some simple sanity checks such as `cpu.Cpus != ""` then handle the error as you know it's invalid configuration?

egernst: Ah, I think I should update the comment.  The simple sanity check is handled within Parse itself, and if we do receive an invalid string (ie, something where Atoi fails), I think we should error out/stop. 

My initial thought is I should just remove then comment/1945.  WDYT @crosbymichael

crosbymichael: ok, sounds good

egernst:
(sorry to keep thrashing the CI).

/test

jodh-intel (approved):
Thanks @egernst.

lgtm

devimc (approved):
thanks @egernst

PR #2980, opened 2020-09-22
shimv2: Add tracing

cmaf:
Added tracing with tracer being created in Create, which should ideally
happen in New. Currently getting no trace or span output when the tracer
is created in New or StartShim, or Create before create() is called.
Also no log output from these functions.

To do:
 - Determine which context is more appropriate to associate with spans
(ctx passed in or ctx saved in service struct - currently implemented
using service ctx)
 - Find out what in create() allows tracing to work and implement it in
New if possible

Signed-off-by: Chelsea Mafrica <chelsea.e.mafrica@intel.com>

### DIFF
--- a/containerd-shim-v2/service.go
+++ b/containerd-shim-v2/service.go
@@ -25,6 +25,7 @@ import (
 	"github.com/containerd/typeurl"
 	ptypes "github.com/gogo/protobuf/types"
 	"github.com/opencontainers/runtime-spec/specs-go"
+	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
@@ -72,6 +73,21 @@ func New(ctx context.Context, id string, publisher events.Publisher) (cdshim.Shi
 	vci.SetLogger(ctx, logger)
 	katautils.SetLogger(ctx, logger, logger.Logger.Level)
 
+	// load runtime config so that tracing can start if enabled
+	_, runtimeConfig, err := katautils.LoadConfiguration("", false, true)
+	if err != nil {
+		return nil, err
+	}
+
+	// create tracer
+	_, err = katautils.CreateTracer("kata")
+	if err != nil {
+		return nil, err
+	}
+	// create span
+	span, ctx := trace(ctx, "New")
+	defer span.Finish()
+
 	ctx, cancel := context.WithCancel(ctx)
 
 	s := &service{
@@ -79,6 +95,7 @@ func New(ctx context.Context, id string, publisher events.Publisher) (cdshim.Shi
 		pid:        uint32(os.Getpid()),
 		ctx:        ctx,
 		containers: make(map[string]*container),
+		config:     &runtimeConfig,
 		events:     make(chan interface{}, chSize),
 		ec:         make(chan exit, bufferSize),
 		cancel:     cancel,
@@ -161,6 +178,13 @@ func newCommand(ctx context.Context, containerdBinary, id, containerdAddress str
 // StartShim willl start a kata shimv2 daemon which will implemented the
 // ShimV2 APIs such as create/start/update etc containers.
 func (s *service) StartShim(ctx context.Context, id, containerdBinary, containerdAddress string) (string, error) {
+	// Stop tracing here since a new tracer will be created the next time New()
+	// is called again after StartShim()
+	defer katautils.StopTracing(s.ctx)
+
+	span, _ := trace(s.ctx, "StartShim")
+	defer span.Finish()
+
 	bundlePath, err := os.Getwd()
 	if err != nil {
 		return "", err
@@ -274,7 +298,22 @@ func getTopic(e interface{}) string {
 	return cdruntime.TaskUnknownTopic
 }
 
+func trace(ctx context.Context, name string) (opentracing.Span, context.Context) {
+	if ctx == nil {
+		logrus.WithField("type", "bug").Error("trace called before context set")
+		ctx = context.Background()
+	}
+
+	span, ctx := opentracing.StartSpanFromContext(ctx, name)
+	span.SetTag("source", "runtime")
+
+	return span, ctx
+}
+
 func (s *service) Cleanup(ctx context.Context) (_ *taskAPI.DeleteResponse, err error) {
+	span, _ := trace(s.ctx, "Cleanup")
+	defer span.Finish()
+
 	//Since the binary cleanup will return the DeleteResponse from stdout to
 	//containerd, thus we must make sure there is no any outputs in stdout except
 	//the returned response, thus here redirect the log to stderr in case there's
@@ -330,6 +369,9 @@ func (s *service) Cleanup(ctx context.Context) (_ *taskAPI.DeleteResponse, err e
 
 // Create a new sandbox or container with the underlying OCI runtime
 func (s *service) Create(ctx context.Context, r *taskAPI.CreateTaskRequest) (_ *taskAPI.CreateTaskResponse, err error) {
+	span, _ := trace(s.ctx, "Create")
+	defer span.Finish()
+
 	defer func() {
 		err = toGRPC(err)
 	}()
@@ -381,6 +423,9 @@ func (s *service) Create(ctx context.Context, r *taskAPI.CreateTaskRequest) (_ *
 
 // Start a process
 func (s *service) Start(ctx context.Context, r *taskAPI.StartRequest) (_ *taskAPI.StartResponse, err error) {
+	span, _ := trace(s.ctx, "Start")
+	defer span.Finish()
+
 	defer func() {
 		err = toGRPC(err)
 	}()
@@ -427,6 +472,9 @@ func (s *service) Start(ctx context.Context, r *taskAPI.StartRequest) (_ *taskAP
 
 // Delete the initial process and container
 func (s *service) Delete(ctx context.Context, r *taskAPI.DeleteRequest) (_ *taskAPI.DeleteResponse, err error) {
+	span, _ := trace(s.ctx, "Delete")
+	defer span.Finish()
+
 	defer func() {
 		err = toGRPC(err)
 	}()
@@ -474,6 +522,9 @@ func (s *service) Delete(ctx context.Context, r *taskAPI.DeleteRequest) (_ *task
 
 // Exec an additional process inside the container
 func (s *service) Exec(ctx context.Context, r *taskAPI.ExecProcessRequest) (_ *ptypes.Empty, err error) {
+	span, _ := trace(s.ctx, "Exec")
+	defer span.Finish()
+
 	defer func() {
 		err = toGRPC(err)
 	}()
@@ -507,6 +558,9 @@ func (s *service) Exec(ctx context.Context, r *taskAPI.ExecProcessRequest) (_ *p
 
 // ResizePty of a process
 func (s *service) ResizePty(ctx context.Context, r *taskAPI.ResizePtyRequest) (_ *ptypes.Empty, err error) {
+	span, _ := trace(s.ctx, "ResizePty")
+	defer span.Finish()
+
 	defer func() {
 		err = toGRPC(err)
 	}()
@@ -541,6 +595,9 @@ func (s *service) ResizePty(ctx context.Context, r *taskAPI.ResizePtyRequest) (_
 
 // State returns runtime state information for a process
 func (s *service) State(ctx context.Context, r *taskAPI.StateRequest) (_ *taskAPI.StateResponse, err error) {
+	span, _ := trace(s.ctx, "State")
+	defer span.Finish()
+
 	defer func() {
 		err = toGRPC(err)
 	}()
@@ -584,11 +641,13 @@ func (s *service) State(ctx context.Context, r *taskAPI.StateRequest) (_ *taskAP
 		Terminal:   execs.tty.terminal,
 		ExitStatus: uint32(execs.exitCode),
 	}, nil
-
 }
 
 // Pause the container
 func (s *service) Pause(ctx context.Context, r *taskAPI.PauseRequest) (_ *ptypes.Empty, err error) {
+	span, _ := trace(s.ctx, "Pause")
+	defer span.Finish()
+
 	defer func() {
 		err = toGRPC(err)
 	}()
@@ -623,6 +682,9 @@ func (s *service) Pause(ctx context.Context, r *taskAPI.PauseRequest) (_ *ptypes
 
 // Resume the container
 func (s *service) Resume(ctx context.Context, r *taskAPI.ResumeRequest) (_ *ptypes.Empty, err error) {
+	span, _ := trace(s.ctx, "Resume")
+	defer span.Finish()
+
 	defer func() {
 		err = toGRPC(err)
 	}()
@@ -655,6 +717,9 @@ func (s *service) Resume(ctx context.Context, r *taskAPI.ResumeRequest) (_ *ptyp
 
 // Kill a process with the provided signal
 func (s *service) Kill(ctx context.Context, r *taskAPI.KillRequest) (_ *ptypes.Empty, err error) {
+	span, _ := trace(s.ctx, "Kill")
+	defer span.Finish()
+
 	defer func() {
 		err = toGRPC(err)
 	}()
@@ -707,6 +772,9 @@ func (s *service) Kill(ctx context.Context, r *taskAPI.KillRequest) (_ *ptypes.E
 // Since for kata, it cannot get the process's pid from VM,
 // thus only return the Shim's pid directly.
 func (s *service) Pids(ctx context.Context, r *taskAPI.PidsRequest) (_ *taskAPI.PidsResponse, err error) {
+	span, _ := trace(s.ctx, "Pids")
+	defer span.Finish()
+
 	var processes []*task.ProcessInfo
 
 	defer func() {
@@ -725,6 +793,9 @@ func (s *service) Pids(ctx context.Context, r *taskAPI.PidsRequest) (_ *taskAPI.
 
 // CloseIO of a process
 func (s *service) CloseIO(ctx context.Context, r *taskAPI.CloseIORequest) (_ *ptypes.Empty, err error) {
+	span, _ := trace(s.ctx, "CloseIO")
+	defer span.Finish()
+
 	defer func() {
 		err = toGRPC(err)
 	}()
@@ -761,6 +832,9 @@ func (s *service) CloseIO(ctx context.Context, r *taskAPI.CloseIORequest) (_ *pt
 
 // Checkpoint the container
 func (s *service) Checkpoint(ctx context.Context, r *taskAPI.CheckpointTaskRequest) (_ *ptypes.Empty, err error) {
+	span, _ := trace(s.ctx, "Checkpoint")
+	defer span.Finish()
+
 	defer func() {
 		err = toGRPC(err)
 	}()
@@ -770,6 +844,9 @@ func (s *service) Checkpoint(ctx context.Context, r *taskAPI.CheckpointTaskReque
 
 // Connect returns shim information such as the shim's pid
 func (s *service) Connect(ctx context.Context, r *taskAPI.ConnectRequest) (_ *taskAPI.ConnectResponse, err error) {
+	span, _ := trace(s.ctx, "Connect")
+	defer span.Finish()
+
 	defer func() {
 		err = toGRPC(err)
 	}()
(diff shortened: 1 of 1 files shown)


cmaf:
@egernst @jodh-intel could either of you test locally? also maybe @amshinde wants to take a look

cmaf:
The answer to the 2nd To Do item is that the tracer must be created after [`katautils.LoadConfiguration`](https://github.com/kata-containers/runtime/blob/master/pkg/katautils/config.go#L1179). This happens [relatively early](https://github.com/kata-containers/runtime/blob/master/cli/main.go#L337) in the cli, but doesn't happen in shimv2 until [`create`](https://github.com/kata-containers/runtime/blob/master/containerd-shim-v2/create.go#L34) (from [`Create`](https://github.com/kata-containers/runtime/blob/master/containerd-shim-v2/service.go#L332) in service.go) calls [`loadRuntimeConfig`](https://github.com/kata-containers/runtime/blob/master/containerd-shim-v2/create.go#L162). 

We could try to call `LoadConfiguration` earlier in shimv2; I'm not sure if there was a reason for calling it after `New` and `StartShim`.

cmaf:
Maybe @bergwolf or @lifupan have advice on whether or not `LoadConfiguration` should be called earlier?

lifupan:
> Maybe @bergwolf or @lifupan have advice on whether or not `LoadConfiguration` should be called earlier?

Hi @cmaf 

Since sihmv2  support loading config files from different places which can be passed from containerd through Create call, thus it's better to load it in Create service.

cmaf:
@jodh-intel @amshinde @egernst this traces from `New` which works when you load a runtime config in `New`.

I still need to: (1) add check for error after initial `LoadRuntimeConfig`, (2) check for tracing enabled before creating tracer, (3) ensure runtime config gets overwritten in `Create`, (4) test additional scenarios.

egernst:
Thanks @cmaf!
What's next here then? Will you start at the new call now if the config allows? What if you create  tracer in new but the loaded config has tracing disabled?

containerd-shim-v2/service.go
```diff
@@ -274,7 +329,23 @@ func getTopic(e interface{}) string {
 	return cdruntime.TaskUnknownTopic
 }
 
+func trace(ctx context.Context, tracer opentracing.Tracer, name string) opentracing.Span {
+	if tracer != opentracing.Tracer(nil) {
+		span, _ := opentracing.StartSpanFromContext(ctx, name)
+		span.SetTag("subsystem", "runtime")
```
jodh-intel, line 335: I think you could set this tag in the root span to avoid having to set every time here.

jodh-intel (commented):
Thanks @cmaf. A few comments.

cmaf:
> What's next here then? Will you start at the new call now if the config allows? What if you create tracer in new but the loaded config has tracing disabled?

@egernst yes it starts at `New` now if the config allows. I haven't tested cases yet where it is enabled in the first config and disabled in the second (or vice versa), but with the former it should create a trace containing `New` and `StartShim` with no other effects, and with the latter tracing still should not work at all since the tracer is created in `New` and not `Create` (using the original place where the config is loaded requires the tracer to be created after `create()` in `Create`).

containerd-shim-v2/service.go
cmaf: @jodh-intel I tried setting it only at `New` where the tracer is created and don't see a subsystem for the other functions, so I'm leaving it in for now.

containerd-shim-v2/service.go
```diff
@@ -72,17 +73,62 @@ func New(ctx context.Context, id string, publisher events.Publisher) (cdshim.Shi
...
 		pid:        uint32(os.Getpid()),
 		ctx:        ctx,
+		tracer:     tracer,
 		containers: make(map[string]*container),
+		config:     &runtimeConfig,
 		events:     make(chan interface{}, chSize),
 		ec:         make(chan exit, bufferSize),
 		cancel:     cancel,
 	}
+	/*// load runtime config so that tracing can start if enabled
```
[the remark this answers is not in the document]
cmaf: thanks! removed

containerd-shim-v2/service.go
```diff
@@ -161,6 +208,14 @@ func newCommand(ctx context.Context, containerdBinary, id, containerdAddress str
 // StartShim willl start a kata shimv2 daemon which will implemented the
 // ShimV2 APIs such as create/start/update etc containers.
 func (s *service) StartShim(ctx context.Context, id, containerdBinary, containerdAddress string) (string, error) {
+	// Stop tracing because a new tracer will be created when New is called again after StartShim
+	defer katautils.StopTracing(s.ctx)
+
+	span := trace(s.ctx, s.tracer, "StartShim")
+	if span != opentracing.Span(nil) {
```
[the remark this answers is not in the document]
cmaf: Simplified the trace function and the logic

containerd-shim-v2/service.go
```diff
@@ -161,6 +208,14 @@ func newCommand(ctx context.Context, containerdBinary, id, containerdAddress str
 // StartShim willl start a kata shimv2 daemon which will implemented the
 // ShimV2 APIs such as create/start/update etc containers.
 func (s *service) StartShim(ctx context.Context, id, containerdBinary, containerdAddress string) (string, error) {
+	// Stop tracing because a new tracer will be created when New is called again after StartShim
+	defer katautils.StopTracing(s.ctx)
```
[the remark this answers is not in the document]
cmaf: @jodh-intel I'm not sure I understand this. I could call `StopTracing()` before the tracer is created in `New()`, but `s.tracer` doesn't exist until halfway through the function. A new context is used every time `New()` is called. `New()` is called twice per container, initially followed by `StartShim()`, so there is one context used between those two, and a different context for the second `New()` and subsequent functions.

cmaf:
@jodh-intel Thanks, I updated it based on your feedback and am testing a few locations for `StopTracing()`.

containerd-shim-v2/service.go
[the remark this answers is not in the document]
cmaf: @jodh-intel I thought about this some more and tracing can't be stopped earlier in `New()` because stopping tracing depends on using a specific context, which isn't available since a new one is used in that function. It is possible I'm missing functions in between `New()` and `StartShim()`, but `StartShim()` is the only function that I've seen called after `New()` with the same context where we can end the first trace.

containerd-shim-v2/service.go
```diff
@@ -818,6 +910,9 @@ func (s *service) Stats(ctx context.Context, r *taskAPI.StatsRequest) (_ *taskAP
 
 // Update a running container
 func (s *service) Update(ctx context.Context, r *taskAPI.UpdateTaskRequest) (_ *ptypes.Empty, err error) {
+	span := trace(s.ctx, "Update")
```
egernst, line 913: @jodh-intel  I'm curious for your input here. 

AFAIU, for shimv2, we'll get a context for the life of the shim itself, which is what we store in s.ctx (I think it may be wise to rename this to be shimCtx to make it more clear). In this case, when we're called, we have a context associated with the given request, which has a 'shorter' scope, ctx.

1. Should we use ctx in the 'short lived' maintenance requests rather than s.ctx? 
2. Should we 'plumb' ctx through to the actual virtcontainers impl here? 

Basically, I am starting to think that we should do this per 'request' into the VC API rather than having a sandbox level ctx that is stored in the struct. The struct makes sense maybe when we're just doing traces, but I think we should be able to also handle contexts per request from shim-v2, in case of timeouts, cancellations, etc. First step would at least be to plumb this through?

egernst: /cc @lifupan for input as well

amshinde:
@cmaf I didnt get a chance to complete the review today, will take a look at this tomorrow.

cmaf:
forward port pr: https://github.com/kata-containers/kata-containers/pull/904

containerd-shim-v2/service.go
```diff
@@ -72,13 +73,33 @@ func New(ctx context.Context, id string, publisher events.Publisher) (cdshim.Shi
...
 	katautils.SetLogger(ctx, logger, logger.Logger.Level)
 
+	// load runtime config so that tracing can start if enabled
+	_, runtimeConfig, err := katautils.LoadConfiguration("", false, true)
+	if err != nil {
+		return nil, err
+	}
+
+	// create tracer
+	tracer := opentracing.Tracer(nil)
```
amshinde, line 83: Why are you instantiating a tracer here?

containerd-shim-v2/service.go
```diff
@@ -274,7 +302,20 @@ func getTopic(e interface{}) string {
 	return cdruntime.TaskUnknownTopic
 }
 
+func trace(ctx context.Context, name string) opentracing.Span {
+	if ctx == nil {
+		logrus.WithField("type", "bug").Error("trace called before context set")
+		ctx = context.Background()
+	}
+	span, _ := opentracing.StartSpanFromContext(ctx, name)
```
amshinde: Shoudnt you be fetching the context returned from StartSpanFromContext here? This function should return the context which should be used henceforth in the calling function.

containerd-shim-v2/service.go
```diff
@@ -86,10 +86,10 @@ func New(ctx context.Context, id string, publisher events.Publisher) (cdshim.Shi
 		if err != nil {
 			return nil, err
 		}
-		// create span
-		span := trace(ctx, "New")
-		defer span.Finish()
 	}
+	// create span
+	span, ctx := trace(ctx, "New")
```
amshinde, line 92: ok nevermind, I see you have addressed my previous comment here.

containerd-shim-v2/service.go
cmaf: @amshinde New is the first function called in service.go, it must happen after the runtime config is loaded (also added to New in this PR), and it must be started after the logging is set or you get errors running a container.

amshinde: I meant this could be changed to a variable declaration  `var tracer opentracing.Tracer` as thats the approach we follow.
I also feel that you need not check for runtimeConfig.Trace here, as looking at the CreateTracer function here:
https://github.com/kata-containers/runtime/blob/d937c067f223324ef2e428adc5e0f807d74c4855/pkg/katautils/tracing.go#L38

it uses a NOP trace implementation if tracing is disabled.
cc @jodh-intel to clarify why we took this approach

cmaf: @amshinde Oh I see; I misunderstood.

Please correct me if I'm misunderstanding your point about the runtime config, but I think we do need to check if we want a complete trace. If we don't check for `runtimeConfig.Trace` here, we won't be able to create the tracer and start tracing until `Create` and will miss spans for `New`, `StartShim`, and `New` again, because they will be treated as NOP. See [this comment](https://github.com/kata-containers/runtime/pull/2980#issuecomment-698466878). We can skip a few spans and start tracing later, but it is my understanding that the runtime config must be loaded (or at least runtimeConfig.Trace must be set) before tracing can begin.

cmaf: @amshinde I see (again), the actual check for tracing being enabled doesn't need to happen.

containerd-shim-v2/service.go
amshinde: Can you merge this commit with the previous one?

jodh-intel:
Hi @cmaf - I've atlast had a chance to try out this PR. Although I'm seeing trace spans, they are a couple of problems:

- I'm not seeing all the spans I'd expect to.
- The spans are not grouped into a single transaction.

Adding a bit of debug shows an initially somewhat confusing set of shimv2 calls being made:

```
New
StartShim
newCommand
New
Create
Wait
Wait
Start
Start
State
State
State
State
Delete
Delete
Shutdown
Shutdown
```

However, adding more shows the real picture: the Kata v2 shim (which is loaded for the lifetime of a single container) is forked twice. The group of API calls divides into two sets:


```
# Initial Kata shimv2 pid
New
StartShim
newCommand

# Double-forked Kata shimv2 pid
New
Create
Wait
Wait
Start
Start
State
State
State
State
Delete
Delete
Shutdown
Shutdown
```

This now makes more sense and leads to the following possibilities:

| Option | Create Tracer in function | Stop tracing function | Notes |
|-|-|-|-|
| 1 | `New()`<br/>(when `ppid != 1`) | `Shutdown()`<br/>(when `ppid == 1`) | Earliest opportunity to trace<br/>(and "widest" root span possible as it spans both pids). |
| 2 | `New()`<br/>(when `ppid == 1`) | `Shutdown()`<br/>(when `ppid == 1`) | "Sweet spot"? Only trace the double-forked shimv2 pid.<br/>Simpler. Nice wide root span. |
| 3 | `Create()`<br/>(when `ppid == 1`) | `Delete()`<br/>(when `ppid == 1`) | The narrowest root span that would make sense. |

I think Option 2 would be the best. And if you changed `pkg/katautils/tracing.go::StopTracing()` to set `tracing=false` at the end of the function, the fact that `Shutdown()` gets called twice by the same pid shouldn't be a problem (I think it's not unreasonable to end tracing at the first call to `Shutdown()` ;)

I appreciate that we're not tracing as early as possible with Option 2 (we wouldn't be tracing `StartShim()`), but we are atleast "spanning" the duration of a container.

containerd-shim-v2/service.go
```diff
@@ -72,13 +73,31 @@ func New(ctx context.Context, id string, publisher events.Publisher) (cdshim.Shi
...
+	defer span.Finish()
+
 	ctx, cancel := context.WithCancel(ctx)
 
 	s := &service{
 		id:         id,
 		pid:        uint32(os.Getpid()),
 		ctx:        ctx,
+		tracer:     tracer,
```
pmores, line 98: @cmaf Could you help me understand how this Tracer instance is used?  I'm probably overlooking something but I don't see it referenced anywhere.

cmaf: @pmores, thanks for catching this. You're not overlooking anything, I previously used `s.tracer` in the code but that had been removed. I removed this instance.

cmaf:
> I think Option 2 would be the best. And if you changed `pkg/katautils/tracing.go::StopTracing()` to set `tracing=false` at the end of the function, the fact that `Shutdown()` gets called twice by the same pid shouldn't be a problem (I think it's not unreasonable to end tracing at the first call to `Shutdown()` ;)

@jodh-intel Updated with a ppid check for scenario 2. If you generate a trace now you should see several ungrouped spans not specific to shimv2, and then a single `New` span with the spans traced in `service.go` grouped under it. Without the check, you will see the several ungrouped spans, and two `New` spans, one with `StartShim` grouped under it and the other with the remaining spans. I wasn't able to replicate the functions being called twice so I did not add the tracing check to `pkg/katautils/tracing.go::StopTracing()`.

The test for tracing in shimv2 will be in the 2.0-dev branch: https://github.com/kata-containers/tests/pull/3038.

containerd-shim-v2/service.go
pmores: @cmaf Thanks!

containerd-shim-v2/service.go
```diff
@@ -556,8 +571,10 @@ func (s *service) Exec(ctx context.Context, r *taskAPI.ExecProcessRequest) (_ *p
 
 // ResizePty of a process
 func (s *service) ResizePty(ctx context.Context, r *taskAPI.ResizePtyRequest) (_ *ptypes.Empty, err error) {
-	span, _ := trace(s.ctx, "ResizePty")
-	defer span.Finish()
+	if os.Getppid() == 1 {
+		span, _ := trace(s.ctx, "ResizePty")
+		defer span.Finish()
+	}
```
pmores, line 577: @cmaf Are the ppid checks useful though outside of New(), Shutdown(), Create() and Delete()?  I would assume that the first fork exits after spawning the second one...

amshinde:
Hi @cmaf, I just tried the latest version of your PR to test out the tracing output, but I am not seeing any traces related to CreateSandbox as I was seeing earlier. I do see that I have enabled tracing for the runtime in the configuration.toml file, but the only traces I am seeing are traces consisting of a single span for `grpc.Health/Check`. Can you check if the tracing works for you?

cmaf:
@amshinde yes it is working for me. I see `createSandbox` under `New`

containerd-shim-v2/service.go
cmaf: @pmores you're right, they actually don't need to be in all functions. However I am removing all the checks because we decided to trace all functions irrespective of ppid.

containerd-shim-v2/service.go
```diff
@@ -274,7 +298,20 @@ func getTopic(e interface{}) string {
 	return cdruntime.TaskUnknownTopic
 }
 
+func trace(ctx context.Context, name string) (opentracing.Span, context.Context) {
```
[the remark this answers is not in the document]
cmaf: @amshinde `Trace()` in `pkg/katautils/tracing.go` sets a tag for the cli, which we don't want in this case. I noticed virtcontainers had its own implementation of `trace()` so I implemented one for shimv2. I think we should fix this in a subsequent PR so that we're not replicating code.

pmores:
With the latest changes, this PR looks good to me at this point.

containerd-shim-v2/service.go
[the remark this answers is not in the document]
c3d: Agreed, but this seems like a simple fix, and probably easier to do in this commit than at a later time when the two functions may have diverged (they already began diverging here)

containerd-shim-v2/service.go
```diff
@@ -274,7 +298,20 @@ func getTopic(e interface{}) string {
...
+	if ctx == nil {
+		logrus.WithField("type", "bug").Error("trace called before context set")
+		ctx = context.Background()
+	}
+	span, ctx := opentracing.StartSpanFromContext(ctx, name)
+	span.SetTag("subsystem", "runtime")
```
c3d, line 307: The `Trace()` function sets the `source` tag to `runtime`, not the `subsystem` one. What is the reason for the difference?

c3d: The `Trace()` function has an additional `created span` message, which according to the comment, is essential for tracing testing. Thag begs the question: how do you plan to test this newly added tracing?

containerd-shim-v2/service.go
```diff
@@ -330,6 +367,9 @@ func (s *service) Cleanup(ctx context.Context) (_ *taskAPI.DeleteResponse, err e
 
 // Create a new sandbox or container with the underlying OCI runtime
 func (s *service) Create(ctx context.Context, r *taskAPI.CreateTaskRequest) (_ *taskAPI.CreateTaskResponse, err error) {
+	span, _ := trace(s.ctx, "Create")
```
c3d, line 370: Don't you want to have a few `span.LogField()` added giving more information about what is being created?
(This same comment applies to practically all functions below)

Or do you plan to add that later?

egernst: I think at least to start, the basic spans would be good -- it shows the trace of execution very clearly. if feasible, I think we should look at adding in follow on PR?

cmaf: @c3d @egernst we can consider this for a later PR

containerd-shim-v2/service.go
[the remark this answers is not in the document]
cmaf: @c3d virtcontainers has 13 different `trace()` implementations which exist to add different tags to spans (@jodh-intel, correct me if I'm wrong). I think this should be changed so that all use a general `Trace()` function that takes a list of tags, but since it is not essential to shimv2 tracing functionality and applies to many other functions, it should have its own issue & PR.

containerd-shim-v2/service.go
```diff
@@ -274,7 +298,20 @@ func getTopic(e interface{}) string {
 	return cdruntime.TaskUnknownTopic
 }
 
+func trace(ctx context.Context, name string) (opentracing.Span, context.Context) {
+	if ctx == nil {
```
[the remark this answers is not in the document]
cmaf: @c3d you're right, it does not exist there but it exists in the [sandbox trace function](https://github.com/kata-containers/runtime/blob/master/virtcontainers/sandbox.go#L139) and a few other ones in virtcontainers. The CI fails with a SIGSEGV without it so I kept it in.

containerd-shim-v2/service.go
cmaf: @c3d I switched it to `source`, it should be consistent.

Regarding the `created span` message, the test is undergoing significant changes in a PR (https://github.com/kata-containers/tests/pull/2129) and will not check for that string.

cmaf:
@c3d, thanks for the feedback! I think I responded to everything, please let me know if you still have concerns.

containerd-shim-v2/service.go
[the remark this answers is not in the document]
cmaf: See https://github.com/kata-containers/runtime/issues/3093

egernst (approved):

amshinde (approved):

Issue #2664, opened 2020-05-06
[qemu] q35 VFIO passthrough fails on both bridge and pcie-root-port


amorenoz:
I have recently tried VFIO passthrough with and without SR-IOV and detected a number of problems that make it fail. 
Reporting them as a single issue since they all contribute to "VFIO passthrough not working". Let me know if you prefer to split them.

The below tests were performed with the following device:
`65:00.0 Ethernet controller: Intel Corporation Ethernet Controller X710 for 10GbE SFP+ (rev 02)`

# Description of the problem
Generically, trying to add a PF or VF via VFIO passthrough fails
## Problem 1: pci-rescan
The pci-rescan triggered by kata-agent makes shpchp fail. Basically, the same as reported in #2460. 
@devimc pointed me to the origin [origin of the rescan](https://github.com/clearcontainers/agent/pull/139) which seems to be related to lack of ACPI hotplug support in old versions of qemu.

Question: Could we implement a mechanism to tell the agent whether a rescan is needed?
Should we keep the rescan as the default behaviour in that case?

Alternatively, could the mechanism that was implemented in to fix #2460  be extended to support not only devices with large BARs.

Failing dmesg:
```
[    2.826270] shpchp 0000:00:02.0: Latch close on Slot(1)
[    2.826282] shpchp 0000:00:02.0: Button pressed on Slot(1)
[    2.826290] shpchp 0000:00:02.0: Card present on Slot(1)
[    2.827662] shpchp 0000:00:02.0: PCI slot #1 - powering on due to button press
[    2.827970] IPv6: ADDRCONF(NETDEV_CHANGE): eth0: link becomes ready
[    2.834453] pci 0000:01:01.0: [8086:1572] type 00 class 0x020000
[    2.834706] pci 0000:01:01.0: reg 0x10: [mem 0x00000000-0x00ffffff 64bit pref]
[    2.834909] pci 0000:01:01.0: reg 0x1c: [mem 0x00000000-0x00007fff 64bit pref]
[    2.836799] pci 0000:01:01.0: PME# supported from D0 D3hot D3cold
[    2.846095] pci 0000:01:01.0: BAR 0: no space for [mem size 0x01000000 64bit pref]
[    2.846100] pci 0000:01:01.0: BAR 0: failed to assign [mem size 0x01000000 64bit pref]
[    2.846106] pci 0000:01:01.0: BAR 3: assigned [mem 0xfe800000-0xfe807fff 64bit pref]
[...]

PR #2981, opened 2020-09-23
Provide information to agent to let it safely wait for VFIO devices to complete hotplug

dgibson:
This is the runtime part of the fix for #2664 and https://github.com/kata-containers/agent/issues/781.

The agent part is https://github.com/kata-containers/agent/pull/850

### DIFF
--- a/virtcontainers/clh.go
+++ b/virtcontainers/clh.go
@@ -444,15 +444,33 @@ func (clh *cloudHypervisor) hotplugAddBlockDevice(drive *config.BlockDrive) erro
 	return err
 }
 
-func (clh *cloudHypervisor) hotPlugVFIODevice(device config.VFIODev) error {
+func (clh *cloudHypervisor) hotPlugVFIODevice(device *config.VFIODev) error {
 	cl := clh.client()
 	ctx, cancel := context.WithTimeout(context.Background(), clhHotPlugAPITimeout*time.Second)
 	defer cancel()
 
-	_, _, err := cl.VmAddDevicePut(ctx, chclient.VmAddDevice{Path: device.SysfsDev, Id: device.ID})
+	pciInfo, _, err := cl.VmAddDevicePut(ctx, chclient.VmAddDevice{Path: device.SysfsDev, Id: device.ID})
 	if err != nil {
-		err = fmt.Errorf("Failed to hotplug device %+v %s", device, openAPIClientError(err))
+		return fmt.Errorf("Failed to hotplug device %+v %s", device, openAPIClientError(err))
 	}
+
+	// clh doesn't use bridges, so the PCI path is simply the slot
+	// number of the device.  This will break if clh starts using
+	// bridges (including PCI-E root ports), but so will the clh
+	// API, since there's no way it can reliably a guest Bdf when
+	// bridges are present.
+	tokens := strings.Split(pciInfo.Bdf, ":")
+	if len(tokens) != 3 || tokens[0] != "0000" || tokens[1] != "00" {
+		return fmt.Errorf("Unexpected PCI address %q from clh hotplug", pciInfo.Bdf)
+	}
+
+	tokens = strings.Split(tokens[2], ".")
+	if len(tokens) != 2 || tokens[1] != "0" || len(tokens[0]) != 2 {
+		return fmt.Errorf("Unexpected PCI address %q from clh hotplug", pciInfo.Bdf)
+	}
+
+	device.GuestPciPath, err = vcTypes.PciPathFromString(tokens[0])
+
 	return err
 }
 
@@ -466,7 +484,7 @@ func (clh *cloudHypervisor) hotplugAddDevice(devInfo interface{}, devType device
 		return nil, clh.hotplugAddBlockDevice(drive)
 	case vfioDev:
 		device := devInfo.(*config.VFIODev)
-		return nil, clh.hotPlugVFIODevice(*device)
+		return nil, clh.hotPlugVFIODevice(device)
 	default:
 		return nil, fmt.Errorf("cannot hotplug device: unsupported device type '%v'", devType)
 	}
--- a/virtcontainers/device/drivers/vfio.go
+++ b/virtcontainers/device/drivers/vfio.go
@@ -210,7 +210,16 @@ func (device *VFIODevice) Load(ds persistapi.DeviceState) {
 // It should implement GetAttachCount() and DeviceID() as api.Device implementation
 // here it shares function from *GenericDevice so we don't need duplicate codes
 func getVFIODetails(deviceFileName, iommuDevicesPath string) (deviceBDF, deviceSysfsDev string, vfioDeviceType config.VFIODeviceType, err error) {
-	vfioDeviceType = GetVFIODeviceType(deviceFileName)
+	tokens := strings.Split(deviceFileName, ":")
+	vfioDeviceType = config.VFIODeviceErrorType
+	if len(tokens) == 3 {
+		vfioDeviceType = config.VFIODeviceNormalType
+	} else {
+		tokens = strings.Split(deviceFileName, "-")
+		if len(tokens) == 5 {
+			vfioDeviceType = config.VFIODeviceMediatedType
+		}
+	}
 
 	switch vfioDeviceType {
 	case config.VFIODeviceNormalType:
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -77,6 +77,7 @@ var (
 	kataSCSIDevType             = "scsi"
 	kataNvdimmDevType           = "nvdimm"
 	kataVirtioFSDevType         = "virtio-fs"
+	kataVfioVMDevType           = "vfio-vm"
 	sharedDir9pOptions          = []string{"trans=virtio,version=9p2000.L,cache=mmap", "nodev"}
 	sharedDirVirtioFSOptions    = []string{}
 	sharedDirVirtioFSDaxOptions = "dax"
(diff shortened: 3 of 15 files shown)


virtcontainers/qemu.go
```diff
@@ -1275,6 +1275,69 @@ func (q *qemu) hotplugVhostUserDevice(vAttr *config.VhostUserDeviceAttrs, op ope
...
+	}
+
+	return addri >> 3, nil
+}
+
+// Query QMP to find a device's PCI path given its QOM path or ID
+// "PCI path" means the PCI slot number of each bridge leading to the
+// device, and finally the device itself, as a '/' separated list
+func (q *qemu) qomGetPciPath(qemuID string) (string, error) {
+	// XXX: For now we assume there's exactly one bridge, since
```
devimc, line 1304: what about root ports ? can this implementation support more than one root port?

devimc (commented):
thanks @dgibson - I have some questions

virtcontainers/device/drivers/vfio.go
```diff
@@ -209,8 +209,18 @@ func (device *VFIODevice) Load(ds persistapi.DeviceState) {
 
 // It should implement GetAttachCount() and DeviceID() as api.Device implementation
 // here it shares function from *GenericDevice so we don't need duplicate codes
+
```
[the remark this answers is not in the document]
dgibson: Corrected, though, that's actually from your patch :).

virtcontainers/qemu.go
dgibson: For the purposes here, a root port's virtual PCI to PCI bridge is basically the same as an actual PCI to PCI bridge.  Either way we have one bridge (either a root port or PCIE->PCI bridge) and then the device.
It can certainly handle multiple root ports - which root port is encoded into that bridge slot.

What it can't handle is a bridge under a bridge, including a bridge under a root port.  That also includes PCI-E switches, which incorporate 2 stacked virtual P2P bridges within them.  But we don't allow any of those configurations for the time being.

I hope to generalize the handling of PCI paths at some point, but I'm hoping that can wait until Kata2.

virtcontainers/device/drivers/vfio.go
```diff
@@ -209,8 +209,18 @@ func (device *VFIODevice) Load(ds persistapi.DeviceState) {
...
 func getVFIODetails(deviceFileName, iommuDevicesPath string) (deviceBDF, deviceSysfsDev string, vfioDeviceType config.VFIODeviceType, err error) {
-	vfioDeviceType = GetVFIODeviceType(deviceFileName)
+	tokens := strings.Split(deviceFileName, ":")
+	vfioDeviceType = config.VFIODeviceErrorType
+	if len(tokens) == 3 {
+		vfioDeviceType = config.VFIODeviceNormalType
+	} else {
+		tokens = strings.Split(deviceFileName, "-")
+		if len(tokens) == 5 {
```
jodh-intel, line 220: This test isn't exhaustive - what if there aren't 3 or 5 tokens? Error? I think a set of unit tests for this function would be useful:

- https://github.com/kata-containers/tests/blob/master/Unit-Test-Advice.md#table-driven-tests

virtcontainers/kata_agent.go
```diff
@@ -1253,6 +1254,40 @@ func (k *kataAgent) appendVhostUserBlkDevice(dev ContainerDevice, c *Container)
 	return kataDevice
 }
 
+func (k *kataAgent) appendVfioDevice(dev ContainerDevice, c *Container) *grpc.Device {
+	device := c.sandbox.devManager.GetDeviceByID(dev.ID)
```
jodh-intel, line 1258: This function can return `nil`, but you are deferencing the return value below.

virtcontainers/qemu.go
```diff
@@ -1275,6 +1275,69 @@ func (q *qemu) hotplugVhostUserDevice(vAttr *config.VhostUserDeviceAttrs, op ope
...
+	addr, err := q.qmpMonitorCh.qmp.ExecQomGet(q.qmpMonitorCh.ctx, qomPath, "addr")
+	if err != nil {
+		return 0xff, err
+	}
+	addrf, ok := addr.(float64)
+	// XXX going via float makes no real sense, but that's how
+	// JSON works, and we'll get away with it for the small values
+	// we have here
+	if !ok {
+		return 0xff, fmt.Errorf("addr QOM property of %s is %t not a number", qomPath, addr)
```
jodh-intel, line 1289: This looks wrong - do you mean `%T` maybe? I'd also consider using `%q` rather than `%s` here.

virtcontainers/qemu.go
```diff
@@ -1275,6 +1275,69 @@ func (q *qemu) hotplugVhostUserDevice(vAttr *config.VhostUserDeviceAttrs, op ope
...
+	}
+	addri := uint8(addrf)
+
+	if (addri & 0x7) != 0 {
+		return 0xff, fmt.Errorf("Unexpected non-zero PCI function on %s", qomPath)
```
jodh-intel, line 1294: Might be worth logging what you're testing here (potentially with `%b`? :smile:)

jodh-intel (commented):
Thanks @dgibson - a few comments.

virtcontainers/device/drivers/vfio.go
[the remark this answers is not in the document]
devimc: touché :smile:

virtcontainers/qemu.go
dgibson: I did indeed mean `%T`, corrected.  `%q` seems reasonable too, changed.

virtcontainers/qemu.go
dgibson: I think %b would be not be helpful, but I've updated this with some more information.

virtcontainers/kata_agent.go
dgibson: That also appears to be the case for all the existing `append*Device` functions.

virtcontainers/device/drivers/vfio.go
dgibson: This isn't new logic, it's just been moved here from `GetVFIODeviceType`, so I don't think fixing it is in scope for this patch.

That said, I think it's actually ok.  If neither of the conditions hit, then vfioDeviceType remains as config.VFIODeviceErrorType which gets reported a little below.

jodh-intel: Ack - thanks for checking.

virtcontainers/kata_agent.go
jodh-intel: Ok. Not strictly required for this PR, but it would be great if you could add a `nil` check for all the handlers on a separate commit.

jodh-intel (approved):
Thanks @dgibson.

lgtm

virtcontainers/qemu.go
```diff
@@ -1275,6 +1275,71 @@ func (q *qemu) hotplugVhostUserDevice(vAttr *config.VhostUserDeviceAttrs, op ope
...
+		return 0xff, fmt.Errorf("Unexpected non-zero PCI function (%02x.%1x) on %q",
+			slotNum, funcNum, qomPath)
+	}
+
+	return slotNum, nil
+}
+
+// Query QMP to find a device's PCI path given its QOM path or ID
+// "PCI path" means the PCI slot number of each bridge leading to the
+// device, and finally the device itself, as a '/' separated list
```
likebreath, line 1304: @dgibson Looks like this is the missing bit from `clh.go` ( and maybe cloud-hypervisor) that prevents your PR from working with cloud-hypervisor.  

Excuse my limited knowledge in the area. Can we derive the `PCI path` from the information already provided by cloud-hypervisor (predicted guest `BDF` for the hotplugged VFIO device and CLH only has one PCI bus)? Or you think we will need more information from cloud-hypervisor?

/cc @sboeuf Please feel free to share your thoughts. Thank you.

virtcontainers/clh.go
```diff
@@ -443,15 +443,16 @@ func (clh *cloudHypervisor) hotplugAddBlockDevice(drive *config.BlockDrive) erro
...
 	cl := clh.client()
 	ctx, cancel := context.WithTimeout(context.Background(), clhHotPlugAPITimeout*time.Second)
 	defer cancel()
 
-	_, _, err := cl.VmAddDevicePut(ctx, chclient.VmAddDevice{Path: device.SysfsDev, Id: device.ID})
+	pdi, _, err := cl.VmAddDevicePut(ctx, chclient.VmAddDevice{Path: device.SysfsDev, Id: device.ID})
 	if err != nil {
 		err = fmt.Errorf("Failed to hotplug device %+v %s", device, openAPIClientError(err))
 	}
+	device.GuestPciPath = "###" + pdi.Bdf
```
likebreath, line 455: I don't follow how it (the predicted guest BDF) would fit the definition of `PCI path` you described earlier. May be I am missing some context.

virtcontainers/qemu.go
dgibson: Yes, this is the spot.  So, in general we cannot derive a PCI path from a BDF, but in this case we can.  The only reason clh is able to get away with using a bdf here is because it always places the device on the root bus, so domain and bus are always zero.  So, the PCI path is simply the slot number from the BDF.

If clh ever starts using bridges, we'll have to change this... but in that case clh will have to change its interface anyway, since it won't be able to reliably report a BDF in that case anyway.

virtcontainers/clh.go
dgibson: Oops, that's some debugging code I left in, while forgetting to change this PR to a draft.

virtcontainers/qemu.go
likebreath: @dgibson Thanks for the explanation as always. Let's derive the `PCI path` from the BDF returned by clh for now. We will make it more generic when cloud-hypervisor changes its interface later. 

Can you please give it a try by updating the `clh.go` as you did for the debugging? Please also include a note on the assumptions we have here on cloud-hypervisor. Let's see whether that fixes the VFIO CI failure. Any other thoughts, please let me know.

dgibson: Yes, I plan to do that after rebasing on the PCI path generalization patches.

dgibson:
Bother.  I've fixed the failure with clh, but now there's another failure in the VFIO check which I don't have any theory to explain.

devimc:
@dgibson this is the error:

```
rpc error: code = DeadlineExceeded desc = Timeout reached after 10s waiting for device 0000:00:02.0/0000:01:01.0
```

I have restarted the job and fixed the teardown process, The runtime logs will be available here http://jenkins.katacontainers.io/job/kata-containers-runtime-vfio-PR/261/ once the job finish

devimc:
@dgibson this is the log http://jenkins.katacontainers.io/job/kata-containers-runtime-vfio-PR/261/artifact/artifacts/kata-runtime_00.gz - I hope it can help you to debug this issue

dgibson:
> @dgibson this is the error:
> 
> ```
> rpc error: code = DeadlineExceeded desc = Timeout reached after 10s waiting for device 0000:00:02.0/0000:01:01.0
> ```

Yeah, I got that far.  I don't know why that error is occuring - it doesn't happen when I try it by hand, it's not obvious what's different about the CI environment, and it's far from easy to replicate exactly what the CI is doing.

jodh-intel:
@dgibson - there are references to that device in http://jenkins.katacontainers.io/job/kata-containers-runtime-vfio-PR/261/artifact/artifacts/kernel_00.gz if that helps?

dgibson:
I'm no longer planning to pursue this in Kata1, I'll be following up in Kata 2 instead.

virtcontainers/kata_agent.go
dgibson: It turns out the caller has already nil checked this value, though it's not very obvious here.  I'm making use of that for my 2.x version of this series.

PR #2988, opened 2020-09-24
hypervisor: don't enforce a minimum memory setting

egernst:
Currently, we enforce a lower limit of 256MB for the defaultMemorySize.

In general case with QEMU, this isn't a major problem, since we'll just
eat the extra page table overhead, and only consume pages when needed.
The memcgroup in the guest should make sure it only utilizes requested
amount, not what is actually available.

However, this becomes very problematic when you use preallocated memory.
In the k8s case, the VMM will get OOM killed very quickly since the host's
memory cgroup (created by kubelet) will limit the entire sandbox to the
requests + pod overhead (on the order of 140 MB). We should allow the administrator
of kata to set a better default value, which should be much closer aligned
with what's used for PodOverhead (in the kube case).

Let's remove the artifical limit in kata, and leave it up to the end
user to pick an appropriate non-default value, if desired.

### DIFF
--- a/virtcontainers/hypervisor.go
+++ b/virtcontainers/hypervisor.go
@@ -75,9 +75,6 @@ const (
 	// Port where the agent will send the logs. Logs are sent through the vsock in cases
 	// where the hypervisor has no console.sock, i.e firecracker
 	vSockLogsPort = 1025
-
-	// MinHypervisorMemory is the minimum memory required for a VM.
-	MinHypervisorMemory = 256
 )
 
 // In some architectures the maximum number of vCPUs depends on the number of physical cores.
--- a/pkg/katautils/config.go
+++ b/pkg/katautils/config.go
@@ -326,11 +326,11 @@ func (h hypervisor) defaultMaxVCPUs() uint32 {
 }
 
 func (h hypervisor) defaultMemSz() uint32 {
-	if h.MemorySize < vc.MinHypervisorMemory {
-		return defaultMemSize // MiB
+	if h.MemorySize > 0 {
+		return h.MemorySize
 	}
 
-	return h.MemorySize
+	return defaultMemSize
 }
 
 func (h hypervisor) defaultMemSlots() uint32 {
--- a/virtcontainers/pkg/oci/utils.go
+++ b/virtcontainers/pkg/oci/utils.go
@@ -477,11 +477,7 @@ func addHypervisorMemoryOverrides(ocispec specs.Spec, sbConfig *vc.SandboxConfig
 	if value, ok := ocispec.Annotations[vcAnnotations.DefaultMemory]; ok {
 		memorySz, err := strconv.ParseUint(value, 10, 32)
 		if err != nil {
-			return fmt.Errorf("Error encountered parsing annotation for default_memory: %v, please specify positive numeric value greater than 8", err)
-		}
-
-		if memorySz < vc.MinHypervisorMemory {
-			return fmt.Errorf("Memory specified in annotation %s is less than minimum required %d, please specify a larger value", vcAnnotations.DefaultMemory, vc.MinHypervisorMemory)
+			return fmt.Errorf("Error encountered parsing annotation for default_memory: %v", err)
 		}
 
 		sbConfig.HypervisorConfig.MemorySize = uint32(memorySz)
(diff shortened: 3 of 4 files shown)


virtcontainers/hypervisor.go
```diff
@@ -77,7 +77,7 @@ const (
 	vSockLogsPort = 1025
 
 	// MinHypervisorMemory is the minimum memory required for a VM.
-	MinHypervisorMemory = 256
+	MinHypervisorMemory = 8
```
bergwolf, line 80: Maybe we should make it per-hypervisor default?

egernst: I’m not sure if we should even have a minimum requirement, to be honest. 8 seemed reasonably small for each of them.

crosbymichael: I think you could remove the min.  Finding min values are hard and can always be tricky. You can just let the underlying system fail and propagate the error up to the user and not be in the game of trying to find a min value that works for everyone.

egernst: Let me checkout the error that'll end up propagating. Agreed that a minimum doesn't buy us much;  the admin should be able to 'do the right thing'

amshinde: @egernst I am in agreement with @crosbymichael here. Lets remove the min memory here in light of Pod overhead feature now.

crosbymichael:
@c3d ya, I agree, last I checked it’s around 15mb.  256mb is way large if we are considering this as the reasoning for the minimum value.  

Overall, I think the real question is, should there be an hardcoded check or should you just rely on the underlying system to fail and propagate the error to the user?

likebreath (approved):
Thanks for the PR. One more thing learnt on k8s w/ kata. :)

jcvenegas (approved):

jodh-intel (approved):
Thanks @egernst.

jodh-intel:
@egernst - this still needs porting to 2.x I think?

c3d:
> @c3d 
> Overall, I think the real question is, should there be an hardcoded check or should you just rely on the underlying system to fail and propagate the error to the user?

The problem is that the failure mode is all but obvious, and not easily reported to the runtime. If I manually boot a VM with qemu and 8M of memory, here is what I get on the console:

```
error: ../../grub-core/loader/i386/pc/linux.c:152:cannot allocate kernel buffer
errror: ../../grub-core/loader/i386/pc/linux.c:416:you need to load the kernel first
```

If I try a direct kernel boot from qemu the error is even worse:

```
# qemu-kvm -m 8M -kernel /boot/vmlinuz 
KVM internal error. Suberror: 1
emulation failure
EAX=00a05240 EBX=00000000 ECX=00000000 EDX=00000000
ESI=00013e50 EDI=00000000 EBP=00100000 ESP=00a0523c
EIP=009e27c0 EFL=00010002 [-------] CPL=0 II=0 A20=1 SMM=0 HLT=0
ES =0018 00000000 ffffffff 00c09300 DPL=0 DS   [-WA]
CS =0010 00000000 ffffffff 00c09b00 DPL=0 CS32 [-RA]
SS =0018 00000000 ffffffff 00c09300 DPL=0 DS   [-WA]
DS =0018 00000000 ffffffff 00c09300 DPL=0 DS   [-WA]
FS =0018 00000000 ffffffff 00c09300 DPL=0 DS   [-WA]
GS =0018 00000000 ffffffff 00c09300 DPL=0 DS   [-WA]
LDT=0000 00000000 ffffffff 00c00000
TR =0020 00001000 00000067 00008b00 DPL=0 TSS32-busy
GDT=     00013830 00000027
IDT=     00000000 00000000
CR0=00000011 CR2=00000000 CR3=00000000 CR4=00000000
DR0=0000000000000000 DR1=0000000000000000 DR2=0000000000000000 DR3=0000000000000000 
DR6=00000000ffff0ff0 DR7=0000000000000400
EFER=0000000000000000
Code=00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 <00> 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
```

Even more annoying, `qemu` does not exit in that case, it remains stuck there.

So I would definitely recommend we "own" that at least for the time being, and set a minimum value that we know at least loads the kernel.

PR #2991, opened 2020-09-28
shimv2: handle ctx passed by containerd

YvesChan:
Sometimes shim process cannot be shutdown because of container list
is not empty. This container list is written in shim service, while
creating container. We find that if containerd cancel its Create
Container Request due to timeout, but runtime didn't handle it properly
and continue creating action, then this container cannot be deleted at
all. So we should make sure the ctx passed to Create Service rpc call
is effective.

Fixes: #2992

### DIFF
--- a/containerd-shim-v2/service.go
+++ b/containerd-shim-v2/service.go
@@ -337,34 +337,46 @@ func (s *service) Create(ctx context.Context, r *taskAPI.CreateTaskRequest) (_ *
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	var c *container
-
-	c, err = create(ctx, s, r)
-	if err != nil {
-		return nil, err
+	type Result struct {
+		container *container
+		err       error
 	}
+	ch := make(chan Result, 1)
+	go func() {
+		container, err := create(ctx, s, r)
+		ch <- Result{container, err}
+	}()
 
-	c.status = task.StatusCreated
+	select {
+	case <-ctx.Done():
+		return nil, errors.Errorf("create container timeout: %v", r.ID)
+	case res := <-ch:
+		if res.err != nil {
+			return nil, res.err
+		}
+		container := res.container
+		container.status = task.StatusCreated
 
-	s.containers[r.ID] = c
+		s.containers[r.ID] = container
 
-	s.send(&eventstypes.TaskCreate{
-		ContainerID: r.ID,
-		Bundle:      r.Bundle,
-		Rootfs:      r.Rootfs,
-		IO: &eventstypes.TaskIO{
-			Stdin:    r.Stdin,
-			Stdout:   r.Stdout,
-			Stderr:   r.Stderr,
-			Terminal: r.Terminal,
-		},
-		Checkpoint: r.Checkpoint,
-		Pid:        s.pid,
-	})
+		s.send(&eventstypes.TaskCreate{
+			ContainerID: r.ID,
+			Bundle:      r.Bundle,
+			Rootfs:      r.Rootfs,
+			IO: &eventstypes.TaskIO{
+				Stdin:    r.Stdin,
+				Stdout:   r.Stdout,
+				Stderr:   r.Stderr,
+				Terminal: r.Terminal,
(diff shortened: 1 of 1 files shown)


devimc:
cc @lifupan @liubin

fidencio:
How does it affect other CRIs?

c3d:
@YvesChan do you think you could create a test case that specifically tests that scenario?

YvesChan:
> How does it affect other CRIs?

Since I'm using containerd in my environment, I can only confirm this patch works for containerd right now. But I think this patch will only enhance the robustness of runtime/shimv2 but not do any harm?

YvesChan:
> @YvesChan do you think you could create a test case that specifically tests that scenario?

I have reproduced this issue in my environment but it's a little bit complicated to configure. Let me try to simplify it first.

lifupan (approved):

lifupan:
> How does it affect other CRIs?

This fix will work for any other CRIs which compatible with shimv2  api.

jodh-intel:
Thanks @YvesChan. Please review the CI failures though - shimv2 tests are failing with SEGV...

containerd-shim-v2/service.go
```diff
@@ -337,34 +337,46 @@ func (s *service) Create(ctx context.Context, r *taskAPI.CreateTaskRequest) (_ *
...
-	c.status = task.StatusCreated
-
-	s.containers[r.ID] = c
+	type Result struct {
+		container *container
+		err       error
+	}
+	ch := make(chan Result, 1)
+	go func() {
+		container, err := create(ctx, s, r)
```
crosbymichael, line 346: I think one thing that you need to be careful of is to make sure that calls within this `create` func handles context.Cancel() correctly.  Running this in a goroutine is fine but you run the risks of this hanging forever on a cancel if the calls within this func do not also handle a canceled context.

jodh-intel:
@YvesChan - there are a lot of failing CI's - have you looked at the logs to see why?

YvesChan:
> @YvesChan - there are a lot of failing CI's - have you looked at the logs to see why?

Working on it. It seems the TestCreateContainerFail did not pass correctly.

containerd-shim-v2/service.go
YvesChan: I suppose the original code doesn't handle cancel ctx as well? The fact is this `Create` service does NOT handle any ctx at all... :(
So do you think this goroutinue will make it more risky?

jodh-intel:
@crosbymichael - could you respond to @YvesChan's comment so we can progress this? It's unclear if your comment are an ack or a nack?

containerd-shim-v2/service.go
crosbymichael: I guess since this is a create, a hanging go routine is fine as the shim will be shutdown following the error.

I think working on the create to fully handle the context can come in a separate PR to make this even more robust :)

YvesChan: ping @crosbymichael
Anything left for me to do? :)

fidencio:
I've opened an issue for the failure on jenkins-ci-podman-clh-fedora-31: https://github.com/kata-containers/tests/issues/2946
It's not related to this series.

jodh-intel (approved):

containerd-shim-v2/service.go
crosbymichael: nope, all good

jodh-intel:
Hi @YvesChan - It looks like this hasn't been ported to 2.x (https://github.com/kata-containers/kata-containers/tree/2.0-dev/src/runtime) yet. Would you be able to look at doing that?

YvesChan:
> Hi @YvesChan - It looks like this hasn't been ported to 2.x (https://github.com/kata-containers/kata-containers/tree/2.0-dev/src/runtime) yet. Would you be able to look at doing that?

OK, I'll do it later

cmaf:
forward port pr: https://github.com/kata-containers/kata-containers/pull/1166

PR #2998, opened 2020-10-05
virtiofs: Enable xattr by default

jcvenegas:
xattr is used by programs like dnf or apt.

Be cause this can be used by containers.
To build and image or by interactive containers
it may be useful to have it by default.

Note that this is not part of the virtiofsd driver
as some users still want to remove this use case.

Fixes: github.com/kata-containers/tests#2008

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,7 +183,7 @@ DEFVIRTIOFSCACHE ?= auto
 #
 # see `virtiofsd -h` for possible options.
 # Make sure you quote args.
-DEFVIRTIOFSEXTRAARGS ?= []
+DEFVIRTIOFSEXTRAARGS ?= [\"-o\", \"xattr\"]
 DEFENABLEIOTHREADS := false
 DEFENABLEMEMPREALLOC := false
 DEFENABLEHUGEPAGES := false


fidencio:
To me, it does make sense, otherwise we'll be set to be face failures on our CI & users may face failures on their workload for a non good specific reason.

It does LGTM, but I also would like to her from @dagrh.

jodh-intel (approved):

c3d:
I am not very comfortable with this. I will let @dagrh confirm this, but my understanding is that this would allow a container to set things like a setuid bit on the host. In my opinion, this largely defeats a key mantra of kata containers, "the security of VMs".

I understand that some containers need that, but for those containers that need it (and that you trust), you can set this option e.g. using annotations. Maybe we can provide slightly better integration than that.

One thing I could see as a default is to enable that for container-specific portions of the file system. In other words, while I'm very uncomfortable having `-o xattr` by default for a container that has a `/usr:/usr` mount (or worse `/sbin:/sbin`), I could very well understand that this would be OK as long as the in-guest `/usr` is private to that container. Could we add some logic along those lines directly in the runtime?

c3d (changes requested):
I'm not comfortable having that for a container that could possible have a `/usr:/usr` mapping and change the setuid bit of some stuff on the host.

dagrh:
> I'm not comfortable having that for a container that could possible have a `/usr:/usr` mapping and change the setuid bit of some stuff on the host.

Yeh, not quite setuid, but similar by setting a capability.  I've got a series that allows the xattr names to be changed so that you can (optionally) prefix xattr names e.g. make all attrs that the guest interacts with be user.virtiofs.something    - the upside to that is that the guest can't change or see capabiltiies/security attributes on the host and wont' clash with trusted. attributes used by overlayfs, so you should be able to stack them.  The downside is that you can't change or see capabilities/security attributes on the host, so if you wanted to unpack a fs on the host and see the capabilities in the guest you can't - which is why my series lets you chose what you want.

jcvenegas:
@c3d  thanks for rice the concern about it :)
@dagrh so to keep safe we should wait for your series. Do you have some patchset somewhere we can take for early integration? or you suggest wait until are in upstream?

jcvenegas:
@c3d 
> I understand that some containers need that, but for those containers that need it (and that you trust), you can set this option e.g. using annotations. Maybe we can provide slightly better integration than that.
An annotation sounds a more longer term solution as I think we need to promote one annotation to be pass  by the caller in kata, I or is there a semantic already in docker, podman, container, k8s? for this use case.  


>
> One thing I could see as a default is to enable that for container-specific portions of the file system. In other words, while I'm very uncomfortable having `-o xattr` by default for a container that has a `/usr:/usr` mount (or worse `/sbin:/sbin`), I could very well understand that this would be OK as long as the in-guest `/usr` is private to that container. Could we add some logic along those lines directly in the runtime?

Let me see If I get the idea:

### Case 1:
user shares its  host files 
e.g
`docker run -ti -v /usr:/usr busybox`
Disable xttars
### Case 2:
User not share any path to the container `/usr` `/sbin` or probably others we have in a filter list.
e.g
`docker run -ti busybox`
enable xattrs as container rootfs is private

So we only should care about `Case 1` correct ?

There are additional thoughts I have.
If we xattrs are limited  ....  Is it just for that path like `/sbin` or to the complete guest?
If  In the case of the host user bind mount `/sbin` -> `/tmp/sbin` and the the container is started as `/tmp/sbin:/sbin` is this a case that should be handled?
Lets take the pods use cases, when we initially create the VM  we may not have check the container info so is there a way for virtiofsd to disable dynamically xattrs, similar to this is when a new container is created in an exiting pod.

jcvenegas:
Based in previous comments I feel that if we want a safe and flexible solution, use the series that @dagrh have is the way to go.

fidencio:
@jcvenegas, I'm thinking here, in order to unblock CI, would be possible to **always** run Fedora containers from our tests with "-o xattr" passed to virtiofsd?

By doing this we don't set it as a default, but we ensure the faulty behaviour doesn't keep breaking our tests.

c3d:
@jcvenegas You correctly identified what I meant, and the problem with an "easy" filtering based on naive path checking.

dagrh:
> @c3d thanks for the concern about it :)
> @dagrh so to keep safe we should wait for your series. Do you have some patchset somewhere we can take for early integration? or you suggest wait until are in upstream?

https://lists.gnu.org/archive/html/qemu-devel/2020-08/msg07194.html  is the current version, but there are a bunch of review comments I intend to attack in the next few days.

jodh-intel:
Any update on this folks?

fidencio:
Folks, this PR has been stale and I'm closing it as we're walking towards the **final** 1.13.0 release of the kata-containers-1.x.
From now on, all the development will happen as part of https://github.com/kata-containers/kata-containers/ and fixes should be provided to that repo.

Thanks a lot for your contribution, we're sorry that for some reason it got stale, and we hope to see you around for new adventures in the kata-containers-2.x land.

Issue #3002, opened 2020-10-08
PCI paths handling is confusing in the code


dgibson:
This is a code clarity and extensibility issue, rather than a problem in behaviour.

# Description of problem

A number of places in the runtime code use values described as a "PCI address" but which are not actuall in PCI address format (DDDD:BB:DD.F).  Rather they are "PCI paths" - that is a description of the location of a device based on its slot number and the slot number of all the bridges traversed to reach it.

For now, the PCI paths we generate always have exactly two elements (one bridge + the device itself), but we will want to use other cases in future.  https://github.com/kata-containers/agent/issues/854 is tracking making that generalization on the agent side.

# Expected result

PCI paths are handled and named consistently distinct from PCI addresses in the normal format.

# Actual result

"PCI address" in the code frequently refers to (a limited form of) PCI paths.

# Further information

See also https://github.com/kata-containers/agent/issues/854 and https://github.com/kata-containers/agent/pull/855

PR #3003, opened 2020-10-08
Handle PCI paths more consistently

dgibson:
This addresses issue #3002


virtcontainers/qemu.go
```diff
@@ -1152,8 +1153,7 @@ func (q *qemu) hotplugAddBlockDevice(drive *config.BlockDrive, op operation, dev
 			}
 		}()
 
-		// PCI address is in the format bridge-addr/device-addr eg. "03/02"
-		drive.PCIAddr = fmt.Sprintf("%02x", bridge.Addr) + "/" + addr
+		drive.PCIPath = vcTypes.PciPath{fmt.Sprintf("%02x", bridge.Addr) + "/" + addr}
```
jodh-intel, line 1156: How about creating a `New()` function for `PciPath`? That would avoid the duplication here. Yes, it's trivial, but it's still duplication and thus could go stale if someone forgets to update one of the multiple calls at some point in the future.

virtcontainers/pkg/types/types.go
```diff
@@ -5,6 +5,18 @@
...
 
+// PciPath describes a PCI path which identifies a PCI device by where
+// it sits in the PCI topology.
+//
+// Has the format "xx/.../yy/zz" Here, zz is the slot of the device on
+// its PCI bridge, yy is the slot of the bridge on its parent bridge
+// and so forth until xx is the slot of the "most upstream" bridge on
+// the root bus.  If a device is connected directly to the root bus,
+// its PciPath is just "zz"
+type PciPath struct {
```
jodh-intel, line 16: How about creating a `PciPath.Valid() bool` function which would handle a check on `""`, and weird values like `/`, `//`, `xx///zz`, *etc*.

jodh-intel (commented):
Thanks @dgibson. This looks much better. A couple of suggestions inline. Also, any way to add some further unit tests for this change?

virtcontainers/pkg/types/types.go
```diff
@@ -5,6 +5,18 @@
...
 package types
 
+// PciPath describes a PCI path which identifies a PCI device by where
+// it sits in the PCI topology.
+//
+// Has the format "xx/.../yy/zz" Here, zz is the slot of the device on
+// its PCI bridge, yy is the slot of the bridge on its parent bridge
+// and so forth until xx is the slot of the "most upstream" bridge on
+// the root bus.  If a device is connected directly to the root bus,
+// its PciPath is just "zz"
```
likebreath, line 15: Looks like, we are going to simply put the `DD` part of the device as the `PciPath` for cloud-hypervisor, right?

virtcontainers/pkg/types/types.go
```diff
@@ -5,6 +5,18 @@
```
likebreath, line 5: Nit: typo in the commit message `descrptions`. Don't forget to run type checker over commit message. :)

likebreath (approved):
@dgibson LGTM. Thanks for the diligent refactoring that clarifies the confusions of `PCIAddr` vs `PCI path`. As the agent side PR is landed, you should be able to rework the first commit and make the TravisCI happy.

virtcontainers/pkg/types/types.go
dgibson: That's right.

virtcontainers/pkg/types/types.go
dgibson: Fixed, thanks.

amshinde:
Havent been able to spend time reviewing PR's related to device handling lately, but will like to take a closer look at the changes introduced here.
Will schedule some time tomorrow for a review.

dgibson:
> Havent been able to spend time reviewing PR's related to device handling lately, but will like to take a closer look at the changes introduced here.
> Will schedule some time tomorrow for a review.

@amshinde maybe hold off for a bit, I'm looking at doing some reworks based on @jodh-intel's comments next week.

dgibson:
Reworking this completely, it now makes much more thorough use of typing.

virtcontainers/pkg/types/pcipath.go
```diff
@@ -0,0 +1,73 @@
...
+		slot, err := strconv.ParseUint(t, 16, 5)
+		if err != nil {
+			return PciPath{}, err
+		}
+		slots[i] = uint8(slot)
+	}
+	return PciPath{path: slots}, nil
+}
+
+func MakePciPath(slots ...uint8) (PciPath, error) {
```
jodh-intel, line 61: Why is this variadic? It seems we only ever pass two values...?

virtcontainers/pkg/types/pcipath.go
```diff
@@ -0,0 +1,73 @@
...
+// device, followed by the slot of the device itself
+//
+// When formatted into a string is written as "xx/.../yy/zz" Here, zz
+// is the slot of the device on its PCI bridge, yy is the slot of the
+// bridge on its parent bridge and so forth until xx is the slot of
+// the "most upstream" bridge on the root bus.  If a device is
+// connected directly to the root bus, its PciPath is just "zz"
+
+type PciPath struct {
+	path []uint8
```
jodh-intel, line 29: Maybe `slots` rather than `path`?

virtcontainers/pkg/types/pcipath.go
```diff
@@ -0,0 +1,73 @@
...
+	if s == "" {
+		return PciPath{}, nil
+	}
+
+	tokens := strings.Split(s, "/")
+	slots := make([]uint8, len(tokens))
+	for i, t := range tokens {
+		slot, err := strconv.ParseUint(t, 16, 5)
+		if err != nil {
+			return PciPath{}, err
```
jodh-intel, line 54: I'd be tempted to mark this as a bug since if the PCI path was already validated in `MakePciPath()`, this should be "impossible" to fail:

```go
return PciPath{}, fmt.Errorf("BUG: %q resulted in invalid PCI slot %v", s, err)
```

virtcontainers/pkg/types/pcipath.go
```diff
@@ -0,0 +1,73 @@
...
+
+func PciPathFromString(s string) (PciPath, error) {
+	if s == "" {
+		return PciPath{}, nil
+	}
+
+	tokens := strings.Split(s, "/")
+	slots := make([]uint8, len(tokens))
+	for i, t := range tokens {
+		slot, err := strconv.ParseUint(t, 16, 5)
```
jodh-intel, line 52: This seems trivial but there are three `strconv.ParseUint()` calls. I wonder if we should create a `strToSlot(s string) (uint8, error)` function to encapsulate this? As a bonus that function to also check the value against `maxPciSlot`.

jodh-intel:
:keyboard: :smile:

jodh-intel (commented):
Thanks @dgibson. A few comments.

virtcontainers/pkg/types/pcipath.go
dgibson: For now, but I have future plans that will change that.  With the agent side changes, it will now accept and work with paths of an arbitrary number of elements.

virtcontainers/pkg/types/pcipath.go
dgibson: Fair enough, changed.

virtcontainers/pkg/types/pcipath.go
dgibson: The input to this is a plain string though, it need not have come from MakePciPath().  So it's not necessarily a bug... or at least, not necessarily a bug in this component.

virtcontainers/pkg/types/pcipath.go
jodh-intel: Hmm. Conventionally PRs shouldn't do this as you are spreading a feature across multiple PRs which, is cause for potential confusion. Why not just make the function accept 2 params and then the follow-on PR (when written) can change this function to be a variadic since that change is part of the future work?

dgibson: Hmm.  (a) that seems unnecessarily bureaucratic.  (b) I don't think the division you propose is actually conceptually sounder than this one.

What this series is doing - and the first patch in particular - is introducing a data structure which will handle general pci paths to replace the ad-hoc and limited form of pci paths we deal with now.  At present we only use one special case of pci paths, but it doesn't seem sensible to artificially limit the data structure just because we don't need the general variant yet.

virtcontainers/pkg/types/pcipath.go
dgibson: I can't really see a sensible way to do this, because 2 of the `ParseUint` calls are outside the code specifically for the new data structure.  `MakePciPath()` still needs to validate against `maxPciSlot`, so checking in a helper before passing into there would be redundant.

This is kind of unavoidably ugly, because those callers have one of their path components as a string, and another as a number.  I'd like to change that, but that involves refactoring of a bunch of other things that's not really in scope for this PR.

likebreath (approved):
Thanks @dgibson. PR looks good to me. I will send follow-up PRs to remove special code path on handling virtio-blk device w/ cloud-hypervisor.

likebreath:
Note: The failure on `podman-clh` CI is a known issue, and is irrelevant to the PR itself.

virtcontainers/pkg/types/pcipath.go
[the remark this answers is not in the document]
dgibson: Exactly.  PCI slot numbers are 5 bits long, 0..31.

virtcontainers/pkg/types/pcipath.go
```diff
@@ -0,0 +1,73 @@
...
+
+package types
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+const maxPciSlot = 31
```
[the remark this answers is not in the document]
dgibson: That's not chosen by us, that's from the PCI specification.  Slot numbers are 5 bits, and that appears in the "over the wire" protocol, so it's essentially fixed forever.

dgibson:
@jodh-intel I've reorganized how I structured the datatype in a way I think should address some of your comments.  As a bonus it should be easier to extend for multifunction devices in future.

virtcontainers/pkg/types/pcipath.go
dgibson: I think I found a way which is done in the new push.

virtcontainers/pkg/types/pcipath.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+const (
+	pciSlotBits = 5
+	maxPciSlot = (1 << pciSlotBits) - 1
```
jodh-intel, line 16: Nice! :smile:

virtcontainers/pkg/types/pcipath.go
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) 2020 David Gibson, Red Hat Inc.
```
jodh-intel, line 1: Just wondering if this needs your name here (and in the tests file)? All the other copyright headers just specify the date and name for contributors working for a company.

jodh-intel (approved):
Thanks @dgibson - just a query on the copyright but other than that...

lgtm

virtcontainers/pkg/types/pcipath.go
dgibson: I don't think it's necessary, but it is the convention I've been used to.

But.. I checked our internal guidelines for this, and it looks like I'm a bit out of date, so I've updated accordingly.

dgibson: Well.. mostly.  Our guidelines recommend against including "(c)" and year info, but checkpatch objects to their absence.

jodh-intel: Thanks @dgibson.

virtcontainers/pkg/types/pcipath.go
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) 2020 Red Hat
```
devimc, line 1: now the header is different

`Copyright (c) 2020 David Gibson, Red Hat Inc.`

dgibson: Uh... I'm confused, the one @devimc quotes should be the old version, not the new one.

dgibson:
Is the `jenkins-metrics-ubuntu-18-04` failure a known problem?  I can't see why it's failing from the logs.

virtcontainers/pkg/types/pcipath_test.go
```diff
@@ -0,0 +1,110 @@
+// Copyright (c) 2020 David Gibson, Red Hat Inc.
```
jodh-intel, line 1: @dgibson - I think @devimc was referring to this one :smile:

devimc: @dgibson yes, both headers are different...which one is the correct?

devimc:
restarting metrics-ci, there is a bug that is messing up docker

virtcontainers/pkg/types/pcipath_test.go
dgibson: Bother.  I forgot to update that one.  Done now.

devimc (approved):
thanks @dgibson - lgtm

virtcontainers/kata_agent.go
```diff
@@ -673,7 +673,10 @@ func (k *kataAgent) listInterfaces() ([]*vcTypes.Interface, error) {
 	}
 	resultInterfaces, ok := resultingInterfaces.(*grpc.Interfaces)
 	if ok {
-		return k.convertToInterfaces(resultInterfaces.Interfaces), err
+		ifaces, err := k.convertToInterfaces(resultInterfaces.Interfaces)
+		if err == nil {
+			return ifaces, nil
+		}
 	}
 	return nil, err
```
amshinde, line 681: @dgibson Didnt get a chance to get to this PR earlier.
If the intention was to capture the error from `convertToInterfaces`, I am afraid you are not quite doing that here, since the err on line 677  has scope only within the if block and not outside it. So line 681 will always return nil for err.
You could just change the previous return statement to this `return k.convertToInterfaces(resultInterfaces.Interfaces)` for things to work correctly.

dgibson: Drat, good point.  The interaction of `:=` with variable shadowing is a real Go gotcha :/.

dgibson: Filed #3032 and #3033 to address this.

jodh-intel:
Hi @dgibson - It looks like this hasn't been ported to 2.x (https://github.com/kata-containers/kata-containers/tree/2.0-dev/src/runtime) yet. Would you be able to look at doing that?

Issue #3030, opened 2020-10-27
Pod annotations for selecting a custom hypervisor binary not working


gkunz:
# Problem
We have a use case in which we'd like to select custom components of the the Kata sandbox (kernel, image, hypervisor) using pod annotations. Following the corresponding documentation [1], setting the kernel and image via pod annotations works fine, but the custom hypervisor binary gets ignored and instead the default hypervisor as configured in the config file is used.

Example workload definition:
```
  apiVersion: v1
  kind: Pod
  metadata:
    name: ubuntu-untrusted-custom-kernel
    annotations:
      # works
      io.katacontainers.config.hypervisor.kernel: "/tmp/kata_custom_sandbox/kernel/vmlinuz.container"
      # works
      io.katacontainers.config.hypervisor.image: "/tmp/kata_custom_sandbox/image/kata-containers.img"
      # gets ignored
      io.katacontainers.config.hypervisor.path: /tmp/kata_custom_sandbox/hypervisor/qemu-virtiofs-system-x86_64
  spec:
    runtimeClassName: kata-qemu
    containers:
      - name: ubuntu
        image: ubuntu
        command: ['bash', '-c', 'sleep infinity']
```

I am running Kata 1.12-alpha as deployed (some time ago) by kata-deploy.

# Expected result

The sandbox should be run by the custom hypervisor specified by the pod annotation.

# Actual result

The sandbox is being run by the default hypervisor according to the runtime class definition.

# Further information

Looking at the code, it seems that the custom "hypervisor asset" is not considered here [2] and here [3] and so the pod annotations are not passed down to sandbox. Those links point to the Kata 1.x repos, but it seems the same situation is true in the 2.0 repo.

[1] https://github.com/kata-containers/documentation/blob/master/how-to/how-to-set-sandbox-config-kata.md
[2] https://github.com/kata-containers/runtime/blob/master/virtcontainers/pkg/oci/utils.go#L353
[3] https://github.com/kata-containers/runtime/blob/master/virtcontainers/sandbox.go#L441

jodh-intel:
Thanks for raising @gkunz. I can confirm this behaviour.

c3d:
I had noticed the discrepancy while working on #3005. I should have asked, I assumed there was something intentional in the list given in `addAssetAnnotations`.

PR #3031, opened 2020-10-28
annotations: Improve asset annotation handling

jodh-intel:
Make `asset.go` the arbiter of asset annotations by removing all asset
annotations lists from other parts of the codebase.

This makes the code simpler, easier to maintain, and more robust.

Specifically, the previous behaviour was inconsistent as the following
ways:

- `createAssets()` in `sandbox.go` was not handling the following asset
  annotations:

    - firmware:
      - `io.katacontainers.config.hypervisor.firmware`
      - `io.katacontainers.config.hypervisor.firmware_hash`

    - hypervisor:
      - `io.katacontainers.config.hypervisor.path`
      - `io.katacontainers.config.hypervisor.hypervisor_hash`

    - hypervisor control binary:
      - `io.katacontainers.config.hypervisor.ctlpath`
      - `io.katacontainers.config.hypervisor.hypervisorctl_hash`

    - jailer:
      - `io.katacontainers.config.hypervisor.jailer_path`
      - `io.katacontainers.config.hypervisor.jailer_hash`

- `addAssetAnnotations()` in the `oci` package was not handling the
  following asset annotations:

    - hypervisor:
      - `io.katacontainers.config.hypervisor.path`
      - `io.katacontainers.config.hypervisor.hypervisor_hash`

    - hypervisor control binary:
      - `io.katacontainers.config.hypervisor.ctlpath`
      - `io.katacontainers.config.hypervisor.hypervisorctl_hash`

    - jailer:
      - `io.katacontainers.config.hypervisor.jailer_path`
      - `io.katacontainers.config.hypervisor.jailer_hash`

This change fixes the bug where specifying a custom hypervisor path via an
asset annotation was having no effect.

Fixes: #3030.


jodh-intel:
Minimal manual test (based on https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#running-standalone):

```bash
$ bundle="/tmp/bundle"
$ rootfs="$bundle/rootfs"
$ mkdir -p "$rootfs" && (cd "$bundle" && kata-runtime spec)
$ sudo docker export $(sudo docker create busybox) | tar -C "$rootfs" -xvf -
$ (cd $bundle && jq '. |= . + {"annotations": {"io.katacontainers.config.hypervisor.path": "/does/not/exist"}}' < config.json > tmp && mv tmp config.json)
$ sudo kata-runtime --log=/dev/stdout run --bundle "$bundle" foo
```

The above will work (incorrectly) with current releases, but will fail with this PR (assuming `/does/not/exist` really doesn't exist ;)

/cc @gkunz

gkunz:
@jodh-intel thanks for the quick fix! I'll test this PR as well in my lab.

jodh-intel:
@GabyCT - `jenkins-ci-podman-clh-fedora-31` is failing with:

```
16:08:18 ./scripts/dev_cli.sh: line 122: docker: command not found
```

amshinde (approved):

GabyCT:
@jodh-intel https://github.com/kata-containers/ci/issues/340 /cc @likebreath @jcvenegas

virtcontainers/pkg/oci/utils.go
```diff
@@ -347,15 +347,19 @@ func addAnnotations(ocispec specs.Spec, config *vc.SandboxConfig) error {
 
 func addAssetAnnotations(ocispec specs.Spec, config *vc.SandboxConfig) {
 	assetAnnotations := []string{
-		vcAnnotations.KernelPath,
-		vcAnnotations.ImagePath,
-		vcAnnotations.InitrdPath,
+		vcAnnotations.AssetHashType,
```
gkunz, line 350: are you not missing `vcAnnotations.HypervisorPath` here? In my local test of this PR, the hypervisor annotations still gets ignored. Upon including `vcAnnotations.HypervisorPath` here, it works.

gkunz (changes requested):

virtcontainers/pkg/oci/utils.go
jodh-intel: @gkunz - apologies - `git stash` fail :) I've actually found a way to make the code more robust, so marking as do-not-merge for now. I'll push an update tomorrow...

gkunz: @jodh-intel no problem. Thanks again for addressing the issue so quickly!

devimc (approved):
thanks @jodh-intel

virtcontainers/hypervisor.go
```diff
@@ -520,26 +520,12 @@ func (conf *HypervisorConfig) assetPath(t types.AssetType) (string, error) {
...
-	case types.FirmwareAsset:
-		return conf.FirmwarePath, nil
-	default:
-		return "", fmt.Errorf("Unknown asset type %v", t)
+	assetPath, _, err := t.Annotations()
+	if err != nil {
+		return "", err
 	}
+
+	return assetPath, nil
```
jodh-intel, line 528: This is incorrect as it changes the function to return the asset annotation *name*, not the config *value*. I'm about to push a fix and a new `TestAssetPath()` test to assert the expected behaviour. However, I'm going to have to revert this change meaning we will still have a "list of asset annotations" outside of `asset.go`. This is annoying as I'd like to totally encapsulate that list in `asset.go`.

One possible improvement would be to annotate `HypervisorConfig` with custom struct tags like this:

```go
type HypervisorConfig struct {
    FirmwarePath      string `annotation:"io.katacontainers.config.hypervisor.firmware"`
    HypervisorCtlPath string `annotation:"io.katacontainers.config.hypervisor.ctlpath"`
    HypervisorPath    string `annotation:"io.katacontainers.config.hypervisor.path"`
    ImagePath         string `annotation:"io.katacontainers.config.hypervisor.image"`
    InitrdPath        string `annotation:"io.katacontainers.config.hypervisor.initrd"`
    JailerPath        string `annotation:"io.katacontainers.config.hypervisor.jailer_path"`
    KernelPath        string `annotation:"io.katacontainers.config.hypervisor.kernel"`
}
```

That may make the code clearer and less error prone. However, there would be a runtime impact as we'd need to use `reflect` to map the struct tag (annotation) back to the config value.

jodh-intel:
HI @gkunz - I've updated the branch now if you'd like to give the latest a try?

gkunz:
@jodh-intel works nicely for me!

gkunz (approved):

jodh-intel:
Thanks for confirming @gkunz !

jodh-intel:
Added dnm label to allow for a few more eyes to review before this lands.

jodh-intel:
@amshinde - any further thoughts on this? I've updated since you approved I think.

virtcontainers/types/asset.go
```diff
@@ -18,6 +18,40 @@ import (
...
+		aPath, aHash, err := at.Annotations()
+		if err != nil {
+			return []string{}, err
+		}
+
+		result = append(result, []string{aPath, aHash}...)
+	}
+
+	// Add the odd one out; it isn't part of a (path+hash) pair.
+	result = append(result, annotations.AssetHashType)
```
amshinde, line 50: We dont need to address this in this PR. 
But raising my concern about value of supporting this. Maybe we should just support sha512 hash and document it so, rather than providing this.

virtcontainers/hypervisor.go
amshinde: @jodh-intel I like the idea of encapsulating all the assets constants and logic in assets.go, feel that it is all over the place right now. Maybe using reflection is not such a bad idea if we can cleanup the code and make sure that we do this once in the CreateSandbox code path.
We can address that in a separate PR.

What I did see is that the hypervisor.go has a lot of functions `CustomFirmwareAsset`, `CustomJailerAsset` etc which are not being called at all. 
We need to assess if these functions should be called from `getHypervisorDetails` or we can just get rid of them.

virtcontainers/types/asset.go
jodh-intel: Fixed.

virtcontainers/types/asset.go
```diff
@@ -18,6 +18,40 @@ import (
 // AssetType describe a type of assets.
 type AssetType string
 
+// AssetTypes returns a list of all known asset types.
+//
+// XXX: New asset types *MUST* be added here.
+func AssetTypes() []AssetType {
```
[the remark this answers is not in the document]
jodh-intel: Fixed.

virtcontainers/types/asset.go
```diff
@@ -18,6 +18,40 @@ import (
...
+	return []AssetType{
+		FirmwareAsset,
+		HypervisorAsset,
+		HypervisorCtlAsset,
+		ImageAsset,
+		InitrdAsset,
+		JailerAsset,
+		KernelAsset,
+	}
+}
```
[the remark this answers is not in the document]
jodh-intel: Fixed.

jodh-intel:
@amshinde - I've removed the unused functions. On reflection (:smile:), I'm not totally convinced about my reflection idea: there 
 is another problem of using that technique - we can't use variables in the struct tags so we'd end up with another maintenance issue as it would be easy to rename an annotation variable *value*, but forget to update the corresponding static struct tag annotation *string*. Sigh.

Let's ponder this some more and make further improvements if possible on a follow-up PR.

For now, I'd like to get this landed so it can be ported to 2.x.

amshinde:
> Let's ponder this some more and make further improvements if possible on a follow-up PR.
> 
> For now, I'd like to get this landed so it can be ported to 2.x.

Agree. SGTM.

jodh-intel:
forward port PR: https://github.com/kata-containers/kata-containers/pull/1086

Issue #2314, opened 2019-12-03
docker 19.03: pod information left on /run/vc/ when killing kata hypervisor


chavafg:
# Description of problem

While trying to update our supported docker version to 19.03, I see that the [`hypervisor_stability_kill_test`](https://github.com/kata-containers/tests/blob/master/integration/stability/hypervisor_stability_kill_test.sh) fails as pod information is left under ` /run/vc/sbs`. This does not happen when using docker 18.06.

To reproduce, follow next steps or run the test script above.

```
fuentess@newdocker:~$ sudo docker run --runtime kata-runtime -tid ubuntu bash
a4ddc2808ed7acf4883397024251fd122990025cf6a7e400a9c166958c010092
fuentess@newdocker:~$ sudo docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
a4ddc2808ed7        ubuntu              "bash"              20 seconds ago      Up 16 seconds                           dreamy_williamson
fuentess@newdocker:~$ pidof qemu-system-x86_64
865
fuentess@newdocker:~$ sudo kill -KILL 865
fuentess@newdocker:~$ sudo docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
fuentess@newdocker:~$ sudo kata-runtime list 
WARNING: failed to get container a4ddc2808ed7acf4883397024251fd122990025cf6a7e400a9c166958c010092 rootfs: stat /var/lib/docker/overlay2/f9314de2f862faf78cb3011921c7124d2af6e3e1e4d23c28c5aae791056f96d4/merged: no such file or directory
ID          PID         STATUS      BUNDLE      CREATED     OWNER
fuentess@newdocker:~$ sudo ls /run/vc/sbs
a4ddc2808ed7acf4883397024251fd122990025cf6a7e400a9c166958c010092
```

# Expected result

pod information on `/run/vc/sbs` should be deleted by the kata-runtime. 

/cc @bergwolf who added the test.
/cc @devimc @jodh-intel @egernst in case they have any idea.

raravena80:
@chavafg any updates on this PR? Thx.

GabyCT:
ping @bergwolf

raravena80:
ping @bergwolf @chavafg

PR #3037, opened 2020-10-30
qemu: fix stopSandbox() to work with newer docker

rhafer:
When connecting to the qemu socket fails with a connection refused error
assume that qemu died and handle that as if the Sandbox was already
stopped. So that deleteSandbox() can still succeed.

This is needed when working with newer docker releases (19.03). As the
containerd included in there does no longer use the "<runtime> kill"
command before deleting the Sandboxes.
(https://github.com/containerd/containerd/commit/b5ccc66c2c814cfc21d58678c854272427814b59#diff-6a93af2fa6a81ffa3592c6ba4c7f628976e9ac0363a9532589b0fc96f3059ad1)

Fixes: #2611, #2314 

Signed-off-by: Ralf Haferkamp <rhafer@suse.com>

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -931,6 +931,22 @@ func (q *qemu) waitSandbox(timeout int) error {
 	return nil
 }
 
+func isConnectionRefusedError(err error) bool {
+	if err == nil {
+		return false
+	}
+	if neterr, ok := err.(*net.OpError); ok {
+		if syserr, ok := neterr.Err.(*os.SyscallError); ok {
+			if errno, ok := syserr.Err.(syscall.Errno); ok {
+				if errno == unix.ECONNREFUSED {
+					return true
+				}
+			}
+		}
+	}
+	return false
+}
+
 // stopSandbox will stop the Sandbox's VM.
 func (q *qemu) stopSandbox() error {
 	span, _ := q.trace("stopSandbox")
@@ -962,6 +978,13 @@ func (q *qemu) stopSandbox() error {
 
 	err := q.qmpSetup()
 	if err != nil {
+		// Ignore any "connection refused" error and assume that the Sandbox
+		// is already stopped in that case.
+		if isConnectionRefusedError(err) {
+			q.Logger().WithError(err).Warn(
+				"Failed to connect to qemu. Assuming Sandbox is already gone.")
+			return nil
+		}
 		return err
 	}
 


rhafer:
I am not really sure if this is actually the right fix. But I want to bring it up for discussion anyway.

Also I am not sure if there is even still interest in getting this addressed for kata 1.x given that `kata-runtime` in 2.0 does not provide a drop-in replacement for `runc` at all.

c3d:
Hi @rhafer, thanks a lot for the fix, and welcome. The general idea looks good to me.

Could you please open an issue, and then add a `Fixes: #<issuenumber>` to your commit? Our CI requires it.

As for whether we need that in 2.0, I'd say chances are the same issue may happen there, even if it is not as easy to add. Also, we don't want too many differences in the code. So I added a label `needs-forward-port`.

virtcontainers/qemu.go
```diff
@@ -961,6 +961,21 @@ func (q *qemu) stopSandbox() error {
 
 	err := q.qmpSetup()
 	if err != nil {
+		// Ignore any "connection refused" error and assume that the Sandbox
```
c3d, line 964: Would you mind putting this code in a separate function `isConnexionRefusedError(err)` to streamline the logic here and clarify the intent?

amshinde: @rhafer Thanks for raising this issue that you are seeing with newer containerd versions. 
Its not clear to me though why the sandbox has already gone at this stage given the behaviour your referenced in the link. Will need to take a closer look at this. 
We will probably need to have the fix in such a way that it applies to all supported hypervisors and not just qemu.

rhafer: > Would you mind putting this code in a separate function isConnexionRefusedError(err) to streamline the logic here and clarify the intent?

:+1:

> Its not clear to me though why the sandbox has already gone at this stage given the behaviour your referenced in the link. Will need to take a closer look at this.

The issue happens when the qemu process gets killed (or crashes). I used https://github.com/kata-containers/tests/blob/master/integration/stability/hypervisor_stability_kill_test.sh as a reproducer. (Basically what is described in #2611)

> We will probably need to have the fix in such a way that it applies to all supported hypervisors and not just qemu.

Hm, yeah. I haven't checked the other hypervisors. Not sure if their implementations suffer from the same issue. A way to fix this in a more general way might be always call `deleteSandbox` with  `force=true` here https://github.com/kata-containers/runtime/blob/master/cli/delete.go#L121. However I think this might be ignore a few real errors.

virtcontainers/qemu.go
```diff
@@ -931,6 +931,17 @@ func (q *qemu) waitSandbox(timeout int) error {
 	return nil
 }
 
+func isConnectionRefusedError(err error) bool {
+	if neterr, ok := err.(*net.OpError); ok {
+		if syserr, ok := neterr.Err.(*os.SyscallError); ok {
+			if syserr.Error() == "connect: connection refused" {
```
jodh-intel, line 937: Wow, they don't make it easy do they? :smile: `SyscallError` is an odd beast - why we cannot check the `errno` value directly I have no idea. What worries me is that the design of `SyscallError` looks rather fragile to me: if that package suddenly supported i18n, this would break. But I've found a solution:

```go
if errno, ok := syserr.Err.(syscall.Errno); ok {
    if errno == unix.ECONNREFUSED {
        return true
    }
}
```

virtcontainers/qemu.go
```diff
@@ -931,6 +931,17 @@ func (q *qemu) waitSandbox(timeout int) error {
 	return nil
 }
 
+func isConnectionRefusedError(err error) bool {
+	if neterr, ok := err.(*net.OpError); ok {
```
jodh-intel, line 935: If `err` is `nil` this will panic I think.

jodh-intel (changes requested):
Thanks @rhafer. I've found a way to avoid the error string comparison.

virtcontainers/qemu.go
rhafer: >  What worries me is that the design of `SyscallError` looks rather fragile to me: if that package suddenly supported i18n, this would break.

Yeah, true. I didn't particularly like comparing the error string either :-/

> But I've found a solution

Cool, thanks. I'll rework that.

virtcontainers/qemu.go
rhafer: Hm, true. Though currently it'll be only called with non-nill values. Anyway, I'll just add another check here.

virtcontainers/qemu.go
jodh-intel: /cc @likebreath.

virtcontainers/qemu.go
```diff
@@ -962,6 +978,13 @@ func (q *qemu) stopSandbox() error {
 
 	err := q.qmpSetup()
 	if err != nil {
+		// Ignore any "connection refused" error and assume that the Sandbox
+		// is already stopped in that case.
+		if isConnectionRefusedError(err) {
+			q.Logger().WithError(err).Warn(
+				"Failed to connect to qemu. Assuming Sandbox is already gone.")
```
jodh-intel, line 985: Nit: "QEMU".

jodh-intel (approved):
Thanks @rhafer.

liubin (approved):

Issue #3039, opened 2020-11-01
Logging level is broken in some cases


marcel-apf:
# Description of problem

Run CRI-O  in normal mode and set enable_debug = true in  configuration.toml.

# Expected result

The runtime logging level should be INFO, which from KATA's perspective behaves like DEBUG.

# Actual result

The runtime logging level is WARN, which from KATA's perspective behaves like INFO.

# Further information

The issue can be traced to  commit eee4d7012 "katutils: logging: default log level to Warn".
Before the mentioned patch the default logging level was "Info" which matches the comment:
https://github.com/kata-containers/runtime/blob/c54378d7768f8f7e0c613ee331fe9a63b89d65af/pkg/katautils/config.go#L1193,L1194

However after the patch the default logging level is not "Debug" anymore.

fidencio:
---

This issue is being automatically closed as Kata Containers 1.x has now reached EOL (End of Life). This means it is **no longer being maintained**.

> **Important:**
>
> All users should switch to the latest [Kata Containers 2.x release](https://github.com/kata-containers/kata-containers/blob/main/docs/install/README.md) to ensure they are using a maintained release that contains the latest security fixes, performance improvements and new features.

This decision was discussed by the @kata-containers/architecture-committee and has been announced via the [Kata Containers mailing list](http://lists.katacontainers.io/pipermail/kata-dev/):
* http://lists.katacontainers.io/pipermail/kata-dev/2020-November/001601.html
* http://lists.katacontainers.io/pipermail/kata-dev/2021-April/001843.html
* http://lists.katacontainers.io/pipermail/kata-dev/2021-May/001896.html

If you believe this issue still applies to Kata Containers 2.x, please open an issue against the [Kata Containers 2.x repository](https://github.com/kata-containers/kata-containers), pointing to this one, providing details to allow us to migrate it.

---

PR #3040, opened 2020-11-01
logging: Fix Logging level

marcel-apf:
When running CRI-O in normal mode and setting enable_debug = true
in configuration.toml, the logging level should be info
(which from KATA's perspective behaves like DEBUG).
However the runtime logging level will be  WARN (which from KATA's
perspective behaves like INFO).

The issue can be traced to commit
    eee4d70 "katutils: logging: default log level to Warn".

Fix it by setting the logging level whether the debugging mode is
enabled or not.

Fixes: #3039

Signed-off-by: Marcel Apfelbaum <marcel@redhat.com>

### DIFF
--- a/pkg/katautils/config.go
+++ b/pkg/katautils/config.go
@@ -1189,7 +1189,10 @@ func LoadConfiguration(configPath string, ignoreLogging, builtIn bool) (resolved
 	}
 
 	config.Debug = tomlConf.Runtime.Debug
-	if !tomlConf.Runtime.Debug {
+	if tomlConf.Runtime.Debug {
+		// logrus "Info" level is used as "Debug"
+		kataUtilsLogger.Logger.Level = logrus.InfoLevel
+	} else {
 		// If debug is not required, switch back to the original
 		// default log priority, otherwise continue in debug mode.
 		kataUtilsLogger.Logger.Level = originalLoggerLevel


fidencio:
@marcel-apf, thanks for the patch. We've discussed this offline and I'd like to know the opinion of the other developers as this patch is the first piece of something that will be a bigger work.

Long-story-short, we should start dealing with our logs in a more standardized way, then we'll allow the admin setting the log_level via the configuration.toml file. This would allow us to have fine grain what we want to see and will help with debuggability.

Anyways, I couldn't wrap my mind around whether this is a valid effort for 1.x or whether it'd be too intrusive at this point. And here we're, getting back to http://lists.katacontainers.io/pipermail/kata-dev/2020-October/001561.html, which would be really good to have more opinions there.

devimc:
thanks @marcel-apf 

> When running CRI-O in normal mode and setting enable_debug = true in configuration.toml, the logging level should be info

what's the expected logging level in containerd?

fidencio:
> thanks @marcel-apf
> 
> > When running CRI-O in normal mode and setting enable_debug = true in configuration.toml, the logging level should be info
> 
> what's the expected logging level in containerd?

https://github.com/containerd/containerd/blob/master/docs/ops.md says that containerd has the following log levels: "debug, info, warn, error, fatal, panic", meaning it does match with CRI-O. Their manpage mentions that "info" in the default log level (when in debug mode).

Does this answer your question, @devimc?

marcel-apf:
@fidencio thank you for you help! Please let me know if there is anything I have missed or I should further do.

jodh-intel (changes requested):
Hi @marcel-apf - thanks for raising, but I'm afraid this isn't clear to me.

If I'm understanding correctly, this change is intended to be cri-o specific, but the change will affect all container managers surely?

The change also looks confusing and fwics means you can not enable true debug output?

marcel-apf:
Hi @jodh-intel, 
> Hi @marcel-apf - thanks for raising, but I'm afraid this isn't clear to me.
> 
> If I'm understanding correctly, this change is intended to be cri-o specific, but the change will affect all container managers surely?

Not really, this change is intended to give "priority" to the log-level settings in the runtime configuration (if set to "debug"). 
Also is more about the runtime behavior when it does not receive the "debug" flag. However, the error is easily reproducible when using cri-o, and since I proposed the PR as a fix, I thought is best If I "connect" the fix to the way I found it.

> 
> The change also looks confusing and fwics means you can not enable true debug output?

I have to agree with you about the confusing part, but I think it may be related to the way the logging is initialized. 
To answer your question, you cannot enable true debug output if the "-debug" flag is not passed as parameter to the runtime, that is at least how I understand the situation .

Zooming in, the code only addresses the configuration debug field only if is set to "false", assuming that otherwise the default configuration will set the logging to debug mode. However, that is not true after commit  eee4d70.

fidencio:
Folks, this PR has been stale and I'm closing it as we're walking towards the **final** 1.13.0 release of the kata-containers-1.x.
From now on, all the development will happen as part of https://github.com/kata-containers/kata-containers/ and fixes should be provided to that repo.

Thanks a lot for your contribution, we're sorry that for some reason it got stale, and we hope to see you around for new adventures in the kata-containers-2.x land.

PR #3042, opened 2020-11-02
readonly volume should be bind mounted readonly on the host

bergwolf:
So that even if the guest is compromised, we still ensure that readonly volumes cannot be modified.

backport kata-containers/kata-containers#1062

### DIFF
--- a/virtcontainers/mount.go
+++ b/virtcontainers/mount.go
@@ -273,6 +273,11 @@ func remount(ctx context.Context, mountflags uintptr, src string) error {
 	return nil
 }
 
+// remount a mount point as readonly
+func remountRo(ctx context.Context, src string) error {
+	return remount(ctx, syscall.MS_BIND|syscall.MS_RDONLY, src)
+}
+
 // bindMountContainerRootfs bind mounts a container rootfs into a 9pfs shared
 // directory between the guest and the host.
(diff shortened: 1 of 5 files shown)


jodh-intel (approved):
Thanks @bergwolf

awprice:
Thanks for raising this @bergwolf - code works in my environment on 1.11.x and fixes the issue.

virtcontainers/mount.go
```diff
@@ -273,6 +273,11 @@ func remount(ctx context.Context, mountflags uintptr, src string) error {
 	return nil
 }
 
+// remount a mount point as readonly
+func remount_ro(ctx context.Context, src string) error {
```
devimc, line 277: rust is affecting us :smile: 

```
virtcontainers/mount.go:277:6: don't use underscores in Go names; func remount_ro should be remountRo (golint)

func remount_ro(ctx context.Context, src string) error {
```

devimc (approved):
thanks @bergwolf - just one nit

amshinde:
@chavafg Is the `jenkins-metrics-ubuntu-18-04` CI failing?
I don't see this PR causing a failure in the metrics. Can you take a look.

chavafg:
The job looks stable, only failing in this PR with blogbench metric going out over the limit with a percentage value of 110.7%.

amshinde:
I dont see how making volumes as read-only could impact the metrics. I'll rerun the the metrics CI to double check.

PR #3051, opened 2020-11-10
Backports: Read-only mount fixes for 1.11

amshinde:
3f0e61c runtime: mount shared mountpoint readonly
228e6eb runtime: readonly mounts should be readonly bindmount on the host


fidencio (approved):
lgtm, thanks!

chavafg (approved):

Issue #2982, opened 2020-09-23
build with go 1.15.2, createContainer failed: Setctty set but Ctty not valid in child 


xiayuhai:
kata-runtime is built with go 1.15.2, and there were some modifications made in the syscall package in Go 1.15.0
Error message: `createContainer failed: Setctty set but Ctty not valid in child`,  
and it looks like the same with [https://github.com/cespare/reflex/issues/77]

PR #3082, opened 2020-11-18
virtcontainers: Don't set CTTY

fidencio:
The https://go-review.googlesource.com/c/go/+/231638/ commit on Golang
introduced a failure on Kata Containers when the runtime is built with
golang 15.2+.

Fixes: #2982

### DIFF
--- a/virtcontainers/shim.go
+++ b/virtcontainers/shim.go
@@ -208,9 +208,6 @@ func startShim(args []string, params ShimParams) (int, error) {
 		cmd.Stderr = f
 		// Create Session
 		cmd.SysProcAttr.Setsid = true
-		// Set Controlling terminal to Ctty
-		cmd.SysProcAttr.Setctty = true
-		cmd.SysProcAttr.Ctty = int(f.Fd())
 	}
 	defer func() {
 		if f != nil {


fidencio:
I must admit I'm not exactly sure whether we actually do need that call, but it seems to be safe to remove (when using with podman).

Let's see what CI tells us.

virtcontainers/shim.go
```diff
@@ -210,7 +210,6 @@ func startShim(args []string, params ShimParams) (int, error) {
 		cmd.SysProcAttr.Setsid = true
 		// Set Controlling terminal to Ctty
 		cmd.SysProcAttr.Setctty = true
```
sameo, line 212: Shouldnt we set that one to false then?

fidencio: Ack, updated!

fidencio: Hmmm. It seems that I actually we shouldn't remove the `Setctty = true` call.

Removing it results in:
```
DEBU[0003] ExitCode msg: "container create failed (no logs from conmon): eof" 
Error: container create failed (no logs from conmon): EOF
```

fidencio: Ah, no, I tested the wrong branch. :-)
Let me mark this one as resolved again.

fidencio:
- **jenkins-ci-ARM-ubuntu-18.04**:
```
21:50:36 Stderr: Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```
Not sure if we can do something on our side to avoid this.

fidencio:
@devimc, can you help me to understand the vfio failures? (I restarted the vfio CI yesterday, just to be sure it wasn't a flake).

sameo (approved):

devimc (approved):
thanks @fidencio

Issue #3083, opened 2020-11-19
split qemu support in stable and experimental


jcvenegas:
**Which feature do you think can be improved?**

Update: After some talk on this issue: No more runtime configs will be created for qemu-virtiofs experimental, instead some job that modify the configuration toml that to point to modify anything we need to test.

Today virtiofs config uses cache=auto and no dax, both are supported in stable kernel releases and qemu.

The config could change in the future looking for the DAX bennefits, but because more testing and stablization is needed there, specially kata keep testing and give feedback to the virtiofs project. It requiered to split the testing in two:

- Move kata-qemu-virtiofs to qemu-vanilla(the one that we track in versions.yaml) and to our lts kernel.

Keep testing dev branch updates from virtiofs

- Create a job that qemu-viritiofs modifies the default qemu-viritofs-config to
  * qemu-vanilla + virtiofs newer patches
  * Any kernel recommended by virtiofs 
  * DAX
  * Unless wanted by users there is not runtimeclass or kata-runtime wrapper for this config.

**Additional Information**

Anything else to add?

**Before raising this feature request**

Have you looked at the [limitations document](https://github.com/kata-containers/documentation/blob/master/Limitations.md)?

wainersm:
@jodh-intel @jcvenegas ,

I always understood QEMU "experimental" as synonymous of qemu-virtiofs. Also I don't know how to call the non-experimental QEMU so usually I say QEMU "standard". Now with this change we have the experimental of QEMU experimental ("qemu-viritiofs experimental" and, using my nomenclature standard of QEMU experimental. Don't you think this is too confusing?

I know that giving name is hard, I am not good at that at all. But let me give a few suggestions:
* experimental stable vs unstable
* experimental stable vs master (often used in Git, but "master" nowadays is avoided)
* experimental stable vs experimental next (I like that nomenclature. In the end it seems what we try to achieve is the "next" stable version of QEMU experimental) 

Disregard if I misunderstood this completely. :)

jcvenegas:
@wainersm 
> I always understood QEMU "experimental" as synonymous of qemu-virtiofs.

Yes, experimental all this time has been just for virtiofs usecase, as part of this CI refactor I was looking to have a more meaningful name.

- It points to virtiofs out-of-tree dev repository from virtiofs dev team.
- This help us to  easily use the tests new patches  that virtiofs team recommend. 

> Also I don't know how to call the non-experimental QEMU so usually I say QEMU "standard"

[...]

PR #3084, opened 2020-11-19
build: Split virtiofs config into stable and experimental

jcvenegas:
Virtiofs config today uses stable features. But in order to
keep testing newer changes and be more agresive.

stable:
- Stable kernel
- qemu vanilla
- no dax
- cache auto

experimental:
- Non lts kernel
- Patched kernel
- Patched qemu
- DAX support

Depends-on: github.com/kata-containers/tests#3068

Fixes: #3083

Signed-off-by: Carlos Venegas <jos.c.venegas.munoz@intel.com>


jodh-intel:
Hi @jcvenegas - I like the concept, but I wonder if we need this in 1.x at this stage? Is there a 2.x equivalent PR too?

Issue #2785, opened 2020-06-19
remove "default hypervisor" message from build


egernst:
AFAIU, there really isn't a notion of "default" VMM for Kata. Let's update the build summary to remove this:

```
• hypervisors:
	Default: qemu
	Known: acrn cloud-hypervisor firecracker qemu qemu-virtiofs
	Available for this architecture: acrn cloud-hypervisor firecracker qemu qemu-virtiofs
```

We should be able to remove the `Default` line, and keep the remaining.

bporter816:
Hi! @tommyrousey, @maomarcus, and @yasv123 and I are students at UT Austin, and we're looking to make some open source contributions over the course of the rest of the year as a part of our virtualization class. Could we be assigned to this issue?

bporter816:
@egernst  
@jodh-intel @ariel-adam @fidencio We will be completing this issue soon and were wondering if there are any functional changes that need to happen outside of removing references to DEFAULT_HYPERVISOR from the Makefile.

egernst:
Hey there @bporter816! I don't think there are any functional changes required; just cleanup/removal for this one.

egernst:
And, sorry for not responding in September! Thanks for the help!

egernst:
If you're interested, we can see if there are more interesting issues within the code base for you :)

bporter816:
@egernst Yes, that would be great! We would love to pick up another issue. Please let us know if you're aware a good one we could scope out that would take about a week to complete.

egernst:
> @egernst Yes, that would be great! We would love to pick up another issue. Please let us know if you're aware a good one we could scope out that would take about a week to complete.

Yes - will do! TAL @ https://github.com/kata-containers/runtime/issues/3093

liubin:
@bporter816 Thanks, you can also filter issues by label `good-first-issue`.

PR #3092, opened 2020-12-02
build: Remove default hypervisor message from build

bporter816:
The default hypervisor variable will no longer be printed from
the Makefile.

Fixes: #2785

Signed-off-by: Benjamin Porter <bporter816@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -767,7 +767,6 @@ else
 endif
 	@printf "\n"
 	@printf "• hypervisors:\n"
-	@printf "\tDefault: $(DEFAULT_HYPERVISOR)\n"
 	@printf "\tKnown: $(sort $(HYPERVISORS))\n"
 	@printf "\tAvailable for this architecture: $(sort $(KNOWN_HYPERVISORS))\n"
 	@printf "\n"


bporter816:
Could someone add the `no-backport-needed` and `no-forward-port-needed` labels?

Also wondering if just removing this printout is sufficient or if we should also remove all references to `DEFAULT_HYPERVISOR` in the Makefile completely.

devimc (approved):
thanks @bporter816 - lgtm

egernst:
Thanks @bporter816.

Looking at it a bit closer, I now understand what default really means in this case. Kata supports multiple VMMs, and ultimately the same runtime-binary does this by utilizing a VMM specific configuration file. Today this is done by things like `/etc/kata-containers/configuration-qemu.toml` and `/etc/kata-containers/configuration-clh.toml`.  The 'default' hypervisor (VMM) ends up creating `/etc/kata-containers/configuration.toml`, which is a symlink today to configuration-qemu.toml.

I think we'll still want this in place, since the runtime will default to looking @ configuration.toml if config isn't otherwise specified.

As a user, I end up specifying the configuration toml I will be using per runtimeclass anyway, so I never consider this 'default' one. @jodh-intel, @sameo  - you have thoughts on this? I think we need to keep this default configuration toml, and we can consider if we want to change the default value in the future from QEMU?

egernst (approved):

PR #3101, opened 2020-12-09
qemu: no state to save if QEMU isn't running

egernst:
On pod delete, we were looking to read files that we had just deleted. In particular,
stopSandbox for QEMU was called (we cleanup up vmpath), and then QEMU's
save function was called, which immediately checks for the PID file.

Let's only update the persist store for QEMU if QEMU is actually
running. This'll avoid Error messages being displayed when we are
stopping and deleting a sandbox:

```
level=error msg="Could not read qemu pid file"
```

I reviewed CLH, and it looks like it is already taking appropriate
action, so no changes needed.

Ideally we won't spend much time saving state to persist.json unless
there's an actual error during stop/delete/shutdown path, as the persist will
also be removed after the pod is removed. We may want to optimize this,
as currently we are doing a persist store when deleting each container
(after the sandbox is stopped, VM is killed), and when we stop the sandbox.
This'll require more rework... tracked in:
  https://github.com/kata-containers/kata-containers/issues/1181

Fixes: #1179

Signed-off-by: Eric Ernst <eric.g.ernst@gmail.com>

### DIFF
--- a/virtcontainers/qemu.go
+++ b/virtcontainers/qemu.go
@@ -2227,6 +2227,11 @@ func (q *qemu) toGrpc() ([]byte, error) {
 }
 
 func (q *qemu) save() (s persistapi.HypervisorState) {
+	// If QEMU isn't even running, there isn't any state to save
+	if q.stopped {
+		return
+	}
+
 	pids := q.getPids()
 	if len(pids) != 0 {
 		s.Pid = pids[0]


egernst:
Hmm:
```
1..1
not ok 1 Block Storage Support
# (in test file k8s-block-volume.bats, line 46)
#   `kubectl wait --for condition=ready "pod/${pod_name}"' failed
# INFO: k8s configured to use runtimeclass
# storageclass.storage.k8s.io/local-storage created
# persistentvolume/block-loop-pv created
# persistentvolumeclaim/block-loop-pvc created
# pod/pod-block-pv created
# error: timed out waiting for the condition on pods/pod-block-pv
# pod "pod-block-pv" deleted
# persistentvolumeclaim "block-loop-pvc" deleted
# persistentvolume "block-loop-pv" deleted
# storageclass.storage.k8s.io "local-storage" deleted
Failed at 82: bats "${K8S_TEST_ENTRY}"
```

jodh-intel:
The CentOS CI is failing with:

```
Warning: failed loading '/etc/yum.repos.d/CentOS-Base.repo', skipping.
Error: No matching repo to modify: PowerTools.
Failed at 46: sudo yum-config-manager --enable PowerTools
```

/cc @GabyCT

jodh-intel (approved):
lgtm - I guess this needs to be backported to the 1.x stable branches?

egernst:
I think it already landed in 1.12 @jodh-intel

jodh-intel:
> I think it already landed in 1.12 @jodh-intel

I doesn't appear to have:

- https://github.com/kata-containers/runtime/blob/stable-1.12/virtcontainers/qemu.go#L2203.

But it shouldn't have hit a stable release before hitting `master` anyway. :confused:

btw - your "Fixes #" number looks wrong (#1179 = "The virtcontainers network interface is not needed" and is already closed)

chavafg (approved):

PR #3109, opened 2020-12-22
runtime: Fix hugepages support

### LINKED ISSUE #2353: hugepage usage with kubernetes and kata-container
# Description of problem
I summarized some usage scenarios of hugepages
## 1. Configuration
The following two options have an impact on huge pages, where ```enable_hugepages``` is always configured as ```true```, the ```default_memory``` always configured as ```2048```and ```sandbox_cgroup_only``` is divided into two cases: ```(1) true (2) false```
```shell
sandbox_cgroup_only = true / false
enable_hugepages = true
default_memory = 2048
```
***config hugepage on host***
if the ```nr_hugepages```has changed,the ```kubelet```should restart.
```shell
[root@host-69 yaml]# sysctl vm.nr_hugepages=1024
vm.nr_hugepages = 1024
[root@host-69 yaml]# cat /proc/meminfo | grep -i huge
AnonHugePages:    280576 kB
HugePages_Total:    1024
HugePages_Free:     1024
HugePages_Rsvd:        0
HugePages_Surp:        0
Hugepagesize:       2048 kB
[root@host-69 yaml]# systemctl restart kubelet
```
### 2.1 pod does not configure huge page parameters and sandbox_cgroup_only=true　
The ```yaml```file:
```shell
apiVersion: v1
kind: Pod
metadata:
  name: busybox-one
spec:
  runtimeClassName: kata-shimv1
  containers:
  - name: busybox-1
    image: busybox:latest
    stdin: true
    tty: true
    imagePullPolicy: IfNotPresent
``` 
Restart ```kubelet```service
```shell
systemctl restart kubelet
```
Start ```pod```,and then i get the error:
```shell
Warning  FailedCreatePodSandBox  10s        kubelet, host-69   Failed create pod sandbox: rpc error: code = Unknown desc = failed to create containerd task: OCI runtime create failed: Failed to check if grpc server is working: rpc error: code = Unavailable desc = transport is closing: unknown
```
***In this case, the usage scenario is incorrect. Because the huge page parameter is not configured in the yaml file, the limit at the pod level is 0. At this time, the qemu process is placed in a subdirectory under the pod level because sandbox_cgroup_only = true.Thus the qemu porcess will be restricted.More detailed analysis can refer to the following link:***
[https://github.com/kata-containers/runtime/issues/2172]()

### 2.2 pod does not configure huge page parameters and sandbox_cgroup_only = false

```shell
sandbox_cgroup_only = false
```
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: busybox-one
spec:
  runtimeClassName: kata-shimv1
  containers:
  - name: busybox-1
    image: busybox:latest
    stdin: true
    tty: true
    imagePullPolicy: IfNotPresent
```
Start ```pod```:
```shell
[root@host-69 yaml]# systemctl restart kubelet
[root@host-69 yaml]# kubectl apply -f busybox-one_no_hugepage.yaml
root@host-69 yaml]# kubectl get pods
NAME          READY   STATUS    RESTARTS   AGE
busybox-one   1/1     Running   0          60s
```
***At this point, the pod can be successfully started, because when sandbox_cgroup_only = false, although the pod level limit is still 0, the qemu process is not restricted because it is not placed in the control group under the pod.***

#### 2.3.1 pod started failed
```pod```config:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: busybox-one
spec:
  runtimeClassName: kata-shimv1
  containers:
  - name: busybox-1
    image: busybox:latest
    stdin: true
    tty: true
    imagePullPolicy: IfNotPresent
    volumeMounts:
    - mountPath: /dev/hugepages
      name: hugepage
    resources:
      limits:
        hugepages-2Mi: 4096Mi
        memory: 4096Mi
      requests:
        memory: 4096Mi
  volumes:
  - name: hugepage
    emptyDir:
      medium: HugePages
```
modify host hugepage configuration:
```shell
[root@host-69 yaml]# sysctl vm.nr_hugepages=2048
vm.nr_hugepages = 2048
[root@host-69 yaml]#
```
Restart ```kubelet```service:
```shell
systemctl restart kubelet
```
Run ```pod```,and then i get the error:
```shell
[root@host-69 yaml]# kubectl apply -f busybox-one.yaml 
pod/busybox-one created
[root@host-69 yaml]# kubectl get pods
NAME          READY   STATUS              RESTARTS   AGE
busybox-one   0/1     RunContainerError   1          5s
[root@host-69 yaml]# kubectl describe pods busybox-one
 Warning  Failed     25s (x3 over 41s)  kubelet, host-69   Error: failed to create containerd task: OCI runtime create failed: QMP command failed: unable to map backing store for guest RAM: Cannot allocate memory: unknown
```
```shell
k8s+qemu = 4096Mi + 2048Mi = needHugePages
HostHugepages = 2048 * 2Mi
```
***The above error is because there is only 4096M huge page memory on the host, but when kata calculates, 4096Mi in the pod and 2048Mi configured by kata itself are accumulated. When the virtual machine is started, the qemu process detects that the requested huge page memory exceeds Host configuration value, so an error is reported.Also,k8s does not know that kata itself has configured huge pages for virtual machines.***

#### 2.3.2 pod started successfully
```pod```config:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: busybox-one
spec:
  runtimeClassName: kata-shimv1
  containers:
  - name: busybox-1
    image: busybox:latest
    stdin: true
    tty: true
    imagePullPolicy: IfNotPresent
    volumeMounts:
    - mountPath: /dev/hugepages
      name: hugepage
    resources:
      limits:
        hugepages-2Mi: 128Mi
        memory: 128Mi
      requests:
        memory: 128Mi
  volumes:
  - name: hugepage
    emptyDir:
      medium: HugePages
```
Host hugepage conig:
modify host hugepage configuration:
```shell
[root@host-69 yaml]# sysctl vm.nr_hugepages=1088
vm.nr_hugepages = 1088
[root@host-69 yaml]#
```
Restart kubeletservice:
```shell
systemctl restart kubelet
```
Run pod, pod started successfully:
```shell
root@host-69 yaml]# kubectl apply -f busybox-one.yaml 
pod/busybox-one created
[root@host-69 yaml]# kubectl get pods
NAME          READY   STATUS    RESTARTS   AGE
busybox-one   1/1     Running   0          4s
[root@host-69 yaml]# cat /proc/meminfo | grep -i huge
AnonHugePages:    331776 kB
HugePages_Total:    1088
HugePages_Free:     1034
HugePages_Rsvd:     1034
HugePages_Surp:        0
Hugepagesize:       2048 kB
```
***Although the above pods can be successfully launched, there are also hidden problems. Because the ```hugepages-2Mi: 128Mi``` value is written in the pod-level control group, the virtual machine is configured with a huge page of ```2Gi```; when the business in the virtual machine uses a large amount of memory, at the same time the qemu process is placed in the pod's child control group which will exceed the limit of the pod control group. This issue may cause the virtual machine to go down.And i guess,the error is same as ```2.1```***
```shell
[root@host-69 pod5e2d430f-1ebe-4d80-a6fc-5ae52d897b6e]# pwd
/sys/fs/cgroup/hugetlb/kubepods/burstable/pod5e2d430f-1ebe-4d80-a6fc-5ae52d897b6e
[root@host-69 pod5e2d430f-1ebe-4d80-a6fc-5ae52d897b6e]# cat hugetlb.2MB.limit_in_bytes
134217728
[root@host-69 pod5e2d430f-1ebe-4d80-a6fc-5ae52d897b6e
```
```shell
[root@host-69 pod5e2d430f-1ebe-4d80-a6fc-5ae52d897b6e]# cd ./kata_7f3be98b62b18f29acbe93aea60c04e0168d9edafce698ff4184bda8a0c1c38b/
[root@host-69 kata_7f3be98b62b18f29acbe93aea60c04e0168d9edafce698ff4184bda8a0c1c38b]# cat cgroup.procs 
43436
43438
43444
43458
[root@host-69 kata_7f3be98b62b18f29acbe93aea60c04e0168d9edafce698ff4184bda8a0c1c38b]# ps aux | grep 43436
root      4052  0.0  0.0  16340  1112 pts/2    S+   09:49   0:00 grep --color=auto 43436
root     43436  0.0  0.0 2759216 67008 ?       Sl   12月12   0:42 /usr/bin/qemu-vanilla-system-x86_64 -name sandbox-7f3be98b62b18f29acbe93aea60c04e0168d9edafce698ff4184bda8a0c1c38b -uuid c13cf4c5-21a3-45bf-ba0b-77dbdef716b4 -machine pc,accel=kvm,kernel_irqchip,nvdimm -cpu host -qmp unix:/run/vc/vm/7f3be98b62b18f29acbe93aea60c04e0168d9edafce698ff4184bda8a0c1c38b/qmp.sock,server,nowait -m 2048M,slots=10,maxmem=97558M -device pci-bridge,bus=pci.0,id=pci-bridge-0,chassis_nr=1,shpc=on,addr=2,romfile= -device virtio-serial-pci,disable-modern=false,id=serial0,romfile= -device virtconsole,chardev=charconsole0,id=console0 -chardev socket,id=charconsole0,path=/run/vc/vm/7f3be98b62b18f29acbe93aea60c04e0168d9edafce698ff4184bda8a0c1c38b/console.sock,server,nowait -device nvdimm,id=nv0,memdev=mem0 -object memory-backend-file,id=mem0,mem-path=/usr/share/kata-containers/kata-containers-image_clearlinux_1.10.0-rc0_agent_8a4d901772.img,size=134217728 -device virtio-scsi-pci,id=scsi0,disable-modern=false,romfile= -object rng-random,id=rng0,filename=/dev/urandom -device virtio-rng,rng=rng0,romfile= -device virtserialport,chardev=charch0,id=channel0,name=agent.channel.0 -chardev socket,id=charch0,path=/run/vc/vm/7f3be98b62b18f29acbe93aea60c04e0168d9edafce698ff4184bda8a0c1c38b/kata.sock,server,nowait -device virtio-9p-pci,disable-modern=false,fsdev=extra-9p-kataShared,mount_tag=kataShared,romfile= -fsdev local,id=extra-9p-kataShared,path=/run/kata-containers/shared/sandboxes/7f3be98b62b18f29acbe93aea60c04e0168d9edafce698ff4184bda8a0c1c38b,security_model=none -netdev tap,id=network-0,vhost=on,vhostfds=3,fds=4 -device driver=virtio-net-pci,netdev=network-0,mac=26:e5:a9:23:9e:9a,disable-modern=false,mq=on,vectors=4,romfile= -global kvm-pit.lost_tick_policy=discard -vga none -no-user-config -nodefaults -nographic -daemonize -object memory-backend-file,id=dimm1,size=2048M,mem-path=/dev/hugepages -numa node,memdev=dimm1 -kernel /usr/share/kata-containers/vmlinuz-4.19.86.59-3.1.container -append tsc=reliable no_timer_check rcupdate.rcu_expedited=1 i8042.direct=1 i8042.dumbkbd=1 i8042.nopnp=1 i8042.noaux=1 noreplace-smp reboot=k console=hvc0 console=hvc1 iommu=off cryptomgr.notests net.ifnames=0 pci=lastbus=0 root=/dev/pmem0p1 rootflags=dax,data=ordered,errors=remount-ro ro rootfstype=ext4 quiet systemd.show_status=false panic=1 nr_cpus=8 agent.use_vsock=false systemd.unit=kata-containers.target systemd.mask=systemd-networkd.service systemd.mask=systemd-networkd.socket -pidfile /run/vc/vm/7f3be98b62b18f29acbe93aea60c04e0168d9edafce698ff4184bda8a0c1c38b/pid -smp 1,cores=1,threads=1,sockets=8,maxcpus=8
root     43438  0.0  0.0      0     0 ?        S    12月12   0:00 [vhost-43436]
root     43442  0.0  0.0      0     0 ?        S    12月12   0:00 [kvm-pit/43436]
[root@host-69 kata_7f3be98b62b18f29acbe93aea60c04e0168d9edafce698ff4184bda8a0c1c38b]#
```

bpradipt:
Fixes: #2353

Signed-off-by: Pradipta Banerjee <pradipta.banerjee@gmail.com>

### DIFF
--- a/virtcontainers/kata_agent.go
+++ b/virtcontainers/kata_agent.go
@@ -18,6 +18,7 @@ import (
 	"syscall"
 	"time"
 
+	units "github.com/docker/go-units"
 	"github.com/gogo/protobuf/proto"
 	aTypes "github.com/kata-containers/agent/pkg/types"
 	kataclient "github.com/kata-containers/agent/protocols/client"
@@ -33,6 +34,7 @@ import (
 	"github.com/kata-containers/runtime/virtcontainers/pkg/uuid"
 	"github.com/kata-containers/runtime/virtcontainers/store"
 	"github.com/kata-containers/runtime/virtcontainers/types"
+	"github.com/kata-containers/runtime/virtcontainers/utils"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/sirupsen/logrus"
@@ -140,6 +142,17 @@ var kataHostSharedDir = func() string {
 		return filepath.Join(rootless.GetRootlessDir(), defaultKataHostSharedDir) + "/"
 	}
 	return defaultKataHostSharedDir
+}
+
+func getPagesizeFromOpt(fsOpts []string) string {
+	//example options array: "rw", "relatime", "seclabel", "pagesize=2M"
+
+	for _, opt := range fsOpts {
+		if strings.HasPrefix(opt, "pagesize=") {
+			return strings.TrimPrefix(opt, "pagesize=")
+		}
+	}
+	return ""
 }
 
 // Shared path handling:
@@ -1429,6 +1442,13 @@ func (k *kataAgent) createContainer(sandbox *Sandbox, c *Container) (p *Process,
 	epheStorages := k.handleEphemeralStorage(ociSpec.Mounts)
 	ctrStorages = append(ctrStorages, epheStorages...)
 
+	k.Logger().WithField("ociSpec Hugepage Resources", ociSpec.Linux.Resources.HugepageLimits).Debug("ociSpec HugepageLimit")
+	hugepages, err := k.handleHugepages(ociSpec.Mounts, ociSpec.Linux.Resources.HugepageLimits)
+	if err != nil {
+		return nil, err
+	}
+	ctrStorages = append(ctrStorages, hugepages...)
+
 	localStorages := k.handleLocalStorage(ociSpec.Mounts, sandbox.id, c.rootfsSuffix)
 	ctrStorages = append(ctrStorages, localStorages...)
 
@@ -1548,6 +1568,71 @@ func (k *kataAgent) handleEphemeralStorage(mounts []specs.Mount) []*grpc.Storage
 		}
 	}
 	return epheStorages
+}
+
+// handleHugePages handles hugepages storage by
+// creating a Storage from corresponding source of the mount point
+func (k *kataAgent) handleHugepages(mounts []specs.Mount, hugepageLimits []specs.LinuxHugepageLimit) ([]*grpc.Storage, error) {
+	//Map to hold the total memory of each type of hugepages
+	optionsMap := make(map[int64]string)
+
+	for _, hp := range hugepageLimits {
+		if hp.Limit != 0 {
+			k.Logger().WithFields(logrus.Fields{
+				"Pagesize": hp.Pagesize,
+				"Limit":    hp.Limit,
+			}).Info("hugepage request")
+			//example Pagesize 2MB, 1GB etc. The Limit are in Bytes
+			pageSize, err := units.RAMInBytes(hp.Pagesize)
+			if err != nil {
+				k.Logger().Error("Unable to convert pagesize to bytes")
+				return nil, err
+			}
+			totalHpSizeStr := strconv.FormatUint(hp.Limit, 10)
+			optionsMap[pageSize] = totalHpSizeStr
+		}
+	}
+
+	var hugepages []*grpc.Storage
+	for idx, mnt := range mounts {
+		if mnt.Type != KataLocalDevType {
+			continue
+		}
+		//HugePages mount Type is Local
+		if _, fsType, fsOptions, _ := utils.GetDevicePathAndFsTypeOptions(mnt.Source); fsType == "hugetlbfs" {
+			k.Logger().WithField("fsOptions", fsOptions).Debug("hugepage mount options")
+			//Find the pagesize from the mountpoint options
+			pagesizeOpt := getPagesizeFromOpt(fsOptions)
+			if pagesizeOpt == "" {
+				return nil, fmt.Errorf("No pagesize option found in filesystem mount options")
+			}
+			pageSize, err := units.RAMInBytes(pagesizeOpt)
+			if err != nil {
+				k.Logger().Error("Unable to convert pagesize from fs mount options to bytes")
+				return nil, err
+			}
+			//Create mount option string
+			options := fmt.Sprintf("pagesize=%s,size=%s", strconv.FormatInt(pageSize, 10), optionsMap[pageSize])
+			k.Logger().WithField("Hugepage options string", options).Debug("hugepage mount options")
+			// Set the mount source path to a path that resides inside the VM
+			mounts[idx].Source = filepath.Join(ephemeralPath(), filepath.Base(mnt.Source))
+			// Set the mount type to "bind"
+			mounts[idx].Type = "bind"
+
+			// Create a storage struct so that kata agent is able to create
+			// hugetlbfs backed volume inside the VM
+			hugepage := &grpc.Storage{
+				Driver:     KataEphemeralDevType,
+				Source:     "nodev",
+				Fstype:     "hugetlbfs",
+				MountPoint: mounts[idx].Source,
+				Options:    []string{options},
+			}
+			hugepages = append(hugepages, hugepage)
+		}
+
+	}
+	return hugepages, nil
 }
 
 // handleLocalStorage handles local storage within the VM
--- a/virtcontainers/kata_agent_test.go
+++ b/virtcontainers/kata_agent_test.go
@@ -408,6 +408,61 @@ func TestHandleLocalStorage(t *testing.T) {
 	assert.Equal(t, localMountPoint, expected)
 }
 
+func TestHandleHugepages(t *testing.T) {
+	if os.Getuid() != 0 {
+		t.Skip("Test disabled as requires root user")
+	}
+
+	assert := assert.New(t)
+
+	dir, err := ioutil.TempDir("", "hugepages-test")
+	assert.Nil(err)
+	defer os.RemoveAll(dir)
+
+	k := kataAgent{}
+	var mounts []specs.Mount
+	var hugepageLimits []specs.LinuxHugepageLimit
+
+	hugepageDirs := [2]string{"hugepages-1Gi", "hugepages-2Mi"}
+	options := [2]string{"pagesize=1024M", "pagesize=2M"}
+
+	for i := 0; i < 2; i++ {
+		target := path.Join(dir, hugepageDirs[i])
+		err := os.MkdirAll(target, 0777)
+		assert.NoError(err, "Unable to create dir %s", target)
+
+		err = syscall.Mount("nodev", target, "hugetlbfs", uintptr(0), options[i])
+		assert.NoError(err, "Unable to mount %s", target)
+
+		defer syscall.Unmount(target, 0)
+		defer os.RemoveAll(target)
+		mount := specs.Mount{
+			Type:   KataLocalDevType,
+			Source: target,
+		}
+		mounts = append(mounts, mount)
+
+	}
+
+	hugepageLimits = []specs.LinuxHugepageLimit{
+		{
+			Pagesize: "1GB",
+			Limit:    1073741824,
+		},
+		{
+			Pagesize: "2MB",
+			Limit:    134217728,
+		},
+	}
+
+	hugepages, err := k.handleHugepages(mounts, hugepageLimits)
+
+	assert.NoError(err, "Unable to handle hugepages %v", hugepageLimits)
+	assert.NotNil(hugepages)
+	assert.Equal(len(hugepages), 2)
+
+}
+
 func TestHandleDeviceBlockVolume(t *testing.T) {
 	k := kataAgent{}
 
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -1873,8 +1873,9 @@ func (s *Sandbox) updateResources() error {
 	sandboxVCPUs += s.hypervisor.hypervisorConfig().NumVCPUs
 
 	sandboxMemoryByte := s.calculateSandboxMemory()
+
 	// Add default / rsvd memory for sandbox.
-	sandboxMemoryByte += int64(s.hypervisor.hypervisorConfig().MemorySize) << utils.MibToBytesShift
+	sandboxMemoryByte += uint64(s.hypervisor.hypervisorConfig().MemorySize) << utils.MibToBytesShift
 
 	// Update VCPUs
 	s.Logger().WithField("cpus-sandbox", sandboxVCPUs).Debugf("Request to hypervisor to update vCPUs")
@@ -1894,7 +1895,9 @@ func (s *Sandbox) updateResources() error {
 
 	// Update Memory
 	s.Logger().WithField("memory-sandbox-size-byte", sandboxMemoryByte).Debugf("Request to hypervisor to update memory")
-	newMemory, updatedMemoryDevice, err := s.hypervisor.resizeMemory(uint32(sandboxMemoryByte>>utils.MibToBytesShift), s.state.GuestMemoryBlockSizeMB, s.state.GuestMemoryHotplugProbe)
+	newMemoryMB := uint32(sandboxMemoryByte >> utils.MibToBytesShift)
+
+	newMemory, updatedMemoryDevice, err := s.hypervisor.resizeMemory(newMemoryMB, s.state.GuestMemoryBlockSizeMB, s.state.GuestMemoryHotplugProbe)
 	if err != nil {
 		return err
 	}
@@ -1912,8 +1915,8 @@ func (s *Sandbox) updateResources() error {
 	return nil
 }
 
-func (s *Sandbox) calculateSandboxMemory() int64 {
-	memorySandbox := int64(0)
+func (s *Sandbox) calculateSandboxMemory() uint64 {
+	memorySandbox := uint64(0)
 	for _, c := range s.config.Containers {
 		// Do not hot add again non-running containers resources
 		if cont, ok := s.containers[c.ID]; ok && cont.state.State == types.StateStopped {
@@ -1921,10 +1924,18 @@ func (s *Sandbox) calculateSandboxMemory() int64 {
 			continue
 		}
 
-		if m := c.Resources.Memory; m != nil && m.Limit != nil {
-			memorySandbox += *m.Limit
+		if m := c.Resources.Memory; m != nil && m.Limit != nil && *m.Limit > 0 {
+			memorySandbox += uint64(*m.Limit)
+			s.Logger().WithField("memory limit", memorySandbox).Info("Memory Sandbox + Memory Limit ")
+		}
+
+		//Add hugepages memory
+		//HugepageLimit is uint64 - https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L242
+		for _, l := range c.Resources.HugepageLimits {
+			memorySandbox += uint64(l.Limit)
 		}
 	}
+
 	return memorySandbox
 }
 
--- a/virtcontainers/sandbox_test.go
+++ b/virtcontainers/sandbox_test.go
@@ -149,13 +149,14 @@ func TestCalculateSandboxMem(t *testing.T) {
 	sandbox.config = &SandboxConfig{}
 	unconstrained := newTestContainerConfigNoop("cont-00001")
 	constrained := newTestContainerConfigNoop("cont-00001")
-	limit := int64(4000)
-	constrained.Resources.Memory = &specs.LinuxMemory{Limit: &limit}
+	mlimit := int64(4000)
+	limit := uint64(4000)
+	constrained.Resources.Memory = &specs.LinuxMemory{Limit: &mlimit}
 
 	tests := []struct {
 		name       string
 		containers []ContainerConfig
-		want       int64
+		want       uint64
 	}{
 		{"1-unconstrained", []ContainerConfig{unconstrained}, 0},
 		{"2-unconstrained", []ContainerConfig{unconstrained, unconstrained}, 0},
@@ -171,6 +172,45 @@ func TestCalculateSandboxMem(t *testing.T) {
 			assert.Equal(t, got, tt.want)
 		})
 	}
+}
+
+func TestSandboxHugepageLimit(t *testing.T) {
+	contConfig1 := newTestContainerConfigNoop("cont-00001")
+	contConfig2 := newTestContainerConfigNoop("cont-00002")
+	limit := int64(4000)
+	contConfig1.Resources.Memory = &specs.LinuxMemory{Limit: &limit}
+	contConfig2.Resources.Memory = &specs.LinuxMemory{Limit: &limit}
+	hConfig := newHypervisorConfig(nil, nil)
+
+	defer cleanUp()
+	// create a sandbox
+	s, err := testCreateSandbox(t,
+		testSandboxID,
+		MockHypervisor,
+		hConfig,
+		NoopAgentType,
+		NetworkConfig{},
+		[]ContainerConfig{contConfig1, contConfig2},
+		nil)
+
+	assert.NoError(t, err)
+
+	hugepageLimits := []specs.LinuxHugepageLimit{
+		{
+			Pagesize: "1GB",
+			Limit:    322122547,
+		},
+		{
+			Pagesize: "2MB",
+			Limit:    134217728,
+		},
+	}
+
+	for i := range s.config.Containers {
+		s.config.Containers[i].Resources.HugepageLimits = hugepageLimits
+	}
+	err = s.updateResources()
+	assert.NoError(t, err)
 }
 
 func TestCreateSandboxEmptyID(t *testing.T) {
--- a/virtcontainers/utils/utils_linux_test.go
+++ b/virtcontainers/utils/utils_linux_test.go
@@ -6,7 +6,10 @@
 package utils
 
 import (
+	"bytes"
 	"errors"
+	"os/exec"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -36,16 +39,27 @@ func TestFindContextID(t *testing.T) {
 
 func TestGetDevicePathAndFsTypeEmptyMount(t *testing.T) {
 	assert := assert.New(t)
-	_, _, err := GetDevicePathAndFsType("")
+	_, _, _, err := GetDevicePathAndFsTypeOptions("")
 	assert.Error(err)
 }
 
 func TestGetDevicePathAndFsTypeSuccessful(t *testing.T) {
 	assert := assert.New(t)
 
-	path, fstype, err := GetDevicePathAndFsType("/proc")
+	cmdStr := "grep ^proc  /proc/mounts"
+	cmd := exec.Command("sh", "-c", cmdStr)
+	output, err := cmd.Output()
+	assert.NoError(err)
+
+	data := bytes.Split(output, []byte(" "))
+	fstypeOut := string(data[2])
+	optsOut := strings.Split(string(data[3]), ",")
+
+	path, fstype, fsOptions, err := GetDevicePathAndFsTypeOptions("/proc")
 	assert.NoError(err)
 
 	assert.Equal(path, "proc")
 	assert.Equal(fstype, "proc")
+	assert.Equal(fstype, fstypeOut)
+	assert.Equal(fsOptions, optsOut)
 }
--- a/virtcontainers/container.go
+++ b/virtcontainers/container.go
@@ -1358,7 +1358,7 @@ func (c *Container) hotplugDrive() error {
 			c.rootfsSuffix = ""
 		}
 		// If device mapper device, then fetch the full path of the device
-		devicePath, fsType, err = utils.GetDevicePathAndFsType(dev.mountPoint)
+		devicePath, fsType, _, err = utils.GetDevicePathAndFsTypeOptions(dev.mountPoint)
 		if err != nil {
 			return err
 		}
--- a/virtcontainers/device/config/pmem.go
+++ b/virtcontainers/device/config/pmem.go
@@ -75,7 +75,7 @@ func PmemDeviceInfo(source, destination string) (*DeviceInfo, error) {
 		return nil, fmt.Errorf("backing file %v has not PFN signature", device.HostPath)
 	}
 
-	_, fstype, err := utils.GetDevicePathAndFsType(source)
+	_, fstype, _, err := utils.GetDevicePathAndFsTypeOptions(source)
 	if err != nil {
 		pmemLog.WithError(err).WithField("mount-point", source).Warn("failed to get fstype: using ext4")
 		fstype = "ext4"
--- a/virtcontainers/mount.go
+++ b/virtcontainers/mount.go
@@ -400,7 +400,7 @@ func IsEphemeralStorage(path string) bool {
 		return false
 	}
 
-	if _, fsType, _ := utils.GetDevicePathAndFsType(path); fsType == "tmpfs" {
+	if _, fsType, _, _ := utils.GetDevicePathAndFsTypeOptions(path); fsType == "tmpfs" {
 		return true
 	}
 
@@ -415,7 +415,7 @@ func Isk8sHostEmptyDir(path string) bool {
 		return false
 	}
 
-	if _, fsType, _ := utils.GetDevicePathAndFsType(path); fsType != "tmpfs" {
+	if _, fsType, _, _ := utils.GetDevicePathAndFsTypeOptions(path); fsType != "tmpfs" {
 		return true
 	}
 	return false
--- a/virtcontainers/utils/utils_linux.go
+++ b/virtcontainers/utils/utils_linux.go
@@ -99,11 +99,12 @@ const (
 	procDeviceIndex = iota
 	procPathIndex
 	procTypeIndex
+	procOptionIndex
 )
 
-// GetDevicePathAndFsType gets the device for the mount point and the file system type
-// of the mount.
-func GetDevicePathAndFsType(mountPoint string) (devicePath, fsType string, err error) {
+// GetDevicePathAndFsTypeOptions gets the device for the mount point, the file system type
+// and mount options
+func GetDevicePathAndFsTypeOptions(mountPoint string) (devicePath, fsType string, fsOptions []string, err error) {
 	if mountPoint == "" {
 		err = fmt.Errorf("Mount point cannot be empty")
 		return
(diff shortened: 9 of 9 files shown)


bpradipt:
Sending this to kick start discussion on how to support hugepage for containers with Kata. 
A google document explaining the details can be accessed from the following link - https://docs.google.com/document/d/1Oo_B2RjkI6pm1Sd8PXL6CzwPv9u0WL-RP6G2-k5__mw/edit

cc @amshinde @fidencio @dgibson @c3d

bpradipt:
Related agent side code - https://github.com/kata-containers/agent/pull/872

virtcontainers/kata_agent.go
```diff
@@ -1550,6 +1556,61 @@ func (k *kataAgent) handleEphemeralStorage(mounts []specs.Mount) []*grpc.Storage
...
+// creating a Storage from corresponding source of the mount point
+func (k *kataAgent) handleHugepages(mounts []specs.Mount, hugepageLimits []specs.LinuxHugepageLimit) []*grpc.Storage {
+	//Map to hold the total memory of each type of hugepages
+	optionsMap := make(map[string]string)
+	var HugePageSizeUnitList = []string{"B", "K", "M", "G", "T", "P"}
+
+	for _, hp := range hugepageLimits {
+		if hp.Limit != 0 {
+			k.Logger().WithField("Pagesize", hp.Pagesize).Info("hugepage request")
+			k.Logger().WithField("Limit", hp.Limit).Info("hugepage request")
```
jodh-intel, line 1569: Please don't split logical log entries across multiple log lines, just do something like:

```go
k.Logger().WithFields(logrus.Fields{
    "Pagesize": hp.Pagesize,
    "Limit" : hp.Limit.
}).Info("hugepage request")
```

virtcontainers/kata_agent.go
```diff
@@ -1550,6 +1556,61 @@ func (k *kataAgent) handleEphemeralStorage(mounts []specs.Mount) []*grpc.Storage
...
+			//example Pagesize 2MB, 1GB etc. The Limit are in Bytes
+			pageSize, _ := units.RAMInBytes(hp.Pagesize)
+			pageSizeStr := units.CustomSize("%g%s", float64(pageSize), 1024.0, HugePageSizeUnitList)
```
jodh-intel, line 1572: I don't understand this - why not use the `units.*` package to deal with the suffixes rather than specifying your own list...?

virtcontainers/kata_agent.go
```diff
@@ -1550,6 +1556,61 @@ func (k *kataAgent) handleEphemeralStorage(mounts []specs.Mount) []*grpc.Storage
...
+			totalHpSize := fmt.Sprintf("%v", hp.Limit)
+			optionsMap[pageSizeStr] = totalHpSize
+		}
+	}
+
+	var hugepages []*grpc.Storage
+	for idx, mnt := range mounts {
+		//HugePages mount Type is Local
+		if mnt.Type == KataLocalDevType {
```
jodh-intel, line 1581: This code is starting to get quite long and deeply indented. I'd consider inverting the logic to simplify it:

```go
if mnt.Type != KataLocalDevType {
    continue
}

if _, fsType, fsOptions, _ := ...
```

virtcontainers/sandbox.go
```diff
@@ -1923,6 +1923,14 @@ func (s *Sandbox) calculateSandboxMemory() int64 {
 
 		if m := c.Resources.Memory; m != nil && m.Limit != nil {
 			memorySandbox += *m.Limit
+			s.Logger().WithField("memory limit", memorySandbox).Info("Memory Sandbox + Memory Limit ")
+		}
+
+		//Add hugepages memory
+		//HugepageLimit is uint64 - https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L242
+		for _, l := range c.Resources.HugepageLimits {
+			memorySandbox += int64(l.Limit)
```
jodh-intel, line 1932: This is confusing: is it `unit64` or `int64`?

virtcontainers/kata_agent.go
bpradipt: Sure.. I'll change it

virtcontainers/kata_agent.go
bpradipt: I used the CustomSize method to explicitly indicate the acceptable unit options for hugepage size. I'll recheck if there is a different method without the need to specify the custom format.

virtcontainers/kata_agent.go
bpradipt: Sure. Will change it

virtcontainers/sandbox.go
bpradipt: the HugepageLimit is in uint64 :-(

jodh-intel: Hmm. I guess you should check that `l.Limit <= math.MaxInt64` herethen. If it's bigger, set it to `math.MaxInt64` and make a logger call (but ideally update the function to return an error).

bpradipt: You are right. Sorry for the missing this.   Do you recommend returning error (which will change the current function) or just make a logger call ?

jodh-intel: I'd make `calculateSandboxMemory()` return an `error`. It's only called from one other function (`updateResources()`) and that function already returns an `error` so not too disruptive to update it. Also, by returning an `error` you could add a unit test to exercise the `EOVERFLOW` scenario.

bpradipt: sure.. I'll make the changes

virtcontainers/kata_agent.go
```diff
@@ -1550,6 +1556,61 @@ func (k *kataAgent) handleEphemeralStorage(mounts []specs.Mount) []*grpc.Storage
...
+			if _, fsType, fsOptions, _ := utils.GetDevicePathAndFsTypeOptions(mnt.Source); fsType == "hugetlbfs" {
+				k.Logger().WithField("fsOptions", fsOptions).Debug("hugepage mount options")
+				//Find the pagesize from the mountpoint to use the right set of options
+				//pagesize is the last element in the mount option. Can it change ?
+				pagesizeOpt := fsOptions[len(fsOptions)-1]
```
dgibson, line 1586: Can you really count on the pagesize being the last option in the list?

virtcontainers/kata_agent.go
dgibson: I don't really understand pageSizeStr at all.  You need to convert the (underspecified) string from the spec into a pagesize in bytes, but why do you then need a string.  You should be able to do your comparisons just based on the absolute size from then onwards.

virtcontainers/sandbox.go
dgibson, line 1932: Is the limit from the spec in bytes, or in # of (huge)pages.  If the latter, don't you need to scale this up by the hugepage size?

bpradipt: Its in bytes

virtcontainers/kata_agent.go
bpradipt: In my tests I always found it as the last option hence avoided searching the option. But I'm not sure hence I added a comment as well. Any ideas?

virtcontainers/sandbox.go
bpradipt: @dgibson regarding your comment on pageSizeStr, I'm unable to provide inline response. Not sure why. So putting it here.
The mount options uses something like `pagesize=<pageSizeStr>` eg. `pagesize=1G`.  So using pageSizeStr to refer to the pagesize (1G, 2M etc.)

virtcontainers/kata_agent.go
bpradipt: @dgibson regarding your comment on pageSizeStr, I'm unable to provide inline response. Not sure why. So putting it here.
The mount options uses something like pagesize=<pageSizeStr> eg. pagesize=1G. So using pageSizeStr to refer to the pagesize (1G, 2M etc.). Further, I use this to get the actual size from the map to create the final mount option string.

bpradipt:
closing and re-opening to check if CI works

virtcontainers/kata_agent.go
bpradipt: @dgibson I looked at the kubernetes code and the kubernetes code only provides `pagesize` option - https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/emptydir/empty_dir.go#L381 which is then used for the actual hugepages mount on the host.
Checked the linux kernel (fs/hugetlbfs/inode.c) code as well. 
We should be able to safely assume that `pagesize` is the last option when it comes to Kubernetes. It doesn't use the `size` option. 
Let me know if you are ok, or want me to change the code to search for `pagesize` option ?

virtcontainers/sandbox.go
```diff
@@ -1923,9 +1926,30 @@ func (s *Sandbox) calculateSandboxMemory() int64 {
...
+			s.Logger().WithField("hugepages memory limit", memorySandbox).Info("Memory Sandbox + Hugepages Memory Limit ")
 		}
 	}
-	return memorySandbox
+
+	//Check overflow
+	if memorySandbox < 0 {
+		return 0, errors.New("Integer overflow")
+	} else {
+		return memorySandbox, nil
```
egernst, line 1951: From taking a look, I think it may be worth clarifying here that we don't differentiate between 'normal pages' and hugePages, since the entire VM will be backed by huge pages.  

Do I understand it correctly?  I think this is probably a higher-level way to validate that we're doing the right thing here, too (ie, do we error if the VM isn't backed by huge pages).

egernst (commented):
Some questions below, but stepping back I'm wondering how this will work in Kubernetes. To _actually_ provide huge pages, we want to volume to be backed by huge pages from the host itself. To facilitate that, AFAICT we'd want the entire VM to be backed by huge pages?

Assuming that....  huge pages are a finite resource, and is tracked by kubelet and the Kube-scheduler. If we back the entire VM in huge pages, we are using a lot more than just what was requested in the pod spec? Worse, I would think we should be killed, since we are bound by the pod's hugetlb cgroup, right?

I'd like to be able to properly support huge pages, but I'm not sure we can do it just in the runtime.

virtcontainers/sandbox.go
```diff
@@ -1923,9 +1926,30 @@ func (s *Sandbox) calculateSandboxMemory() int64 {
 
 		if m := c.Resources.Memory; m != nil && m.Limit != nil {
 			memorySandbox += *m.Limit
+			s.Logger().WithField("memory limit", memorySandbox).Info("Memory Sandbox + Memory Limit ")
+		}
+
+		//Add hugepages memory
+		//HugepageLimit is uint64 - https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L242
```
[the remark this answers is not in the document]
jodh-intel: This relates to https://github.com/kata-containers/runtime/pull/3109#discussion_r551415801. However, I'm looking at the code again and wondering why we cannot simply change `calculateSandboxMemory()` to return `uint64` rather than `int64`?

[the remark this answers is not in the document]
bpradipt: > This relates to [#3109 (comment)](https://github.com/kata-containers/runtime/pull/3109#discussion_r551415801). However, I'm looking at the code again and wondering why we cannot simply change `calculateSandboxMemory()` to return `uint64` rather than `int64`?

Memory limit is in int64 whereas hugepage limit is in unit64 .  One of the entities will need conversion.
https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L298
https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L242

virtcontainers/sandbox.go
```diff
@@ -1923,9 +1926,30 @@ func (s *Sandbox) calculateSandboxMemory() int64 {
...
+		for _, l := range c.Resources.HugepageLimits {
+			if l.Limit <= math.MaxInt64 {
+				memorySandbox += int64(l.Limit)
+			} else {
+				//Overflow
+				s.Logger().WithField("hugepages memory limit overflow", l.Limit).Warn("Memory Sandbox + Hugepages Memory Limit ")
```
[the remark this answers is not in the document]
bpradipt: > s/Warn/Error in this case, right?

ah yes.. Thanks

virtcontainers/kata_agent.go
```diff
@@ -1550,6 +1556,64 @@ func (k *kataAgent) handleEphemeralStorage(mounts []specs.Mount) []*grpc.Storage
...
+	for idx, mnt := range mounts {
+		if mnt.Type != KataLocalDevType {
+			continue
+		}
+		//HugePages mount Type is Local
+		if _, fsType, fsOptions, _ := utils.GetDevicePathAndFsTypeOptions(mnt.Source); fsType == "hugetlbfs" {
+			k.Logger().WithField("fsOptions", fsOptions).Debug("hugepage mount options")
+			//Find the pagesize from the mountpoint to use the right set of options
+			//pagesize is the last element in the mount option. Can it change ?
+			pagesizeOpt := fsOptions[len(fsOptions)-1]
```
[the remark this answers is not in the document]
bpradipt: > this seems too magical. I'd rather assume it can change.

@egernst I can go ahead and change the logic here. However can you please take a look at the following comment once and share your thoughts - https://github.com/kata-containers/runtime/pull/3109#discussion_r553878031

virtcontainers/sandbox.go
bpradipt: > From taking a look, I think it may be worth clarifying here that we don't differentiate between 'normal pages' and hugePages, since the entire VM will be backed by huge pages.
> 
> Do I understand it correctly? I think this is probably a higher-level way to validate that we're doing the right thing here, too (ie, do we error if the VM isn't backed by huge pages).

Yeah true.. We don't care if the VM is backed by hugepages or not.  Now question is should we allow hugepages inside the Kata VM only if the entire VM is backed by hugepages or we keep the two separately ?  

Although having hugepages inside the VM without the VM memory itself being backed by hugepages will not be of much use other than ensuring the apps requiring hugepages continues to run on Kata (eg. dpdk apps).

virtcontainers/sandbox.go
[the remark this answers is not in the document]
jodh-intel: Can't we do the following on top of your PR?:

```diff
diff --git a/virtcontainers/sandbox.go b/virtcontainers/sandbox.go
index 9f0a2f45..c87d4c4f 100644
--- a/virtcontainers/sandbox.go
+++ b/virtcontainers/sandbox.go
@@ -1877,7 +1877,7 @@ func (s *Sandbox) updateResources() error {
 		return err
 	}
 	// Add default / rsvd memory for sandbox.
-	sandboxMemoryByte += int64(s.hypervisor.hypervisorConfig().MemorySize) << utils.MibToBytesShift
+	sandboxMemoryByte += uint64(s.hypervisor.hypervisorConfig().MemorySize) << utils.MibToBytesShift
 
 	// Update VCPUs
 	s.Logger().WithField("cpus-sandbox", sandboxVCPUs).Debugf("Request to hypervisor to update vCPUs")
@@ -1897,7 +1897,9 @@ func (s *Sandbox) updateResources() error {
 
 	// Update Memory
 	s.Logger().WithField("memory-sandbox-size-byte", sandboxMemoryByte).Debugf("Request to hypervisor to update memory")
-	newMemory, updatedMemoryDevice, err := s.hypervisor.resizeMemory(uint32(sandboxMemoryByte>>utils.MibToBytesShift), s.state.GuestMemoryBlockSizeMB, s.state.GuestMemoryHotplugProbe)
+	newMemoryMB := uint32(sandboxMemoryByte >> utils.MibToBytesShift)
+
+	newMemory, updatedMemoryDevice, err := s.hypervisor.resizeMemory(newMemoryMB, s.state.GuestMemoryBlockSizeMB, s.state.GuestMemoryHotplugProbe)
 	if err != nil {
 		return err
 	}
@@ -1915,8 +1917,8 @@ func (s *Sandbox) updateResources() error {
 	return nil
 }
 
-func (s *Sandbox) calculateSandboxMemory() (int64, error) {
-	memorySandbox := int64(0)
+func (s *Sandbox) calculateSandboxMemory() (uint64, error) {
+	memorySandbox := uint64(0)
 	for _, c := range s.config.Containers {
 		// Do not hot add again non-running containers resources
 		if cont, ok := s.containers[c.ID]; ok && cont.state.State == types.StateStopped {
@@ -1924,8 +1926,8 @@ func (s *Sandbox) calculateSandboxMemory() (int64, error) {
 			continue
 		}
 
-		if m := c.Resources.Memory; m != nil && m.Limit != nil {
-			memorySandbox += *m.Limit
+		if m := c.Resources.Memory; m != nil && m.Limit != nil && *m.Limit > 0 {
+			memorySandbox += uint64(*m.Limit)
 			s.Logger().WithField("memory limit", memorySandbox).Info("Memory Sandbox + Memory Limit ")
 		}
 
@@ -1933,7 +1935,7 @@ func (s *Sandbox) calculateSandboxMemory() (int64, error) {
 		//HugepageLimit is uint64 - https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L242
 		for _, l := range c.Resources.HugepageLimits {
 			if l.Limit <= math.MaxInt64 {
-				memorySandbox += int64(l.Limit)
+				memorySandbox += uint64(l.Limit)
 			} else {
 				//Overflow
 				s.Logger().WithField("hugepages memory limit overflow", l.Limit).Warn("Memory Sandbox + Hugepages Memory Limit ")
```

virtcontainers/kata_agent.go
dgibson: The mount options allow "pagesize=1G" as syntactic sugar, but "pagesize=1073741824" will work too.  It makes sense to allow user input to use suffixes for brevity, but internally you should be able to use pagesize values in bytes everywhere.

virtcontainers/kata_agent.go
dgibson: Relying on exactly how external software (k8s in this case) presents the options is incredibly fragile.  You absolutely should search for the option.

virtcontainers/kata_agent.go
bpradipt: @dgibson addressed your comment, ptal

virtcontainers/kata_agent.go
bpradipt: addressed this in latest iteration

virtcontainers/sandbox.go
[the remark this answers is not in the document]
bpradipt: @jodh-intel the current iteration incorporates your above suggestion ..  ptal

virtcontainers/kata_agent.go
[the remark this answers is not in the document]
bpradipt: @egernst pushed a new iteration with logic for searching the option. ptal

virtcontainers/kata_agent.go
```diff
@@ -142,6 +144,17 @@ var kataHostSharedDir = func() string {
 	return defaultKataHostSharedDir
 }
 
+func getPagesizeFromOpt(fsOpts []string) string {
+	//example options array: "rw", "relatime", "seclabel", "pagesize=2M"
+
+	for _, opt := range fsOpts {
+		if strings.Contains(opt, "pagesize") {
```
dgibson, line 151: This is still sloppy, since it will get confused by an option which contains the works "pagesize" but not at the beginning.  I really think we should do this properly: check explicitly for a prefix of "pagesize=" and nothing else.

virtcontainers/kata_agent.go
```diff
@@ -1550,6 +1570,72 @@ func (k *kataAgent) handleEphemeralStorage(mounts []specs.Mount) []*grpc.Storage
...
+				return nil, fmt.Errorf("No pagesize option found in filesystem mount options")
+			}
+			pageSize, err := units.RAMInBytes(pagesizeOpt)
+			if err != nil {
+				k.Logger().Error("Unable to convert pagesize from fs mount options to bytes")
+				return nil, err
+			}
+			pageSizeStr := strconv.FormatInt(pageSize, 10)
+			//Create mount option string
+			options := "pagesize=" + pageSizeStr + "," + "size=" + optionsMap[pageSize]
```
dgibson, line 1616: This would probably be more readable if you combine this expression and the FormatInt into a single Sprintf.

virtcontainers/utils/utils_linux_test.go
```diff
@@ -36,16 +36,18 @@ func TestFindContextID(t *testing.T) {
...
 	assert := assert.New(t)
+	opts := []string{"rw", "nosuid", "nodev", "noexec", "relatime"}
 
-	path, fstype, err := GetDevicePathAndFsType("/proc")
+	path, fstype, fsOptions, err := GetDevicePathAndFsTypeOptions("/proc")
 	assert.NoError(err)
 
 	assert.Equal(path, "proc")
 	assert.Equal(fstype, "proc")
+	assert.True(DeepCompare(fsOptions, opts))
```
dgibson, line 52: Can we actually rely on the options for /proc being the same on every distro?

virtcontainers/kata_agent.go
bpradipt: sure, I'll change it

virtcontainers/utils/utils_linux_test.go
bpradipt: > Can we actually rely on the options for /proc being the same on every distro?

On Fedora and Ubuntu I didn't find any difference. But you are right. Any suggestions on how to handle this for the specific unit test ?

virtcontainers/kata_agent.go
bpradipt: > This would probably be more readable if you combine this expression and the FormatInt into a single Sprintf.

ok

virtcontainers/utils/utils_linux_test.go
jodh-intel: I think @dgibson is right - it's going to be safer (and more future-proof) to make `TestGetDevicePathAndFsTypeSuccessful()` effectivley do `grep /proc /proc/mounts` and compare it to the output of `GetDevicePathAndFsTypeOptions("/proc")`.

virtcontainers/sandbox.go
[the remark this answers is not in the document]
bpradipt: The function is modified based on inputs from James. Ptal

virtcontainers/kata_agent.go
```diff
@@ -142,6 +144,17 @@ var kataHostSharedDir = func() string {
 	return defaultKataHostSharedDir
 }
 
+func getPagesizeFromOpt(fsOpts []string) string {
+	//example options array: "rw", "relatime", "seclabel", "pagesize=2M"
+
+	for _, opt := range fsOpts {
+		if strings.Contains(opt, "pagesize=") {
```
jodh-intel, line 151: This should be `strings.HasPrefix()` to avoid matching things like "`foo_bar_pagesize=`" (and for symmetry with the `TrimPrefix()` call).

virtcontainers/kata_agent_test.go
```diff
@@ -408,6 +408,62 @@ func TestHandleLocalStorage(t *testing.T) {
 	assert.Equal(t, localMountPoint, expected)
 }
 
+func TestHandleHugepages(t *testing.T) {
+	if os.Getuid() != 0 {
+		t.Skip("Test disabled as requires root user")
+	}
+
+	dir, err := ioutil.TempDir("", "hugepages-test")
+	assert.Nil(t, err)
```
jodh-intel, line 417: Nit: You can avoid passing the `t` param (easier to read ;-) if you use the following idiom:

```go
assert := assert.New(t)

// ...

assert.NoError(err)
```

virtcontainers/kata_agent_test.go
```diff
@@ -408,6 +408,62 @@ func TestHandleLocalStorage(t *testing.T) {
...
+	k := kataAgent{}
+	var mounts []specs.Mount
+	var hugepageLimits []specs.LinuxHugepageLimit
+
+	hugepageDirs := [2]string{"hugepages-1Gi", "hugepages-2Mi"}
+	options := [2]string{"pagesize=1024M", "pagesize=2M"}
+
+	for i := 0; i < 2; i++ {
+		target := path.Join(dir, hugepageDirs[i])
+		err := os.MkdirAll(target, 0777)
```
jodh-intel, line 429: Missing check on `err`.

virtcontainers/kata_agent_test.go
```diff
@@ -408,6 +408,62 @@ func TestHandleLocalStorage(t *testing.T) {
...
+		if err = syscall.Mount("nodev", target, "hugetlbfs", uintptr(0), options[i]); err != nil {
+			fmt.Printf("Unable to mount %s: %v\n", target, err)
+		}
```
jodh-intel, line 432: This looks like it might need a slight rework (since it's test code) to something like this:

```go
err = syscall.Mount("nodev", target, "hugetlbfs", uintptr(0), options[i])
assert.NoError(err, "Unable to mount %s", target")

```

virtcontainers/kata_agent_test.go
```diff
@@ -408,6 +408,62 @@ func TestHandleLocalStorage(t *testing.T) {
...
+		defer syscall.Unmount(target, 0)
+		defer os.RemoveAll(target)
+		mount := specs.Mount{
+			Type:   KataLocalDevType,
+			Source: target,
+		}
+		mounts = append(mounts, mount)
+
+	}
+	fmt.Printf("mounts: %v\n", mounts)
```
jodh-intel, line 443: Do we really need these printf calls in the tests? They should ideally be as quiet as possible, but on assertion failure dump full details.

virtcontainers/utils/utils_linux_test.go
```diff
@@ -36,16 +39,27 @@ func TestFindContextID(t *testing.T) {
...
-	_, _, err := GetDevicePathAndFsType("")
+	_, _, _, err := GetDevicePathAndFsTypeOptions("")
 	assert.Error(err)
 }
 
 func TestGetDevicePathAndFsTypeSuccessful(t *testing.T) {
 	assert := assert.New(t)
 
-	path, fstype, err := GetDevicePathAndFsType("/proc")
+	cmdStr := "grep '/proc '  /proc/mounts "
```
jodh-intel, line 49: It would be safer to anchor this using something like:

```suggestion
	cmdStr := "grep ^proc /proc/mounts"
```

An alternative to the `exec` + `grep` dance is to use `ioutil.ReadFile()`.

jodh-intel (commented):
Thanks @bpradipt. A few comments.

virtcontainers/kata_agent_test.go
bpradipt: Makes sense. I have updated it

virtcontainers/utils/utils_linux_test.go
bpradipt: Let me know if you prefer ReadFile. For now I have kept the `exec + grep` combination with your suggested changes.

jodh-intel (approved):
Thanks @bpradipt.

lgtm

virtcontainers/sandbox_test.go
```diff
@@ -173,6 +174,45 @@ func TestCalculateSandboxMem(t *testing.T) {
...
+		NetworkConfig{},
+		[]ContainerConfig{contConfig1, contConfig2},
+		nil)
+
+	assert.NoError(t, err)
+
+	hugepageLimits := []specs.LinuxHugepageLimit{
+		{
+			Pagesize: "1GB",
+			Limit:    322122547,
```
dgibson, line 201: This limit isn't a multiple of 1GiB, which doesn't seem sensible.

dgibson (approved):
LGTM, excepting one nit which I've commented.

virtcontainers/sandbox_test.go
bpradipt: @dgibson good catch. I have updated it. Thanks

PR #3115, opened 2021-01-08
blk-dev: hotplug read only if applicable

egernst:
If a volume is backed by a block device, let's make sure we take into account whether it should be hot plugged RO or not.

While updating this, I made some minor non-functional changes to correct comments, make storage handling a bit more clear.


jodh-intel:
Marking as dnm due to vendor change.

@egernst - I'm slightly surprised this is being raised on 1.x - is this correct?

egernst:
It is correct. I'll be back porting to stable and forward to 2.0 as well. I think we'll want to update our vendoring to be consistent between these.

bergwolf (approved):
lgtm!

egernst:
Minor update to fix pmem test failures.

mxpv (approved):
LGTM

PR #3121, opened 2021-01-17
cli/config: move the pflash annotation to hypervisor part

jongwu:
pflash annotation is placed under factory which should be
hypervisor.

Fixes: #3120
Signed-off-by: <jianyong.wu@arm.com>

@jodh-intel @GabyCT

### DIFF
--- a/cli/config/configuration-qemu.toml.in
+++ b/cli/config/configuration-qemu.toml.in
@@ -243,6 +243,10 @@ valid_vhost_user_store_paths = @DEFVALIDVHOSTUSERSTOREPATHS@
 # Your distribution recommends: @DEFVALIDFILEMEMBACKENDS@
 valid_file_mem_backends = @DEFVALIDFILEMEMBACKENDS@
 
+# -pflash can add image file to VM. The arguments of it should be in format
+# of ["/path/to/flash0.img", "/path/to/flash1.img"]
+pflashes = []
+
 # Enable swap of vm memory. Default false.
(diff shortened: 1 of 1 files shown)


jodh-intel (approved):
Thanks @jongwu. You may want to make it clearer that this is a breaking change?

jongwu:
thanks @jodh-intel -, it is a remedy for my mistake of putting the "pflash" under the factory. part. by the way, I'm not clear with the "breaking change". what's that mean here❓

jodh-intel:
Hi @jongwu - I mean that a user who has set `pflashes` in their current configuration file (and maybe copied it to `/etc/kata-containers/` needs to be aware that this will no longer work when *this* PR is included in a release.

We don't use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary), but you could add a simple "`BREAKING-CHANGE`" tag to the first line of the commit somewhere so that it appears in the release notes for the next release maybe?

jongwu:
Hi @jodh-intel , I think we don't need that tag. pflash does not work for now as it has been put in the wrong place and will work after include this PR.

jongwu:
hello @lifupan @GabyCT, any comments?

lifupan (approved):

lifupan:
@jongwu  

Did  the 2.0 branch need this fix?

jongwu:
thanks @lifupan , I have just raised a [PR](https://github.com/kata-containers/kata-containers/pull/1318) to enable acpi for 2.x

PR #3144, opened 2021-02-25
Backports to stable 1.12

bergwolf:
c5447462 cpuset: when creating container, don't pass cpuset details
607e96dd virtcontainers: Fix the memory leak caused by supportedQemuMachines


fidencio:
@bergwolf, can this one be closed? Or do we still need it?

Issue #3157, opened 2021-04-09
cli: the default hostname should not be "runc"


LiangZhou-CTY:
When use "kata-runtime spec" to generate config.json, 
we see that the default hostname is "runc", which is not correct.
 
I understand that it's caused by the call chain spec.go -> specconv.Example(),
which is from runc package, but anyway the default hostname of kata container 
should not be "runc". 
I think "kata" is more proper for the default hostname. 

Use the commands below can expose this issue:
```
mkdir rootfs
docker export $(docker create busybox) | tar -xf - -C rootfs
kata-runtime spec
kata-runtime run busybox
/ # hostname
runc
```

LiangZhou-CTY:
I'm going to work on it.

PR #3158, opened 2021-04-09
cli: Fix the incorrect default hostname "runc"

LiangZhou-CTY:
When use "kata-runtime spec" to generate config.json, 
the default hostname is "runc", which is not correct. 
The proper one should be "kata", instead of "runc".

Fixes: #3157

Signed-off-by: Liang Zhou <zhoul110@chinatelecom.cn>

### DIFF
--- a/cli/spec.go
+++ b/cli/spec.go
@@ -83,6 +83,8 @@ generate a proper rootless spec file.`,
 
 		spec := specconv.Example()
 
+		spec.Hostname = "kata"
+
 		checkNoFile := func(name string) error {
 			_, err := os.Stat(name)
 			if err == nil {


devimc:
@LiangZhou-CTY thanks for the PR, silly question: are you still using kata 1.x ?  may be you want to raise this in the [2.x repo](https://github.com/kata-containers/kata-containers/) ?

LiangZhou-CTY:
> @LiangZhou-CTY thanks for the PR, silly question: are you still using kata 1.x ? may be you want to raise this in the [2.x repo](https://github.com/kata-containers/kata-containers/) ?

Yes I will check if 2.x has the similar issue or not but AFAIK seems that 2.x does not support “spec” command anymore...

fidencio (approved):
lgtm, thanks @LiangZhou-CTY!

devimc (approved):

